Issue #30, opened 2016-08-15
Error message "File type is not supported" when uses .nunit configuration file


CharliePoole:
@NikolayPianikov commented on [Mon Aug 15 2016](https://github.com/nunit/nunit/issues/1748)

Version dev-03141
[714679585.zip](https://github.com/nunit/nunit/files/418109/714679585.zip)
NUnit.Extension.NUnitProjectLoader.3.5.0-dev-03141 exists in the "packages" directory.

NikolayPianikov:
@CharliePoole I can't reproduce it on the last master branch, could we publish new packages to myget?

CharliePoole:
Every time we build master branch in AppVeyor it is published to MyGet.

NikolayPianikov:
So it looks very strange. It works only when I replace nunit\NUnit.ConsoleRunner.3.5.0-dev-03141\tools\nunit.engine.dll by my nunit.engine.dll compiled from a master branch.

NikolayPianikov:
I see new packages but it does not help. I am going to loook tomorrow deeper

NikolayPianikov:
I found a problem. The package NUnit.ConsoleRunner.3.5.0-dev-03141 has nunit.engine.dll This assembly contains class NUnit.Engine.Runners.MasterTestRunner. Reflected code of this class is not a last version from the master brunch and it contains bugs.
[Reflected.zip](https://github.com/nunit/nunit-console/files/424830/Reflected.zip)

NikolayPianikov:
Also I found a strange code in the class NUnit.Engine.Runners.DirectTestRunner, method LoadPackage()
`
// DirectRunner may be called with a single-assembly package
// or a set of assemblies as subpackages.
var packages = TestPackage.SubPackages;
if (packages.Count == 0)
    packages.Add(TestPackage);
`
This code makes a cyclic tree of packages when the SubPackages collection is empty

CharliePoole:
@NikolayPianikov Well, the package NUnit.ConsoleRunner.3.5.0-dev-03141
contains the output of Appveyor job 3141, which was run three days ago. It
was the last build of the nunit repository to contain engine and console
components. The next build, 3144, contains only framework components.

Unfortunately, now that console and engine runner are in a different
repository, AppVeyor started it's build numbers again. In addition, it
seems that the naming convention has reverted to an older pattern. You
should look at 3.5.0-CI-8-master. I'll see about fixing this so that the
[...]

PR #38, opened 2016-08-22
NUnit 3 console should produce xml events for ITestEventListener which contain unique id in the scope of all test agents for NUnit 2 tests

NikolayPianikov:
Use the current process id as a prefix for ID

### DIFF
--- a/src/NUnitEngine/Addins/nunit.v2.driver/NUnit2FrameworkDriver.cs
+++ b/src/NUnitEngine/Addins/nunit.v2.driver/NUnit2FrameworkDriver.cs
@@ -26,9 +26,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Xml;
 using NUnit.Core;
-using NUnit.Core.Filters;
 using NUnit.Engine.Extensibility;
-
 
 namespace NUnit.Engine.Drivers
 {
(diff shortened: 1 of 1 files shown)


CharliePoole:
I see two problems with using the process id as the first part of the test id:

1. Although unlikely, it's theoretically possible for the process id to clash with the id used for an NUnit3 test.

2. When multiple assemblies are run in the same process, they will all have the same id.

I think this should be fixed by using the same approach as the NUnit 3 driver. Each driver takes on the unique test package ID, stored in the package. This will be unique across all packages and subpackages created for a run. See the nunit3 driver code.

NikolayPianikov:
I've done something like that you were talking about. But I am not sure that code looks fine. I mean that now there is some implicit rule - in what order each method should be called. Also there is another rule - `TestPackage.ID` should be a string representation of an integer value because it is used as `runnerId` to call the method `RemoteTestRunner.CreateInstance(_testDomain, runnerId)`. RemoteTestRunner and ProxyTestRunner are located in NUnit.Core.dll. This assembly is used from the "lib" directory as a binary dependency and I think we should not change it.

src/NUnitEngine/Addins/nunit.v2.driver/NUnit2FrameworkDriver.cs
```diff
@@ -68,12 +66,28 @@ public NUnit2FrameworkDriver(AppDomain testDomain)
...
 
+        private TestRunner Runner {
+            get
+            {
+                if (_runner == null)
+                {
+                    int runnerId;
+                    if (ID == null || !int.TryParse(ID, out runnerId))
+                    {
+                        throw new InvalidSuiteException("ID must be defined at first as a string representation of an integer value.");
```
CharliePoole: This is an NUnit V2 Exception, completely unknown to the engine. Let's use NUnitEngineException, which gets special handling by some runners.

CharliePoole:
"Implicit rule - in what order each method should be called"

Do you mean that Load must be called first? That has always been the rule. Or is it something else?

"TestPackage.ID should  be a string representation of an integer value"

That's true. But the only way to eliminate the problem is to limit NUnit V3 forever to using integer values. NUnit V2 uses ints and we can't change that unless the driver intercepts every event and modifies the id using a lookup table. I don't think we want to do that! So I think we are stuck with this bit of awkwardness so long as we are supporting V2.

src/NUnitEngine/Addins/nunit.v2.driver/NUnit2FrameworkDriver.cs
CharliePoole: Also, can we add something to the message indicating that it's the V2 driver that requires an int?

CharliePoole:
@rprouse This looks good to merge, modulo my comments.

NikolayPianikov:
@CharliePoole Yes I mean Load must be called first

CharliePoole:
@NikolayPianikov The DirectRunner takes care that this is done. It probably should be documented somewhere as an assumption that a driver can make. Maybe in the interface comments?

CharliePoole:
@rprouse Ready to merge, I think.

NikolayPianikov:
@CharliePoole comments will never be superfluous

rprouse:
Looks good.

NikolayPianikov:
@CharliePoole @rprouse 

https://github.com/nunit/nunit-console/issues/30 does not work again

PR #54, opened 2016-09-08
Added unescape of unicode characters for console output

null511:
[nunit > issue #1787](https://github.com/nunit/nunit/issues/1787)

Quick and dirty attempt at unescaping unicode characters before printing them to the console.

### DIFF
--- a/src/NUnitConsole/nunit3-console/ConsoleOptions.cs
+++ b/src/NUnitConsole/nunit3-console/ConsoleOptions.cs
@@ -67,6 +67,8 @@ namespace NUnit.Common
 
         public bool LoadUserProfile { get; private set; }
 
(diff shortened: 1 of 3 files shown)


CharliePoole:
I'm going to hold this change while we check to see if there's a more fundamental fix that can be made within the framework. While the particular application of nunit/nunit#1787 is only pertinent to the console, there could easily be other uses of control characters that would affect runners in general. And, in fact, it's even conceivable that we would add colorization to the GUI in the future.

rprouse:
@null511 is this still needed with #63?

ChrisMaddock:
Along with https://github.com/nunit/nunit/pull/1789 - we've lost track of what's needed here.

Closing - if someone else has this issue and wants to pick it up again, would be happy to look at it.

Issue #58, opened 2016-09-19
System.Configuration.ConfigurationErrorsException thrown in multiple domain mode.


mrsneeze:
I have an issue that I was finding when using multiple domain mode and assemblies with user configuration.

If you have both multiple AppDomains specifided in nunit and a project that contains several modules that use the configuration system they can fail with the exception:
System.Configuration.ConfigurationErrorsException : Configuration system failed to initialize
----> System.Configuration.ConfigurationErrorsException : Unrecognized configuration section userSettings. (C:\Users\username\AppData\Local\NUnit_Software\test-domain-Components.Mi_Url_zeii3t5j3suizxytnaaivzomze5k3svw\3.4.1.0\user.config line 3)

The problem is that windows stores user configuration, it using only the first 25 characters of the AppDomain name in the directory. When nunit creates a test domain it's called test-domain-<assembly namespace>, so we're already using 12 characters for test-domain-. This means that the domains are using the same configuration area if they have the same first 13 characters of the assembly namespace - in our case, that's Components.Misc.Utils and Component.Misc.Client (eg) both give the test-domain-Components.Mi_Url_zeii3t5j3suizxytnaaivzomze5k3svw directory in the exception. The two assemblies will thus use the same file, which causes this error if one assembly has no user settings. There's also likely to be some potential unexpected conflicts in settings.

We can work around this using a single domain and a single separate process, but that isn't always possible.

I've made a simple fix that solves this problem in a fork - I've added the hash of the string into the domain name. This should usually be different for different domains - there's obviously some risk that similar domains will be created, but I think that is managably small. Better solutions I'm sure exist, such as a longer hash of the assembly name.

The patch I've used is: 

```
--- a/src/NUnitEngine/nunit.engine/Services/DomainManager.cs
+++ b/src/NUnitEngine/nunit.engine/Services/DomainManager.cs
@@ -57,7 +57,7 @@ namespace NUnit.Engine.Services
         {
             AppDomainSetup setup = CreateAppDomainSetup(package);

-            string domainName = "test-domain-" + package.Name;
+            string domainName = "domain-" + package.Name.GetHashCode().ToString("x") + "-" + package.Name;
             // Setup the Evidence
             Evidence evidence = new Evidence(AppDomain.CurrentDomain.Evidence);
             if (evidence.Count == 0)
```

CharliePoole:
[...]

PR #59, opened 2016-09-20
System.Configuration.ConfigurationErrorsException thrown in multiple domain mode. #58

mrsneeze:
Fix for issue #58

### DIFF
--- a/src/NUnitEngine/nunit.engine/Services/DomainManager.cs
+++ b/src/NUnitEngine/nunit.engine/Services/DomainManager.cs
@@ -48,8 +48,6 @@ namespace NUnit.Engine.Services
 
         private ISettings _settingsService;
 
-        private int _domainIndex = 0;
-
         #region Create and Unload Domains
         /// <summary>
         /// Construct an application domain for running a test package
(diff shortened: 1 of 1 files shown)


rprouse:
In what case was the package.Name null?

mrsneeze:
It happens if you run nunit on itself with no other parameters, or run a single assembly.

rprouse:
What do you mean by "It happens if you run nunit on itself with no other parameters?"

mrsneeze:
as in: nunit3-console nunit3-console.exe

It was the first assembly I could think of to test that it starts :)

rprouse:
@CharliePoole I think this is good to merge once we update from master. We should look at the package being null, but I think that is separate from this PR. Can you give it a quick second review?

@mrsneeze, can you merge latest from master or rebase? If not, we can do it for you.

rprouse:
@mrsneeze skip the update from master. I turned off the requirement to be up to date with master as I already did on the main framework repository. It sounds like a good idea, but every contributor usually has to update before their PR goes in and it ends up being a race condition with multiple PRs 👊 😦

rprouse:
@nunit/core-team this PR needs a second review.

rprouse:
Once again, @nunit/core-team this needs a second review. This PR has been outstanding for a long time, so I would like to get it merged. I will merge soon if no reviewer is available.

CharliePoole:
Looks OK to me

rprouse:
Thanks for the contribution @mrsneeze. Sorry it took so long to get merged.

PR #63, opened 2016-09-28
--encoding option for Console std-out

null511:
\* _This PR is a subset of the changes from PR #54._

Adds an option to nunit3-console for specifying the consoles standard-output encoding. This is required to display non-CP437 unicode characters. Since this is only a visual change, no tests were created.

### DIFF
--- a/src/NUnitConsole/nunit3-console/ConsoleOptions.cs
+++ b/src/NUnitConsole/nunit3-console/ConsoleOptions.cs
@@ -67,6 +67,8 @@ namespace NUnit.Common
 
         public bool LoadUserProfile { get; private set; }
 
+        public string ConsoleEncoding { get; private set; }
+
         private int maxAgents = -1;
         public int MaxAgents { get { return maxAgents; } }
         public bool MaxAgentsSpecified { get { return maxAgents >= 0; } }
@@ -147,6 +149,9 @@ namespace NUnit.Common
 
             this.Add("list-extensions", "List all extension points and the extensions for each.",
                 v => ListExtensions = v != null);
+
+            this.Add("encoding=", "Specifies the encoding to use for Console standard output.",
+                v => ConsoleEncoding = RequiredValue(v, "--encoding"));
 
 #if DEBUG
             this.Add("debug-agent", "Launch debugger in nunit-agent when it starts.",
--- a/src/NUnitConsole/nunit3-console/ConsoleOptions.cs
+++ b/src/NUnitConsole/nunit3-console/ConsoleOptions.cs
@@ -67,8 +67,6 @@ namespace NUnit.Common
 
         public bool LoadUserProfile { get; private set; }
 
-        public string ConsoleEncoding { get; private set; }
-
         private int maxAgents = -1;
         public int MaxAgents { get { return maxAgents; } }
         public bool MaxAgentsSpecified { get { return maxAgents >= 0; } }
@@ -149,9 +147,6 @@ namespace NUnit.Common
 
             this.Add("list-extensions", "List all extension points and the extensions for each.",
                 v => ListExtensions = v != null);
-
-            this.Add("encoding=", "Specifies the encoding to use for Console standard output.",
-                v => ConsoleEncoding = RequiredValue(v, "--encoding"));
 
 #if DEBUG
             this.Add("debug-agent", "Launch debugger in nunit-agent when it starts.",
--- a/src/NUnitConsole/nunit3-console/Program.cs
+++ b/src/NUnitConsole/nunit3-console/Program.cs
@@ -74,7 +74,7 @@ namespace NUnit.ConsoleRunner
                 catch (Exception error)
                 {
                     WriteHeader();
-                    OutWriter.WriteLine(ColorStyle.Error, string.Format("Unsupported Encoding! {0}", error.Message));
+                    OutWriter.WriteLine(ColorStyle.Error, string.Format("Unsupported Encoding, {0}", error.Message));
                     return ConsoleRunner.INVALID_ARG;
                 }
             }
--- a/src/NUnitConsole/nunit3-console/Program.cs
+++ b/src/NUnitConsole/nunit3-console/Program.cs
@@ -24,6 +24,7 @@
 using System;
 using System.IO;
 using System.Reflection;
+using System.Text;
 using NUnit.Common;
 using NUnit.Options;
 using NUnit.Engine;
@@ -62,6 +63,20 @@ namespace NUnit.ConsoleRunner
                 WriteHeader();
                 OutWriter.WriteLine(ColorStyle.Error, string.Format(ex.Message, ex.OptionName));
                 return ConsoleRunner.INVALID_ARG;
+            }
+
+            if (!string.IsNullOrEmpty(Options.ConsoleEncoding))
+            {
+                try
+                {
+                    Console.OutputEncoding = Encoding.GetEncoding(Options.ConsoleEncoding);
+                }
+                catch (Exception error)
+                {
+                    WriteHeader();
+                    OutWriter.WriteLine(ColorStyle.Error, string.Format("Unsupported Encoding! {0}", error.Message));
+                    return ConsoleRunner.INVALID_ARG;
+                }
             }
 
             //ColorConsole.Enabled = !Options.NoColor;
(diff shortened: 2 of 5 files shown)


src/NUnitConsole/nunit3-console/ConsoleOptions.cs
```diff
@@ -67,6 +67,8 @@ public class ConsoleOptions : CommandLineOptions
 
         public bool LoadUserProfile { get; private set; }
 
+        public string ConsoleEncoding { get; private set; }
```
rprouse: This probably belongs in CommandLineOptions, I expect that if we accept this PR, we will want to do the same for NUnitLite and the .NET Core runner.

src/NUnitConsole/nunit3-console/Program.cs
```diff
@@ -64,6 +65,20 @@ public static int Main(string[] args)
                 return ConsoleRunner.INVALID_ARG;
             }
 
+            if (!string.IsNullOrEmpty(Options.ConsoleEncoding))
+            {
+                try
+                {
+                    Console.OutputEncoding = Encoding.GetEncoding(Options.ConsoleEncoding);
```
rprouse: I checked the behaviour, this call ignores the case, so we don't need to do a ToLower or anything like that. 👍

rprouse: Is the root of the problem that the console.encoding is UTF-8, but you need an ASCII encoding (iso-885901) for your scenario?

src/NUnitConsole/nunit3-console/Program.cs
```diff
@@ -64,6 +65,20 @@ public static int Main(string[] args)
...
+                }
+                catch (Exception error)
+                {
+                    WriteHeader();
+                    OutWriter.WriteLine(ColorStyle.Error, string.Format("Unsupported Encoding! {0}", error.Message));
```
null511: Please, nitpick all you want. I put this together rather quickly, so I'm expecting formatting changes.

CharliePoole:
I also have a few code comments, but I'll hold off on those while we discuss why this is needed in the first place. The question could easily be answered by creating a test. Just add a test that displays something incorrectly on the console unless the correct encoding is specified. It doesn't need any asserts, sSince it does nothing that can be tested automatically. If it causes an exception when the correct option is not specified, just mark it as Explicit. Since it's just one test, you could put it with the option tests.

I do have a question about whether you are applying the encoding to the console at the right point, but I'd rather see the problem you are trying to solve before spending time with the details.

src/NUnitConsole/nunit3-console/Program.cs
null511: The default encoding for Windows CMD is [CP437](https://en.wikipedia.org/wiki/Code_page_437), which is very limited. Changing this to an alternate encoding (most often UTF-8) will allow a much broader range of characters that can be printed.

For my particular use case, I am printing a large amount of output from each test to our CI server, which catalogs all test runs & output (automated Selenium testing). The special characters are used to improve the readability of this output for review by non-developers.

src/NUnitConsole/nunit3-console/ConsoleOptions.cs
null511: Ironically, that is where I had originally placed it, but I thought ConsoleOptions was more closely related. Will move.

src/NUnitConsole/nunit3-console/Program.cs
rprouse: If that is the case, then should we just be setting the encoding to UTF-8?

src/NUnitConsole/nunit3-console/ConsoleOptions.cs
```diff
@@ -148,6 +150,9 @@ protected override void ConfigureOptions()
             this.Add("list-extensions", "List all extension points and the extensions for each.",
                 v => ListExtensions = v != null);
 
+            this.Add("encoding=", "Specifies the encoding to use for Console standard output.",
```
null511: I imagine "utf-8", "ascii", & "unicode" will be the most popular encodings. Is there a different name you would prefer for this option, ie `--stdout-encoding`?

rprouse: I think encoding is fine.

null511:
I put together a small test in `ConsoleOutputTests.cs` that simply prints a few unicode characters. Since the console encoding has to be set before the output stream is changed, I was unable to alter the encoding from within the test (since NUnit has already redirected std-out).

Expected Text: `•ÑÜńĭŧ·`

Default Encoding


UTF-8


It is also worth noting that once an encoding has been specified on a console window, it does not reset when the running executable completes. So if you perform a test run that alters the encoding, and then perform an additional test run _without_ specifying an encoding, the first encoding applied will be used, until altered or the console has terminated.

rprouse:
The fact that the encoding does not switch back is worrying. If that is the case, then we should capture the encoding at the start and reset it at the end.

Also, is the ability to set the encoding on the command line necessary, or should we always switch it to UTF-8 since that is the format of the XML we are sending back and forth?

null511:
> The fact that the encoding does not switch back is worrying. If that is the case, then we should capture the encoding at the start and reset it at the end.

Sounds like a good idea. This behavior had me very confused when I first ran through the above scenario and continued seeing unicode characters without specifying encoding (same console session).

> Also, is the ability to set the encoding on the command line necessary, or should we always switch it to UTF-8 since that is the format of the XML we are sending back and forth?

Yes, it must be specified on `System.Console` **before** the output stream has been redirected. All C# strings are stored in-memory as UTF-16/Unicode characters. The encoding only comes into play when these characters are written to disk (or displayed in the console), since a conversion from a `char` to a `byte[]` will occur. Specifying the console encoding allows the program reading the file (or console) to properly decode the byte streams back into the proper characters.

null511:
Batch script from the above photos just in case it's useful.

```
@cd "C:\Dev\Local\nunit-console\bin\Debug"
@ECHO
@ECHO DEFAULT ENCODING
nunit3-console.exe nunit3-console.tests.dll --where="name == 'ConsoleEncoding'"
@ECHO
@ECHO UTF-8
nunit3-console.exe nunit3-console.tests.dll --where="name == 'ConsoleEncoding'" --encoding="utf-8"
pause
```

src/NUnitConsole/nunit3-console/Program.cs
null511: Setting UTF-8 (or any encoding) by default could also create issues. Windows CMD uses bitmap fonts by default ("Raster Fonts"), which do **not** support unicode characters; A TrueType font must be set.

So if UTF-8 is the default, and a user runs NUnit from Windows CMD console with the default settings, they will see random characters in place of their text. Which I believe qualifies this as an opt-in condition.

null511: Here is an example showing how unicode characters look in a Windows CMD console with the default font. The first example shows how unicode characters would be displayed with NUnit currently, and the second shows how they are displayed with UTF-8.

Unicode Text: `•ÑÜńĭŧ·`
CMD Font: "Raster Fonts" (bitmap)

NUnit > Default Encoding (CP437)


NUnit > UTF-8

null511:
Just curious what the status of this request is, or if there are any additional changes requested? I am planning on using this feature in a custom build within my organization, but cannot unless it is scheduled for a future release. Not trying to rush this, just want to make sure I've done all I can on my end.

CharliePoole:
It's waiting for me - or another committer - to find time to review it. Our practice is that it takes two committers to approve a merge. Rob will also need to come back in and OK the changes you made due to his review. 

You already answered my general question about why it's a good idea, and I agree. I think you've done what you needed to do but are just hitting us at a busy time as we prepare the 3.5 release (and as I get my house remodelled). :-) I'll try to do an actual code review today or tonight.

rprouse:
@null511 is your framework pull request, nunit/nunit#1789 required for this or can it be closed?

rprouse:
@null511 I ask about your Framework pull request because I am planning on doing the Framework release tomorrow. I can wait for it though.

CharliePoole:
@rprouse PR nunit/nunit#1789 is still under discussion, although the discussion is actually on nunit/nunit-console#1787, the related console PR. I wouldn't want either of those merged without our acquiring a better understanding of what's happening currently. 

I assumed this one was separate, and the tests seem to demonstrate that. I'm OK to merge it.

null511:
> @null511 is your framework pull request, nunit/nunit#1789 required for this or can it be closed?

It is not required; This PR is actually a subset of those changes.

PR [nunit/nunit#1789](https://github.com/nunit/nunit/pull/1789) is based on enabling control characters to be unescaped before being printed to the console, which requires a UTF-8 encoded console standard-output stream. This enables color information (and much more) to be output by tests and decoded by the console or any report generating tool.

I believe that will require another opt-in option though, because it will require escaping all backslashes in xml output to enable proper decoding. I figured this subset of console changes was useful for a variety of reasons, and served better as a first step towards the overall goal. I will revise PR 1789 if this is merged.

null511:
Just wanted to say thank you for your time and assistance with this PR. Looking forward to the next release.

Issue #74, opened 2016-10-05
Update .nuspec files to use newest version of TeamCityEventListener


NikolayPianikov:
Dependency for TeamCityEventListener:
`<dependency id="NUnit.Extension.TeamCityEventListener" version="[1.0.*, 1.1)" />`

.nuspec files:
`nuget\runners\nunit.console-runner-with-extensions.nuspec`
`nuget\runners\nunit.runners.nuspec`

PR #75, opened 2016-10-05
#74 Update .nuspec files to use newest version of TeamCityEventListener

NikolayPianikov:


NikolayPianikov:
@rprouse could you please verify and merge this PR to 3.5

ChrisMaddock:
@NikolayPianikov - I believe adding a range like this will default to the lowest possible patch version? Was that intentional?

Docs at http://docs.nuget.org/ndocs/consume-packages/dependency-resolution - I may be mis-reading.

NikolayPianikov:
@ChrisMaddock oh, thank you, I see that rules were changed
I will change it to [1.0.1.*, 1.1)

nuget/runners/nunit.console-runner-with-extensions.nuspec
```diff
@@ -34,7 +34,7 @@
         <dependency id="NUnit.Extension.VSProjectLoader" version="$version$" />
         <dependency id="NUnit.Extension.NUnitV2ResultWriter" version="$version$" />
         <dependency id="NUnit.Extension.NUnitV2Driver" version="$version$" />
-        <dependency id="NUnit.Extension.TeamCityEventListener" version="1.0.1" />
+        <dependency id="NUnit.Extension.TeamCityEventListener" version="[1.0.*,1.1)" />
```
CharliePoole: This will cause users installing for the first time to get the lowest version that matches, unless they use options on the command line. See general comment.

NikolayPianikov: As I understand - we are planning to have ability to do independent releases. If I defined `1.0.2` all users would install 1.0.1 according to rule `Lowest applicable version` When I set `[1.0.*,1.1)` all users are able to get hot-fixes in this range, according to rule `Floating Versions`, and we should not make a new package `nunit.runners` each time when we would like to publish hot-fix. Although maybe that something is misunderstood in http://docs.nuget.org/ndocs/consume-packages/dependency-resolution

CharliePoole:
@NikolayPianikov I think we are up against a policy question rather than a technical one here.

NUnit releases the NUnit.Console meta-package, with the intention of installing particular versions of other packages. Some of those other packages are developed by other people - only one right now, of course, but it's potentially a bigger issue.

Before we release a meta-package, we should be testing it with the specific dependency packages that it contains.

If JetBrains - or any other outside group - then releases a new version of an extension, do we want users of our meta-package to get that new version, which we have not tested?

My original thought was that we did not. That we would periodically (e.g. every three months) release new meta-packages using the latest tested version of each meta-package.

However, I understand what you are suggesting and why. It's just a question we have to decide. @nunit/core-team What do you think?

ChrisMaddock:
> do we want users of our meta-package to get that new version, which we have not tested?

Wasn't the intention that the extension author would always test with the latest engine release?

I think if we assume that patch versions are only bug fixes, using the `*` seems ok. The other option would be that NUnit.Console needs re-releasing for every bug fix in every extension. If we expect extension release patterns to diverge, not needing to do that seems preferable.

CharliePoole:
That's our intention, but we are not guaranteed that it will work.

I think there's a third option: wait till we do our next release. That was my original intention and it seems especially reasonable for extensions we don't produce ourselves. In that case, users wanting the fix sooner could simply install the updated package directly.

I realize it may be difficult to imagine the situation where we are dealing with an extension over which we have no control. But the TC extension is moving in that direction and there could be others. Maybe one possible answer is that we don't bundle such extensions at all.

ChrisMaddock:
> In that case, users wanting the fix sooner could simply install the updated package directly.

Sure, but with the `*` I believe users will be notified by NuGet that a new extension package is available? Without this, users get no notification until we release the next version of NUnit.Console.

CharliePoole:
I could be wrong, but I thought that the new package would simply be installed for users of NuGet 3+. For earlier releases, I think it will say the version number is bad, but I haven't tried it. 

I'm somewhat nervous of using a feature that only works if the user has the right version of nuget installed.

Here's one basic philosophical difference you have uncovered as well: I'm assuming that users of a third-party extension are getting there information about updates from that third party and that it's not our job to provide it to them. Perhaps that's too harsh?

NikolayPianikov:
@nunit/core-team so what's the verdict? Should we just close this PR?

CharliePoole:
I've been hoping for more comments on this because I think it's a pretty significant release issue. However, to get the release out, I'm currently intending to stick with our current practice of using a single value rather than a range.

I'm working on the console packaging now and can change the versions we use up until the time of release. @rprouse Have you got any thoughts on this?

rprouse:
My thoughts are that extensions will continue to be backwards compatible and that releasing with >= versioning, that users can get fixes without us releasing a new console package. I think it should be an extension author's choice. If they make the choice to use >= versioning, then it is their responsibility to test before release.

CharliePoole:
@rprouse OK, that makes sense, but see below.

I've been trying to use this question vis-a-vis teamcity to set a policy
for any future bundled extensions. But now that I think about it, If we
bundle - even if only by use of a meta-package - a third-party extension,
we are already giving an great deal of trust to that third party. This
question of accepting bug fixes is really rather minor by comparison and
the focus should be on whose extensions we bundle. I don't think there will
be much more of this anyway, but we can deal with it if it comes up.

I'd like to urge we don't use the term you used: >= versioning. It's
confusing, because __all__ nuget versioning is >= unless you restrict it
via a range. Our current use of `version=1.0.2` for teamcity already allows
the dependency to be satisfied by 1.0.3 (or even 1.1) if that version is
pre-installed. The question here is what you get if nothing is installed.
The use of a floating version causes the latest patch level to be taken if
nothing is already installed.

Let's go with the form of version that Nikolay has proposed and use similar
forms for the other extensions.

CharliePoole:
@NikolayPianikov 
This is failing all builds. It may be due to your adding a bunch of new binary files for Cake and nuget. 

AppVeyor says that your version string is illegal, but it may be a side-effect of the binaries.

I'll duplicate your change in my own PR and see if it passes. If the version string is still treated as illegal, I'll back it out in order to get 3.5 out and we can deal with it subsequently.

CharliePoole:
The change was made through PR #87

NikolayPianikov:
@CharliePoole thank you

PR #78, opened 2016-10-06
Issue24: Update --labels switch

### LINKED ISSUE #24: Update --labels switch with option to show real-time pass/fail results in console runner
@dybs commented on [Fri Aug 12 2016](https://github.com/nunit/nunit/issues/1738)

When running a large number of test cases, it would be nice to see the pass/fail status of each case in the console as it completes.  Using --labels=All, I can see which tests have run, but the results are not available until all cases have completed.  Seeing which cases fail during a long scenario would allow me to investigate the failures while the remaining tests continue to run.  It also has the advantage of knowing which cases passed/failed if the test scenario happens to crash and doesn't get to write the result file.

Per the discussion in #1735 some ideas are to modify --labels with the following options:
- Results -> same behavior as =All, but includes a Pass/Fail/Error status
- Before -> shows the test case before it executes, and appends the status after the test completes.  Similar to On.
- After -> shows the test case and status together after the test case completes.

---

@CharliePoole commented on [Fri Aug 12 2016](https://github.com/nunit/nunit/issues/1738#issuecomment-239549055)

I like this idea, but I would modify/simplify it by preference:
- Off No change
- On No change
- Before / All shows test label when the test starts executing.
- After Shows all tests after they have completed, including status.

Note that All is described as it is supposed to work already. There's currently a bug in that it runs at the end of a test. We should fix that. We should keep All so as not to break existing usage.

The introduction of After means that we have to treat it as also meaning On. That is, if a test produces output, there should be a label before that output and another one after it, together with the test status..

The change should be made in both the console runner and nunitlite. Once we split the repos, this issue will need to be duplicated.

The fix is _relatively_ easy. Depending on who takes it, I can mentor them as needed.

---

@constructor-igor commented on [Sun Aug 14 2016](https://github.com/nunit/nunit/issues/1738#issuecomment-239676310)

Hi @CharliePoole,

Probably, I can start to work on the issue.
I checked all existing options of "--labels" parameter. Please, could you help me to understand suggested changes? I added sample with "--lablels=All". As I understand, the issue suggests to add "status" of each test?

image

thank you,
Igor.

---

@CharliePoole commented on [Sun Aug 14 2016](https://github.com/nunit/nunit/issues/1738#issuecomment-239684890)

Hi @constructor-igor - I've been meaning to poke you to see what you are up to. :-)

As you probably figured out, all the code is in Console's TestEventHandler and in NUnitLite's TestRunner. There are tests for TestEventHandler but not for TestRunner. :-(

I would first off fix the problem with All, which is supposed to be issued at the start of each test, not at the end. I think I broke this when I was fixing something else.

Once All is correct, you've got Before, since it's intended to work the same way.

After is the new thing, which is supposed to show status of the test. For that, I would use a combination of the `result` and `label` attributes to display a set of values like...
- Passed
- Failed
- Error
- Invalid
- Ignored
- Explicit
- Skipped
- Inconclusive
  In NUnitLite, you'll have the ResultState available to use for equivalent logic.

---

@constructor-igor commented on [Mon Aug 15 2016](https://github.com/nunit/nunit/issues/1738#issuecomment-239782084)

I am going to start from "All".
But, right now, I don't understand expected behavior of "All".
So, when I'll investigate code I'll contact with you.

---

@CharliePoole commented on [Mon Aug 15 2016](https://github.com/nunit/nunit/issues/1738#issuecomment-239783383)

You might try this sequence - a bit different order from my first suggestion:
- Change "All" in existing code everywhere to "After" because All is now incorrectly running after the test.
- Add status info to After
- Add "Before" to run before the test
- Add back "All" as a synyonym for Before

---

@constructor-igor commented on [Mon Aug 15 2016](https://github.com/nunit/nunit/issues/1738#issuecomment-239785294)

I try to build nunit by "build.cmd":
image

VS2013:
image

---

@CharliePoole commented on [Mon Aug 15 2016](https://github.com/nunit/nunit/issues/1738#issuecomment-239790081)

Are you using the latest master? Can you build in the IDE?

---

@constructor-igor commented on [Mon Aug 15 2016](https://github.com/nunit/nunit/issues/1738#issuecomment-239796767)

I cloned today and tried to build via build.cmd and IDE VS2013: same errors.

---

@CharliePoole commented on [Mon Aug 15 2016](https://github.com/nunit/nunit/issues/1738#issuecomment-239798656)

Using VS 2015?

---

@rprouse commented on [Mon Aug 15 2016](https://github.com/nunit/nunit/issues/1738#issuecomment-239803785)

I just got latest from master and ran `.\build.cmd` and it builds fine for me. Maybe do a full clean including deleting your packages folder and rebuild?

---

@constructor-igor commented on [Mon Aug 15 2016](https://github.com/nunit/nunit/issues/1738#issuecomment-239894381)

I could build nunit solution on other computer.
I tried to build on my computer again (downloaded master to new folder) and  see the error.

I investigate the issue and found, probably, same issues 
- https://github.com/nunit/nunit3-vs-adapter/issues/194
- https://github.com/nunit/docs/issues/44

I continue my investigation.

constructor-igor:
Do Not Merge
Fixes #24


src/NUnitConsole/nunit3-console.tests/CommandLineTests.cs
```diff
@@ -101,7 +101,7 @@ public void CanRecognizeBooleanOptions(string propertyName, string pattern)
         [TestCase("OutFile",            "output|out", new string[] { "output.txt" },                     new string[0])]
         [TestCase("ErrFile",            "err",        new string[] { "error.txt" },                      new string[0])]
         [TestCase("WorkDirectory",      "work",       new string[] { "results" },                        new string[0])]
-        [TestCase("DisplayTestLabels",  "labels",     new string[] { "Off", "On", "All" },               new string[] { "JUNK" })]
+        [TestCase("DisplayTestLabels",  "labels",     new string[] { "Off", "On", "Before", "After" },   new string[] { "JUNK" })]
```
CharliePoole: Need to put back All.

src/NUnitConsole/nunit3-console.tests/TestEventHandlerTests.cs
```diff
@@ -60,32 +60,45 @@ public void EventsWriteExpectedOutput(string report, string labels, string expec
             // Start Events
             new TestCaseData("<start-test/>", "Off", ""),
             new TestCaseData("<start-test/>", "On", ""),
-            new TestCaseData("<start-test/>", "All", ""),
+            new TestCaseData("<start-test/>", "Before", ""),
```
CharliePoole: Keep All as well - synonym for Before or Before + After

src/NUnitConsole/nunit3-console/TestEventHandler.cs
```diff
@@ -127,6 +128,21 @@ private void WriteLabelLine(string label)
...
             }
         }
+        private void WriteLabelLine(string label, string resultStatus)
+        {
+            if (label != _currentLabel)
+            {
+                ColorStyle colorByResultStatus = resultStatus == "Passed" ? ColorStyle.Pass : resultStatus == "Failed"
+                            ? ColorStyle.Failure : resultStatus == "Warning"
+                    ? ColorStyle.Warning
+                    : ColorStyle.Output;
```
CharliePoole: Hard to read  this without proper nesting

CharliePoole:
Aside from the line comments, the current status by label setting seems to be...

* Off - Working
* All - Not yet working
* Before - Not yet working
* After - Working pending definition of further statuses
* On - Working

@constructor-igor Do you agree with this summary? What shall we make work next?

src/NUnitConsole/nunit3-console/TestEventHandler.cs
constructor-igor: I am going to move the code to separated private method (with switch/case), but probably it should be moved to general "utility/helper"?

constructor-igor:
please, could you explain (or add test) why "After" doesn't work. 
Also, to implement "Before", please, could you send sample of test? "Before" should write info before test's execution (all other write after).
As I understand "All" should write "Before" + "After" data. Correct?

CharliePoole:
@constructor-igor 
1. After seems to be working, although as I stated we probably want to refine what statuses we use and that's complicated enough that you might want to break it out into a separate method. I think we have code for this somewhere - I'll check and post it.

2. I'll add a comment with a sample Before test.

3.  We previously defined All as equivalent to Before. I realize that's not logical since one would expect All to mean something like "both before and after" but it's a matter of backward compatibility. (Remember, it's a bug that All was running after the test and not before.)

CharliePoole:
Here's how we display status in the error listing:

```
            string status = result.GetAttribute("label");
            if (status == null)
                status = result.GetAttribute("result");
```

This code makes the assumption that all labels are unique - i.e. we never have the same label value for two different result values. That's true at this time and will probably stay true. At any rate, if it should change we can worry about it then.

If you use this logic, you'll have to change your color calculation to be something like...

```
result == "Failed" 
    ? ColorStyle.Error
    :  label == "Ignored"
        ? ColorStyle.Warning
        : ColorStyle.Output;
```

Add in result==Passed if we decide to color that one.

constructor-igor:
Implemented minor refactoring and added ```--labels=Before```

CharliePoole:
This is actually looking pretty close to ready. Here's what I see as remaining issues:

1. The color logic is no longer correct for non-runnable tests or errors. Best logic is to look at result rather than status field. If the result is Skipped, then look at status or label and use warning color if it is equal to Ignored.

2. We need tests of After for all valid combos of result and label, currently
 * Failed
 * Failed+Invalid
 * Failed+Error
 * Failed+Cancelled
 * Passed
 * Skipped
 * Skipped+Ignored
 * Inconclusive

Beyond that... it's just a matter of making sure that everyone is OK with what we have done at a feature level. @nunit/core-team @nunit/contributors Are you OK with our use of After in particular?
It's slightly unintuitive since the label is displayed before the output, not after it. But it's created after the test has finished. Let's do a functional review after this is all working.

If we could accomplish all that in the next 24 hours, this could be in the release.

constructor-igor:
Please, could you help me to understand complex state, like "**Failed+Invalid**"
I see tag "result" with "Failed", but I don't see any other information.

```xml
<test-case id="0-1082" name="Test(3)" fullname="NUnit_v3_samples.NUnitIssue1738.Test(3)" methodname="Test" classname="NUnit_v3_samples.NUnitIssue1738" runstate="Runnable" seed="323558644" result="Failed" start-time="2016-10-09 20:30:51Z" end-time="2016-10-09 20:30:53Z" duration="2.033294" asserts="0">
            <failure>
              <message><![CDATA[test #03]]></message>
              <stack-trace><![CDATA[at NUnit_v3_samples.NUnitIssue1738.Test(Int32 index) in d:\My\@gh\TechSugar\TestingLab\NUnitSamples\NUnit_v3_samples\NUnitIssue1738.cs:line 30
]]></stack-trace>
            </failure>
          </test-case>
```

CharliePoole:
@constructor-igor Sorry, I thought my terminology was "obvious" which it _obviously_ was not. :-)

**Failed+Invalid** means `result="Failed" label="Invalid"` in your test XML element.

"Failed" by itself means `result="Failed"` without a label attribute present.

Note that there can only be four values for the result attribute, because it comes from an internal enumeration. Label is available for extension by users, so it's open-ended in principle.

You couldn't be expected to know any of this at the start, of course. :-)

CharliePoole:
Master has now been updated with the fix for `--labels=All`. The labels are now displayed at the start of the test.

constructor-igor:
Is it correct?

```c#
            new TestCaseData(
                "<test-case fullname='SomeName' result='Failed' label='Invalid'><output>OUTPUT</output></test-case>", 
                "After", 
                "INVALID => SomeName\r\nOUTPUT\r\n"),
            new TestCaseData(
                "<test-case fullname='SomeName' result='Failed' label='Error'><output>OUTPUT</output></test-case>", 
                "After", 
                "ERROR => SomeName\r\nOUTPUT\r\n"),
            new TestCaseData(
                "<test-case fullname='SomeName' result='Failed' label='Cancelled'><output>OUTPUT</output></test-case>", 
                "After", 
                "CANCELLED => SomeName\r\nOUTPUT\r\n"),

            new TestCaseData(
                "<test-case fullname='SomeName' result='Skipped' label='Ignored'><output>OUTPUT</output></test-case>", 
                "After", 
                "IGNORED => SomeName\r\nOUTPUT\r\n"),

```

constructor-igor:
All implemented tests are passed. 
Right now I don't see opened issues, exclude color. Please, could you review and help me to implement correct color selection?

CharliePoole:
@nunit/core-team @nunit/contributors 
This seems to be technically complete and ready to merge if we agree with what it is doing. We need one more code review but also a general review of the user-facing part of this:

1. Do we agree with what Before, After and All do. Are the names OK?
2. Should After labels be colored red, yellow and green? In particular, is green for passing (which is normally almost everything) OK?
3. Do we need the ability to combine any of these styles (e.g. Before + After)
4. Are we happy with how After works: any immediate output displays before the label while standard output appears after it.

Sebazzz:
Regarding 3: I would like to combine labels, especially when running longer integration tests.

General question: Does this feature work or conflict with the TeamCity output of NUnit? Since the TeamCity add-in implements an event listener, I guess it doesn't matter?

CharliePoole:
@sebazzz What specific combos do you think are needed?

I don't know how labels interact with teamcity but adding new forms shouldn't change what it already does. Both tc and the console runner use event listeners operating in parallel and each doing their own thing.

Sebazzz:
@CharliePoole That would be the combo Before/After.

CharliePoole:
We could call it Both or use + in the option string.

CharliePoole:
Let's imagine a test that produces two immediate output lines and two standard output lines. Currently, they will produce output as follows:

**--labels:Off**
```
Immediate Line 1
Immediate Line 2
Standard Output Line 1
Standard Output Line 2
```

**--labels:On** (default)
```
=> Name.Of.The.Test
Immediate Line 1
Immediate Line 2
Standard Output Line 1
Standard Output Line 2
```

**--labels:Before / --labels=All**
```
=> Name.Of.The.Test
Immediate Line 1
Immediate Line 2
Standard Output Line 1
Standard Output Line 2
```

**--labels:After**
```
Immediate Line 1
Immediate Line 2
PASSED => Name.Of.The.Test
Standard Output Line 1
Standard Output Line 2
```

The `Off`` output is as expected. `On` and `Before` look the same here, but `Before` displays the label even if there is no output from the test, including if the test hangs. `On` has the added benefit that the label will be displayed again if output from parallel tests comes between the lines of this test.

`After` frankly looks Odd to me.

I think I would lean toward modifying `After` to display output first, before the label. I would also make `After` imply `On` as well. Then the output would look like this:

**--labels:After**
```
=> Name.Of.The.Test
Immediate Line 1
Immediate Line 2
Standard Output Line 1
Standard Output Line 2
PASSED => Name.Of.The.Test
```

This is __almost__ like Before + After, except that the initial label won't display if there is no output. That's what you want most of the time, I think, but in complex situations, you may need it to display as even with no output. We would have to define something like `Both` to cover that case.

As to naming: we could consider renaming After as Result.

At this point, I think the main question is whether we want to get this into the release, which I want to do today if possible. If we postpone, we can discuss this at leisure. If we merge it, I would not want to make changes later, only additions. In particular, the relative position of the output and the After label ought to be decided before merging.

@constructor-igor It's night where you are and morning for me. I can do a few quick modifications if necessary, and if we decide it should go in right away.

@rprouse What are your thoughts?

constructor-igor:
@CharliePoole what next steps?

CharliePoole:
I was waiting for feedback on the suggestions I outlined. What do you think? Other folks?

constructor-igor:
@CharliePoole probably, publish your decision and we'll continue implementation of the feature?

CharliePoole:
@constructor-igor Well, I was hoping for input but you and I can just decide for now. :-)

Here's what I think:

1. For "After" make the output appear first. So your test case would be (one example):

```C#
            new TestCaseData(
                "<test-case fullname='SomeName' result='Failed' label='Invalid'><output>OUTPUT</output></test-case>", 
                "After", 
                "OUTPUT\r\nINVALID => SomeName\r\n"),
```

2. Make Use of BEFORE, ALL or AFTER supersets of On. By that I mean the labels will re-appear as needed each time tere is output from a different test. We will need tests of this, maybe based on the examples I gave in my earlier comment. One way to do it would be to replace the test for "On" with a boolean and set the boolean to true for all values of labels except OFF. I can write a test to show what I mean if it isn't clear.

constructor-igor:
> Make Use of BEFORE, ALL or AFTER supersets of On. By that I mean the labels will re-appear as needed each time tere is output from a different test. We will need tests of this, maybe based on the examples I gave in my earlier comment. One way to do it would be to replace the test for "On" with a boolean and set the boolean to true for all values of labels except OFF. I can write a test to show what I mean if it isn't clear.

@CharliePoole, please, could you add test for the request?

ChrisMaddock:
Appreciate this is a WIP - is there a reason for turning off warnings as errors, or is that just temporary?

CharliePoole:
One more day of guests here and then I'll be able to do that. :-)

constructor-igor:
> Make Use of BEFORE, ALL or AFTER supersets of On. By that I mean the labels will re-appear as needed each time tere is output from a different test. We will need tests of this, maybe based on the examples I gave in my earlier comment. One way to do it would be to replace the test for "On" with a boolean and set the boolean to true for all values of labels except OFF. I can write a test to show what I mean if it isn't clear.

@CharliePoole, I want to continue, but I need help.
Please, could you add test for the request?

rprouse:
@constructor-igor I am removing the hacktoberfest label. PRs and Issues don't need to be labeled to get the T-Shirt 😛

CharliePoole:
@constructor-igor Sorry for keeping you waiting. I'll try to get you a test case today.

CharliePoole:
@constructor-igor I created a new branch under the nunit org and a new PR #124. I think I can give you feedback more easily and quickly if we work in that branch.

Instead of working on this in your own fork, you should clone nunit/nunit-console locally and work in the issue-24 branch there. This PR can be closed as the new one has all your code.

I added tests for handling multiple events, which should explain what we need to do to move this along. I'll add some additional tests in a few days for multiple events from multiple, interleaved tests.

Send me a note if you have any questions.

PR #85, opened 2016-10-08
Ensure agent is stopped even if AppDomain unload times out

### LINKED ISSUE #43: Agent's process stays in memory when it was failed to unload appdomain
@NikolayPianikov commented on [Mon Jun 27 2016](https://github.com/nunit/nunit/issues/1628)

---

@NikolayPianikov commented on [Mon Jun 27 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-228848913)

From my point of view it could be critical issue

---

@CharliePoole commented on [Mon Jun 27 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-228858014)

Is this new with 3.4?

---

@NikolayPianikov commented on [Mon Jun 27 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-228861624)

Yes, looks like a regression from 3.2.1

---

@CharliePoole commented on [Mon Jun 27 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-228863909)

This is a bit strange. The code at https://github.com/nunit/nunit/blob/master/src/NUnitEngine/nunit.engine/Services/DomainManager.cs#L153 throws an exception if the domain unload throws or if it times out. I have run under console and verified that the exception comes back to the console and is displayed. Maybe the Join is timing out and the Thread.Abort is hanging.

---

@NikolayPianikov commented on [Tue Jun 28 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-229007788)

Anyway we should terminate all related agents when the console's process was finished

---

@CharliePoole commented on [Tue Jun 28 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-229019041)

Certainly... but we have to  figure out how to do that. :-) The code I pointed to is part of the termination and it did change in 3.4. 

Ideas:

 * Look at what changed in 3.4. Of course the change was intended to solve a problem, so we will have to make sure that still gets solved.

 * Use code from https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Internal/ThreadUtility.cs#L55 which may do a better job of killing the thread.

 * Agent termination has always been cooperative - that is, we tell it to terminate and it does. We could add some code to TestAgency to actually kill the process if it doesn't  terminate within a certain time.

What do you think?

---

@NikolayPianikov commented on [Tue Jun 28 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-229023984)

Ok. I will try to find. 

---

@rprouse commented on [Tue Jun 28 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-229026260)

I need to look at the code, but a timeout in TestAgency and then killing the process seems like the right place. 

Better would be identifying why it is happening and fix that ;) 

---

@CharliePoole commented on [Tue Jun 28 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-229137803)

I reviewed the 3.4 changes and don't see anything that should cause this. Moving on to the other bullet points.

---

@CharliePoole commented on [Thu Jun 30 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-229655480)

@NikolayPianikov We really need to get the release out, so we are going with the changes we have so far. We can follow up further on this problem if it continues.

---

@NikolayPianikov commented on [Thu Jun 30 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-229666698)

Issue was reproduced on "master", see http://win10nik.cloudapp.net/viewLog.html?buildId=94&buildTypeId=NUnit_NUnit3IntegrationTests&tab=buildResultsDiv

---

@NikolayPianikov commented on [Thu Jun 30 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-229673005)

User creates new AppDomain and does not finish thread there. We can reproduce it using [mocks.zip](https://github.com/nunit/nunit/files/341635/mocks.zip)

```
internal class UnloadingDomainUtil
{
    public static void Create()
    {
        var newDomain = System.AppDomain.CreateDomain(System.Guid.NewGuid().ToString(), System.AppDomain.CurrentDomain.Evidence);
        newDomain.CreateInstanceFrom(typeof(UnloadingDomainUtil).Assembly.Location, typeof(UnloadingDomainUtil).FullName);
    }

    public UnloadingDomainUtil()
    {
        new System.Threading.Thread(() => { while(true); }).Start();
    }   
}


[Test]
        public void MyTest()
        {
            UnloadingDomainUtil.Create();
        }
```

---

@CharliePoole commented on [Thu Jun 30 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-229677504)

Can you run it using the release/3.4.1 branch?

---

@NikolayPianikov commented on [Thu Jun 30 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-229678411)

One moment

---

@CharliePoole commented on [Thu Jun 30 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-229678698)

And you say we were successfully terminating such a case using 3.2.1?

Since this is a user-specific situation, caused by a bad test it could be postponed. The most critical issue is that 3.4 won't run the teamcity extension at all! I can only give this a very short timebox before I do the 3.4.1 release without it. 

---

@NikolayPianikov commented on [Thu Jun 30 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-229702831)

It is reproduced on 3.2.1 too. May be it is a bit another case (we have an exception "Error while unloading appdomain"), but it could have a similar solution.

---

@CharliePoole commented on [Thu Jun 30 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-229741204)

When I add your test to a console run, the only effect is a brief pause before it terminates. This works if I run in process or in a separate process.

I'm going to move ahead with 3.4.1 as it is. After it's out, we can try to address this if it's still a problem.

---

@tfabris commented on [Tue Aug 23 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-241909852)

Regarding Nunit issue #1628, "Agent's process stays in memory when it was failed to unload appdomain":

This issue is a blocking issue for us at the moment, due to the fact that we recently converted our company's tests over from Nunit 2.6.4 to Nunit 3.4.1. There are some details about the issue which were not revealed on the original bug report and discussion. We have some new information about this bug. We would like to request that the bug be re-opened because it is a blocking issue.

New details:

- Bug definitely did not occur in Nunit 2.6.4.

- Bug occurs in 3.4.1 but we do not know in which build it was introduced.

- Bug occurs only on certain test DLLs which induce the situation which cause the message to appear: "Unable to unload AppDomain, Unload thread timed out."

- Important: Bug only occurs when the output of the console is redirected. For example, in test automation situations where console output is captured to memory, such as when Team City is running the Nunit Console runner to execute tests. This bug causes Team City to hang and not complete the build. For ease of repro, you can also reproduce in a powershell script which redirects console output.

- Important: Bug does not occur when you run the Nunit Console at a plain console (output is not redirected).

- Example PowerShell command which does NOT cause a hang (bug still occurs, but the output appears normally and drops back to powershell prompt):

    & nunit3-console.exe MyBadTest.dll --x86 --framework=net-4.0
                
- Exmple PowerShell command which DOES INDUCE the hang because it is redirecting the output (this is the same thing that happens to Team City):

    $result = & nunit3-console.exe MyBadTest.dll --x86 --framework=net-4.0
    write-host "$result"

- If you watch the Windows Task Manager in the two cases above, you see that there is a launch of Nunit3-Console.exe and one or more Nunit-agent-x86.exe instances. Only the Nunit3-Console.exe closes at the end of the test, the agent instance(s) remain open and do not close. When there is no console redirection, control returns to the console, but when there is redirection, control does not return to the console until the agent process is forcefully terminated.

- This is a blocking issue for any instances where Nunit Console is being launched by any type of automation, because it causes the automation to hang, and fail to process any more commands. The test runner should close down when it is done, regardless of whether the tests had an issue unloading the appdomain or not.

---

@tfabris commented on [Tue Aug 23 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-241911180)

Clarifying my last message:
- The same bug occurs whether or not you redirect the console output: The nunit agent remains in memory and does not leave. 
- However, you only notice a problem (it's only a blocking issue) in situations where the console output is redirected, such as part of an automation system. That is the situation in which the automation hangs because of the Nunit agent which does not close.

---

@CharliePoole commented on [Tue Aug 23 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-241943236)

Some work is ongoing in this area. I'll reopen this and we can test when the changes are in.

---

@tfabris commented on [Tue Aug 23 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-241943836)

Sweet. We have an easy repro case, so let me know if you want me to verify a fixed build.

---

@CharliePoole commented on [Sat Aug 27 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-242932213)

@tfabris Can you test this against the latest MyGet build? It's at https://www.myget.org/feed/Packages/nunit and the latest version at the moment is 3.5.0-dev-03158.

---

@tfabris commented on [Mon Aug 29 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-243200690)

I'm very interested in trying out this build of Nunit against our repro case. 

However I am unable to obtain it at the link you provided. Following the link brought me to a signup page for the MyGet service. I had never used MyGet before, and, even after I set up an account there and created a feed, the link you provided ( https://www.myget.org/feed/Packages/nunit ) still does not result in me obtaining the file; it redirects me to a MyGet help/config page. Even after I made sure my profile was filled out and I created a "Feed" in MyGet and tried to add "Nunit" to my feed, same thing: I can't obtain this new build of Nunit that you're trying to link me to. The version of Nunit that is gives me in MyGet is 3.4.1 and I don't see any option to have it give me work-in-progress builds.

Do you have any tips on how to obtain this 3.5.0 dev build?

Thanks!

---

@CharliePoole commented on [Mon Aug 29 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-243208285)

@tfabris Sorry - I guess that's only a link for our account to use. There is a Gallery feature in MyGet, but we have not set that up for nunit yet.

Use the following url in Visual Studio (under Nuget sources) to get our myget feed:
   https://www.myget.org/F/nunit/api/v2

Alternatively, you can grab the build from AppVeyor:
   https://ci.appveyor.com/project/CharliePoole/nunit-console/build/3.5.0-dev-03158/artifacts

---

@tfabris commented on [Mon Aug 29 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-243219355)

Thanks so much, Charlie!!!!

I obtained the package from AppVeyor:  package\NUnit.ConsoleRunner.3.5.0-dev-03158.nupkg

I extracted it and ran the console runner contained within that package at the command prompt, pointing it to our repro DLL. 

Result: Same problem. Steps:

- When the tests in our repro test DLL are complete and all tests have passed, there is the final report of all test scores at the console. All tests pass. Then, after the test scores are displayed, the error message "Unable to unload AppDomain, Unload thread timed out" appears on the console (this is the bad condition that the test DLL must induce in order for the problem below to repro).

- At that point in time, "Nunit3-Console.exe" closes and is no longer in memory, as expected. However  "nunit-agent-x86" remains in memory and does not leave memory. (This is the bug.)

- If you are running directly at the DOS console, then control returns to the console, despite the agent remaining in memory.

- If you are redirecting the console output of the Nunit runner (for example, if you are running a script which pipes the output somewhere else for capture/logging, or, you are running in Team City which pipes the output into memory for capture/logging), then the progress hangs and control is not returned to the calling program. It waits forever and does not return. If your test is running in Team City, the the build hangs forever and never finishes. (This is the most critical failure mode of the bug as it blocks production.)

Details, from an analysis by our escalation developer, using the Nunit 3.4.1 build:

The process nunit-agent-x86.exe is kept alive because its main thread is waiting forever for a manual Event to be set. This is the call stack of the main thread:

OS Thread Id: 0x6ff0 (0)
Child SP       IP Call Site
00afede8 7745718c [HelperMethodFrame_1OBJ: 00afede8] System.Threading.WaitHandle.WaitOneNative(System.Runtime.InteropServices.SafeHandle, UInt32, Boolean, Boolean)
00afeecc 723a49d1 System.Threading.WaitHandle.InternalWaitOne(System.Runtime.InteropServices.SafeHandle, Int64, Boolean, Boolean)
00afeee4 723a4998 System.Threading.WaitHandle.WaitOne(Int32, Boolean)      <<< Waits forever here…
00afeef8 723a496e System.Threading.WaitHandle.WaitOne()
00afef00 0539707e NUnit.Engine.Agents.RemoteTestAgent.WaitForStop()
00afef08 00d2173e NUnit.Agent.NUnitTestAgent.Main(System.String[]) [C:\Users\xxxxxxxxxxx\AppData\Local\JetBrains\Shared\v04\DecompilerCache\decompiler\9E876AE6-521E-4A30-9583-EE5D668CC2CC\76\160e3076\NUnitTestAgent.cs @ 92]
00aff0e8 73301376 [GCFrame: 00aff0e8] 

This is the NUnit source code relative to the methods above:
 
public void WaitForStop()
    {
      this.stopSignal.WaitOne();
    }

---

@CharliePoole commented on [Mon Aug 29 2016](https://github.com/nunit/nunit/issues/1628#issuecomment-243221404)

This issue should have been moved to the new nunit-console repository when we split repos. Doing so now.

CharliePoole:
Fixes #43

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/HangingAppDomainFixture.cs
+++ b/src/NUnitEngine/nunit.engine.tests/HangingAppDomainFixture.cs
@@ -1,0 +1,44 @@
+﻿// ***********************************************************************
+// Copyright (c) 2016 Charlie Poole
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System;
+using System.Threading;
+using NUnit.Framework;
+
+namespace NUnit.Engine.Tests
+{
+    [Explicit]
+    public class HangingAppDomainFixture
+    {
+        [Test]
+        public void PassingTest()
+        {
+            Assert.Pass();
+        }
+
+        ~HangingAppDomainFixture()
+        {
+            Thread.Sleep(TimeSpan.FromDays(1));
+        }
+    }
+}
--- a/src/NUnitEngine/nunit.engine.tests/nunit.engine.tests.csproj
+++ b/src/NUnitEngine/nunit.engine.tests/nunit.engine.tests.csproj
@@ -70,6 +70,7 @@
     <Compile Include="AsyncTestEngineResultTests.cs" />
     <Compile Include="Drivers\NUnit3FrameworkDriverTests.cs" />
     <Compile Include="DummyExtensions.cs" />
+    <Compile Include="HangingAppDomainFixture.cs" />
     <Compile Include="Internal\AssemblyHelperTests.cs" />
     <Compile Include="Internal\DirectoryFinderTests.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
--- a/src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
+++ b/src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
@@ -215,27 +215,58 @@ namespace NUnit.Engine.Runners
 
         protected override void Dispose(bool disposing)
         {
-            base.Dispose(disposing);
+            // Disposal has to perform two actions, unloading the runner and
+            // stopping the agent. Both must be tried even if one fails so
+            // there can be up to two independent errors to be reported
+            // through an NUnitEngineException. We do that by combining messages.
+            if (!_disposed && disposing)
+            {
+                _disposed = true;
 
-            try
-            {
-                if (disposing && _agent != null)
+                string unloadError = null;
+
+                try
                 {
-                    log.Debug("Stopping remote agent");
-                    _agent.Stop();
-                    _agent = null;
+                    Unload();
                 }
-            }
-            catch (Exception e)
-            {
-                log.Error("Failed to stop the remote agent. {0}", e.Message);
-                _agent = null;
+                catch(Exception ex)
+                {
+                    // Save and log the unload error
+                    unloadError = ex.Message;
+                    log.Error(unloadError);
+                }
+
+                if (_agent != null)
+                {
+                    try
+                    {
+                        log.Debug("Stopping remote agent");
+                        _agent.Stop();
+                        _agent = null;
+                    }
+                    catch (Exception e)
+                    {
+                        string stopError = string.Format("Failed to stop the remote agent. {0}", e.Message);
+                        log.Error(stopError);
+                        _agent = null;
+
+                        // Stop error with no unload error, just rethrow
+                        if (unloadError == null)
+                            throw;
+
+                        // Both kinds of errors, throw exception with combined message
+                        throw new NUnitEngineException(unloadError + Environment.NewLine + stopError);
+                    }
+                }
+
+                if (unloadError != null) // Add message line indicating we managed to stop agent anyway
+                    throw (new NUnitEngineException(unloadError + "\nAgent Process was terminated successfully after error."));
             }
         }
 
-        #endregion
+#endregion
 
-        #region Helper Methods
+#region Helper Methods
 
         private void CreateAgentAndRunner()
         {
@@ -283,6 +314,6 @@ namespace NUnit.Engine.Runners
             return new TestEngineResult(suite);
         }
 
-        #endregion
+#endregion
     }
 }
--- a/src/NUnitEngine/nunit.engine/Services/DomainManager.cs
+++ b/src/NUnitEngine/nunit.engine/Services/DomainManager.cs
@@ -157,7 +157,7 @@ namespace NUnit.Engine.Services
 
                 if (!_unloadThread.Join(30000))
                 {
-                    string msg = "Unable to unload AppDomain, Unload thread timed out";
+                    string msg = "Unable to unload AppDomain, Unload thread timed out.";
 
                     log.Error(msg);
                     Kill(_unloadThread);


rprouse:
Looks good

Trass3r:
Will this end up in 3.5?

CharliePoole:
Yes... see milestone on the issue itself, #43

PR #87, opened 2016-10-09
Udpate scripts; use floating version for teamcity extension

CharliePoole:
The main purpose of this PR was to get the dependency version [1.0.*,1.0.1) to work for the teamcity extension. Getting it to work involved updating the bootstrap scripts and updating Cake as well.


CharliePoole:
@nunit/core-team @nunit/contributors Please review for the 3.5 release.

CharliePoole:
@rprouse @ChrisMaddock @NikolayPianikov 
As far as I can tell, this PR is not working correctly - at least on my machine.

The TC dependency contains `version="[1.0.*,1.1)"` but the nuspec file inside the nupkg has `version="[1.0.0,1.1.0)"`. Can anyone explain?

CharliePoole:
@rprouse @ChrisMaddock @NikolayPianikov 
As far as I can tell, this PR is not working correctly.

The TC dependency contains `version="[1.0.*,1.1)"` but the nuspec file inside the nupkg has `version="[1.0.0,1.1.0)"`. This is so both locally and on the AppVeyor build. Can anyone explain?

CharliePoole:
Some research seems to indicate that we have to convert from using packages.config to project.json in order to use this feature. I'm inclined to roll back the change, and use `version=1.0.2` until we have time to complete that migration and test it.

CharliePoole:
Still more reading... I'm not sure we can ever use this feature, since it won't work for users of older versions of VS. Not sure it will work with other IDEs either.

CharliePoole:
And still more... I'm not convinced that floating version is even intended to work for second-level dependencies such as those we include in our nuspec files. I think this needs a lot more research.

CharliePoole:
Removed from milestone since we backed out the change.

PR #90, opened 2016-10-10
Back out use of floating version for teamcity

CharliePoole:
This restores teamcity to depending on version 1.0.2 rather than using a range with a floating version. See discussion on PR #74.


CharliePoole:
@NikolayPianikov @rprouse This is the last fix required on before I can move to the release branch.

If either of you see an approach I'm missing, please let me know. Otherwise, I'd like to merge this and revisit the question of dependencies for 3.6.

rprouse:
Looks good, we can figure this out when we have more time. Merging.

NikolayPianikov:
@CharliePoole @rprouse ok, let's go back to this in 3.6, It is not principal for the TeamCity extension, but it could be useful for all components in the future

PR #92, opened 2016-10-11
Update README.md for NUnit-console

### LINKED ISSUE #81: Update README
README.md needs to be updated so that the badges at the top point to the console builds and NuGet and so that the information is correct for this repository.

Sebazzz:
Fixes #81

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# NUnit 3.0 #
+-# NUnit 3.0 #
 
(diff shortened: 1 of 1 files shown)


CharliePoole:
Seems good to me with one question. Do we want to single out a particular nuget package for download without having the user understand the options? @rprouse What do you think?

rprouse:
I would like to see more of an explanation of the various NuGet packages along with links/badges for each. I would also like to see the text of the readme changed to better describe the differences between this repository and the framework repository, possibly with cross-links in both. I also think we should have links to the documentation from each. That said, I think @CharliePoole or myself should handle that, so I am okay merging this pull request.

Before we merge however, @Sebazzz, could you update the contribution graph link to point to this repository instead of the main NUnit repository?

Sebazzz:
Sure, done. Happy to help.

rprouse:
@Sebazzz looks good, thanks for the contribution. I am going to merge as-is. @CharliePoole I will update the readme to include some information about the NuGet packages. I will create an issue.

Issue #94, opened 2016-10-12
Allow users to pass arguments via file(s) instead parameters of a command line


NikolayPianikov:
It's important because arguments containing a bunch of assemblies with full paths have a significant length. And users should split these assemblies to run tests in several sessions or use .nunit configuration files. It could be implemented like:
`--arguments=filename` and `-a=filename`
File name contains the text similar a command line.

For example a command line could be:
`nunit3-console.exe -a=myArgs`
where the `myArgs` file contains one line:
`assembly1 "assembly 2" assembly3 --noheader --teamcity --x86`
or several lines:
`assembly1 "assembly 2"`
`assembly3`
`--noheader --teamcity --x86`

Also in the simplest implementation we can use one command line argument per line, because we could avoid the question how to deal with symbol ".

Theoretically:
- The `-a` argument could be used several times in the command line.
- Each file with arguments could contain other arguments `-a`

CharliePoole:
In the past, I was accustomed to the syntax "@filename" to represent a file whose contents were interpolated into the command-line. Is that too archaic a convention to use today? It would be a convenient way to pre-process args before handing them to Mono.Options to parse.

I'm for this as a feature if we can agree on a syntax. I'm up for repeating it on the command-line but I don't like the idea of nesting.

NikolayPianikov:
I like idea `@filename`

I agree that the nesting is not useful in general scenarios, but for this argument we should check a source of the argument each time and we should raise an error if the argument was defined in a file.

Also I spent some time to see how settings are working now and to write tests for settings before making changes. I found that we have no any unit/integration tests for it. I started from prototyping and found that we have no way to write "real" unit tests - we are not using any frameworks for mocks so far. It is possible to choose any and use it everywhere?

CharliePoole:
If we treat it as a pre-processing step

CharliePoole:
If we treat it as a pre-processing step, then the options code won't even understand the format and should give an error.

We have unit tests for all our settings.  We don't have real integration tests for anything. Maybe you mean something else?

CharliePoole:
We use nsubstitute

NikolayPianikov:
Ok I will use `nsubstitute` if it is needed for tests.

I mean unit tests for `OptionSet` class that looks quite complex for me. Yes I found unit tests for the `ConsoleOptions` class.

What do you think should we use a file's format like for the command line - the single line in a file? Or we parse the each argument from the own line?

[...]

PR #98, opened 2016-10-13
Simple implementation for "Allow users to pass arguments via file(s) instead parameters of a command line"

NikolayPianikov:
See Allow users to pass arguments via file(s) instead parameters of a command line #94

### DIFF
--- a/src/NUnitConsole/nunit3-console.tests/ArgumentsFileParserTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/ArgumentsFileParserTests.cs
@@ -1,0 +1,63 @@
+﻿// ***********************************************************************
+// Copyright (c) 2011 Charlie Poole
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System;
+using NUnit.Framework;
+
+namespace NUnit.ConsoleRunner.Tests
+{
+    [TestFixture]
+    public class ArgumentsFileParserTests
+    {
+        [Test]
+        [TestCase("--arg", "--arg")]
+        [TestCase("--arg1 --arg2", "--arg1\n--arg2")]
+        [TestCase("--arg1\n--arg2", "--arg1\n--arg2")]
+        [TestCase("", "")]
+        [TestCase("   ", "")]
+        [TestCase("--arg1 --arg2", "--arg1\n--arg2")]
+        [TestCase("\"--arg 1\" --arg2", "--arg 1\n--arg2")]
+        [TestCase("\"--arg 1\"\n--arg2", "--arg 1\n--arg2")]
+        [TestCase("--arg1 \"--arg 2\"", "--arg1\n--arg 2")]
+        [TestCase("\"--arg 1\" \"--arg 2\"", "--arg 1\n--arg 2")]
+        [TestCase("\"--arg 1\" \"--arg 2\" arg3 \"arg 4\"", "--arg 1\n--arg 2\narg3\narg 4")]
+        [TestCase("--arg1 \"--arg 2\" arg3 \"arg 4\"", "--arg1\n--arg 2\narg3\narg 4")]
+        [TestCase("\"--arg 1\" \"--arg 2\" arg3 \"arg 4\"\n\"--arg 1\" \"--arg 2\" arg3 \"arg 4\"", "--arg 1\n--arg 2\narg3\narg 4\n--arg 1\n--arg 2\narg3\narg 4")]
+        [TestCase("\"--arg\"", "--arg")]
+        [TestCase("\"--arg 1\"", "--arg 1")]
+        [TestCase("\"--arg\"1\"", "--arg\n1")]
+        public void ShouldParseArgumentsFromLines(string linesStr, string expectedArgsStr)
+        {
+            // Given
+            var parser = new ArgumentsFileParser();
+            var lines = linesStr.Split(new[] { '\n' }, StringSplitOptions.RemoveEmptyEntries);
+            var expectedArgs = expectedArgsStr.Split(new[] { '\n' }, StringSplitOptions.RemoveEmptyEntries);
+
+            // When
+            var actualArgs = parser.Convert(lines);
+
+            // Then
+            Assert.AreEqual(expectedArgs, actualArgs);
+        }
+    }
+}
--- a/src/NUnitConsole/nunit3-console.tests/ArgumentsFileParserTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/ArgumentsFileParserTests.cs
@@ -45,7 +45,6 @@ namespace NUnit.ConsoleRunner.Tests
         [TestCase("\"--arg 1\" \"--arg 2\" arg3 \"arg 4\"\n\"--arg 1\" \"--arg 2\" arg3 \"arg 4\"", "--arg 1\n--arg 2\narg3\narg 4\n--arg 1\n--arg 2\narg3\narg 4")]
         [TestCase("\"--arg\"", "--arg")]
         [TestCase("\"--arg 1\"", "--arg 1")]
-        [TestCase("\"--arg\"1\"", "--arg\n1")]
         [TestCase("\"--arg abc\"", "--arg abc")]
         [TestCase("\"--arg   abc\"", "--arg   abc")]
         [TestCase("\" --arg   abc \"", " --arg   abc ")]
--- a/src/NUnitConsole/nunit3-console.tests/ArgumentsFileParserTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/ArgumentsFileParserTests.cs
@@ -29,8 +29,8 @@ namespace NUnit.ConsoleRunner.Tests
     [TestFixture]
     public class ArgumentsFileParserTests
     {
-        [Test]
         [TestCase("--arg", "--arg")]
+        [TestCase("--ArG", "--ArG")]
         [TestCase("--arg1 --arg2", "--arg1\n--arg2")]
         [TestCase("--arg1\n--arg2", "--arg1\n--arg2")]
         [TestCase("", "")]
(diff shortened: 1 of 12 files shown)


CharliePoole:
This looks very good to me as a start but it's difficult to review. It looks as if you didn't update your local master before creating the new branch. Can you tease on current master and push -f?

NikolayPianikov:
Oh I see, done

src/NUnitConsole/nunit3-console/IFileSystem.cs
```diff
@@ -0,0 +1,34 @@
...
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System.Collections.Generic;
+
+namespace NUnit.Common
+{
+    public interface IFileSystem
```
rprouse: It is about time we started abstracting the filesystem and stuff like that. Good job. "Poor man's dependency injection will do for now, but eventually I would like us to pull in our own small dependency injection framework for this stuff. Maybe SimpleIOC?

That is a total aside from this PR, just wanted to say 👍

CharliePoole: Totally agree! I thought the same thing when I saw this.

src/NUnitConsole/nunit3-console/ConsoleOptions.cs
```diff
@@ -33,9 +32,19 @@ namespace NUnit.Common
     /// </summary>
     public class ConsoleOptions : CommandLineOptions
     {
+        private readonly IFileSystem fileSystem;
```
CharliePoole: Our standard naming is to use `_fileSystem` for this and not to use `this` in references.

src/NUnitConsole/nunit3-console/IFileSystem.cs
CharliePoole: Regarding a DI framework, don't you have an issue on that? I'd be interested to see how you feel a framework would fit into a small example like this - either on that issue or via email.

rprouse: @CharliePoole there is an issue, nunit/nunit#346. I don't think that a framework would fit into a small example like this since there is only one dependency being injected, so just constructing it inline is fine. My thinking is that if we could start abstracting all of our usages of things like the file system, dates, etc, then constructing and testing those things starts becoming more difficult.

Once we get to the point where we have multiple interfaces and their implementation constructors also have dependencies injected, I find it easier to bootstrap a dependency injection container at startup and then only resolve interfaces in the code.

NikolayPianikov:
Should we allow duplication for files of arguments?

CharliePoole:
We don't check for duplicate arguments, so it probably doesn't matter if we check for duplicate files.

If we ever had any arguments in the future that were position dependent, allowing duplicates might even be a feature. :-)

So I'd say keep it simple.

NikolayPianikov:
Ok KISS is a good principle.

NikolayPianikov:
@oznetmaster @rprouse @CharliePoole do you have any updates/ideas about this feature? I've done all things which were discussed. I am going to add integration tests for this, but in the nunit-teamcity-event-listener repo :)

src/NUnitConsole/nunit3-console/ConsoleOptions.cs
```diff
@@ -23,19 +23,34 @@
...
-        internal ConsoleOptions(IDefaultOptionsProvider provider, params string[] args) : base(provider, args) { }
+        internal ConsoleOptions(
+            IDefaultOptionsProvider provider,
+            IFileSystem fileSystem,
+            IConverter<IEnumerable<string>, IEnumerable<string>> argumentsFileParser,
+            params string[] args)
+            : base(provider, args)
+        {
+            if (fileSystem == null) throw new ArgumentNullException("fileSystem");
+            if (fileSystem == null) throw new ArgumentNullException("argumentsFileParser");
```
rprouse: This line should be argumentsFileParser

src/NUnitConsole/nunit3-console.tests/ArgumentsFileParserTests.cs
```diff
@@ -0,0 +1,63 @@
...
+        [TestCase("\"--arg 1\"\n--arg2", "--arg 1\n--arg2")]
+        [TestCase("--arg1 \"--arg 2\"", "--arg1\n--arg 2")]
+        [TestCase("\"--arg 1\" \"--arg 2\"", "--arg 1\n--arg 2")]
+        [TestCase("\"--arg 1\" \"--arg 2\" arg3 \"arg 4\"", "--arg 1\n--arg 2\narg3\narg 4")]
+        [TestCase("--arg1 \"--arg 2\" arg3 \"arg 4\"", "--arg1\n--arg 2\narg3\narg 4")]
+        [TestCase("\"--arg 1\" \"--arg 2\" arg3 \"arg 4\"\n\"--arg 1\" \"--arg 2\" arg3 \"arg 4\"", "--arg 1\n--arg 2\narg3\narg 4\n--arg 1\n--arg 2\narg3\narg 4")]
+        [TestCase("\"--arg\"", "--arg")]
+        [TestCase("\"--arg 1\"", "--arg 1")]
+        [TestCase("\"--arg\"1\"", "--arg\n1")]
+        public void ShouldParseArgumentsFromLines(string linesStr, string expecdetArgsStr)
```
rprouse: Fix spelling of `expecdetArgsStr`. I wouldn't worry about this normally, but I am asking for other changes, so may as well fix it 😄

src/NUnitConsole/nunit3-console/ConsoleOptions.cs
NikolayPianikov: fixed

src/NUnitConsole/nunit3-console.tests/ArgumentsFileParserTests.cs
NikolayPianikov: fixed

NikolayPianikov:
@rprouse thank you for code review
May be we will move a common logic into nunitlite and will use it everywhere via dependency on nunitlite?

rprouse:
> May be we will move a common logic into nunitlite and will use it everywhere via dependency on nunitlite?

NUnitLite depends on NUnit, so we would introduce a chain of dependencies that we don't want to see. I think that eventually we may create something like an `NUnit.Common` or `NUnit.Utils` to hold common code between multiple projects, but that isn't a discussion for here 😄 

This is the first time where multiple people are implementing the same feature in multiple projects at the same time, so we are learning how to manage it. Normally, one person will do the work in every project, or people will port from one to the other.

For now, I would like to see the structure and naming be as similar between the two projects as possible.

NikolayPianikov:
@rprouse nunitlite depends on NUnit but not on NUnit engine/console I think. I mean that if nunitlite provides some subset of full NUnit, thus logically if we wanted to have the common code we could put it in nunitlite. We could avoid creating additional packages/assemblies if they are not really needed. It just my reasoning and I agree it is wrong place to discuss this.

CharliePoole:
@NikolayPianikov What you describe with the console and nunitlite is what we were doing before we split the repositories. We sacrificed the ability to do that because we wanted the framework and console to be able to evolve separately and nunitlite is considered a part of the framework because it actually references the framework assembly using the same version. This allows a very efficient way of running tests in a single assembly - which is all we officially support.

Since we just made this major change, I don't think we want to reverse it right now. We should keep along the path we started until/unless something arises to make us change it. So that leaves two options:

1. NUnitLite and the console (or engine) both reference a common assembly as Rob suggests.
2. We use a Git submodule for sharing at the source level.

The second approach is something I'm trying to get familiar with in order to see if it would work for us. If you have any experience in your projects using submodules, let's talk separately and maybe we can come up with something. In particular, I'm thinking about some utility methods and our command-line option processing as things we might handle this way.

The first approach would still be available as a later option. I just wanted to explore this option that I know very little about first.

src/NUnitConsole/nunit3-console.tests/ArgumentsFileParserTests.cs
```diff
@@ -0,0 +1,63 @@
...
+
+using System;
+using NUnit.Framework;
+
+namespace NUnit.ConsoleRunner.Tests
+{
+    [TestFixture]
+    public class ArgumentsFileParserTests
+    {
+        [Test]
```
NikolayPianikov: done

src/NUnitConsole/nunit3-console.tests/ArgumentsFileParserTests.cs
```diff
@@ -0,0 +1,63 @@
...
+        [TestCase("\"--arg 1\"\n--arg2", "--arg 1\n--arg2")]
+        [TestCase("--arg1 \"--arg 2\"", "--arg1\n--arg 2")]
+        [TestCase("\"--arg 1\" \"--arg 2\"", "--arg 1\n--arg 2")]
+        [TestCase("\"--arg 1\" \"--arg 2\" arg3 \"arg 4\"", "--arg 1\n--arg 2\narg3\narg 4")]
+        [TestCase("--arg1 \"--arg 2\" arg3 \"arg 4\"", "--arg1\n--arg 2\narg3\narg 4")]
+        [TestCase("\"--arg 1\" \"--arg 2\" arg3 \"arg 4\"\n\"--arg 1\" \"--arg 2\" arg3 \"arg 4\"", "--arg 1\n--arg 2\narg3\narg 4\n--arg 1\n--arg 2\narg3\narg 4")]
+        [TestCase("\"--arg\"", "--arg")]
+        [TestCase("\"--arg 1\"", "--arg 1")]
+        [TestCase("\"--arg\"1\"", "--arg\n1")]
+        public void ShouldParseArgumentsFromLines(string linesStr, string expectedArgsStr)
```
NikolayPianikov: done

src/NUnitConsole/nunit3-console/ArgumentsFileParser.cs
```diff
@@ -0,0 +1,53 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+
+namespace NUnit.ConsoleRunner
+{
+    internal class ArgumentsFileParser: IConverter<IEnumerable<string>, IEnumerable<string>>
+    {
+        private static readonly Regex ArgsRegex = new Regex("(\\\"(?<arg>.+?)\\\"|(?<arg>[^\\s\"]+))", RegexOptions.Compiled | RegexOptions.CultureInvariant);
```
NikolayPianikov: @CharliePoole Regarding DI. I try making a code in accordance with the principle SOLID and "test driven" if it possible. I see that the most simplest way to do it is to use DI/IoC/IoC containers. Also it allows to avoid an inheritance and a code duplication, it makes my code simpler and it becomes a source of really unit tests without combinatorial explosion to cover all possible cases. I mean the each "if" statement increases the number of combinations of 2 times, but the each class inheritance makes it many times more. But is is only my view and I do not want to promote it to anyone :)

NikolayPianikov: @CharliePoole I've checked Regex which was suggested by @oznetmaster and it works fine too. I could use it if it looks better.

CharliePoole: @NikolayPianikov I haven't even looked at the other regex. Just wanted you to check it out and do what you think best.

src/NUnitConsole/nunit3-console/ConsoleOptions.cs
```diff
@@ -155,5 +170,57 @@ protected override void ConfigureOptions()
...
+                if (arg.Length == 0 || arg[0] != '@')
+                {
+                    yield return arg;
+                    continue;
+                }
+
+                var fileName = arg.Substring(1, arg.Length - 1);
+                if (string.IsNullOrEmpty(fileName))
+                {
+                    ErrorMessages.Add("The file name should not be empty.");
```
CharliePoole: This doesn't tell the user much if it's a long command-line... can we say "...after @"?

CharliePoole:
Looks OK to me except for that one error message that seems a bit unclear. @rprouse ?

src/NUnitConsole/nunit3-console/ConsoleOptions.cs
rprouse: Can you change the error to "You must include a file name after @"?

With that, we can merge. Thanks.

NikolayPianikov: @rprouse I've changed the error message

Issue #100, opened 2016-10-20
class NUnit.Engine.Services.ResultWriters.Tests.SchemaValidator is never used


constructor-igor:
probably, it's necessary to remove the class (and file)?

CharliePoole:
I agree. I had left it in from NUnit V2 with the idea we might use it for testing, but we never did. It's always recoverable if we later need it and simplifying the  code is always a good idea.

PR #107, opened 2016-10-20
#100 removed file SchemaValidator.cs

constructor-igor:
#100 removed file SchemaValidator.cs


CharliePoole:
Just an unused file... merging.

Issue #110, opened 2016-10-22
Writability check could give a friendlier message.


CharliePoole:
@CharliePoole commented on [Fri Oct 16 2015](https://github.com/nunit/nunit/issues/901)

We introduced a writability check for result files, which simply tries to write an abbreviated file to the directory, throwing an exception and aborting the runner if it is unable.

We should see if we can give a better message that explains the directory is not writable.

When the directory used was chosen by default (i.e. the current directory is not writable), we should consider having a backup location instead.

---

@rvignesh89 commented on [Fri Oct 21 2016](https://github.com/nunit/nunit/issues/901#issuecomment-255506467)

@CharliePoole could you give a little more details about this enhancement?

---

@CharliePoole commented on [Fri Oct 21 2016](https://github.com/nunit/nunit/issues/901#issuecomment-255507631)

The console runner calls CheckWritability on each output path specified for saving results, e.g.: TestResult.xml. That method is implemented by various result writer extensions and generally throws some System.IO type of exception. The user has to interpret the exception to figure out what is wrong.

We should catch the exception and make it the inner exception of an NUnitEngineException that has a message saying specifically that the output path specified is not writable. If you want to do it, get that part working and we can consider a backup location after that.

Since this is related to the console runner, it's in the wrong repo. I will move it.

octoberclub:
What is the intention of implementing CheckWritability in each of the implementations? They all seem to do the same - open a StreamWriter to check the file path is writable.

CharliePoole:
The implementation may choose to create an output file at this point, which is later overwritten. That leaves some info behind in the case where the run terminates prematurely. The nunit 3 writer does this.

octoberclub:
Thanks, so I guess it needs to catch SystemException here https://github.com/nunit/nunit-console/blob/master/src/NUnitConsole/nunit3-console/ConsoleRunner.cs#L172 and rethrow a NUnitEngineException with the system exception set as the inner exception ?

CharliePoole:
Yes, I think that will do the job.

rvignesh89:
@CharliePoole I couldn't find any type of tests for ConsoleRunner. Could you lead me to the direction if I've missed it?

CharliePoole:
[...]

PR #119, opened 2016-10-26
fix Issue #110 Writability check could give a friendlier message

octoberclub:
Fix Issue #110 Writability check could give a friendlier message 
Throw a NUnitEngineException with the system exception set as the inner exception

### DIFF
--- a/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
+++ b/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
@@ -169,7 +169,17 @@ namespace NUnit.ConsoleRunner
             foreach (var spec in _options.ResultOutputSpecifications)
             {
                 var outputPath = Path.Combine(_workDirectory, spec.OutputPath);
-                GetResultWriter(spec).CheckWritability(outputPath);
+                try
+                {
+                    GetResultWriter(spec).CheckWritability(outputPath);
+                }
+                catch (SystemException ex)
+                {
+                    throw new NUnitEngineException(
+                        String.Format(
+                            "The path specified in --result {0} could not be written to",
+                            spec.OutputPath), ex);
+                }
             }
 
             // TODO: Incorporate this in EventCollector?


CharliePoole:
Two people with PRs for the same issue! Never seen that before. I'll review both and pick one. 😄

CharliePoole:
Thanks for the contribution! It does everything needed for the moment and we can consider adding tests later.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -169,7 +169,17 @@ private int RunTests(TestPackage package, TestFilter filter)
...
             {
                 var outputPath = Path.Combine(_workDirectory, spec.OutputPath);
-                GetResultWriter(spec).CheckWritability(outputPath);
+                try
+                {
+                    GetResultWriter(spec).CheckWritability(outputPath);
+                }
+                catch (SystemException ex)
+                {
+                    throw new NUnitEngineException(
```
rvignesh89: @CharliePoole IMO throwing NUnitEngine or any Exception here without suppressing the stacktrace in Program.cs is not improving the readability of the error message because the resultant exception looks like this, And once you read the error message in the first line the stacktrace is basically redundant.

```
C:\Users\vignesh.rajasekaran3\gitspace\nunit-console\bin\Debug [master]
λ .\nunit3-console.exe .\mock-assembly.dll --work=c:\windows\
NUnit Console Runner 3.5.0 (Debug)
Copyright (C) 2016 Charlie Poole

Runtime Environment
   OS Version: Microsoft Windows NT 10.0.14393.0
  CLR Version: 4.0.30319.42000

Test Files
    .\mock-assembly.dll

NUnit.Engine.NUnitEngineException: The path specified in --result TestResult.xml could not be written to ---> System.UnauthorizedAccessException: Access to the path 'c:\windows\TestResult.xml' is denied.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.StreamWriter.CreateFile(String path, Boolean append, Boolean checkHost)
   at System.IO.StreamWriter..ctor(String path, Boolean append, Encoding encoding, Int32 bufferSize, Boolean checkHost)
   at System.IO.StreamWriter..ctor(String path, Boolean append, Encoding encoding)
   at NUnit.Engine.Services.NUnit3XmlResultWriter.WriteResultFile(XmlNode resultNode, String outputPath) in c:\Users\vignesh.rajasekaran3\gitspace\nunit-console\src\NUnitEngine\nunit.engine\Services\ResultWriters\NUnit3XmlResultWriter.cs:line 53
   at NUnit.Engine.Services.NUnit3XmlResultWriter.CheckWritability(String outputPath) in c:\Users\vignesh.rajasekaran3\gitspace\nunit-console\src\NUnitEngine\nunit.engine\Services\ResultWriters\NUnit3XmlResultWriter.cs:line 48
   at NUnit.ConsoleRunner.ConsoleRunner.RunTests(TestPackage package, TestFilter filter) in c:\Users\vignesh.rajasekaran3\gitspace\nunit-console\src\NUnitConsole\nunit3-console\ConsoleRunner.cs:line 174
   at NUnit.ConsoleRunner.ConsoleRunner.RunTests(TestPackage package, TestFilter filter) in c:\Users\vignesh.rajasekaran3\gitspace\nunit-console\src\NUnitConsole\nunit3-console\ConsoleRunner.cs:line 178
   at NUnit.ConsoleRunner.ConsoleRunner.Execute() in c:\Users\vignesh.rajasekaran3\gitspace\nunit-console\src\NUnitConsole\nunit3-console\ConsoleRunner.cs:line 129
   at NUnit.ConsoleRunner.Program.Main(String[] args) in c:\Users\vignesh.rajasekaran3\gitspace\nunit-console\src\NUnitConsole\nunit3-console\Program.cs:line 129
```

Whilst I think what could make it easier to identify this issue is an output something like this. I'm thinking a little bigger here in terms of throwing the exception only when trying to write the results into the file. Basically separating of concerns of running the tests and saving the results. Not sure if this is aligning with the design but just my 2 cents 😄 

```
C:\Users\vignesh.rajasekaran3\gitspace\nunit-console\bin\Debug [master]
λ .\nunit3-console.exe .\mock-assembly.dll --work=c:\windows\
NUnit Console Runner 3.5.0 (Debug)
Copyright (C) 2016 Charlie Poole

Runtime Environment
   OS Version: Microsoft Windows NT 10.0.14393.0
  CLR Version: 4.0.30319.42000

Test Files
    .\mock-assembly.dll

<results of tests actually run>

Run Settings
    DisposeRunners: True
    WorkDirectory: c:\windows\
    ImageRuntimeVersion: 2.0.50727
    ImageRequiresX86: False
    ImageRequiresDefaultAppDomainAssemblyResolver: False
    NumberOfTestWorkers: 8

Test Run Summary
  Overall result: Failed
  Test Count: 28, Passed: 15, Failed: 5, Inconclusive: 1, Skipped: 7
    Failed Tests - Failures: 1, Errors: 1, Invalid: 3
    Skipped Tests - Ignored: 4, Explicit: 3, Other: 0
  Start time: 2016-10-27 04:16:03Z
    End time: 2016-10-27 04:16:03Z
    Duration: 0.504 seconds

Unable to save results to c:\windows\TestResults.xml due to lack of permission.
```

CharliePoole:
@rvignesh89 It's late here, so I won't look over your PRs until tomorrow but I thought I would mention that I see you do have a partial point. I mistook the location of the throw and was thinking it would be caught and handled before going out to Program.cs.

However, I don't think we want the test to run if the writability check fails. Imagine that you have just started a test run that will take two hours, specifying a certain location for saving the result file. If that location can't be used, you want to be notified immediately so you can correct the command, not two hours later after it has run but failed to save the result.

More tomorrow.

rvignesh89:
@CharliePoole I get your point about failing early. Considering that here are some of my thoughts.

- If the exception has to be thrown before then we won't be able to show summary since AFAIK summary requires the tests to be run.
- The change implemented in this PR should also be implemented in ConsoleRunner.Explorer() to maintain consistency. Because I noticed Explore also writes results to a file.
- If Explore also has to do this check I think the writability check itself can be moved outside of RunTests.

PR #120, opened 2016-10-26
Fixes #110 Writability check could give a friendlier message

### LINKED ISSUE #110: Writability check could give a friendlier message.

rvignesh89:
This fixes #110.

I've added a test which uses `c:\windows` as the working directory in system which should not be authorized unless you are running from an elevated prompt. 

My PR is slightly different from #119 where I catch and throw an UnAuthorizedAccessException because,

1. I figured CheckWritability could fail for multiple reasons but this issue was more focused on permission issues.
2. I also modified Program.cs to catch UnAuthorizedException to suppress the stacktrace which I thought was unnecessary.
3. I've tried to follow the error message similar to other ones.

I also noticed that throwing an exception as soon as CheckWritability fails causes the Test Summary to be not printed. I feel this could be improved upon. @CharliePoole what do you think?

### DIFF
--- a/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
+++ b/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
@@ -156,7 +156,7 @@ namespace NUnit.ConsoleRunner
             {
                 foreach (OutputSpecification spec in _options.ExploreOutputSpecifications)
                 {
-                    _resultService.GetResultWriter(spec.Format, new object[] {spec.Transform}).WriteResultFile(result, spec.OutputPath);
+                    _resultService.GetResultWriter(spec.Format, new object[] { spec.Transform }).WriteResultFile(result, spec.OutputPath);
                     _outWriter.WriteLine("Results ({0}) saved as {1}", spec.Format, spec.OutputPath);
                 }
             }
@@ -169,7 +169,14 @@ namespace NUnit.ConsoleRunner
             foreach (var spec in _options.ResultOutputSpecifications)
             {
                 var outputPath = Path.Combine(_workDirectory, spec.OutputPath);
-                GetResultWriter(spec).CheckWritability(outputPath);
+                try
+                {
+                    GetResultWriter(spec).CheckWritability(outputPath);
+                }
+                catch (UnauthorizedAccessException e)
+                {
+                    throw new UnauthorizedAccessException("Error: Unauthorized working directory",e);
+                }
             }
 
             // TODO: Incorporate this in EventCollector?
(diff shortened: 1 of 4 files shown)


rvignesh89:
Okay, using `c:\windows\` seems to be a bad idea. Let me try to work around that in the test.

rvignesh89:
After some digging up of Mono framework it seems like it doesn't share a consistent API with windows when setting Directory permissions based on this [code](https://github.com/mono/mono/blob/8994ce8a09a101e7965a820ed7dc148d9685f89f/mcs/class/corlib/System.Security.AccessControl/NativeObjectSecurity.cs) from Mono.

So I'll have to figure out other ways to test this piece of code. Unix systems should have a similar restricted folder like `c:\windows\` that I can use. I'll check this out.

CharliePoole:
Two people with PRs for the same issue! Never seen that before. I'll review both and pick one. 😄

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -169,7 +169,14 @@ private int RunTests(TestPackage package, TestFilter filter)
             foreach (var spec in _options.ResultOutputSpecifications)
             {
                 var outputPath = Path.Combine(_workDirectory, spec.OutputPath);
-                GetResultWriter(spec).CheckWritability(outputPath);
+                try
+                {
+                    GetResultWriter(spec).CheckWritability(outputPath);
+                }
+                catch (UnauthorizedAccessException e)
```
CharliePoole: Better to catch any Exception here. I understand why you did this: there could be some unrelated exception in the extension, like a null reference. However extensions are supposed to have been tested separately, so at this point we have to trust the extension and assume the path is non-writable.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -169,7 +169,14 @@ private int RunTests(TestPackage package, TestFilter filter)
...
+                {
+                    throw new UnauthorizedAccessException("Error: Unauthorized working directory",e);
```
CharliePoole: This should be an NUnitEngineException.

rvignesh89:
I've commented on the other PR for some questions that I have. Since you have merged that one I'll close this PR and open a new one for the test on ConsoleRunner as you suggested. I think I get what you mean by using NSubstitute, let me work on that!

PR #121, opened 2016-10-27
Add ConsoleRunnerTest for writability check

rvignesh89:
@CharliePoole I wrote a simple test to assert PR #119 

Let me know if this is what you were aiming for. If so then I'll refactor the Mocks using NSubstitue to simplify the code. And I'll also give it a good name 😆

### DIFF
--- a/src/NUnitConsole/nunit3-console.tests/ConsoleRunnerTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/ConsoleRunnerTests.cs
@@ -1,0 +1,60 @@
+﻿using System;
+using System.IO;
+using System.Xml;
+using NUnit.Common;
+using NUnit.Engine;
+using NUnit.Engine.Extensibility;
+using NUnit.Engine.Services;
+using NUnit.Framework;
+
+namespace NUnit.ConsoleRunner.Tests
+{
+    class ConsoleRunnerTests
+    {
+        [Test]
+        public void ThrowsNUnitEngineExceptionWhenTestResultsAreNotWriteable()
+        {
+            var testEngine = new TestEngine();
+            testEngine.Services.Add(new FakeResultService());
+
+            var consoleRunner = new ConsoleRunner(testEngine, new ConsoleOptions("mock-assembly.dll"), new ColorConsoleWriter());
+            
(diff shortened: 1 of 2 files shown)


src/NUnitConsole/nunit3-console.tests/CosoleRunnerTests.cs
```diff
@@ -0,0 +1,57 @@
...
+using NUnit.Engine.Extensibility;
+using NUnit.Engine.Services;
+using NUnit.Framework;
+
+namespace NUnit.ConsoleRunner.Tests
+{
+    class CosoleRunnerTests
+    {
+        [Test]
+        public void XXX()
```
rprouse: Could you rename this to something more meaningful? Maybe `ThrowsNUnitEngineExceptionWhenTestResultsAreNotWriteable`?

src/NUnitConsole/nunit3-console.tests/CosoleRunnerTests.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            var testEngine = new TestEngine();
+            testEngine.Services.Add(new TestResultService());
+            var consoleRunner = new ConsoleRunner(testEngine, new ConsoleOptions("mock-assembly.dll", "--work=c:\\windows\\"), new ColorConsoleWriter());
+            Assert.Throws<NUnitEngineException>(() => { consoleRunner.Execute(); });
```
rprouse: I would like to check the message in the exception. You can check that the message contains a substring.

rvignesh89:
@rprouse I've made the changes as requested. Although, I couldn't use NSubstitute in my tests as previously planned. This was because the tests are still in .NET 2.0. Any plans to upgrade?

rprouse:
We don't plan to upgrade the .NET version any time soon. A feature of NUnit is that it works on older frameworks, and if you want to run your tests under .NET 2.0, the runner needs to be compiled to the lowest version supported, 2.0.

rprouse:
Looks good, thanks for the help. 👍

rprouse:
@nunit/core-team this needs a second review and merge.

ChrisMaddock:
Looks good to me!

PR #131, opened 2016-11-10
NUnit3-console to use AssertionResult collection for Multiple results.

### LINKED ISSUE #127: Modify console runner to display multiple assert information
Currently, the console runner uses the single failure message created for compatibility in order to display the result of a test using multiple asserts. It should use the actual result of each assert to display both the message and stack frame.

CharliePoole:
Fixes #127

### DIFF
--- a/src/NUnitConsole/nunit3-console/ResultReporter.cs
+++ b/src/NUnitConsole/nunit3-console/ResultReporter.cs
@@ -260,6 +260,47 @@ namespace NUnit.ConsoleRunner
 
         private void WriteSingleResult(XmlNode result, ColorStyle colorStyle)
         {
+            string status = GetStatus(result);
+
+            string reportID = (++_reportIndex).ToString();
+            string fullName = result.GetAttribute("fullname");
+            var assertions = result.SelectNodes("assertions/assertion");
+            int numAsserts = assertions.Count;
+
+            //if (numAsserts > 0)
+            //{
+            //    int assertionCounter = 0;
+            //    string assertID = reportID;
+            //    foreach (XmlNode assertion in assertions)
+            //    {
+            //        if (numAsserts > 1)
+            //            assertID = string.Format("{0}-{1}", reportID, ++assertionCounter);
+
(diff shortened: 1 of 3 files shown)


src/NUnitConsole/nunit3-console/ResultReporter.cs
```diff
@@ -260,6 +260,47 @@ private void WriteSummaryCount(string label, int count, ColorStyle color)
...
         private void WriteSingleResult(XmlNode result, ColorStyle colorStyle)
         {
+            string status = GetStatus(result);
+
+            string reportID = (++_reportIndex).ToString();
+            string fullName = result.GetAttribute("fullname");
+            var assertions = result.SelectNodes("assertions/assertion");
+            int numAsserts = assertions.Count;
+
+            //if (numAsserts > 0)
```
rprouse: Is this still a Do not merge because of this?

CharliePoole:
It is and also due to some refactoring! I forgot to mark it in the description.

CharliePoole:
Should be ready to merge now. I created a ConsoleTestResult class that ought to be useful in other runners that use the XML format, although the xml code itself will need to be changed. If you look at how it is used, you'll see there is a compromise between efficiency and convenience. Specifically, the class is not used until we have determined that a particular XML node represents a test to be displayed in the report.

src/NUnitConsole/nunit3-console/nunit3-console.csproj
```diff
@@ -45,6 +45,7 @@
     <ApplicationIcon>..\..\..\nunit.ico</ApplicationIcon>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Microsoft.VisualStudio.QualityTools.UnitTestFramework, Version=10.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL" />
```
rprouse: This looks wrong. Isn't this MSTest?

rprouse:
Other than the questionable reference, looks good.

rprouse:
Looks good now. Merging...

Issue #133, opened 2016-11-16
NUnit downloadable packages zip file naming is confusing and non-intuitive


hiredgunhouse:
Here https://github.com/nunit/nunit-console/releases/tag/3.5 you can download a package that is called "NUnit Console 3.5" but the main zip and msi files are named NUnit.3.5.0.zip and NUnit.3.5.0.msi which is very confusing if you consider that here https://github.com/nunit/nunit/releases/tag/3.5 you can download the main package called "NUnit 3.5" which gives you the main zip file named "NUnit-3.5.0.zip".
Can you spot the difference?
NUnit.3.5.0.zip vs NUnit-3.5.0.zip
Very confusing after you download the files cause you can tell which one is which.
Wouldn't it be better if the console package was named xxxxx-console or something similar?

rprouse:
Agreed, the ZIP files that we create in this repo should be renamed.

rprouse:
This is left over from when we split the repositories. It is an easy fix, we just need to update the names in `build.cake`

ChrisMaddock:
Just a note - the zip with the release here is (I think!) built in nunit-distribution, and contains all the extensions too.

Whether that's how it should be or not...not sure!

ChrisMaddock:
I wanted to do some tidying up of nunit-distribution before the next release anyway. The zip packages is made in that repo now, so I'm going to move the issue over there. @hiredgunhouse - feel free to subscribe to the new issue. 😄

ChrisMaddock:
Issue moved to [nunit/nunit-distribution #9](https://github.com/nunit/nunit-distribution/issues/9) via [**ZenHub**](https://www.zenhub.com/)

CharliePoole:
I'm re-opening this because I think the immediate issue is critical. The long-term fix belongs in the nunit-distribution project, but the badly named files can't wait for the next release. I've renamed files in the past when it was necessary, so I'll look at doing that.

CharliePoole:
I should add, this is also blocking the creation of a Chocolatey console package.

CharliePoole:
Direct renaming is possible for both the zip and the msi. WRT the msi, however, the dialog shows "NUnit 3.5.0" in the title bar. Of course, it would be better to show NUnit Console but I think that's acceptable. I'll spend a little time seeing if I can do a better fix, but not much.

CharliePoole:
[...]

PR #137, opened 2016-11-24
Remove all zip tasks

ChrisMaddock:
From tidying up https://github.com/nunit/nunit-console/issues/133

`ZIP_PACKAGE` is no longer user. `SRC_PACKAGE` is no longer used in the release process, and I think keeping the task in the cake file causes more confusion than removing it. GitHub provides 1-click functionality if a source zip is still required for any reason.


rprouse:
Good by me.

PR #124, opened 2016-10-31
Update --labels switch (continued)

CharliePoole:
**Do not merge**
Fixes #24 

I created this new branch under the nunit organization, merging @constructor-igor 's PR #78. This will make it easier to collaborate, particularly by providing tests.

I added new tests to the code for multiple events. They are written so that the expected output is what the current code prodcues but there are comments regarding what the output should be.

In so doing, I uncovered a linux bug. The travis CI is failing because of unexpected `\r\n` strings in the output. IOW, we don't convert newline characters provided by the user. This may actually be a feature rather than a bug, in which case the test can simply be changed.

PR #144, opened 2016-12-05
Revert "Update --labels switch (continued)"

rprouse:
Reverts nunit/nunit-console#124

I haven't heard back from @constructor-igor yet, so to be safe, I am reverting the pull request.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -180,4 +180,3 @@ MockAssemblyResult.xml
 PortabilityAnalysis*.html
 tools
 !tools/packages.config
-/issue_16.cmd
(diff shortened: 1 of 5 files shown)


CharliePoole:
@constructor-igor and I have been working on this using the approach where I write a test and he implements it. This may not be working so well. :-) Or at least it slows things down.

@constructor-igor I looked at the existing display with each available option and compared it with what you have implemented. At this point, I think the old display looks better in some cases, so I went ahead and reverted. I think the choice is whether to continue as we have been doing (in which case I'll try to be more responsive) or for me to finish it up. In either case, we can subsequently do further issues to improve it of course. What's your preference?

PR #146, opened 2016-12-08
Fix MasterTestRunner.RunAsync

### LINKED ISSUE #145: MasterTestRunner.RunAsync no longer provides start-run and test-run events
In 3.4.1 they were provided. This is critical to the GUI.

CharliePoole:
Fixes #145 

Latest changes since 3.4.1 broke RunAsync so that `<start-run>` and `<test-run>` events are never sent. The gui requires these events to know when a test run is starting and ending.

### DIFF
--- a/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
+++ b/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
@@ -29,6 +29,7 @@ using System.Reflection;
 using System.Xml;
 using NUnit.Engine.Internal;
 using NUnit.Engine.Services;
+using System.ComponentModel;
 
 namespace NUnit.Engine.Runners
 {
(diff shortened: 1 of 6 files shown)


CharliePoole:
Also added a line to suppress creating a window when launching the agent from the Gui.

CharliePoole:
Marking this do not merge for the moment. The tests all pass but it turns out that the `<start-run>` event is coming with a count of zero. The problem is that the console doesn't need the count, while the gui does need it.

CharliePoole:
@nunit/engine-team @nunit/core-team I could really use a review of this!

src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
```diff
@@ -374,6 +377,24 @@ private TestEngineResult RunTests(ITestEventListener listener, TestFilter filter
             return result;
         }
 
+        private AsyncTestEngineResult RunTestsAsync(ITestEventListener listener, TestFilter filter)
```
CharliePoole: This method is copied from AbstractTestRunner. In 3.4.1, MasterTestRunner inherited from AbstractTestRunner and the method was called. We eliminated that in separating MasterTestRunner from the internal engine runners. Since we don't call RunAsync in the console or VS adapter, the error went unnoticed until we updated the Gui to use the 3.5 engine.

rprouse:
Looks good, merging.

PR #150, opened 2016-12-14
Add option to skip non-test assemblies

### LINKED ISSUE #62: NUnit3 Fails on DLL with no Tests, Unlike NUnit2
@hypersw commented on [Fri Jun 17 2016](https://github.com/nunit/nunit/issues/1586)

Some projects of our codebase are for test purposes only, though not all of them have test cases (some got only mocking infrastructure code).

With NUnit1-NUnit2, we used to submit all of the DLLs from the test area, and NUnit would run all the tests it could find.

With NUnit3, it fails with an error like this:

```
Errors and Failures
1) Invalid : xxx.dll
No suitable tests found in 'xxx.dll'.
Either assembly contains no tests or proper test driver has not been found.
```

And returns the `ERRORLEVEL` of `FFFFFFFE`.

This is a breaking change from NUnit2, from how I see it.
We didn't think up a workaround yet.
Is it possible to get the previous version behavior, with an option or something?

---

@hypersw commented on [Fri Jun 17 2016](https://github.com/nunit/nunit/issues/1586#issuecomment-226791460)

Errorlevel looks like `NUnit.ConsoleRunner.ConsoleRunner::INVALID_ASSEMBLY`.

---

@hypersw commented on [Fri Jun 17 2016](https://github.com/nunit/nunit/issues/1586#issuecomment-226792272)

The error is probably triggered in `NUnit.Engine.Services.DriverService::GetDriver`.

---

@hypersw commented on [Fri Jun 17 2016](https://github.com/nunit/nunit/issues/1586#issuecomment-226796952)

Looks like it actually checks by referenced assemblies only rather than presence of the actual test cases.
That's something I could also check in the build script before spawning tests for each DLL.
Might be useful to be able to batch-submit them though, as you'd not always have a build script capable of reading thru assembly references like in our case.

---

@hypersw commented on [Fri Jun 17 2016](https://github.com/nunit/nunit/issues/1586#issuecomment-226803083)

I've prototyped a fix at https://github.com/JetBrains/nunit/commit/ad29539dca76c87dd11e2491cf8537f01a9e38e3 for us to try nunit3 full-scale without missing test errors.
Might not be a good idea doing this by default (even though this seems to be a breaking change from nunit2), but could prove useful if under an option.

---

@CharliePoole commented on [Fri Jun 17 2016](https://github.com/nunit/nunit/issues/1586#issuecomment-226806058)

Well, it's 3.x versus 2.x, so it has a ton of breaking changes. Minimum "fix" for this issue would be to make sure the change is listed on the Breaking Changes page. :-)

That said, I think this is a correct change. In most usage, passing a non-test assembly to NUnit is an error of some kind. It could be a simple, ignorable error as in your case. OTOH, it could be that you are using some third-party framework without installing the driver extension for that framework. There is no way for NUnit to know what is the cause.

This has already given us a lot of trouble with the VS adapter because VS passes us **all** the assemblies in a solution and it's also a problem for users who try to run the solution from the command line. We deal with the first by special checks in the adapter. The second is the subject of a separate issue (#668), which we will probably take care of by passing in some info about the source.

So, for various reasons, a package setting of "IgnoreNonTestAssemblies" is probably in the future. Once that's available, we could add a console option to activate it.

Note: This should be worked on in conjunction with #668.

---

@DanHarman commented on [Tue Sep 27 2016](https://github.com/nunit/nunit/issues/1586#issuecomment-249884710)

Is there any further thinking on this? My particular scenario is passing in a glob to my teamcity nunit build step for:

`**\bin\Release\*Tests.dll`

As this means we don't need to keep adding test projects into the build server manually (I've already failed to do this once...) However we also have mstest projects in the build which also have *Tests.dll extension, and nunit console test runner errors out as soon as it encounters one. Ideally I'd like a command line param to not error out if it is passed a dll without any tests it understands.

---

@CharliePoole commented on [Tue Sep 27 2016](https://github.com/nunit/nunit/issues/1586#issuecomment-249896696)

Well, as a result of earlier discussion, we accepted this as a feature request and added it to our backlog.

The next thing you can watch for is for somebody to take it on as an assignment. That could be a team member or an outside contributor who wants it badly. :smile:

---

@rprouse commented on [Tue Sep 27 2016](https://github.com/nunit/nunit/issues/1586#issuecomment-249917378)

If we decide to change this, we should do the same in the .NET Core runner. There is a duplicate issue there with a couple of requests, nunit/dotnet-test-nunit#77

---

@rprouse commented on [Tue Sep 27 2016](https://github.com/nunit/nunit/issues/1586#issuecomment-249917929)

This is also a Console issue, I am going to move this issue there.

CharliePoole:
Fixes #62 

I followed the choices laid out in the discussion on the issue. The use of NonTestAssemblyAttribute was tested on an adhoc basis but would still need to be added to the framework.

### DIFF
--- a/src/NUnitEngine/nunit.engine/Services/DriverService.cs
+++ b/src/NUnitEngine/nunit.engine/Services/DriverService.cs
@@ -48,26 +48,36 @@ namespace NUnit.Engine.Services
         /// <param name="domain">The AppDomain to use for the tests</param>
         /// <param name="assemblyPath">The full path to the test assembly</param>
         /// <param name="targetFramework">The value of any TargetFrameworkAttribute on the assembly, or null</param>
+        /// <param name="skipNonTestAssemblies">True if non-test assemblies should simply be skipped rather than reporting an error</param>
         /// <returns></returns>
-        public IFrameworkDriver GetDriver(AppDomain domain, string assemblyPath, string targetFramework)
+        public IFrameworkDriver GetDriver(AppDomain domain, string assemblyPath, string targetFramework, bool skipNonTestAssemblies)
         {
             if (!File.Exists(assemblyPath))
-                return new NotRunnableFrameworkDriver(assemblyPath, "File not found: " + assemblyPath);
+                return new InvalidAssemblyFrameworkDriver(assemblyPath, "File not found: " + assemblyPath);
 
             if (!PathUtils.IsAssemblyFileType(assemblyPath))
-                return new NotRunnableFrameworkDriver(assemblyPath, "File type is not supported");
+                return new InvalidAssemblyFrameworkDriver(assemblyPath, "File type is not supported");
 
             if (targetFramework != null)
             {
                 var platform = targetFramework.Split(new char[] { ',' })[0];
                 if (platform == "Silverlight" || platform == ".NETPortable" || platform == ".NETCompactFramework")
-                    return new NotRunnableFrameworkDriver(assemblyPath, platform + " test assemblies are not yet supported by the engine");
+                    return new InvalidAssemblyFrameworkDriver(assemblyPath, platform + " test assemblies are not yet supported by the engine");
             }
 
             try
             {
+                var assemblyDef = AssemblyDefinition.ReadAssembly(assemblyPath);
+
+                if (skipNonTestAssemblies)
+                {
+                    foreach (var attr in assemblyDef.CustomAttributes)
+                        if (attr.AttributeType.FullName == "NUnit.Framework.NonTestAssembly")
+                            return new SkippedAssemblyFrameworkDriver(assemblyPath);
+                }
+
                 var references = new List<AssemblyName>();
-                foreach (var cecilRef in AssemblyDefinition.ReadAssembly(assemblyPath).MainModule.AssemblyReferences)
+                foreach (var cecilRef in assemblyDef.MainModule.AssemblyReferences)
                     references.Add(new AssemblyName(cecilRef.FullName));
 
                 foreach (var factory in _factories)
@@ -81,10 +91,13 @@ namespace NUnit.Engine.Services
             }
             catch (BadImageFormatException ex)
             {
-                return new NotRunnableFrameworkDriver(assemblyPath, ex.Message);
+                return new InvalidAssemblyFrameworkDriver(assemblyPath, ex.Message);
             }
 
-            return new NotRunnableFrameworkDriver(assemblyPath, string.Format("No suitable tests found in '{0}'.\n" +
+            if (skipNonTestAssemblies)
+                return new SkippedAssemblyFrameworkDriver(assemblyPath);
+            else
+                return new InvalidAssemblyFrameworkDriver(assemblyPath, string.Format("No suitable tests found in '{0}'.\n" +
                                                                               "Either assembly contains no tests or proper test driver has not been found.", assemblyPath));
         }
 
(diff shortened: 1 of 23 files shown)


src/NUnitEngine/nunit.engine/Services/DriverService.cs
```diff
@@ -48,26 +48,36 @@ public class DriverService : Service, IDriverService
...
             }
 
             try
             {
+                var assemblyDef = AssemblyDefinition.ReadAssembly(assemblyPath);
+
+                if (skipNonTestAssemblies)
+                {
+                    foreach (var attr in assemblyDef.CustomAttributes)
+                        if (attr.AttributeType.FullName == "NUnit.Framework.NonTestAssembly")
```
rprouse: Shouldn't this be `"NUnit.Framework.NonTestAssemblyAttribtute"`?

CharliePoole:
I was waiting to see whether we all liked the attribute idea before creating an issue.

rprouse:
> I was waiting to see whether we all liked the attribute idea before creating an issue.

I know this has been discussed at length, but the attribute worries me a bit. The one use case I see is for a stub-test suite which would seem to be fairly uncommon usage. My worry is that it will be added to a suite by one developer, then forgotten. Other developers on the team will not understand why their tests are not running.

Also, if we do accept the attribute, shouldn't that suite always be skipped? I don't think it should be command line driven to enable it.

CharliePoole:
I too went back and forth on this. I feel that there is only one valid use case: a third-party framework extension that references nunit.framework is copied to the output directory. There have been several problems reported with these.

The attribute would be available for the third-party developer to use, to indicate that the assembly is not an nunit test assembly __even though it looks like one to the engine__. I don't think we want users to make use of it at all, but it has to be public so that the third-party dev can use it. Perhaps it should be put into the Internal or Extensions namespace and maybe even marked as non-browsable.

Originally, I thought that we should skip such assemblies, without warning, even if the user did not use the `--skipnontestassemblies` option. However, I decided that the user should see the message if the assembly were included in the set of assemblies being run and then choose whether to exclude it or use the option.

If you like, I can create a PR for the framework as well, so we can review and merge both bits of code at the same time.

CharliePoole:
@nunit/core-team This one has been waiting. Any reviews?

rprouse:
@CharliePoole I am fine with the attribute based on your comments and I am okay with the command line option. I don't see a commit for the attribute name, have you checked it?

Other than confirming the attribute code I commented on earlier works, this is good to merge.

CharliePoole:
@rprouse I had shelved this while waiting to be sure that the overall approach was OK with everyone. I'll fix that attribute spelling, add some tests and create an issue for the framework.

rprouse:
Looks good. Feel free to merge now if you are ready or we can wait for the Framework issue to be added if you prefer to do them both at once.

rprouse:
I have merged the related framework issue and resolved merge conflicts in this PR. Waiting for CI to pass to ensure my conflict changes are correct, then we can merge this.

CharliePoole:
I'm updating to the 3.6.0-dev-3507 framework and adding a further test. Will merge after.

CharliePoole:
@rprouse I pulled this and it gives a lot of errors running the tests locally. I'll try cloning another copy to see what happens.

CharliePoole:
@nunit/core-team I'll merge when CI passes. The last commit may be useful to examine for an understanding of how drivers are selected in the absence or presence of the --skipNonTestAssemblies option.

PR #152, opened 2016-12-20
Fix --framework option

### LINKED ISSUE #151: Unexpected behaviour from --framework flag
This isn't an option I've used before - so I think this is a bug, but can someone confirm what I'm expected is whats intended?

We just updated R#, and tests started failing. It appears the current resharper is forcing things to run under .NET 4.5, regardless of the framework specified - which is breaking some of our (questionable) tests.

To prove this was the case, I wanted to use `--framework=net-4.5` with the console runner, to force our tests which target .NET 4.0, to run on .NET 4.5. Surprisingly they passed. It looks like the 3.5 engine chooses the project's targeted .NET version, over what was specified on the command line.

The logs showed this:
```
16:55:36.477 Debug [ 1] RuntimeFrameworkService: Current framework is net-4.5
16:55:36.477 Debug [ 1] RuntimeFrameworkService: Requested framework is net-4.5
16:55:36.482 Debug [ 1] RuntimeFrameworkService: Assembly [redacted] uses version 4.0.30319
16:55:36.486 Debug [ 1] RuntimeFrameworkService: Test will use net-4.0 framework
```

I can debug this further, but before I do - am I expecting the right behaviour? If I specify a later framework than targeted on the command line, should the tests run under that?

ChrisMaddock:
Fixes #151. 

Think this was just something that was caught in a refactor, and untested. This PR also adds InternalsVisibleTo for the test assembly, so the package setting constants could be used.

### DIFF
--- a/src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs
+++ b/src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs
@@ -96,7 +96,10 @@ namespace NUnit.Engine.Services
         /// <returns>A string representing the selected RuntimeFramework</returns>
         public string SelectRuntimeFramework(TestPackage package)
         {
-            // Start by examining the provided settings
+            // Evaluate package target framework
+            ApplyImageData(package);
+
+            // Examine the provided settings
             RuntimeFramework currentFramework = RuntimeFramework.CurrentFramework;
             string frameworkSetting = package.GetSetting(EnginePackageSettings.RuntimeFramework, "");
             RuntimeFramework requestedFramework = frameworkSetting.Length > 0
@@ -115,13 +118,10 @@ namespace NUnit.Engine.Services
             if (targetRuntime == RuntimeType.Any)
                 targetRuntime = currentFramework.Runtime;
 
-            // Examine the package 
-            ApplyImageData(package);
+            if (targetVersion == RuntimeFramework.DefaultVersion)
+                targetVersion = package.GetSetting(InternalEnginePackageSettings.ImageRuntimeVersion, currentFramework.FrameworkVersion);
 
-            // Modify settings if necessary
-            targetVersion = package.GetSetting(InternalEnginePackageSettings.ImageRuntimeVersion, targetVersion);
-            RuntimeFramework checkFramework = new RuntimeFramework(targetRuntime, targetVersion);
-            if (!checkFramework.IsAvailable)
+            if (!new RuntimeFramework(targetRuntime, targetVersion).IsAvailable)
             {
                 log.Debug("Preferred version {0} is not installed or this NUnit installation does not support it", targetVersion);
                 if (targetVersion < currentFramework.FrameworkVersion)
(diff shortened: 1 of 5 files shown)


src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs
```diff
@@ -96,7 +96,10 @@ private static bool FrameworksMatch(RuntimeFramework f1, RuntimeFramework f2)
         /// <returns>A string representing the selected RuntimeFramework</returns>
         public string SelectRuntimeFramework(TestPackage package)
         {
-            // Start by examining the provided settings
+            // Evaluate package target framework
+            ApplyImageData(package);
+
```
ChrisMaddock: Where `ApplyImageData` is done should have no effect - I just moved it up to the top, so it was out of the way.

ChrisMaddock:
InternalsVisibleTo hasn't worked on Travis. I'll have to come back to it later...

ChrisMaddock:
Ok, this is ready for review now.

The problem with `InternalsVisibleTo` on mono was caused by `ExtensionAttribute` being available to the `nunit.engine.test` project from both `nunit.engine` and `nunit.linq`.

I just suppressed the warning here, but that's not ideal. I wonder if `nunit.linq` should all be internal, and use `InternalsVisibleTo` with the 2.0 framework assembly, so that the linq classes there don't leak out to user code?

CharliePoole:
Maybe... or maybe just make ExtensionsAttribute internal in NUnit.System.Linq?

rprouse:
> I wonder if nunit.linq should all be internal, and use InternalsVisibleTo with the 2.0 framework assembly, so that the linq classes there don't leak out to user code?

I believe that our intention is that people can use our LINQ implementation within their own tests. We could try making the extension attribute `InternalsVisibleTo`, but I expect that would cause the extensions to fail to compile? We should test it though.

CharliePoole:
I think the original intention was just to use it for ourselves but that I then advertised it as available in one of the release notes - IMO, not a feature we have to maintain.

Frankly, I could have done without NUnit.System.Linq, continuing to limit the framework code to what works under .NET 2.0. @oznetmaster was particularly wanting to have it and we all agreed to make the change. If it's seeming like it's more trouble than it's worth we should check to see how much reliance we actually have on it.

oznetmaster:
@rprouse was really in favor of having it available as well.

I have no problem with it being internal, since 2.0 tests should not use it anyway.

ChrisMaddock:
Yeah - I wouldn't like to loose it, I previously really missed being able to use LINQ.

> I believe that our intention is that people can use our LINQ implementation within their own tests.

This seems to be the question - let's decide if that is a 'feature'. Personally - I think it should be internal, providing "LINQ" for user use is out of scope for NUnit. It also introduces collision risk for users importing from elsewhere (although evidently not very common - given this is the first issue reported!)

CharliePoole:
I agree. Use of this by users was only mentioned in a release note IIRC and never documented in any other way.

ChrisMaddock:
I've created https://github.com/nunit/NUnit.System.Linq/issues/12 for this.

PR #155, opened 2016-12-28
Test parameter parsing out of framework

### LINKED ISSUE #23: In nunit3-console you cannot pass parameters containing ';' because they always get splitted
@SilverXXX commented on [Fri Jul 22 2016](https://github.com/nunit/nunit/issues/1694)

When running some tests with nunit3-console.exe, if you pass with --param there is no way to pass inside a value a ';'.

It seems from [CommandLineOptions.cs](https://github.com/nunit/nunit/blob/master/src/NUnitConsole/nunit3-console/CommandLineOptions.cs#L296) the you can't escape them in any way (it's not documented so i read the source directly).
A regex to capture only not escaped char would solve the problem (i'm not a regex expert so i cant suggest exact one)

---

@rprouse commented on [Fri Jul 22 2016](https://github.com/nunit/nunit/issues/1694#issuecomment-234521017)

See our docs, there is general information there on quoting for various operating systems. You can also have multiple --param in the command line, one per variable. 

I am closing this as a question, but feel free to add questions, we will see them and try to help. 

---

@ChrisMaddock commented on [Fri Jul 22 2016](https://github.com/nunit/nunit/issues/1694#issuecomment-234526296)

@rprouse - I think this is a potential 'issue' within NUnit, not command line usage? Looks like the options will always split on semi-colon, and then throw if it's not a valid `A=B` format afterwards? So I don't think there is any way pass in a test parameter with a semi-colon currently. (If that's to be a supported thing to do!)

---

@SilverXXX commented on [Fri Jul 22 2016](https://github.com/nunit/nunit/issues/1694#issuecomment-234528492)

Yes @ChrisMaddock, i would have tried to create a pull request to show a possible solution, it actually require only a one line change and a sensible regexp, but i'm quite bad at it (a connection string is a simple example of an impossible paramater to pass right now)
Edit: and since it's all made of key=value it doesnt even return an error

---

@rprouse commented on [Fri Jul 22 2016](https://github.com/nunit/nunit/issues/1694#issuecomment-234555217)

Sorry, reading issues on my phone, I missed the detail :frowning:

jnm2:
Fixes #23

### DIFF
--- a/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
+++ b/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
@@ -22,6 +22,7 @@
 // ***********************************************************************
 
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Xml;
 using NUnit.Common;
@@ -29,6 +30,7 @@ using NUnit.ConsoleRunner.Utilities;
 using NUnit.Engine;
 using NUnit.Engine.Extensibility;
 using System.Runtime.InteropServices;
+using System.Text;
 
 namespace NUnit.ConsoleRunner
 {
@@ -438,10 +440,30 @@ namespace NUnit.ConsoleRunner
             if (options.DefaultTestNamePattern != null)
                 package.AddSetting(FrameworkPackageSettings.DefaultTestNamePattern, options.DefaultTestNamePattern);
 
-            if (options.TestParameters != null)
-                package.AddSetting(FrameworkPackageSettings.TestParameters, options.TestParameters);
+            if (options.TestParameters.Count != 0)
+                AddTestParametersSetting(package, options.TestParameters);
 
             return package;
+        }
+
+        /// <summary>
+        /// Sets test parameters, handling backwards compatibility.
+        /// </summary>
+        private static void AddTestParametersSetting(TestPackage testPackage, IDictionary<string, string> testParameters)
+        {
+            testPackage.AddSetting(FrameworkPackageSettings.TestParametersDictionary, testParameters);
+
+            if (testParameters.Count != 0)
+            {
+                // This cannot be changed without breaking backwards compatibility with old frameworks.
+                // Reserializes the way old frameworks understand, even if this runner's parsing is changed.
+
+                var oldFrameworkSerializedParameters = new StringBuilder();
+                foreach (var parameter in testParameters)
+                    oldFrameworkSerializedParameters.Append(parameter.Key).Append('=').Append(parameter.Value).Append(';');
+
+                testPackage.AddSetting(FrameworkPackageSettings.TestParameters, oldFrameworkSerializedParameters.ToString(0, oldFrameworkSerializedParameters.Length - 1));
+            }
         }
 
         private TestFilter CreateTestFilter(ConsoleOptions options)
--- a/src/NUnitConsole/nunit3-console.tests/CommandLineTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/CommandLineTests.cs
@@ -586,7 +586,7 @@ namespace NUnit.ConsoleRunner.Tests
         {
             var options = new ConsoleOptions("--params=X=5");
             Assert.That(options.errorMessages, Is.Empty);
-            Assert.That(options.TestParameters, Is.EqualTo("X=5"));
+            Assert.That(options.TestParameters, Is.EqualTo(new Dictionary<string, string> { { "X", "5" } }));
         }
 
         [Test]
@@ -594,7 +594,7 @@ namespace NUnit.ConsoleRunner.Tests
         {
             var options = new ConsoleOptions("--params:X=5;Y=7");
             Assert.That(options.errorMessages, Is.Empty);
-            Assert.That(options.TestParameters, Is.EqualTo("X=5;Y=7"));
+            Assert.That(options.TestParameters, Is.EqualTo(new Dictionary<string, string> { { "X", "5" }, { "Y", "7" } }));
         }
 
         [Test]
@@ -602,7 +602,7 @@ namespace NUnit.ConsoleRunner.Tests
         {
             var options = new ConsoleOptions("-p:X=5", "-p:Y=7");
             Assert.That(options.errorMessages, Is.Empty);
-            Assert.That(options.TestParameters, Is.EqualTo("X=5;Y=7"));
+            Assert.That(options.TestParameters, Is.EqualTo(new Dictionary<string, string> { { "X", "5" }, { "Y", "7" } }));
         }
 
         [Test]
@@ -610,7 +610,7 @@ namespace NUnit.ConsoleRunner.Tests
         {
             var options = new ConsoleOptions("--params:X=5;Y=7", "-p:Z=3");
             Assert.That(options.errorMessages, Is.Empty);
-            Assert.That(options.TestParameters, Is.EqualTo("X=5;Y=7;Z=3"));
+            Assert.That(options.TestParameters, Is.EqualTo(new Dictionary<string, string> { { "X", "5" }, { "Y", "7" }, { "Z", "3" } }));
         }
 
         [Test]
--- a/src/NUnitConsole/nunit3-console/CommandLineOptions.cs
+++ b/src/NUnitConsole/nunit3-console/CommandLineOptions.cs
@@ -84,7 +84,8 @@ namespace NUnit.Common
         private List<string> testList = new List<string>();
         public IList<string> TestList { get { return testList; } }
 
-        public string TestParameters { get; private set; }
+        private Dictionary<string, string> testParameters = new Dictionary<string, string>();
+        public IDictionary<string, string> TestParameters { get { return testParameters; } }
 
         public string WhereClause { get; private set; }
         public bool WhereClauseSpecified { get { return WhereClause != null; } }
@@ -295,16 +296,22 @@ namespace NUnit.Common
                 {
                     string parameters = RequiredValue( v, "--params");
 
+                    // This can be changed without breaking backwards compatibility with frameworks.
                     foreach (string param in parameters.Split(new[] { ';' }))
                     {
-                        if (!param.Contains("="))
+                        int eq = param.IndexOf("=");
+                        if (eq == -1 || eq == param.Length - 1)
+                        {
                             ErrorMessages.Add("Invalid format for test parameter. Use NAME=VALUE.");
-                    }
+                        }
+                        else
+                        {
+                            string name = param.Substring(0, eq);
+                            string val = param.Substring(eq + 1);
 
-                    if (TestParameters == null)
-                        TestParameters = parameters;
-                    else
-                        TestParameters += ";" + parameters;
+                            TestParameters[name] = val;
+                        }
+                    }
                 });
 
             this.Add("timeout=", "Set timeout for each test case in {MILLISECONDS}.",
--- a/src/NUnitConsole/nunit3-console/FrameworkPackageSettings.cs
+++ b/src/NUnitConsole/nunit3-console/FrameworkPackageSettings.cs
@@ -106,9 +106,14 @@ namespace NUnit
         public const string DefaultTestNamePattern = "DefaultTestNamePattern";
 
         /// <summary>
-        /// Parameters to be passed on to the test
+        /// Parameters to be passed on to the tests, serialized to a single string which needs parsing. Obsoleted by <see cref="TestParametersDictionary"/>; kept for backward compatibility.
         /// </summary>
         public const string TestParameters = "TestParameters";
+
+        /// <summary>
+        /// Parameters to be passed on to the tests, already parsed into an IDictionary&lt;string, string>. Replaces <see cref="TestParameters"/>.
+        /// </summary>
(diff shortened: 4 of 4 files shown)


jnm2:
Not sure why the errors on Travis:

> System.Runtime.Remoting.RemotingException : Tcp transport error.

> System.Runtime.Remoting.RemotingException : Uri already in use: ProcessRunnerTests.

CharliePoole:
I deleted your "umbrella" comment in the description and added a "Fixes" reference to issue 23. We always want an issue in each project we are fixing, otherwise we have nothing to reference in the release notes. Note that the "Fixes #nnn" format is what causes GitHub to autoclose the issue when the PR is merged - which is what we want to happen 99% of the time.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -433,11 +435,51 @@ public static TestPackage MakeTestPackage(ConsoleOptions options)
                 package.AddSetting(FrameworkPackageSettings.DefaultTestNamePattern, options.DefaultTestNamePattern);
 
             if (options.TestParameters != null)
-                package.AddSetting(FrameworkPackageSettings.TestParameters, options.TestParameters);
+            {
+                var dictionary = new Dictionary<string, string>();
+
+                // Exactly what the framework used to do.
+                // Now can be independently changed without breaking backwards compatibility with old frameworks.
+                foreach (var param in options.TestParameters.Split(';'))
```
CharliePoole: Use a review comment (like this one) if you want to tell the reviewers why you did something. The comment in the code will only have meaning for us while this change is in progress but will just be confusing afterward. Anyway, we understand what you are doing. :-)

That said, It seems to me to make more sense for the ConsoleOptions class to create the dictionary in the first place, rather than making the runner do it. @oznetmaster What do you think?

CharliePoole: @jnm2 Do you intend to come back and add escaping at some point?

CharliePoole:
@jnm2 I restarted the failed job. Infrastructure failures on Travis are all too common. Generally a committer who has the authority will restart the job for you when it looks like that happened.

jnm2:
> Let's clarify what we are trying to do here - i.e. what about escaping semicolon?

All covered in the [umbrella issue](nunit/nunit#1941) :smile:

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
jnm2: [Conversation at https://github.com/nunit/nunit/issues/1941#issuecomment-269597215]

jnm2:
> I deleted your "umbrella" comment in the description and added a "Fixes" reference to issue 23.

> Let's clarify what we are trying to do here - i.e. what about escaping semicolon?

All covered in the [umbrella issue](https://github.com/nunit/nunit/issues/1941) which was its purpose :smile:

jnm2:
My gut instinct is that CommandLineOptions should be List<string>, one for each `--p` argument, for the dictionary to be generated later during the test generation. I don't have a strong feeling on this.

oznetmaster:
That is what I did in my CF implementation.

CharliePoole:
The thing is, "later" means a different place for each runner. If it's in CommandLineOptions (not ConsoleOptions as I first wrote) we can keep the runners closer, leading to an eventual client component. For a single runner at a time, I agree that the two approaches are fairly equal.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
jnm2: > Use a review comment (like this one) if you want to tell the reviewers why you did something.

Thanks for the suggestion. I had this in mind when I did it, but future bug fix es will involve seeing the code comments and will not involve seeing these comments. It was important to me that those commends stay with the code.

jnm2:
> If it's in CommandLineOptions (not ConsoleOptions as I first wrote) we can keep the runners closer, leading to an eventual client component. 

I like this eventual goal. So rather than:

a) both runners getting a CommandLineOptions with `IList<string> TestParameters`, and both runners translating `IList<string>` to `IDictionary<string, string>`  at the point where they translate the options to the package or translate the options to the test suite

We will have:

b) both runners translating `IList<string>` to `IDictionary<string, string>`  at the point where they set up CommandLineOptions, and both runners assigning the dictionary directly at the point where they translate the options to the package or translate the options to the test suite

Does b) get us any closer to that goal than a)?
If not, maybe we should consider that CommandLineOptions has been simply a bunch of strings gathered from the command line without semantic knowledge of what the runner will do with it.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
CharliePoole: @jnm2 In that case, change the comment to make sense long after this issue has merged. It's only at the present time that it's relevant to know what the framework used to do. If the code is not sufficiently self-explanatory, explain why it does what it does. The comment is also misleading in that the framework is going to continue splitting on unescaped semicolons for the foreseeable future.

CharliePoole:
Regarding your (a) and (b) options - in fact, I think we want both those things done in the options class, not in the runner. Right now, the little that we do (combining multiple items to a single string) is done in that class so it makes sense to do the dictionary in the same place. Options should expose both. The runner adds the knowledge that a certain option maps to a certain TestPackage setting.

jnm2:
> Regarding your (a) and (b) options - in fact, I think we want both those things done in the options class, not in the runner.

Not quite following. There is one CommandLineOptions class for every runner and they are not shared. Indeed I think that this is a good situation because it decouples each runner's CLI style from the others.

jnm2:
I think I'm following now. You're saying the CommandLineOptions doesn't serve much purpose so it may as well do semantic parsing.

Don't you think that a better scenario is for CommandLineOptions to be a dumb DTO? It's a great API. I think we should even go the opposite direction with it that you're suggesting and make `TestParameters` _less_ parsed- becoming an `IList<string>`.
Then something with a better name would be responsible for semantically understanding what that means in terms of `IDictionary<string, string>` and any other properties that need semantic translation.

I don't see how this holds us back from keeping the runners close. It's not runners should ever be coupled by sharing the same CommandLineOptions class definition.

oznetmaster:
Actually, they used to, and probably should again (sharing the same CommandLineOptions class definition).

jnm2:
@oznetmaster, this is from @CharliePoole https://github.com/nunit/nunit/pull/1942#issuecomment-269833389:

> OTOH, changing the same thing for the console crosses an arbitrary administrative boundary we have created... between the team of people who work on the framework and the team that works on the engine and console. So it needs to be two issues. If too many things cross that boundary and cause us problems, it's in the realm of the possible that we could re-examine the decision. But for now, it sticks.
>
> For a newcomer, this is probably hard to see because we created the boundary but for the moment the same people are working on both sides of it. We hope that won't last and that more new people will come to each project, allowing them to move ahead independently. IOW it's a bet we are making.

For that type of development, I'd say the less coupled the better.

jnm2:
@oznetmaster 
> Actually, they used to, and probably should again (sharing the same CommandLineOptions class definition).

This is from @CharliePoole https://github.com/nunit/nunit/pull/1942#issuecomment-269833389:

> OTOH, changing the same thing for the console crosses an arbitrary administrative boundary we have created... between the team of people who work on the framework and the team that works on the engine and console. So it needs to be two issues. If too many things cross that boundary and cause us problems, it's in the realm of the possible that we could re-examine the decision. But for now, it sticks.
>
> For a newcomer, this is probably hard to see because we created the boundary but for the moment the same people are working on both sides of it. We hope that won't last and that more new people will come to each project, allowing them to move ahead independently. IOW it's a bet we are making.

For that type of development, I'd say the less coupled the better. It feels like we're in a good place.

CharliePoole:
@jnm2 I think we are miscommunicating. You're referring to the CommandLineOptions class, which is the common class used in both nunitlite and nunit3-console. It derives from the Mono Options class and is inherited by the individual classes in each runner, NUnitLiteOptions and ConsoleOptions.

OTOH I'm talking about the entire hierarchy, which I plan to replace by an NUnitOptions hierarchy. This will get rid of Mono Options, which has things in it we don't need and is missing some things we need - in particular grouping of options in categories for the purpose of the help display.

The options class at the top of the hierarchy is supposed to hand the runner objects of the necessary Type to be used in communicating with the engine. The reason there is no capability in the hierarchy for handling a dictionary is that we never had one before.

@oznetmaster I agree that the commonality is still needed. We originally used simple source code sharing when, as you know,  the two were in the same repository. But that was broken by the decision to separate the framework from the console. We were left with two choices: source-code sharing via a Git submodule and binary sharing via a separate assembly. We spent a fair amount of time debating the two and ended up doing neither.

In the interim, we have tried to maintain commonality by manually keeping the code in both copies of CommandLineOptions in sync. I'm not sure we really succeeded - I haven't compared the two for a while. I am planning to do something using the separate assembly approach at some point in the new year.

jnm2:
@CharliePoole so I should I go ahead and move the parsing code to `NUnitLiteOptions` and `ConsoleOptions`?

CharliePoole:
Since it's a common option, it should go into the base class, CommandLineOptions in both cases. If you do it in the framework project first, we may be able to get it into the release this week. The console release will be next week.

jnm2:
I finished this in all three console runners.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -438,12 +440,32 @@ public static TestPackage MakeTestPackage(ConsoleOptions options)
...
+            if (testParameters.Count != 0)
+            {
+                // This cannot be changed without breaking backwards compatibility with old frameworks.
+                // Reserializes the way old frameworks understand, even if this runner's parsing is changed.
+
+                var oldFrameworkSerializedParameters = new StringBuilder();
+                foreach (var parameter in testParameters)
+                    oldFrameworkSerializedParameters.Append(parameter.Key).Append('=').Append(parameter.Value).Append(';');
+
+                testPackage.AddSetting(FrameworkPackageSettings.TestParameters, oldFrameworkSerializedParameters.ToString(0, oldFrameworkSerializedParameters.Length - 1));
```
CharliePoole: This potentially changes the order in which the parameters appear in the string, since the dictionary is free to return the pairs in any order. The original implementation created the string in the same order as they appeared on the command-line. This may or may not be an issue.

CharliePoole: If we want to preserve order, then the CommandLineOptions class has to provide both representations. Do we?

jnm2: I don't think we do.

[@CharliePoole:](https://github.com/nunit/nunit/issues/1885#issuecomment-268686434)

> That said, I don't particularly want to guarantee order among the differently named parameters. It makes no sense to me. I do understand the desire for a particular parameter name to reference an array of values in a certain order, but that's another (future) thing.
>
> Unless you have a good reason of your own to use IEnumerable<KeyValuePair<string, string>>, I'd go with <IDictionary<string, string>> for the test parameters. That's what they end up as in TestContext anyway.

I agree with that quote. I feel somewhat strongly that the parameter names should be the unit of isolation between tests. If you need ordering, it means you're actually taking what should be a single parameter and spreading it across multiple parameters.

oznetmaster:
It is one of the issues I raised about the use of a Dictionary vs a List, the loss of the order information.

I use it in the CF build, and it is significant there.

CharliePoole:
@jnm2 I agree with you and also with myself in the earlier quote!

However, I'm not asking about the forward-looking solution but the backward compatible solution. The string representation we are passing in 3.5 does guarantee order - although this has never been stated.

@oznetmaster Is your CF build still based on our framework? Will our decision here impact you?

oznetmaster:
@CharliePoole  Not sure what you mean.  I can turn off one conditional and compile the framework exactly as it is released.  So I guess the answer is yes.

oznetmaster:
@jnm2 Yes, that is exactly what I am doing.  I am using the order of the parameters, when they have the same name, to define the order of the values in the IEnumerable representation of that parameter.

CharliePoole:
@jnm2 Based on what @oznetmaster is saying, I don't think we want to go out of our way to break his code. 😉 What if we have CommandLineOptions manage two separate properties from which the two different package settings are created?

oznetmaster:
As I have previously indicated, I already "broke" compatibility when I changed CommandLineOptions to pass an IList instead of a string.  So I am not sure if this is an issue any more or not :(

jnm2:
If that's helpful for @oznetmaster then absolutely. NUnitLite has already been released this way. Since you have a fork @oznetmaster, is it difficult for you to set up the conditional to preserve order yourself so that this repo can do the cleanest thing for us and yours can do the cleanest thing for you? If it's not a ton of overhead, since it seems you're doing this kind of thing already, that would be great since it separates our two concerns.

CharliePoole:
@jnm2 I merged this already based on @oznetmaster 's response. AFAIK he doesn't use this repo.

jnm2:
Sounds good.
(I'll have to refresh my phone's web page if I leave it open overnight -_- it only loaded three new comments. Sorry guys!)

PR #157, opened 2017-01-06
Fix Detection of invalid framework when --inprocess

ChrisMaddock:
When running with an invalid framework and the `--inprocess` flag - this wouldn't be detected, and the tests would be run in the current framework. This was caused by the engine checking for the (obsoleted) ProcessModel "single", whilst the console converts all uses of "single" to "inprocess".

I searched for other uses at the same time, and made one further change to check for both inprocess AND single. (As the current conversion is done in the runner, I assume a different (old) runner could still be using the obsoleted "single" option, so the engine should check for both.)

There's no test, as the check currently relies on the static `CurrentFramework` property - so faking it all out was a little non-trivial. Could revisit that, if people think it's worthwhile? The other option would be to instead compile the engine tests for .NET 3.5, so we have an actual 'invalid framework' to use.

### DIFF
--- a/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
+++ b/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
@@ -229,8 +229,9 @@ namespace NUnit.Engine.Runners
             // be used to determine how to run the assembly.
             _runtimeService.SelectRuntimeFramework(TestPackage);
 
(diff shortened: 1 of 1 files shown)


CharliePoole:
This makes sense. It goes back to the time when we used an enum for ProcessModel in the runner. I stopped doing that at some point, but it was protecting us from any problem like this.

PR #158, opened 2017-01-09
Typo fixes

jnm2:
Same types of changes as https://github.com/nunit/nunit/pull/1960 with one important exception. The default value of test package setting `PrincipalPolicy` is changing from `"UnauthenticatedPricipal"` to  `"UnauthenticatedPrincipal"`.

/cc @CharliePoole @ChrisMaddock

### DIFF
--- a/src/NUnitConsole/nunit3-console.tests/CommandLineTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/CommandLineTests.cs
@@ -158,7 +158,7 @@ namespace NUnit.ConsoleRunner.Tests
         }
 
         [Test]
-        public void CanRegognizeInProcessOption()
+        public void CanRecognizeInProcessOption()
         {
             ConsoleOptions options = new ConsoleOptions("--inprocess");
             Assert.True(options.Validate(), "Should be valid: --inprocess");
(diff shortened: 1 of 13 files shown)


ChrisMaddock:
@jnm2 - What's tool are you using, and do you think we could hook it in to the CI, for pull requests? Or are the results requiring some manual modification?

jnm2:
> What tool are you using, and do you think we could hook it in to the CI, for pull requests? Or are the results requiring some manual modification?

That would be great, but what I'm doing won't work with that. I installed the ReSpeller plugin for ReSharper, and running ReSharper's "Find all issues in solution" with each of the three types of typo (comment, identifier, and string literal).

There are plenty of false positives, mostly due to notusingcamelcase. And I'm fixing each one by hand because it's just as fast as going through the quick actions menu, so I don't know how accurate the auto fixes are. They could be accurate.

ChrisMaddock:
That's a shame - yeah, I imagine something automated might be so error-prone it would become an irritation. I don't think there's a way a CI hook can provide a warning status! Nevermind, just a thought!

jnm2:
It's a really good concept though, as long as you can whitelist things. Maybe create a Cake plugin :-D

CharliePoole:
Looks good. I restarted the failing job, which appears to be a travis infrastructure problem.

PR #161, opened 2017-01-12
Terminate agent if main process has terminated

### LINKED ISSUE #128: When running under teamcity, agent processes are sometimes left running after nunit-console has exited
We are finding that nunit agent processes are sometimes left running forever after a CI build has completed. This is serious issue for a CI server as once an agent is in such a state, build after build can fail and manual intervention is required to recover. 

When logging into the build machine, I typically observe several agent processes consuming 0 CPU and they appear to have unloaded all tests assemblies, leaving the directory and nunit assemblies locked. In this state, the CI system is then blocked from cleaning up the working directory.

This seemed to start happening in nunit 3.4 and is still an issue with nunit 3.5.0

Our initial investigation suggests that the user-invoked cancellation of CI builds might be the source of the issue.  Cancellation will appear to be successful and the error will typically manifest later when the agent picks up another build.

CharliePoole:
Fixes #128 

My preference would have been to fix this by monitoring the remoting interface to the TestAgency but that gets very involved because of timeouts and because the agent doesn't actually intercept the test events that are sent back to the main process. Instead, I'm checking the process itself, which is much quicker although it won't be a solution if we start running agents on other machines. 

I can see that we will need to revamp the communications layer, which I wrote many years ago as my first toe-dipping into the TCP swamp. 😄 For now, this should solve the problem. I've tested by running  manually and canceling the console process. No agent processes remain according to Task Manager.

### DIFF
--- a/src/NUnitEngine/nunit-agent/Program.cs
+++ b/src/NUnitEngine/nunit-agent/Program.cs
@@ -41,7 +41,9 @@ namespace NUnit.Agent
 
         static Guid AgentId;
         static string AgencyUrl;
+        static Process AgencyProcess;
         static ITestAgency Agency;
+        static RemoteTestAgent Agent;
 
         /// <summary>
         /// Channel used for communications with the agency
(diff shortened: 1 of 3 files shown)


rprouse:
Pretty simple solution 👍 

Looks good to merge once CI passes. I will include this in the release which I hope to be able to do during the day tomorrow.

Issue #162, opened 2017-01-13
Add namespace keyword to Test Selection Language


tom-dudley:
Suppose I currently have some tests sitting under `My.Namespace` and some other tests sitting under `My.Namespace.SubNamespace`, I can't identify an easy way to select just the former as using `--where "class =~ My.Namespace"` will result in running all of the tests.

What I suggest is that the keywords `namespace` is added to the Test Selection Language allowing such filtering, such as `--where "namespace == My.Namespace"`

CharliePoole:
Your example can be run a number of ways. The two simplest would be:

```
--where "class =~ My.Namespace and class !~ My.Namespace.SubNamespace"
--where "test == My.Namespace and test != My.Namespace.SubNamespace"
```

Tests are treated as nesting, which is why it's possible to use the equality and inequality operators in the second example as opposed to a regular expression.

Adding namespace is indeed a possible extension and could allow a slightly shorter expression. However, I'm afraid that some users would expect `My.Namespace` to include `My.Namespace.SubNamespace`, leading to more confusion. I'm inclined to make this a "won't fix" item. What do other @nunit/core-team and @nunit/framework-team members think?

jnm2:
If I have five subnamespaces and want to execute only the tests in the outer namespace, I'd like to be able to match an exact namespace.

CharliePoole:
@jnm2 OK... that's a point to consider.
.
Of course, there are workarounds: (1) stop organizing tests that way or (2) use categories. The latter is the classic NUnit answer of course.

Still it would be easy to add this. Again, my main concern is user confusion. I don't think it's obvious that
`namespace == My.Namespace` doesn't include `My.Namespace.SubNamespace`. The docs will explain it of course but even so I wonder if we would be confusing users.

jnm2:
> or (2) use categories. The latter is the classic NUnit answer of course.

Which is what I do, but is less ad-hoc.

> The docs will explain it of course but even so I wonder if we would be confusing users.

The users should think of it as the entire namespace, the same as if they wrote `typeof(x).Namespace ==`. I guess it'll be a learning thing. Unless you like "EntireNamespace ==`, but at some level it seems redundant.

CharliePoole:
I see what you're saying. When I'm looking at a single class, I understand "namespace" as you do. When I'm running tests from the command-line or gui, I think of it as being inclusive. The namespace code element stands for a test suite in the logical NUnit view of tests. We'll have to explain it carefully if we do it.

[...]

PR #165, opened 2017-01-19
Add support for namespace keyword/filter (#162)

tom-dudley:
Implementation of: https://github.com/nunit/nunit-console/issues/162

Added support for the namespace keyword / filter along with exception handling to allow backwards compatibility with versions of the framework which don't support the framework.

I'm not massively happy with parsing the exception message so am open to alternative ways to give some meaningful output without modifying the framework.

The relevant framework change is here: https://github.com/nunit/nunit/pull/1995


src/NUnitEngine/nunit.engine/Drivers/NUnit3FrameworkDriver.cs
```diff
@@ -120,7 +121,24 @@ public string Run(ITestEventListener listener, string filter)
...
-            CreateObject(RUN_ACTION, _frameworkController, filter, handler);
+
+            try
+            {
+                CreateObject(RUN_ACTION, _frameworkController, filter, handler);
+            }
+            catch (System.Reflection.TargetInvocationException ex)
+            {
+                string innerMessage = ex.InnerException.Message;
+                if (ex.InnerException is ArgumentException && innerMessage.StartsWith("Invalid filter element"))
```
rprouse: It looks like the `InnerException` contains enough information to help the user diagnose the problem, it is just the fact that a `TargetInvocationException` is obscuring the inner message. Instead of matching on the message, could we not instead just throw an `NUnitEngineException` with the message from any `InnerException`?

```C#
catch (TargetInvocationException ex)
{
    if (ex.InnerException != null)
    {
        throw new NUnitEngineException(ex.InnerException.Message);
    }
   throw;
}
```

CharliePoole: The driver currently knows nothing whatsoever about filters by design. In fact, it knows nothing about any of the content of the xml and exceptions that pass through it. That's what keeps the driver maintainable.

Additionally, consider that the engine may use some other driver, like the V2 driver. It won't understand the filter either but some kind of an error will propagate back to the engine and possibly to the runner.

tom-dudley:
I've simplified the error handling as @rprouse suggested. Not sure what I was thinking with that!

src/NUnitEngine/nunit.engine/Drivers/NUnit3FrameworkDriver.cs
```diff
@@ -120,7 +121,20 @@ public string Run(ITestEventListener listener, string filter)
...
+            }
+            catch (TargetInvocationException ex)
+            {
+                if (ex.InnerException != null)
+                {
+                    throw new NUnitEngineException(ex.InnerException.Message);
+                }
+
+                throw;
+            }
```
CharliePoole: What problem are you trying to solve here? Why is this better than letting the exception be handled higher in the chain of calls? The driver doesn't have any special knowledge to help it deal with this exception as far as I can see.

src/NUnitEngine/nunit.engine.tests/Drivers/NUnit3FrameworkDriverTests.cs
```diff
@@ -149,11 +147,23 @@ public void RunTestsAction_AfterLoad_ReturnsRunnableSuite()
...
         }
+
+        [Test]
+        public void RunTestsAction_WithInvalidFilterElement_ThrowsNUnitEngineException()
+        {
+            _driver.Load(_mockAssemblyPath, _settings);
+
+            var invalidFilter = "<filter><invalidElement>foo</invalidElement></filter>";
+            var ex = Assert.Catch(() => _driver.Run(new NullListener(), invalidFilter));
+            if (ex is TargetInvocationException)
```
CharliePoole: Are you actually observing TargetInvocationException here? It seems to me that would indicate something is leaking out of the framework that is not supposed to.

src/NUnitEngine/nunit.engine.tests/Drivers/NUnit3FrameworkDriverTests.cs
```diff
@@ -23,10 +23,10 @@
 
 using System;
 using System.Collections.Generic;
-using System.Web.UI;
+using System.Reflection;
 using System.Xml;
 using NUnit.Tests.Assemblies;
-using NUnit.Engine.Internal;
+using NUnit.Engine.Internal.Tests;
```
CharliePoole: Why this change? What are you referring to? The CI is failing because of an ambiguous reference to Assert, between NUnit.Framework.Assert and NUnit.Engine.Internal.Tests.Assert, which I can't see anywhere.

src/NUnitEngine/nunit.engine.tests/Drivers/NUnit3FrameworkDriverTests.cs
tom-dudley: Yes I'm observing TargetInvocationException from any exception thrown. For example [this exception](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Internal/TestFilter.cs#L131) in `Framework.Internal.TestFilter` also gives a TargetInvocationException. What's the behaviour you would expect here? I can provide the call stack paused at the point where it throws the framework exception if that might help.

src/NUnitEngine/nunit.engine.tests/Drivers/NUnit3FrameworkDriverTests.cs
tom-dudley: My mistake, I'll revert this.

src/NUnitEngine/nunit.engine.tests/Drivers/NUnit3FrameworkDriverTests.cs
CharliePoole: TargetInvocationException means we are returning from a call through reflection. I guess that's because we are in the constructor called by the framework driver through reflection. However, if NUnit3FrameworkDriver catches those and wraps the original as an NUnitEngineException with it's own message, we should no longer see the TargetInvocationException.

tom-dudley: There aren't any `try catch` blocks between the exception thrown in the Framework and the catching of the TargetInvocationException. [Here](https://github.com/nunit/nunit-console/blob/master/src/NUnitEngine/nunit.engine/Drivers/NUnit3FrameworkDriver.cs#L164-L170) is the relevant `NUnit3FrameworkDriver` code which has no error handling. Is this where you're expecting to see some?

I wonder if we should have something like the following. (Not sure if the null check is required)
```
        private object CreateObject(string typeName, params object[] args)
        {
            try
            {
                return _testDomain.CreateInstanceAndUnwrap(
                    NUNIT_FRAMEWORK, typeName, false, 0,
#if !NET_4_0
                    null, args, null, null, null);
#else
                null, args, null, null );
#endif
            }
            catch (TargetInvocationException ex)
            {
                if (ex.InnerException != null)
                {
                    throw ex.InnerException;
                }
            }
        }
```

CharliePoole: I'm confused. Isn't that the code you already added? And if the code in the driver is handling the TargetInvocationException, why are you testing for it in the test code? It shouldn't get through.

tom-dudley: Didn't realise I'd already added that. I see what you mean now. Regarding this actual handling - Is there a preference as to whether we just let the TargetInvocationException filter up from the driver, or would a catch and throw (of `NUnitEngineException`) be better? `Load` in `NUnit3FrameworkDriver` currently does the latter for a `SerializationException.`

CharliePoole: A really good question! Since the driver may be user-written or third-party, it could have any kinds of errors in it. If it does throw an unexpected exception, we should be wrapping that in an NUnitEngineException that basically says "the driver did something strange - here's the exception it through" Now that I think about it, we should probably wrap any exception from the driver __except__ an NUnitEngineException, although that doesn't have to be done for this PR unless you want to.

So, what about this case? The driver knows it's talking to the nunit 3 framework. The framework returns an exception, which will almost always be a TargetInvocationException, since we are using reflection. It doesn't know the cause of that exception - it could be any kind of error in the framework. So I guess it should just wrap the exception (inner exception in the case of TIE) in an NUnitEngineException. Does that make sense?

tom-dudley: Yep, that makes sense. I'll see about adding the exception wrapping to `DirectTestRunner`.

tom-dudley:
Any objection to me adding NSubstitute (or similar) to the engine tests project? If not I'll add a few tests for the DirectTestRunner.

Also, I don't think that Travis error was my fault, looks like a genuine myget timeout.

ChrisMaddock:
I restarted the CI for you.

NSubstitue is what we normally use for other NUnit projects - so sounds sensible to me. 🙂

jnm2:
NSubstitute is awesome.

src/NUnitEngine/nunit.engine/Runners/DirectTestRunner.cs
```diff
@@ -109,13 +124,27 @@ protected override TestEngineResult LoadPackage()
...
+        {
+            try
+            {
+                return driver.Load(testFile, subPackage.Settings);
+            }
+            catch (Exception ex)
+            {
+                if (ex is NUnitEngineException)
+                    throw;
+                throw new NUnitEngineException("An exception occurred in the driver while loading itself.", ex);
```
CharliePoole: "...while loading tests."

src/NUnitEngine/nunit.engine/Runners/DirectTestRunner.cs
```diff
@@ -71,7 +71,22 @@ public override TestEngineResult Explore(TestFilter filter)
...
+                string driverResult;
+
+                try
+                {
+                    driverResult = driver.Explore(filter.Text);
+                }
+                catch (Exception ex)
+                {
+                    if (ex is NUnitEngineException)
+                        throw;
```
jnm2: Should be: `catch (Exception ex) when (!(ex is NUnitEngineException))`

src/NUnitEngine/nunit.engine/Runners/DirectTestRunner.cs
```diff
@@ -109,13 +124,27 @@ protected override TestEngineResult LoadPackage()
...
+        private static string LoadDriver(IFrameworkDriver driver, string testFile, TestPackage subPackage)
+        {
+            try
+            {
+                return driver.Load(testFile, subPackage.Settings);
+            }
+            catch (Exception ex)
+            {
+                if (ex is NUnitEngineException)
+                    throw;
```
jnm2: Same [Should be: `catch (Exception ex) when (!(ex is NUnitEngineException))`]

src/NUnitEngine/nunit.engine/Runners/DirectTestRunner.cs
```diff
@@ -129,7 +158,18 @@ public override int CountTestCases(TestFilter filter)
...
-                count += driver.CountTestCases(filter.Text);
+            {
+                try
+                {
+                    count += driver.CountTestCases(filter.Text);
+                }
+                catch (Exception ex)
+                {
+                    if (ex is NUnitEngineException)
+                        throw;
```
jnm2: Same [Should be: `catch (Exception ex) when (!(ex is NUnitEngineException))`]

src/NUnitEngine/nunit.engine/Runners/DirectTestRunner.cs
```diff
@@ -151,7 +191,20 @@ protected override TestEngineResult RunTests(ITestEventListener listener, TestFi
...
+                string driverResult;
+
+                try
+                {
+                    driverResult = driver.Run(listener, filter.Text);
+                }
+                catch (Exception ex)
+                {
+                    if (ex is NUnitEngineException)
+                        throw;
```
jnm2: Same [Should be: `catch (Exception ex) when (!(ex is NUnitEngineException))`]

src/NUnitEngine/nunit.engine/Runners/DirectTestRunner.cs
```diff
@@ -174,8 +227,20 @@ protected override TestEngineResult RunTests(ITestEventListener listener, TestFi
...
+            {
+                try
+                {
+                    driver.StopRun(force);
+
+                }
+                catch (Exception ex)
+                {
+                    if (ex is NUnitEngineException)
+                        throw;
```
jnm2: Same [Should be: `catch (Exception ex) when (!(ex is NUnitEngineException))`]

src/NUnitEngine/nunit.engine/Runners/DirectTestRunner.cs
```diff
@@ -174,8 +227,20 @@ protected override TestEngineResult RunTests(ITestEventListener listener, TestFi
...
         public override void StopRun(bool force)
         {
-            foreach(var driver in _drivers)
-                driver.StopRun(force);
+            foreach (IFrameworkDriver driver in _drivers)
+            {
+                try
+                {
+                    driver.StopRun(force);
+
```
jnm2: Nit: inconsistent empty space

src/NUnitEngine/nunit.engine/Runners/DirectTestRunner.cs
CharliePoole: That would certainly be cleaner, assuming it works on the Mono builds as well. My guess is that it will - the language stuff is usually quite up to date.

tom-dudley: This is a new(ish) framework feature, right? (I hadn't heard of it, looks handy!) I thought NUnit supported back to 2.0 and given this is in the engine it might also be the case. Or is that not the case since this is the 'client side' of NUnit? I guess my question is: what's the minimum supported framework version for the engine? The project is currently set to target 2.0

jnm2: @tom-dudley Exception filters were supported in IL and VB.NET from day one, [back in 2001](https://www.ecma-international.org/publications/files/ECMA-ST-WITHDRAWN/ECMA-335,%201st%20edition,%20December%202001.pdf). I

The C# language access to this feature was added in version 6 in 2015.

jnm2: Incidentally, I was eight years old. O.o

tom-dudley:
OK, all done with this I think. As Charlie noted we should probably wait on nunit/nunit#1995 before merging.

CharliePoole:
This looks great! Just waiting for the CI to pass.

I'm happy you added some tests using mocks. We've been needing to do that for this project for a while.

CharliePoole:
No reason not to merge. My suggestion had to do with the "best" order of working on the problems. We have to support frameworks both with and without "namespace" now and in the future.

rprouse:
@tom-dudley you have been very active in the console project recently. Are you interested in joining the team as a contributor? If you are, please email me at rob@prouse.org so we can discuss your interests and I can explain what it means to be a contributor.

Thanks for your help. 👍

Issue #171, opened 2017-02-04
Socket Exception when stopping Remote Agent


eberlid:
This problem exists using NUnit 3.6.

The TCP channel may throw a Socket Exception (0x80004005) when calling `Stop` on the Agent. We cannot reproduce the problem deterministically, but we are able run tests so that the exception occurs with a high probability.

Our test toolchain consists of a python script which calls the `nunit3-console.exe` (I'll provide more information about the test setup on request). The problem exists _rarely_ when calling the console runner from python with these arguments
`nunit3-console.exe myTestAssembl.dll --work=[pathToWorkDir] --where=Category==Default`
The likelihood that the problem occurs can be _increased_ by creating Threads in Python which will call the console runners with different work directories for each Thread Runner in parallel. However, adding `--trace=Debug` reduces the likelihood that the problem occurs (also logging process activity with procmon reduces the frequency the problem occurs).

Inspecting log results and procmon logs yields that the Agent receives the `Stop` call and sets its `stopSignal`. Once the `stopSignal` is signaled, the Agent terminates successfully. However, the `ProcessRunner` ends up in exception state because the underlying TCP channel to the Agent throws a Socket Exception with error Code 0x80004005.

AFAIK the Process Runner wants to be sure that Agent Process still exists, i.e., it has not terminated due to some unexpected error, and therefore catches all exceptions when calling `Stop`. I changed the behavior in a way that the Process Runner 'pings' the Agent first to make sure the process is still alive (and by catching all exceptions) and then stopping the Agent (and ignoring SocketException). See this change in this commit: https://github.com/eberlid/nunit-console/commit/82685137f9363ffffedf218b94fa212f280e2b4b

(Please let me know if there is a more appropriate way to share this commit with you, rather than just copy paste the URL in here).

With this patch, the problem did not occur so far.

CharliePoole:
Best thing to do is to create a pull request, which allows us to comment more readily, even line-by-line.

It's interesting to think about multiple copies of the console runner executing. I would think that might cause a problem, since there would be multiple TestAgency servers set up. I'll have to look at the code to remind myself how their URIs are assigned. Normally, the engine assumes it is handling all the parallelism, not that it is done from outside.

We'll be happy to look at your PR in more detail. THanks for your work.

BlythMeister:
I have noticed also seeing this exception on occasion using 3.6 which we did not see using 3.5.
[...]

PR #175, opened 2017-02-04
Ping remote Agent before shutdown

eberlid:
Original issue #171 

Stopping the agent may cause a 'SocketException'.
Call 'Ping' before, in order to check that the
agent is still alive and waiting for its shutdown.

### DIFF
--- a/src/NUnitEngine/nunit.engine/Agents/RemoteTestAgent.cs
+++ b/src/NUnitEngine/nunit.engine/Agents/RemoteTestAgent.cs
@@ -105,6 +105,11 @@ namespace NUnit.Engine.Agents
             stopSignal.Set();
         }
 
+        public override void Ping()
+        {
+            log.Info("Ping");
+        }
+
         public bool WaitForStop(int timeout)
         {
             return stopSignal.WaitOne(timeout);
--- a/src/NUnitEngine/nunit.engine/Agents/TestAgent.cs
+++ b/src/NUnitEngine/nunit.engine/Agents/TestAgent.cs
@@ -102,6 +102,11 @@ namespace NUnit.Engine.Agents
         public abstract void Stop();
 
         /// <summary>
+        /// Pings the agent to check that it is still alive.
+        /// </summary>
+        public abstract void Ping();
+
+        /// <summary>
         ///  Creates a test runner
         /// </summary>
         public abstract ITestEngineRunner CreateRunner(TestPackage package);
--- a/src/NUnitEngine/nunit.engine/ITestAgent.cs
+++ b/src/NUnitEngine/nunit.engine/ITestAgent.cs
@@ -52,6 +52,11 @@ namespace NUnit.Engine
         void Stop();
 
         /// <summary>
+        /// Pings the agent to check that it is still alive.
+        /// </summary>
+        void Ping();
+
+        /// <summary>
         ///  Creates a test runner
         /// </summary>
         ITestEngineRunner CreateRunner(TestPackage package);
--- a/src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
+++ b/src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
@@ -22,6 +22,7 @@
 // ***********************************************************************
 
 using System;
+using System.Net.Sockets;
 using NUnit.Engine.Internal;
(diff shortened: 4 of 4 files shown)


BlythMeister:
i'm having some issues with the Nunit console and a socket error. It seems i'm not the only one,
What is the process to get these reviewed and included into a release?
i see aprox 10-20% of builds fail because nunit is returning a -100 error and that socket exception

rprouse:
Interesting workaround here. I wonder if there is a race condition in the agent where it shuts down before the Stop() call returns. Looking at that may be the more correct fix for this issue. This is more a workaround. @CharliePoole, you've looked at this more, thoughts?

ChrisMaddock:
I was about to comment the same as @rprouse - I'd be wary of just suppressing the exception, without fully understanding what's causing it.

Both @BlythMeister and @eberlid have encountered this while running multiple consoles in parallel, as opposed to letting the engine handle the parallelisation. As @CharliePoole wrote on #171 - that's uncommon, and could be bringing it's own issues. Given that two people have now encountered this error in this manner, it seems a likely cause - I don't have the time to dig into it right now, though. 😞 

@BlythMeister @eberlid - is there a particular reason you run multiple consoles in parallel, as opposed to passing multiple assemblies to a single console, and letting the console run them in parallel?

CharliePoole:
@ChrisMaddock I've seen people do this because they want separate output files. It's kind of funny considering how much effort we have put into merging the output!

That may be an option to add at some point.

If folksinger multiple consoles in parallel, I'd suggest using --inprocess if possible.

BlythMeister:
For me, i think its kind of historical as to why.
We are using FAKE which comes with this functionality.

We essentially call nunit console once per test DLL using fake for the parallel.
I could try and call nunit console once and pass all the dll setting workers to the level of parallel we are after.
@harrisonmeister you got any idea why that might not work?

harrisonmeister:
@blythmeister - yes there is a reason. We have more tests / characters per test DLL combined than a command line argument will allow. We hit the max a while back and had to split them out somehow. Single call to nunit console per test was what fake did and fixed the immediate issue

BlythMeister:
Well...There is the reason...Glad i didn't spend a day changing our code 😋
I'm interested in the inprocess you mention @charliepoole does this essentially run the test in console rather than create an agent?

BlythMeister:
@CharliePoole unfortunately we cannot use the "inprocess" flag, as we get this error:
"The --x86 and --inprocess options are incompatible" 

Therefore, i think we are going to need to keep using the agent process, but therefore still have this issue with the socket error.

BlythMeister:
@rprouse I'm not saying this is the solution to the problem either, if there is another way to prevent the socket exception and therefore error we see i'm all for it! :)

ChrisMaddock:
>  We have more tests / characters per test DLL combined than a command line argument will allow.

Have you tried the `file` option to get around this, which would allow you to specify your command line in a file?

Regardless, I think it would be nice if running multiple consoles in parallel _did_ work. Thinking it through - the same could happen to us. We have a single CI server running many builds - if two of those happened to hit the NUnit stage at the same time, this could be (a lesser version!) of the same issue. As the builds are independent, passing the assemblies in one go isn't an option. I'm not sure this is the solution we want to put in for this problem however - although until anyone has any time to work on it, maybe it's all we've got.

eberlid:
@ChrisMaddock we are running multiple TeamCity agents on one single machine. It happens that more than one agent is executing tests at the same time.

ChrisMaddock:
@eberlid @BlythMeister @harrisonmeister - would be great if you guys could try out @jnm2's fix in #223, which should handle this issue more gracefully, and check it still solves the problem for your cases.

You can pull the NuGet package from the Appveyor artefacts.
https://ci.appveyor.com/project/CharliePoole/nunit-console/build/3.7.0-ci-03635-pr-223/artifacts

BlythMeister:
We now run using nunit to handle the parallel, so don't have a setup to test the change in afraid.

eberlid:
I cherry picked the commits of #223 onto version 3.6.1 and awaiting feedback of our test systems.

jnm2:
I'm gonna go ahead and close this since I'm closing #171. Thanks for this; it is appreciated. If anything further turns up please let us know!

PR #177, opened 2017-02-09
Default to multiple agent process when using a single nunit project file

### LINKED ISSUE #116: NUnit 3.5.0 defaults to single agent process when using an nunit project file
NUnit 3.5.0 defaults to single agent process when using an nunit project file. This is an unexpected change in behaviour from 3.4.1.

Reproduction:

> nunit3-console.exe input.nunit

input.nunit:

```
 <NUnitProject>
 <Settings activeconfig="active" appbase="D:\path"/>
 <Config name="active">
 <assembly path="A.Tests.dll" />
 <assembly path="B.Tests.dll"/>
 </Config>
 </NUnitProject>
```

Result;

When 3.4.1 reads an .nunit config file, it defaults to multiple processes.
When 3.5.0 reads an .nunit config file it defaults to single process.
When either 3.4.1 or 3.5.0 are passed a list of assemblies at the command line, both default to multiple processes.

Expected behaviour:

Default to multiple processes when using a project file.

Workaround:

Add --process=Multiple

jnm2:
Fixes #116

@CharliePoole:

> When neither the project file nor the command line specifies running in multiple processes, we should either default to multiple process or base the decision on whether the project file contains more than one assembly.

What I did is default to multiple process.

There is an issue where a project file that specifies the single process model gets a multiple process runner.
There will have to be some parsing of the project file.

Questions:

 * What does `AggregatingTestRunner` do? Is it a way to defer the process model decision? Should these tests that depend on parsing the project all be returning `AggregatingTestRunner` instead of `MultipleTestProcessRunner`?
 * How do I actually write a tests covering the end-to-end scenario in #116? Is testing the runner type good enough? (Probably not, if `AggregatingTestRunner` defers to other runners?)

### DIFF
--- a/src/NUnitEngine/nunit.engine/Services/DefaultTestRunnerFactory.cs
+++ b/src/NUnitEngine/nunit.engine/Services/DefaultTestRunnerFactory.cs
@@ -93,7 +93,7 @@ namespace NUnit.Engine.Services
                 case ProcessModel.Default:
                     if (projectCount > 0)
                         return new AggregatingTestRunner(ServiceContext, package);
-                    else if (package.SubPackages.Count > 1 || projectCount > 0)
+                    else if (package.SubPackages.Count > 1)
                         return new MultipleTestProcessRunner(this.ServiceContext, package);
                     else
                         return new ProcessRunner(this.ServiceContext, package);
--- a/src/NUnitEngine/nunit.engine/Services/DefaultTestRunnerFactory.cs
+++ b/src/NUnitEngine/nunit.engine/Services/DefaultTestRunnerFactory.cs
@@ -79,15 +79,11 @@ namespace NUnit.Engine.Services
             // If we have multiple projects or a project plus assemblies
             // then defer to the AggregatingTestRunner, which will make
             // the decision on a file by file basis so that each project
-            // runs with its own settings.
+            // runs with its own settings. Note that bad extensions are
+            // ignored rather than assumed to be projects. This doesn't
+            // really matter since they will result in an error anyway.
             if (projectCount > 1 || projectCount > 0 && assemblyCount > 0)
                 return new AggregatingTestRunner(ServiceContext, package);
-
-            // If we have a single project by itself, make it the top level project.
-            if (projectCount > 0 && assemblyCount == 0)
-                package = package.SubPackages[0];
-
-            // TODO: What about bad extensions?
 
             ProcessModel processModel = GetTargetProcessModel(package);
 
@@ -95,7 +91,9 @@ namespace NUnit.Engine.Services
             {
                 default:
                 case ProcessModel.Default:
-                    if (package.SubPackages.Count > 1)
+                    if (projectCount > 0)
+                        return new AggregatingTestRunner(ServiceContext, package);
+                    else if (package.SubPackages.Count > 1 || projectCount > 0)
                         return new MultipleTestProcessRunner(this.ServiceContext, package);
                     else
                         return new ProcessRunner(this.ServiceContext, package);
--- a/src/NUnitEngine/nunit.engine.tests/Services/DefaultTestRunnerFactoryTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Services/DefaultTestRunnerFactoryTests.cs
@@ -52,15 +52,15 @@ namespace NUnit.Engine.Services.Tests
         {
             Assert.That(_factory.Status, Is.EqualTo(ServiceStatus.Started));
         }
-
+        
         // Single file
-        [TestCase("EngineTests.nunit", null,        typeof(ProcessRunner))] // Needs to exist because contents are checked
+        [TestCase("x.nunit",           null,        typeof(AggregatingTestRunner))]
         [TestCase("x.dll",             null,        typeof(ProcessRunner))]
-        [TestCase("EngineTests.nunit", "Single",    typeof(TestDomainRunner))]
+        [TestCase("x.nunit",           "Single",    typeof(TestDomainRunner))]
         [TestCase("x.dll",             "Single",    typeof(TestDomainRunner))]
-        [TestCase("EngineTests.nunit", "Separate",  typeof(ProcessRunner))]
+        [TestCase("x.nunit",           "Separate",  typeof(ProcessRunner))]
         [TestCase("x.dll",             "Separate",  typeof(ProcessRunner))]
-        [TestCase("EngineTests.nunit", "Multiple",  typeof(MultipleTestProcessRunner))]
+        [TestCase("x.nunit",           "Multiple",  typeof(MultipleTestProcessRunner))]
         [TestCase("x.dll",             "Multiple",  typeof(MultipleTestProcessRunner))]
         // Two files
         [TestCase("x.nunit y.nunit",   null,        typeof(AggregatingTestRunner))]
@@ -89,9 +89,6 @@ namespace NUnit.Engine.Services.Tests
         [TestCase("x.dll y.dll z.dll",     "Multiple", typeof(MultipleTestProcessRunner))]
         public void CorrectRunnerIsUsed(string files, string processModel, Type expectedType)
         {
-            if (files == "EngineTests.nunit")
-                files = Path.Combine(TestContext.CurrentContext.TestDirectory, files);
-
             var package = new TestPackage(files.Split(new char[] { ' ' }));
             if (processModel != null)
                 package.AddSetting("ProcessModel", processModel);


CharliePoole:
@jnm2 IF you're working on #116, please self-assign it so we know and nobody else starts it. Also, put a line "Fixes #116" in the description of the PR and give the PR a descriptive title that doesn't include the issue number.

CharliePoole:
We already default to multiple process runner if there is more than one assembly.

The problem is that expanding an nunit project is an extension and not part of the basic engine itself. Much as I hate to say it, we may need to create a mock extension of some kind in order to test this.

AggregatingTestRunner has two roles. It serves as the base class for MultipleProcessTestRunner and MultipleDomainTestRUnner. Used by itself, it serves to defer the decision of the model for each of the items entered on the command-line. THis may be what we need to do for all projects, since two projects on the command-line might want to use different settings for process model or domain usage.

However, I think the key issue here is to figure out if we can get projects expanded __before__ we decide what runners will ultimately be used.

Yes, we do need a higher level of testing - something we have been saying for a long time!

You've bitten off a fairly difficult piece of work here. 😄

jnm2:
@CharliePoole where is the NUnit project extension defined?

CharliePoole:
It's in the nunit-project-loader repo. However, you shouldn't need to make any changes there to resolve this issue. 

If you are looking to mock it or otherwise create a fake project loader extension, you should be looking at the interface, which is defined in the nunit.engine.api assembly. It's `IProjectLoader` and is called by the `ProjectService` in the engine.

The engine already uses some fake extensions in it's testing, so you could definitely add one for these tests. I would use an extension that is not likely to be used, like `.fakeproj` and I would not store them on disk at all, but just maintain a set of fake `IProjects` internally, to be returned based on the name that is passed in.

However, I would say that this might be an "extra" on top of actually solving the problem that was reported.

CharliePoole:
@jnm2 Once again, the line comments I had made earlier became pending and I had to submit a "comment" review in order to make them visible to you. I wonder if this could be a bug in GitHub (or a design) due to the fact that you specifically requested my review? @rprouse do you know?

CharliePoole:
Two travis jobs failed in different places, but both due to some error being thrown that caused the agent to terminate. I restarted both of them.

jnm2:
@CharliePoole I got the fake project loader working.  ProjectService itself requires the `TestPackage.Name` to be a file that exists, so a dummy file will have to be created. I was just piggybacking of the existing `.nunit` file for a dummy name and specifying a different IProject each time.

Almost finished the test. This is probably as far as I'll get today.

src/NUnitEngine/nunit.engine/Services/DefaultTestRunnerFactory.cs
```diff
@@ -76,11 +76,13 @@ public override ITestEngineRunner MakeTestRunner(TestPackage package)
...
             }
 
+            ProcessModel processModel = GetTargetProcessModel(package);
+
             // If we have multiple projects or a project plus assemblies
             // then defer to the AggregatingTestRunner, which will make
             // the decision on a file by file basis so that each project
             // runs with its own settings.
-            if (projectCount > 1 || projectCount > 0 && assemblyCount > 0)
+            if (projectCount > 1 || (projectCount > 0 && assemblyCount > 0) || (projectCount > 0 && processModel == ProcessModel.Default))
```
CharliePoole: The problem I  see is that when processmodel is not specified we actually don't know what runner we want to use for a project without looking inside it. That gives us two options:

1) Figure out a way to call the project service early enough to expand the project, so we can use that information to decide on the runner.
2) Assume that the AggregatingTestRunner will be able to figure it out, as you have done.

Can we take a look at whether option 1 is a possibility?

jnm2: @CharliePoole option 1 feels cleaner and I do want to take a look at it. Before anything else though I want the integration tests in place which are close to finished. That will give us feedback for this decision. I'm pretty sure option 2 will be the less disruptive option as far as fixing #116 quickly, assuming the integration tests pass this way. A followup PR could then pursue option 1.

CharliePoole: @jnm2 Could you clarify what you mean by integration tests? We have had a long-standing discussion about creating integration tests separate from the unit tests we already have, possibly in a separate repository. I don't think that should be done as part of a bug fix like this, but as a separate thing, because it's a pretty large project. OTOH, some tests that are slightly more integrated than what we have already, run as a part of the unit tests, would fit in well here.

jnm2: > OTOH, some tests that are slightly more integrated than what we have already, run as a part of the unit tests, would fit in well here.

Okay, my terminology may be off. It's unit testing but it's testing more than one layer at a time.

CharliePoole:
@jnm2 Cool. I'll wait to see what you come up with.

jnm2:
@CharliePoole btw my "bandwidth" varies. If I'm ever in your road and you just want to get a hotfix out I'll never be offended if you take over the work. =D

CharliePoole:
That's not an issue here. I just want to nudge you onto paths were we are most likely to merge what you come up with. So, for example, we wouldn't want the unit tests for the engine to required integrating a real extension, but some future, separate integration tests might actually do that.

jnm2:
I'm backing away from the multi-system unit test because it's requiring too much engine refactoring to use interfaces where it currently depends on concrete implementations (TestAgency, specific runners). Instead I'll do a multi-level test what checks the type of the main package as well as each subpackage.

CharliePoole:
That sounds sensible. We should keep your comments in mind as other work goes on, so that interfaces may be introduced gradually where needed.

jnm2:
New discovery: AggregatingTestRunner cannot be used for project packages because project packages have 0 subpackages and therefore AggregatingTestRunner will have 0 Runners and running tests will be a no-op.

Either that, or the .nunit project service *does* create a subpackage under the .nunit project package, which your current tests aren't doing.

jnm2:
Alright, there is one subproject per assembly in the project. Easy enough. https://github.com/nunit/nunit-project-loader/blob/master/src/extension/NUnitProject.cs#L102

jnm2:
I'm sorry this is taking such a ridiculously long time.

CharliePoole:
@jnm2 When I asked you the status of #116, it was just because I saw it listed. I forgot that you had a PR on it a while back - this one! Anyway, I pulled down your branch, rebased on master and went from there. The new commit is very much the same as yours, with a few changes.

For one thing, the comment in `DefaultTestRunnerFactoryTests` saying that the nunit files must exist because they are examined has been wrong for a long time - ever since we separated the extensions from the engine build, in fact. I removed the comment and the extra files you created.

I kept your logic, just reorganizing it a bit. It's a bit sit simpler as well since I removed some unneeded code. At the same time, while testing, I noticed issue #350. This PR fixes #350 as well.

src/NUnitConsole/nunit3-console/ResultReporter.cs
```diff
@@ -191,7 +191,7 @@ private void WriteErrorsFailuresAndWarnings(XmlNode resultNode)
                         else
                         {
                             var site = resultNode.GetAttribute("site");
-                            if (site == "SetUp" || site == "TearDown")
+                            if (site != "Parent" && site != "Child")
```
jnm2: Is it possible to protect this with a test?

mikkelbu: Actually, I think that this is wrong 😄 as the `site` is `null` for warnings, hence we will write one warning for every step down to the actual warning.

I - more or less - did the opposite change, so we did not have to introduce a null check (also to align the code with NUnitLite). See https://github.com/nunit/nunit-console/commit/d5822fec5af7633a4f1c25caf9585e8d8c3fcbc8#diff-16ed6eb93d79ef9b2ca6bd51950d69ba

Also see here for the discussion #326.

CharliePoole:
@jnm2 Test has been added, please re-review!

As you might expect the test was a __lot__ harder than the code change. Teaches me to do two things in one issue, which I usually don't like when others do it.

The code to create a secondary `MasterTestRunner` is pretty tricky. We may want to encapsulate it somewhere, sometime, but not now please! Also, it __could__ use fake services in this case, since the files don't actually get loaded. But creating all those fakes could also be an exercise for the future.

The technique used here could also be ported to the ResultReportTests, which use the framework and only simulate the action of the engine. When I wrote that, I hadn't thought of the approach used here.

src/NUnitConsole/nunit3-console.tests/BadFileTests.cs
```diff
@@ -0,0 +1,73 @@
...
+            reporter.WriteErrorsFailuresAndWarningsReport();
+            var report = sb.ToString();
+
+            Assert.That(report, Contains.Substring($"1) Invalid : {fullname}"));
+            Assert.That(report, Contains.Substring(message));
+        }
+
+        public void OnTestEvent(string report)
+        {
+        }
```
jnm2: Was this method supposed to be temporary?

jnm2:
@CharliePoole I'm worried now about the change @mikkelbu commented on: https://github.com/nunit/nunit-console/pull/177#discussion_r161982528

CharliePoole:
@mikkelbu I realized I was reversing your change. That's when the assembly-level errors stopped displaying, because there is no site in those cases. But I didn't think about warnings.

You are correct, however, WRT warnings displaying for every nested suite. Should we start setting FailureSite for warnings?

Perhaps I should back out the changes related to issue #350 and just push the ones directly relevant to the original issue.

src/NUnitConsole/nunit3-console.tests/BadFileTests.cs
CharliePoole: No, it's the implementation of `ITestEventListener`, required because I use `this` in the call to `Run`.

jnm2: Oh duh, sorry!

jnm2:
> Perhaps I should back out the changes related to issue #350 and just push the ones directly relevant to the original issue.

Whatever is most expedient, but I hope the same total work stays tracked.

ChrisMaddock:
> Also, it could use fake services in this case, since the files don't actually get loaded. But creating all those fakes could also be an exercise for the future.

My current open PR does some work there. I need to finish it at some point...

jnm2:
If I understandc orrectly, the quick fix would regress https://github.com/nunit/nunit-console/issues/289. If that's the case, can we wait for the full fix?

What's the reason we can't Test, SetUp and TearDown sites? We just need to make sure RecordFailure has access to the same context that an exception handler would.

CharliePoole:
@jnm2 I see no reason why any fix would break #289 "Warnings not displayed" but there are other reasons to do the fix of #350 separately - the way I usually like to do things! 😆 

This appears to be one of those "can of worms" compatibility issues. The "real" fix needed is in the framework, but console the console needs to work with the current and earlier frameworks as well as the next one. I'm backing out the #350-related changes and will push with just those things needed to fix #116, which was the original issue targeted.

jnm2:
Oh, I see. I'm okay with a quick fix in this PR if it comes with tests to prevent 
https://github.com/nunit/nunit-console/pull/177#discussion_r161982528 from happening in the future.

CharliePoole:
@mikkelbu Thanks for pointing out the "worms"! 😸

mikkelbu:
@CharliePoole My pleasure 😄. Too bad we did not think about the assembly-level errors when I did #326. Do we have any detailed description of the XML (schema and assumptions)?

CharliePoole:
Ready for merge now. All CI passed except for travis IOS. I'll do a separate PR for #350.

CharliePoole:
@mikkelbu There's this: https://github.com/nunit/docs/wiki/Test-Result-XML-Format reachable through NUnit | Usage Notes | XML Formats in the menu. It doesn't give any info other than what we have already disussed.

src/NUnitEngine/nunit.engine/Services/DefaultTestRunnerFactory.cs
```diff
@@ -79,23 +79,21 @@ public override ITestEngineRunner MakeTestRunner(TestPackage package)
...
             ProcessModel processModel = GetTargetProcessModel(package);
 
             switch (processModel)
             {
                 default:
                 case ProcessModel.Default:
-                    if (package.SubPackages.Count > 1)
+                    if (projectCount > 0)
+                        return new AggregatingTestRunner(ServiceContext, package);
+                    else if (package.SubPackages.Count > 1 || projectCount > 0)
```
rprouse: The `|| projectCount > 0` will never be true because of the previous if statement. This will still work, but it is redundant.

CharliePoole: Right! Left over from earlier incarnation. I'll remove it for cleanness.

jnm2: I didn't see that! The `else` is redundant too because of the `return`.

CharliePoole:
Trying to comment on my own PR and GitHub thinks it's a review! AFAIC they could dump the entire "review" thing.

jnm2:
> Redundant else for clarity is a "thing." I tend to use it when it feels right, otherwise not. File under "style."

Me too. I didn't see the following statements.

Issue #173, opened 2017-02-04
Refactor RuntimeFramework to ensure thread-safety


ChrisMaddock:
This came up in debugging https://github.com/nunit/nunit-console/issues/168. The inline code comment covers everything, I think.

> //RuntimeFramework was originally created for use in
        // a single-threaded environment. The introduction of parallel
        // execution and especially parallel loading of tests has
        // exposed some weaknesses.
        //
        // Ideally, we should remove all knowledge of the environment
        // from RuntimeFramework. An instance of RuntimeFramework does
        // not need to know, for example, if it is available on the 
        // current system. In the present architecture, that's really
        // the job of the RuntimeFrameworkService. Other functions
        // may actually belong in TestAgency.
        //
        // All the static properties of RuntimeFrameowork need to be
        // examined for thread-safety, particularly CurrentFramework
        // and AvailableFrameworks. The latter caused a problem with
        // parallel loading, which has been fixed for now through a
        // hack added to RuntimeFrameworkService. We may be able to
        // move all this functionality to services, eliminating the
        // use of public static properties here.

ToDo:
- [ ] CurrentFramework
- [x] AvailableFrameworks Interface for GUI dependency
- [ ] AvailableFrameworks
- [ ] MonoVersion/MonoPrefix/MonoExePath (which relate to available versions of mono, not the mono the current RuntimeFramework object represents)
- [ ] GetBestAvailableFramework()

ChrisMaddock:
I've started work on this. RuntimeFramework is a big class, so as well as moving some properties to RuntimeFrameworkService, I also want to move some content out to utility classes. Hopefully then, RuntimeFramework will just end up holding the details of a single framework.

I'm aiming to move the following to instead be accessible via RuntimeFrameworkService:
- CurrentFramework (Plus framework-locating utility class)
- AvailableFrameworks (Plus framework-locating utility classes)
- MonoVersion/MonoPrefix/MonoExePath
- GetBestAvailableFramework

All of this is currently public - but I think it should mostly be internal, excluding CurrentFramework/AvailableFrameworks, which could expose a (enumerable of) IRuntimeFrameworks via the RuntimeFrameworkService. This would potentially be a breaking change for any runners referencing the engine directly. @nunit/gui-team - does the GUI use any of these classes?

@nunit/engine-team - Are we happy with this these (breaking) changes, and in general? My personal thought is that getting the architecture right takes priority here.

ChrisMaddock:
[...]

PR #183, opened 2017-02-25
Add locking to AgentDatabase

### LINKED ISSUE #168: Intermittent errors while running tests after updating to 3.6
@appel1 commented on [Mon Jan 30 2017](https://github.com/nunit/nunit/issues/2012)

Every now and then when we run our tests using the console runner after updating to 3.6 it fails with an error for some of the test assemblies. Tests for assemblies with this error are not discovered nor executed. It is not always the same assemblies that fails, sometimes it is one and sometimes there are more than one that fails. 

We execute the tests with: nunit3-console.exe [path to assemblies] -work=[absolute path to a directory] --result=TestResult.xml;format=nunit2

So far we've observed these errors with similar callstacks:
> Index was outside the bounds of the array.
>    at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
>    at NUnit.Engine.Services.TestAgency.LaunchAgentProcess(TestPackage package)
>    at NUnit.Engine.Services.TestAgency.CreateRemoteAgent(TestPackage package, Int32 waitTime)
>    at NUnit.Engine.Runners.ProcessRunner.CreateAgentAndRunner()
>    at NUnit.Engine.Runners.ProcessRunner.RunTests(ITestEventListener listener, TestFilter filter)

> Object reference not set to an instance of an object.
>    at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
>    at NUnit.Engine.Services.TestAgency.LaunchAgentProcess(TestPackage package)
>    at NUnit.Engine.Services.TestAgency.CreateRemoteAgent(TestPackage package, Int32 waitTime)
>    at NUnit.Engine.Runners.ProcessRunner.CreateAgentAndRunner()
>    at NUnit.Engine.Runners.ProcessRunner.RunTests(ITestEventListener listener, TestFilter filter)

---

@ChrisMaddock commented on [Mon Jan 30 2017](https://github.com/nunit/nunit/issues/2012#issuecomment-276062189)

I'll move this to the console repo, and comment over there. 😄

ChrisMaddock:
Fixes #168 . This is targeting release/3.6 as I think it may need to be a hotfix - although I don't think that decisions been made yet, so will move it over if not.

This adds a lock object to the AgentDatabase, to allow access in parallel. This was introduced by the just-in-time agent loading fixed in 3.5, and will be required if we parallelise load/explore etc. 

Many of the methods in AgentDatabase were no longer used - so I took the liberty of removing them rather than making them safe. The class is now solely additive.

I wanted to add a test to ensure agent creation can be parallelised, but wasn't able to reproduce the error. The test I had (and removed) is below - open to any ideas as to how we might be able to put something in to prevent regression here. (Note the below additionally doesn't work due to #173 - but once I'd hacked around that, I still wasn't able to reproduce the error from #168.)

```
        [Test]
        public void CanCreateAgentsInParallel()
        {
            const int agents = 10;
            var workerPool = new ParallelTaskWorkerPool(agents);

            for (var i = 0; i < agents; i++)
            {
                var task = new LaunchAgentTask(_testAgency);
                workerPool.Enqueue(task);
            }

            TestDelegate createAgents = () =>
            {
                workerPool.Start();
                workerPool.WaitAll();
            };
            Assert.That(createAgents, Throws.Nothing);
        }

        private class LaunchAgentTask : ITestExecutionTask
        {
            private readonly TestAgency _agency;

            public LaunchAgentTask(TestAgency agency)
            {
                _agency = agency;
            }

            public void Execute()
            {
                _agency.GetAgent(new TestPackage("."), Timeout.Infinite);
            }
        }
```

### DIFF
--- a/src/NUnitEngine/nunit.engine/Services/TestAgency.cs
+++ b/src/NUnitEngine/nunit.engine/Services/TestAgency.cs
@@ -1,5 +1,5 @@
 // ***********************************************************************
-// Copyright (c) 2011 Charlie Poole
+// Copyright (c) 2011-2016 Charlie Poole
 //
 // Permission is hereby granted, free of charge, to any person obtaining
 // a copy of this software and associated documentation files (the
@@ -381,58 +381,6 @@ namespace NUnit.Engine.Services
             {
                 Status = ServiceStatus.Error;
                 throw;
-            }
-        }
-
-        #endregion
-
-        #region Nested Class - AgentRecord
-        private class AgentRecord
-        {
-            public Guid Id;
-            public Process Process;
-            public ITestAgent Agent;
-            public AgentStatus Status;
-
-            public AgentRecord( Guid id, Process p, ITestAgent a, AgentStatus s )
-            {
-                this.Id = id;
-                this.Process = p;
-                this.Agent = a;
-                this.Status = s;
-            }
-
-        }
-        #endregion
-
-        #region Nested Class - AgentDataBase
-        /// <summary>
-        ///  A simple class that tracks data about this
-        ///  agencies active and available agents.
-        ///  This class is required to be multi-thread safe.
-        /// </summary>
-        private class AgentDataBase
-        {
-            private readonly Dictionary<Guid, AgentRecord> _agentData = new Dictionary<Guid, AgentRecord>();
-            private readonly object _lock = new object();
-
-            public AgentRecord this[Guid id]
-            {
-                get
-                {
-                    lock (_lock)
-                    {
-                        return _agentData[id];
-                    }
-                }
-            }
-
-            public void Add(AgentRecord r)
-            {
-                lock (_lock)
-                {
-                    _agentData[r.Id] = r;
-                }
             }
         }
 
--- a/src/NUnitEngine/nunit.engine/Services/TestAgency.cs
+++ b/src/NUnitEngine/nunit.engine/Services/TestAgency.cs
@@ -121,9 +121,6 @@ namespace NUnit.Engine.Services
         public ITestAgent GetAgent(TestPackage package, int waitTime)
         {
             // TODO: Decide if we should reuse agents
-            //AgentRecord r = FindAvailableRemoteAgent(type);
-            //if ( r == null )
-            //    r = CreateRemoteAgent(type, framework, waitTime);
             return CreateRemoteAgent(package, waitTime);
         }
 
@@ -139,16 +136,6 @@ namespace NUnit.Engine.Services
             }
         }
 
-        //public void DestroyAgent( ITestAgent agent )
-        //{
-        //    AgentRecord r = agentData[agent.Id];
-        //    if ( r != null )
-        //    {
-        //        if( !r.Process.HasExited )
-        //            r.Agent.Stop();
-        //        agentData[r.Id] = null;
-        //    }
-        //}
         #endregion
 
         #region Helper Methods
@@ -183,7 +170,7 @@ namespace NUnit.Engine.Services
                     string.Format("The {0} framework is not available", targetRuntime),
                     "framework");
 
-            string agentExePath = GetTestAgentExePath(targetRuntime.ClrVersion, useX86Agent);
+            string agentExePath = GetTestAgentExePath(useX86Agent);
 
             if (agentExePath == null)
                 throw new ArgumentException(
@@ -236,26 +223,6 @@ namespace NUnit.Engine.Services
             return agentId;
         }
 
-        //private void OnProcessExit(object sender, EventArgs e)
-        //{
-        //    Process p = sender as Process;
-        //    if (p != null)
-        //        agentData.Remove(p.Id);
-        //}
-
-        //private AgentRecord FindAvailableAgent()
-        //{
-        //    foreach( AgentRecord r in agentData )
-        //        if ( r.Status == AgentStatus.Ready)
-        //        {
-        //            log.Debug( "Reusing agent {0}", r.Id );
-        //            r.Status = AgentStatus.Busy;
-        //            return r;
-        //        }
-
-        //    return null;
-        //}
-
         private ITestAgent CreateRemoteAgent(TestPackage package, int waitTime)
         {
             Guid agentId = LaunchAgentProcess(package);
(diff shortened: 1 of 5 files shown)


src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -408,60 +408,32 @@ public AgentRecord( Guid id, Process p, ITestAgent a, AgentStatus s )
...
-            public void Add( AgentRecord r )
+            public void Add(AgentRecord r)
             {
-                agentData[r.Id] = r;
-            }
-
-            public void Remove(Guid agentId)
-            {
-                agentData.Remove(agentId);
-            }
```
CharliePoole: I agree with the removal of unused methods, but I'm very surprised that this is unused. Seems like we should be removing entries when we stop the runner.

We've known about the problem of thread-safety for a long time - before the latest changes made it impossible to ignore. So it's possible that some of these method have calls that were deleted or commented out. Let's take a look at TestAgency to see if that's so.

ChrisMaddock:
Thanks @CharliePoole - will have a look tomorrow.

Which bits of commented code do you think should be be re-enabled? I'm a little wary of blindly making changes to the agent termination code, given the issues we've have and had have the remoting exceptions here. Would it perhaps be better if this PR just puts back the relevant bits of the AgentDatabase, and the other commented out code is looked at in a separate issue?

CharliePoole:
I'd be OK with leaving the TestAgency changes re-enabling use of the other methods to a separate PR. For this one, I think we should re-enable Remove and Clear and add the lock to them. Ideally, we should add some tests too, but it depends on how much time you have right now. We shouldn't start to actually use those other methods till we have tested them with multiple clients.

ChrisMaddock:
I didn't manage to get back to this last weekend, and I'm not sure I will this weekend either. If anyone wants to take over this branch, you're welcome to - otherwise I'll get back to it when time allows!

CharliePoole:
@ChrisMaddock I pulled this down to give it a try and it's hanging when I try to run the engine tests. Since it was passing in AppVeyor, it must be something different on my system.

CharliePoole:
@ChrisMaddock Resolved my problem. Writing some tests now.

ChrisMaddock:
Great - thanks @CharliePoole and @thinkbeforecoding!

thinkbeforecoding:
What's still needed to make it ready ?

CharliePoole:
I need to decide whether to keep and use the currently commented methods . I'll look at it in the morning.

CharliePoole:
I think this is now ready to merge. Since it has changed substantially, I'd like a @nunit/framework-team member to (re)review it.

`AgentDataBase` is no longer an internal class to `TestAgency`. It's in its own file and has its own tests.

I left some commented-out code in place in `TestAgency`, because I think we may want to re-activate it to fix some other issues. For the same reason, I left in a few `AgentDataBase` methods that are not now called by `TestAgency`.

I'm not completely happy with adding the `Snapshot` facility to `AgentDataBase` since it serves only to enable testing more easily. However, it simplified things enormously when I added it and we can always refactor it in another PR if we want.

CharliePoole:
Well, my "clever" use of `nameof` causes a warning under Mono, which makes the build fail.

I did this...

```C#
private static int[] Counts = new int[] { 1, 3, 10 };

const string COUNTS = nameof(Counts);
```

which allows me to use `TestCaseSource(COUNTS)`.

However, `Counts` is seen as never used! 😠

CharliePoole:
Tests failed once in travis and passed on re-running. Failure was an issue with how we run multiple `TestAgency` instances in our own tests, not something that comes up in normal execution of user tests.

ChrisMaddock:
Thanks for picking this up @CharliePoole. Just to note - I started this PR based of the release-3.6 branch, we'll need to make sure this is either released as a hotfix, or pulled into master.

CharliePoole:
@ChrisMaddock @rprouse 

Well darn it! I knew it was on release/3.6 but forgot it when I was doing the merge. I restored the branch but didn't do anything else yet. We need to decide what we want before I do more.

First thing would be to decide about a hotfix release. I'm not too keen on it. The latest framework hotfix required a lot of work by Rob and I was planning to send you guys a note suggesting we change our view of hotfixes. I'm thinking we should do one of two things:

1. Reduce the time between releases so we have less need of hotfixes. Just releasing what's in master is really quite easy compared to hot fixes.

2. Reduce the time between releases and just __stop__ doing hotfixes. If something comes up that's urgent, just release everything as either a new minor version upgrade or a third-digit update, depending on whether there are any features included.

I prefer the second approach.

If you agree with me, then I think I should back out the merge and redo it into master. I'll wait to see what your opinion is. If we do back it out, I don't think we want an offseting commit. I would just do it locally and force push to the branch since we haven't yet released anything from it beyond 3.6 itself.

I think we need to decide on this rather quickly.

ChrisMaddock:
I have no problem with releasing everything as needed, instead of cherry picking commits.

Does that need to affect the time between main releases?

CharliePoole:
I hope the time gets shorter. Seems to me that if we focus on a single thread of development, we can try to get really good at releasing and automate as much as possible. 

@rprouse I think this lies with you to a great extent, since you have been doing most of the work of releasing.

rprouse:
I am okay with releasing early instead of doing a hotfix, but I don't really want to reduce the normal cadence of releases. No matter how much we automate it, it still takes time to do releases and I think a normal cadence of more than quarterly creates too much churn.

That said, this fix is on the 3.6 branch, so I would be fine with doing a 3.6.1 hotfix with just this fix. If there are not other fixes to be cherry-picked into this hotfix, doing a hotfix on this branch is just as easy as doing a full release. At this point, I think I would prefer to do the hotfix.

thinkbeforecoding:
Yes, not waiting too long to release the fix would be great since it makes build randomly fail.
This causes some unnecessary time waste in the CI tool chaine.

CharliePoole:
Let's do the hotfix for this one then and deal with others as they come up.

rprouse:
I will start the release tonight and hopefully get it out tomorrow or Wed.

PR #186, opened 2017-03-06
Update --labels with additional options.

### LINKED ISSUE #24: Update --labels switch with option to show real-time pass/fail results in console runner

CharliePoole:
Fixes #24 

@constructor-igor spent a lot of time on this issue, but the branch was never merged. As part of cleaning up my old work in the framework, I decided to recreate the branch and try to get it finished.

Since the issue had a lot of comments over time, some conflicting, I'll summarize how the new labels values work in this PR. I use "normal" output to mean output through `Console.Out` or `TestContext.Out` and "immediate" for output through `Console.Error`, `TestContext.Progress` or `TestContext.Error`.

**--labels:Off** 
No labeling is used. Both normal and immediate output appear in the order produced - i.e. immediate first.

**--labels:On**
A label appears before each sequence of output lines from the same test. Since tests may be run in parallel, output from different tests may be intermixed.

**--labels:All**
A label appears at the start of every test, whether it produces output or not. Additional labels are produced as needed if interspersed output takes place, just as for `--labels:On`. Synonym for `--labels:Before`.

**--labels:Before**
A label appears at the start of every test, whether it produces output or not. Additional labels are produced as needed if interspersed output takes place, just as for `--labels:On`. Synonym for `--labels:All`.

**--labels:After**
A label appears at the end of every test, whether it produced output or not. This label includes the pass/fail status of the test in addition to its name. Additional labels are produced as needed if there is any output, just as for `--labels:On`.

### DIFF
--- a/src/NUnitConsole/nunit3-console.tests/CommandLineTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/CommandLineTests.cs
@@ -126,7 +126,7 @@ namespace NUnit.ConsoleRunner.Tests
         [TestCase("OutFile", "output|out", new string[] { "output.txt" }, new string[0])]
         [TestCase("ErrFile", "err", new string[] { "error.txt" }, new string[0])]
         [TestCase("WorkDirectory", "work", new string[] { "results" }, new string[0])]
-        [TestCase("DisplayTestLabels", "labels", new string[] { "Off", "On", "All" }, new string[] { "JUNK" })]
+        [TestCase("DisplayTestLabels", "labels", new string[] { "Off", "On", "Before", "After", "All" }, new string[] { "JUNK" })]
         [TestCase("InternalTraceLevel", "trace", new string[] { "Off", "Error", "Warning", "Info", "Debug", "Verbose" }, new string[] { "JUNK" })]
         [TestCase("DefaultTestNamePattern", "test-name-format", new string[] { "{m}{a}" }, new string[0])]
         [TestCase("ConsoleEncoding", "encoding", new string[] { "utf-8", "ascii", "unicode" }, new string[0])]
@@ -167,7 +167,7 @@ namespace NUnit.ConsoleRunner.Tests
 
         [TestCase("ProcessModel", "process", new string[] { "InProcess", "Separate", "Multiple" })]
         [TestCase("DomainUsage", "domain", new string[] { "None", "Single", "Multiple" })]
-        [TestCase("DisplayTestLabels", "labels", new string[] { "Off", "On", "All" })]
+        [TestCase("DisplayTestLabels", "labels", new string[] { "Off", "On", "Before", "After", "All" })]
         [TestCase("InternalTraceLevel", "trace", new string[] { "Off", "Error", "Warning", "Info", "Debug", "Verbose" })]
         public void CanRecognizeLowerCaseOptionValues(string propertyName, string optionName, string[] canonicalValues)
         {
--- a/src/NUnitConsole/nunit3-console.tests/TestEventHandlerTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/TestEventHandlerTests.cs
@@ -42,8 +42,8 @@ namespace NUnit.ConsoleRunner.Tests
             _writer = new StringWriter(_output);
         }
 
-        [TestCaseSource("EventData")]
-        public void EventsWriteExpectedOutput(string report, string labels, string expected)
+        [TestCaseSource("SingleEventData")]
+        public void SingleEventsWriteExpectedOutput(string report, string labels, string expected)
         {
             var handler = new TestEventHandler(_writer, labels);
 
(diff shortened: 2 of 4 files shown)


CharliePoole:
@nunit/core-team @nunit/engine-team @nunit/framework-team 

This could use a careful review of how it's all defined to work. I'd rather make any changes before it gets released and people are using it. In addition to whatever you come up with, I have few particular questions:

1. Should we deprecate or eliminate `--labels:All`?
2. Should the label that appears at the start of a test have a different appearance from other labels?
3. On After labels, I'm coloring only the status word (Passed, Failed, etc.) according to the result. Would it be better to color the entire line?
4. There is currently no way to get both a Before and an After label. Do we need one?

CharliePoole:
Travis is failing because it can't get an osx agent.

JustinRChou:
1. Doesn't seem to be a point in keeping `--labels:All`.
2. Don't quite understand this.
3. Should be fine. You do mean that it's the status color followed by the blue label?
4. I would think that should actually be the `--labels:All` option?

Should the default for labels be updated to `Before`? And is there still a need for `On`?
I guess it would not make sense for projects for lots of tests and have the default be `Before`.

CharliePoole:
@JustinRChou You're keeping me on my toes - I should have explained some things better.

1. We can't drop it completely without breaking scripts, but I'd definitely like to phase it out.

2. Imagine you see something like this:
```
=> Name.Of.Test
Some output
```

You can't tell if that label was displayed because of the output or if it marks the start of a test. With just one test, it doesn't matter but if many tests are running in parallel and if you are trying to see how they overlap it may be important. Of course, __you__ know what option you used to run the test so you can interpret, but if the report is saved, others may not be able to tell. So there is an issue, but I'm just not sure if it's really important.

3. Exactly.

4. Yes, it's too bad that we already use All to mean something else: display label at the start of __all__ tests. This is one reason I want to phase it out: it looks like it might mean something else. And after we phase it out, I don't think we can re-use it with a different meaning without confusing people.

So basically, there is no way currently to show a label both when the test starts as well as the end. If you use Before, you see the start but you only see where it ended if there is normal output. If you use After, you can see the end but not the start. 

Should we have Both? Or wait till it's asked for?

JustinRChou:
2. For the parallel testcase, would it produce something like
```
=> Test1(1)
=> Test2(2)
Output from Test2(2)
Output from Test1(1)
=> Test3(3)
Output from Test2(3)
```
I don't think I am good at simulating parallel tests. I might be thinking of multi-threading. 
If so, I don't see how to avoid putting the test name as part of the output again.
And when you mean saving the report, you're referring to a redirection to a log file? Or the testresult.xml?

I don't think `--labels:Both` would be critical, looks to me that `--labels:After` would be more useful. Unless 'Both' would work better for parallel tests?
But I am not a frequent user of the `label` option and parallel testing.

JustinRChou:
Would this also be a valid solution for #25?
Looks like using `--labels:After` would provide a decent progress report.

CharliePoole:
@JustinRChou Well, the whole point of this change is to ensure that no output appears without a proper label ahead of it (unless you turn labels off). Did you look at the tests? The ones that show two tests interacting lay out exactly the output you would see.

Yes, After can be used as a progress report of sorts just as All is currently used. But I think #25 is envisioning some kind of continued status feedback such as you often see with downloads.

CharliePoole:
@nunit/core-team Hate to bug you guys, but this has been waiting almost three weeks for a review!

rprouse:
Code looks good. I do worry about the number of options. It seems to me we are adding complexity and flexibility because of a vocal minority. This is also yet another change that needs to be ported across to NUnitLite. Speaking of that, I wonder if NUnitLite should be scaled back to just support simple running, exploring and filters?

CharliePoole:
I've been thinking about scaling back NUnitLite too. It might be something I'd like to do once I get off of the framework and console.

PR #195, opened 2017-03-11
Retrieve settings with invariant culture

### LINKED ISSUE #193: Settings are stored with invariant culture but retrieved with CurrentCulture
The following problem was discovered trying to debug a problem with nunit-gui, hence running `.\nunit-gui.exe --trace=Debug`. That gave me the following lines in the InternalTrace-file
```
...
23:02:50.311 Error [ 1] SettingsGroup: Unable to convert setting Gui.MainForm.Location to Point
23:02:50.311 Error [ 1] SettingsGroup: 10, 10 is not a valid value for Int32.
23:02:50.312 Error [ 1] SettingsGroup: Unable to convert setting Gui.MainForm.Size to Size
23:02:50.312 Error [ 1] SettingsGroup: 504, 301 is not a valid value for Int32.
...
```
and the Nunit30Settings.xml file looked like this 
```xml
<?xml version="1.0"?>
<NUnitSettings>
  <Settings>
    ...
    <Setting name="Gui.MainForm.Location" value="10, 10" />
    ...
    <Setting name="Gui.MainForm.Size" value="504, 301" />
    ...
  </Settings>
</NUnitSettings>
```

The problem is that in danish culture (da-DK) the first separator character (and the one that `TypeConverter` uses to split the string) is not "," but ";", hence `TypeConverter` reads "10, 10" as just one element which then fail to parse as an integer.

The problem can be reproduced with the following test
```csharp
[Test]
//[SetCulture("en-US")] // Works 
[SetCulture("da-DK")] // Fails 
public void SaveAndGetSettingShouldReturnTheOriginalValue()
{
  var settingsStore = new SettingsStore(@"C:\Temp\SomeFolder\MySettings.xml", true);
  var settingName = "MySetting";

  var settingValue = new Point(10, 20);
  settingsStore.SaveSetting(settingName, settingValue);
  settingsStore.SaveSettings();
  settingsStore.LoadSettings();

  var point = settingsStore.GetSetting(settingName, new Point(30, 40));
  Assert.That(point, Is.EqualTo(settingValue));
}
```

The problem can be solved by changing 
`return (T)converter.ConvertFrom(result);`
in https://github.com/nunit/nunit-console/blob/8cb7226c0ad683fe5eb7a4eee8989aff08dc4ccb/src/NUnitEngine/nunit.engine/Internal/SettingsGroup.cs#L82
to
`return (T)converter.ConvertFrom(null, CultureInfo.InvariantCulture, result);`

I can supply a pull request, but I'm a bit unsure whether the test should be included, since it writes to the file system (I will of cause change the location of the file).

mikkelbu:
As settings are saved with ConvertToInvariantString
we also need to read the settings using
CultureInfo.InvariantCulture.

Fixes #193

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/Internal/SettingsGroupTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Internal/SettingsGroupTests.cs
@@ -22,6 +22,8 @@
 // ***********************************************************************
 
 using System;
+using System.ComponentModel;
+using System.Drawing;
 using NUnit.Framework;
 using Microsoft.Win32;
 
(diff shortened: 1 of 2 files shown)


CharliePoole:
Travis failure looks like an unrelated systemic failure. I restarted the build. Pending passing the tests, this looks good to me. It will need one more committer review to be merged.

rprouse:
Looks good to me too. Thanks for the contribution 👍

PR #199, opened 2017-03-15
Calling Unregister on the test agency when stopping

JeffAshton:
This is really just a follow up to the work @ChrisMaddock did recently on AgentDataBase: https://github.com/nunit/nunit-console/commit/6fef1f3574cdfd2e21cd484ca142e34b2a03fd84

Now that it's thread safe, we can unregister the agents.

In our use case, we have lots of assemblies and we've been getting OutOfMemoryException errors from the console runner.  I was doing some profiling with .NET Memory Profiler (http://memprofiler.com/) last night and came across this leak.  This something I'm sure you're aware of @CharliePoole giving your existing comments in RemoteTestAgent.Stop

https://github.com/nunit/nunit-console/blob/dc655ce241994f5a3664fe034b19d7be04035b64/src/NUnitEngine/nunit.engine/Agents/RemoteTestAgent.cs#L99

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/Agents/RemoteTestAgentTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Agents/RemoteTestAgentTests.cs
@@ -1,0 +1,95 @@
+﻿// ***********************************************************************
+// Copyright (c) 2016 Charlie Poole
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System;
+using System.Collections.Generic;
+using NUnit.Framework;
+
+namespace NUnit.Engine.Agents.Tests
+{
+    [TestFixture]
+    public sealed class RemoteTestAgentTests
+    {
+        [Test]
+        public void Stop_ShouldUnregisterWithAgency()
+        {
+            List<Guid> unregisteredAgents = new List<Guid>();
+
+            ITestAgency agency = new MockAgency(
+                register: agent => { Assert.Fail("Should not be invoked"); },
+                unregister: unregisteredAgents.Add
+            );
+
+            Guid agentId = Guid.NewGuid();
+
+            using (RemoteTestAgent testAgent = new RemoteTestAgent(agentId, agency, services: null))
+            {
+                testAgent.Stop();
+                CollectionAssert.AreEqual(new[] { agentId }, unregisteredAgents, "Agent should be unregistered");
+            }
+
+            CollectionAssert.AreEqual(new[] { agentId }, unregisteredAgents, "Agent should only unregistered once");
+        }
+
+        [Test]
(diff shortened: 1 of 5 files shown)


src/NUnitEngine/nunit.engine/ITestAgency.cs
```diff
@@ -34,5 +34,11 @@ public interface ITestAgency
         /// </summary>
         /// <param name="agent"></param>
         void Register(ITestAgent agent);
+
+        /// <summary>
+        /// Unregisters an agent with an agency
+        /// </summary>
+        /// <param name="agent"></param>
+        void Unregister(ITestAgent agent);
```
JeffAshton: This could also take the Guid agent id.

src/NUnitEngine/nunit.engine.tests/Agents/RemoteTestAgentTests.cs
```diff
@@ -0,0 +1,93 @@
...
+                Assert.DoesNotThrow(() => testAgent.Stop(), "Agent should catch and log exception");
+            }
+        }
+
+        #region MockAgency Definition
+
+        private sealed class MockAgency : ITestAgency
+        {
+            private readonly Action<ITestAgent> m_register;
+            private readonly Action<ITestAgent> m_unregister;
```
CharliePoole: Please use our convention: underscore prefix without the 'm'.

src/NUnitEngine/nunit.engine.tests/Agents/RemoteTestAgentTests.cs
```diff
@@ -0,0 +1,93 @@
...
+            }
+        }
+
+        [Test]
+        public void Stop_WhenAgencyThrows_ShouldNotThrow()
+        {
+            ITestAgency agency = new MockAgency(
+                register: agent => { Assert.Fail("Should not be invoked"); },
+                unregister: agent => { throw new Exception("Boom"); }
+            );
```
JeffAshton: Another case where this is super useful is when you're passing things like null or 0 to a method.  If you look below, had I constructed RemoteTestAgent with null for the services, you put the work on the reader to figure out what null means.  But using the named parameter it's clearer, and it calls out my intent.

src/NUnitEngine/nunit.engine/ITestAgency.cs
CharliePoole: The guid seems best to me. No reason to pass more info to the database than it really needs.

That said, let's first make sure we are doing enough with this. Have you checked what happens to a process runner that's holding a test agent and runner when this de-registration happens?

src/NUnitEngine/nunit.engine.tests/Agents/RemoteTestAgentTests.cs
JeffAshton: Sorry, my bad.  Will clean that up.

src/NUnitEngine/nunit.engine/ITestAgency.cs
JeffAshton: Yes, it unfortunately still holding the reference to both the agent and runner which has me a little sad.  The comment in [TestAgency](https://github.com/nunit/nunit-console/blob/master/src/NUnitEngine/nunit.engine/Services/TestAgency.cs#L112) indicate there was intent for agents to be reused, but that doesn't appear to be happening in practice.

Given the current design, we could return a proxy ITestAgent that cleans up on Unregister.  But it might be better to revisit the design between ProcessRunner and TestAgency.  If ProcessRunner always just makes action calls to TestAgency, then we don't have this weird left over reference.

In other words,  ProcessRunner.RunTests => TestAgency.RunTests . TestAgency worries about either running the tests in a new process, or re-use existing.  But it just does actions.  It doesn't return agents like it does currently.

src/NUnitEngine/nunit.engine.tests/Agents/RemoteTestAgentTests.cs
JeffAshton: Fixed

src/NUnitEngine/nunit.engine/ITestAgency.cs
JeffAshton: Switched to take a Guid now

CharliePoole:
@JeffAshton Did you intend to close this?

JeffAshton:
Yes.  I kind of came to the conclusion that nunit-console just isn't great in general for large #s of tests and assemblies.  Couple reasons for this conclusions:

* Remote agents are held on to longer then needed, and the design of [ProcessRunner](https://github.com/nunit/nunit-console/blob/master/src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs) makes it difficult to define this scope.  ITestEngineRunner is a big interface, and having ProcessRunner try and provide all the same functionality complicates things.  For example Load isn't internally called on ProcessRunner, but 3rd party consumers might be using it?
* Results are stored in memory for all assemblies, instead of being streamed one at a time.  You can see this in [AggregatingTestRunner](https://github.com/nunit/nunit-console/blob/master/src/NUnitEngine/nunit.engine/Runners/AggregatingTestRunner.cs#L158), and eventually this approach is going to hit an upper bound

Because of that I decided to abandon this approach an re-purpose the runner I wrote for NUnit2 years ago.  It's main advantage is that it writes tests results per assembly, instead of trying to do a combined result file.  This allows me to only have to store the aggregated totals in memory.

For most uses cases though, this is overkill.  Most projects should have a small number of assemblies, that this memory overhead isn't an issue.

Issue #194, opened 2017-03-10
Better logging or error handling in SettingsStore.SaveSettings


mikkelbu:
Several people in the nunit-gui project have reported problems with corrupt Nunit30Settings.xml files (see for instance nunit/nunit-gui#130 or nunit/nunit-gui#162). 

The code in `SettingsStore.SaveSettings` however makes it difficult to pinpoint these problems as the code catches all exceptions (it also forgets to dispose the XmlTextWriter, but this is problably not relevant in this issue).
https://github.com/nunit/nunit-console/blob/8cb7226c0ad683fe5eb7a4eee8989aff08dc4ccb/src/NUnitEngine/nunit.engine/Internal/SettingsStore.cs

I propose that the method is changed such that it either:
* rethrows the exception after setting `_writeable = false` and then captures it further out the callstack where it can be logged (as far as I can tell the exception will be captured in `ServiceManager.StopServices`)
* logs the exception directly in `SettingsStore.SaveSettings`

CharliePoole:
I agree the code looks strange, but there are some valid occasions where the settings cannot be saved because there is no valid location at which to write them. However, this fact should be logged as you suggest.

CharliePoole:
I just noticed that the original description refers to the gui project. Everyone working on the gui project should be aware that we don't always use the production engine. At this issue was reported, we were not using it so that could lead to some confusion when trying to reproduce errors. 

Recently, I updated the gui to use the latest production engine, so this is a good time to be working on this issue. However, it's almost certain that we will go back to using a special engine after the 0.4 release.

PR #201, opened 2017-03-19
SettingsStore.SaveSettings rethrows exception

mikkelbu:
SettingsStore.SaveSettings now rethrows exceptions.
These exceptions will in turn will be caught by
ServiceManager.

Furthermore, we now use a using statment on XmlTextWriter,
hence forcing a dispose if an exception occur. This also
makes it easier to pinpoint exceptions occuring in
SaveSettings. For instance, the problem in nunit/nunit-gui#130
will give the following `Nunit30Settings.xml`
```xml
<?xml version="1.0"?>
<NUnitSettings>
  <Settings>
    <Setting name="Engine.Options.ReloadOnChange" value="True" />
    <Setting name="Engine.Options.ReloadOnRun" value="True" />
    <Setting name="Engine.Options.RerunOnChange" value="True" />
    <Setting name="Gui.InitialSettingsPage" value="Engine.Assembly Reload" />
    <Setting name="Gui.MainForm.Location" value="10, 10" />
    <Setting name="Gui.MainForm.Maximized" value="False" />
    <Setting name="Gui.MainForm.Size" value="504, 301" />
    <Setting name="Gui.RecentProjects.File1" value="C:\src\Test\some.dll" />
    <Setting name="Gui.TestTree.InitialTreeDisplay" />
  </Settings>
</NUnitSettings>
```
Giving an indication that the error occured at `Gui.TestTree.InitialTreeDisplay`.

Tests have been added in ServiceManagerTests and SettingsStoreTests.

Fixes: #194


mikkelbu:
Then we will have to change the produkction code, i.e. SettingsStore.SaveSettings, some more. As the code is now with this PR, we will write an entry with `name`, but with no `value`, if there is an invalid value. Before this PR we would stop somewhere in the middle of the file (depending on the flushing of the data to the file).

I propose that we only write til file to NUnit30Settings.xml if no error occured. Meaning that we will write to a temporary memorystream, and when we have written all the entries (and surrounding `<NUnitSettings>` and `<Settings>`) then we will write the resulting text directly to a file using for instance `File.WriteAllText();`.

CharliePoole:
I guess it depends on how our reading handles an entry with no value attribute.

mikkelbu:
The current code in `LoadSettings` will fail as it validates that each `Setting`-element has a `name` and a `value` attribute.

The solution I like the most is to keep `LoadSettings` as it is and instead change `SaveSettings` such that it only writes to the file if there are no errors. If there are errors then we instead write the resulting (incomplete) file to another file - in the same folder as `Nunit30Settings.xml`. We could append a timestamp to the filename of the other file to make it possible to record more problems. The filename could be on the form `Nunit30Settings_Error_<timestamp>.xml`

The consequences of this are:
* We will only have `Nunit30Settings.xml` files that satisfies the expected schema.
* If there are no errors then the code will have the same behaviour as before this PR.
* If there are errors then we will have information in `Nunit30Settings_Error_<timestamp>.xml` enabling us to troubleshoot the problem.

ChrisMaddock:
@mikkelbu - your comments sound sensible to me. I'd be tempted to write the corrupt settings out to the log, rather than a separate file. @CharliePoole?

Were you planning to add the extra work to this PR, or should we merge and address the partial case as a separate issue?

mikkelbu:
@ChrisMaddock The problem with writing the corrupt settings to the log is that we do not have access to the log in `SaveSettings`. One solution could be to create a special Exception type that can hold both the original exception and the content of the corrupted settings file and then override ToString() on this type such that we log properly in `ServiceManager.StopServices()`.

I will let you decide whether to add this work to this PR or to address it in another issue. I'm fine with both solutions.

ChrisMaddock:
The log is a static, it should be possible to get hold of it with `InternalTrace.GetLogger("SettingsGroup");` (see SettingsGroup.cs for an example.) I'll defer to @CharliePoole as to if doing that is actually a good idea or not though - seems to make sense to me to have all errors in a single log, but he may disagree!

Tackling this as a single issue for all 'error-state improvements' seems sensible to me - let's leave this PR open for now. Looks good so far though!

mikkelbu:
@CharliePoole Do you have any opinion on whether we should write corrupt settings out to the log or to a separate file? (see discussion above)

Also it seems that I've been overtaken by the addition of .NET Standard in master (which also touches SettingsStore.cs). What is the best approach for resolving this problem? Should I pull master into my topic branch?

CharliePoole:
Frankly, I don't see why we would write corrupt settings anywhere. Since only we write out settings, any corrupt settings represent an error in our code. Why not just throw as soon as we see the problem?

The problem with .NET standard is a result of not merging. If this had been merged, then Rob would have picked up the changes for .NET standard. It's an example of why "Merge early, merge often" is a watchword of open source work. Right now, I would say try to merge in master and fix what needs fixing before merging this PR. GitHub gives us a discounts on new PRs. :smile:

mikkelbu:
I've added two commits:
* The merge of master which was trivial except for `SaveSettings` which needed the new changes for NETSTANDARD1_3.
* The second commit changes my original attempt at SaveSettings such that we only save the file if no errors occurred. I've also added tests for this behaviour.

src/NUnitEngine/nunit.engine.tests/Internal/SettingsStoreTests.cs
```diff
@@ -0,0 +1,101 @@
...
+    [TestFixture]
+    public class SettingsStoreTests
+    {
+        private string settingsFile;
+        private SettingsStore settings;
+
+        [SetUp]
+        public void SetUp()
+        {
+            settingsFile = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
```
ChrisMaddock: How about `Path.GetTempFileName()`? 🙂

src/NUnitEngine/nunit.engine.tests/Internal/SettingsStoreTests.cs
```diff
@@ -0,0 +1,101 @@
...
+using NUnit.Framework;
+
+namespace NUnit.Engine.Internal.Tests
+{
+    [TestFixture]
+    public class SettingsStoreTests
+    {
+        private string settingsFile;
+        private SettingsStore settings;
+
```
ChrisMaddock: Private fields should be pre-fixed with underscores, please.

src/NUnitEngine/nunit.engine.tests/Internal/SettingsStoreTests.cs
```diff
@@ -0,0 +1,101 @@
...
+            var xmlTypeCodeSettingValue = XmlTypeCode.ProcessingInstruction;
+            var dateSettingValue = new DateTime();
+
+            settings.SaveSetting(xmlTypeCodeSettingName, xmlTypeCodeSettingValue);
+            settings.SaveSetting(dateSettingName, dateSettingValue);
+            settings.SaveSettings();
+            settings.LoadSettings();
+
+            var actualXmlTypeCodeValue = settings.GetSetting(xmlTypeCodeSettingName, XmlTypeCode.Comment);
+            var actualDateValue = settings.GetSetting(dateSettingName, DateTime.MinValue);
```
ChrisMaddock: `new DateTime()` == `DateTime.MinValue` I think? I think you need a different test value to check this is working/

src/NUnitEngine/nunit.engine.tests/Internal/SettingsStoreTests.cs
```diff
@@ -0,0 +1,101 @@
...
+            settings.LoadSettings();
+            var wrongValue = -1;
+            var settingValue = (XmlTypeCode)wrongValue;
+            settings.SaveSetting("setting", settingValue);
+            Assert.Throws<ArgumentException>(() => settings.SaveSettings());
+
+            // Assert that the initial version is not overwritten
+            settings.LoadSettings();
+            var actualXmlTypeCodeValue = settings.GetSetting(xmlTypeCodeSettingName, XmlTypeCode.Comment);
+            var actualDateValue = settings.GetSetting(dateSettingName, DateTime.MinValue);
```
ChrisMaddock: As above

ChrisMaddock:
> The travis-ci/pr-build failed due to a timeout. Can I rerun the build?

No, unfortunately GitHub/Travis don't give us that granularity of control. 👎 I've reran it.

src/NUnitEngine/nunit.engine.tests/Internal/SettingsStoreTests.cs
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        [Test]
+        public void SaveSettingsThrowsExceptionOnErrorInSetting()
+        {
+            var wrongValue = -1;
+            XmlTypeCode settingValue = (XmlTypeCode)wrongValue;
+            settings.SaveSetting("setting", settingValue);
+            Assert.Throws<ArgumentException>(() => settings.SaveSettings());
+            Assert.That(File.Exists(settingsFile), Is.False);
```
ChrisMaddock: I'd personally do `Assert.That(settingsFile, Does.Not.Exist);` here - which would give a nicer error message. That's mainly a matter of preference though. ¯\_(ツ)_/¯

src/NUnitEngine/nunit.engine.tests/Internal/SettingsStoreTests.cs
ChrisMaddock: I'd also add a TearDown in this, to delete the temp file at the end - but that's possibly because I'm a little obsessive. 😄

CharliePoole: That's not obsessive, it's basic. However, I would look into using a mock rather than creating a file - at least to evaluate whether that is a practical approach. I can live with the file, however, so long as it's deleted.

src/NUnitEngine/nunit.engine.tests/Internal/SettingsStoreTests.cs
CharliePoole: Feel free to use your birthday. 😄

mikkelbu:
I've made the changes requested except for using `Path.GetTempFileName()` as this creates an empty file.

ChrisMaddock:
>  except for using Path.GetTempFileName()

Ah - good thinking! 

Looks good, will merge once CI has passed - thanks! And good choice of dates. 😉

PR #202, opened 2017-03-27
Allow comments in files included in the command-line

### LINKED ISSUE #196: Allow comments in @FILE files
@somatronicon commented on [Mon Mar 13 2017](https://github.com/nunit/nunit/issues/2080)

In the @FILE files, allow comments. These files could get pretty complex, and having the ability to add comments would be beneficial. Maybe if a line starts with # it's interpreted as a comment? Not sure, just throwing out ideas.

```
# Black list specific tests we don't want to run
--where "method != /.*_(Blah|Blah2|Blah3)$/"
```

Great project by the way :+1: 

---

@JustinRChou commented on [Mon Mar 13 2017](https://github.com/nunit/nunit/issues/2080#issuecomment-286280536)

@CharliePoole 
Is this really a easy fix and wouldn't this involve updating the console runner and the lite runner?

And this is probably extra work but would it make sense to write the comment into the TestResult.xml?

---

@CharliePoole commented on [Mon Mar 13 2017](https://github.com/nunit/nunit/issues/2080#issuecomment-286295357)

I was thinking of this as a console issue but you're right that it applies to NUnitLite as well.

In the console, it's extremely easy to just ignore a full line starting with # - looks like a change of one line. Doing more is possible, but difficult since the individual line can contain # (or any other characters) in quotes. The format was designed to be simple and we probably shouldn't complicate it, since that would require a redesign without a clear benefit.

The NUnitLite implementation is entirely different and treats the entire file as a single string. That looks like an easyfix as well, but more than a single line.

I'll duplicate this under the console as well.

CharliePoole:
Fixes #196 

Thought I'd already created this PR, but apparently not!

### DIFF
--- a/src/NUnitConsole/nunit3-console.tests/CommandLineTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/CommandLineTests.cs
@@ -43,6 +43,8 @@ namespace NUnit.ConsoleRunner.Tests
         [TestCase("--arg1 @file1.txt --arg2 @file2.txt", "file1.txt:--fileArg1\n--fileArg2,file2.txt:--fileArg3", "--arg1 --fileArg1 --fileArg2 --arg2 --fileArg3", "")]
         [TestCase("--arg1 @file1.txt --arg2", "file1.txt:", "--arg1 --arg2", "")]
         [TestCase("--arg1 @file1.txt --arg2", "file1.txt:--fileArg1\n\n\n--fileArg2", "--arg1 --fileArg1 --fileArg2 --arg2", "")]
+        [TestCase("--arg1 @file1.txt --arg2", "file1.txt:--fileArg1\nThis is a COMMENT\n--fileArg2", "--arg1 --fileArg1 This is a COMMENT --fileArg2 --arg2", "")]
+        [TestCase("--arg1 @file1.txt --arg2", "file1.txt:--fileArg1\n#This is a COMMENT\n--fileArg2", "--arg1 --fileArg1 --fileArg2 --arg2", "")]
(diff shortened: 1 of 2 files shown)


src/NUnitConsole/nunit3-console.tests/CommandLineTests.cs
```diff
@@ -43,6 +43,8 @@ public class CommandLineTests
         [TestCase("--arg1 @file1.txt --arg2 @file2.txt", "file1.txt:--fileArg1\n--fileArg2,file2.txt:--fileArg3", "--arg1 --fileArg1 --fileArg2 --arg2 --fileArg3", "")]
         [TestCase("--arg1 @file1.txt --arg2", "file1.txt:", "--arg1 --arg2", "")]
         [TestCase("--arg1 @file1.txt --arg2", "file1.txt:--fileArg1\n\n\n--fileArg2", "--arg1 --fileArg1 --fileArg2 --arg2", "")]
+        [TestCase("--arg1 @file1.txt --arg2", "file1.txt:--fileArg1\nThis is a COMMENT\n--fileArg2", "--arg1 --fileArg1 This is a COMMENT --fileArg2 --arg2", "")]
+        [TestCase("--arg1 @file1.txt --arg2", "file1.txt:--fileArg1\n#This is a COMMENT\n--fileArg2", "--arg1 --fileArg1 --fileArg2 --arg2", "")]
         public void ArgumentsFromFilesTests(string args, string files, string expectedExpandedArgs, string expectedErrorMessages)
```
ChrisMaddock: Should these use `Environment.NewLine` rather than just `\n`? (Not specific to this PR - I know!)

PR #205, opened 2017-03-30
Lightweight .NET Standard version of the Engine

### LINKED ISSUE #10: Create a .NET Standard version of the Engine
@rprouse commented on [Mon Dec 14 2015](https://github.com/nunit/nunit/issues/1138)

This needs design, but could be platform specific agents. If so, it will likely be blocked by #362.

---

@rprouse commented on [Sat Jan 02 2016](https://github.com/nunit/nunit/issues/1138#issuecomment-168407265)

In #1168, @CharliePoole said,

> I'm not sure there is a problem here since this refers (in my mind anyway) to an engine running on the desktop. What we need to find out is whether an engine running under .NET 2.0/3.5 is able to analyze assemblies built for other targets. I don't see why not, since they are not actually loaded. Consider that we are already analyzing .NET 4.5 assemblies. Of course, this needs to be tested on a VM that has only .NET 2.0 or 3.5 installed.

Based on my spikes into this so far, I think this is the way we need to go. The engine can inspect the test assembly and determine it's target platform. We just can't run it directly.

This is also mixed up with #677 and #362. My idea so far is as follows,
- The current NUnit3 engine inspects the test assembly in addition to the framework. It looks for the `TargetFrameworkAttribute` which will tell you for example if it is targeting `.NETCore,Version=v5.0`.
- If the target framework is a non-desktop framework, disallow `--inprocess` since we can't run non-desktop targets. We might also want to disallow `--process=Single` or we make sure all test assemblies target the same platform.
- Modify `TestAgency` to allow it to launch platform specific agents and communicate with them in a manner other than .NET remoting. This is #362. If we are running those agents on the desktop machine, we could even capture Console.In/Out for communications.

The platform specific agents are where decisions still need to be made. It would be nice to have a portable mini-engine (#677) that is capable of loading and running the portable framework. Like the full framework, this would allow us to run test no matter which version of the NUnit 3 framework is used. It would likely be a stripped down version of the engine that can only load from the agent directory and is only intended to be used by agents. This might even boil down to a portable `IFrameworkDriver` and some supporting classes?

A simplified and not mutually exclusive approach might be that the agents are actually the nunitlite version of the tests. We could modify nunitlite to run directly, or be able to communicate with the engine.

---

@CharliePoole commented on [Sat Jan 02 2016](https://github.com/nunit/nunit/issues/1138#issuecomment-168450607)

I think we're pretty much on the same page at a high level. Some details...
- Can we count on TargetFrameworkAttribute being reliably set?
- We can't precisely disallow all those options, although we might be able to in some cases. Basically, we can only look at  explicit options, but if the defaults are used, we have to wait until we analyze the assembly and then reflect an error back to the runner.
- I think we still have two very distinct options for running portable and device-based tests: a scaled down engine versus a scaled up driver. Neil has been pushing us strongly toward an engine for CF but I'm not sure that's the  easiest approach. At some point, we have to spike both approaches. I think the easiest way to do that initially is right on the desktop itself.

Charlie

---

@rprouse commented on [Mon Jan 04 2016](https://github.com/nunit/nunit/issues/1138#issuecomment-168685508)

> Can we count on TargetFrameworkAttribute being reliably set?

I have tested quite a few different assemblies and have not found it unset except for .NET 2.0 assemblies. I think that we can safely assume that if it is not set, it targets a full .NET framework. I assume that these days with so many potential targets, it needs to be set.

> I think we still have two very distinct options for running portable and device-based tests: a scaled down engine versus a scaled up driver. Neil has been pushing us strongly toward an engine for CF but I'm not sure that's the easiest approach. At some point, we have to spike both approaches. I think the easiest way to do that initially is right on the desktop itself.

I installed the [.NET Portability Analyzer](https://visualstudiogallery.msdn.microsoft.com/1177943e-cfb7-4822-a8a6-e56c7905292b) extension for Visual Studio and analyzed the engine against the platforms that we want to support. Based on what is not available, I think approaching it from the driver end is probably the best option. At a minimum, we need to be able to load the framework in a version independent way, explore or run the tests and communicate the results. The majority of the functionality of the engine is used in the startup process, not in the agents themselves.

---

@xied75 commented on [Wed Feb 17 2016](https://github.com/nunit/nunit/issues/1138#issuecomment-185270668)

:+1: to track.

---

@rprouse commented on [Wed Feb 17 2016](https://github.com/nunit/nunit/issues/1138#issuecomment-185272239)

@xied75, for future reference, you can just click the subscribe button at the bottom of the toolbar on the left. That way, all other subscribers don't get an email :wink: 

image

---

@xied75 commented on [Wed Feb 17 2016](https://github.com/nunit/nunit/issues/1138#issuecomment-185274734)

Well this is in fact a new trend on GitHub, since GitHub refuses to add the feature to allow one see what he/she subscribed, so now people do +1 in order to find what they want track.

---

@rprouse commented on [Fri Jun 10 2016](https://github.com/nunit/nunit/issues/1138#issuecomment-225170065)

I am going to move this out of 3.4. The new `dotnet-test-nunit` will run .NET Core and UWP apps and the promise of `netstandard` may change the way that we approach this. I want to see how things go after the release of .NET Core before committing to major changes in the engine.

---

@CharliePoole commented on [Fri Jun 10 2016](https://github.com/nunit/nunit/issues/1138#issuecomment-225171102)

That makes perfect sense.

rprouse:
I would like to start reviewing this approach, but ~~we **should not merge** until end-to end testing with the adapter has been performed.~~ I will add some comments inline with the code as to some of the decisions that I made, but the overall approach is,

We need versions of the VS Adapter that targets the full CLR, .NET Core and UWP to run tests for those platforms in Visual Studio and with the newer version of the `dotnet test` command. Our adapter uses the engine in-process and without many of the extensions, so that is what I am targeting. At first, I attempted to use the `nunit.portable.agent` project, but the majority of that code was unused and I was pulling in too much code from this project, so I switched back to creating a .NET Standard version of the engine.

- This version of the engine is only intended to be used in-process. If we support running .NET Core tests from our current desktop console runner, it will launch a .NET Core agent that uses this engine, but the console runner will not.
- I decided to make it a different NuGet package to differentiate the purpose
- This does not support extensions. There is very little support in .NET Standard for loading assemblies from other directories which makes the plugin model difficult.
- There is no support for AppDomains or separate processes
- This only supports classes needed for the Visual Studio adapter. We can add more as needed, but I think we should keep it lean

For now, I plan on using CI builds of the NuGet package in the adapter, then release an alpha version of this and the adapter so we can get people in the community who are using .NET Core to test and give feedback.

Fixes #10


src/NUnitEngine/nunit.engine.api/TestPackage.cs
```diff
@@ -49,7 +51,14 @@ public TestPackage(string filePath)
 
             if (filePath != null)
             {
+#if NETSTANDARD1_3
+                if (!Path.IsPathRooted(filePath))
```
rprouse: This is one major difference between the full engine and this version. The paths in .NET Standard vary dramatically depending on how you run the project. For example, it you use `dotnet run`, it compiles a project and runs it. The assembly is in a sub-directory, but the current directory is the location of the project file. If you use `dotnet execute` the path is that of the assembly. Because of this, our code that figures out the full path does not work in all cases, so I decided to only allow full paths and leave it up to the calling runner to resolve that path.

src/NUnitEngine/nunit.engine.netstandard/Runners/DirectTestRunner.cs
```diff
@@ -0,0 +1,162 @@
...
+using NUnit.Engine.Drivers;
+using NUnit.Engine.Services;
+
+namespace NUnit.Engine.Runners
+{
+    /// <summary>
+    /// DirectTestRunner is the abstract base for runners 
+    /// that deal directly with a framework driver.
+    /// </summary>
+    public class DirectTestRunner : AbstractTestRunner
```
rprouse: Any engine files that were dramatically different I created a new copy for the .NET Standard engine. They were getting way too complicated with the `#if` statements.

src/NUnitEngine/nunit.engine.tests/Internal/PathUtilTests.cs
```diff
@@ -121,8 +122,9 @@ public void RelativePath()
             Assert.AreEqual(@"..\Folder2\folder3", PathUtils.RelativePath(
                 @"c:\folder1", @"C:\Folder2\folder3"));
         }
+#endif
 
-		[Test]
+        [Test]
```
rprouse: Looks like we have some old mixed tabs and spaces in this file.

src/NUnitEngine/nunit.engine.tests/Internal/PathUtilTests.cs
```diff
@@ -86,7 +86,8 @@ public void Canonicalize()
 				PathUtils.Canonicalize( @"folder1\folder2\..\..\..\file.tmp" ) );
 		}
 
-		[Test]
+#if !NETCOREAPP1_1
```
rprouse: Tests are all a .NET Core app because they are self-executing NUnitLite.

src/NUnitEngine/nunit.engine/Internal/SettingsStore.cs
```diff
@@ -71,18 +73,21 @@ public void LoadSettings()
...
+                using (var stream = new FileStream(_settingsFile, FileMode.Open, FileAccess.Read))
+                {
+                    doc.Load(stream);
+                }
 
                 foreach (XmlElement element in doc.DocumentElement["Settings"].ChildNodes)
                 {
                     if (element.Name != "Setting")
-                        throw new ApplicationException("Unknown element in settings file: " + element.Name);
+                        throw new Exception("Unknown element in settings file: " + element.Name);
```
rprouse: `ApplicationException` is not available 😦

src/NUnitEngine/nunit.engine/Services/ResultService.cs
```diff
@@ -67,6 +77,10 @@ public IResultWriter GetResultWriter(string format, object[] args)
                     return new NUnit3XmlResultWriter();
                 case "cases":
                     return new TestCaseResultWriter();
+#if NETSTANDARD1_3
+                default:
+                    return null;
+#else
                 case "user":
                     return new XmlTransformResultWriter(args);
```
rprouse: None of the XML transform classes we use are available and they aren't used in the VS Adapter.

build.ps1
```diff
@@ -89,7 +89,7 @@ if(!$PSScriptRoot){
 $TOOLS_DIR = Join-Path $PSScriptRoot "tools"
 $NUGET_EXE = Join-Path $TOOLS_DIR "nuget.exe"
 $CAKE_EXE = Join-Path $TOOLS_DIR "Cake/Cake.exe"
-$NUGET_URL = "https://dist.nuget.org/win-x86-commandline/latest/nuget.exe"
+$NUGET_URL = "https://dist.nuget.org/win-x86-commandline/v4.0.0/nuget.exe"
```
rprouse: The .NET Core restore needs the 4.0 version of NuGet, but latest is currently pointing at 3.5. Once latest is updated, this can be reverted.

rprouse:
I can't make much of the build failures on Travis. If anyone has ideas, I would appreciate the help, otherwise I may just not build .NET Core on Linux/Mac 😄 

Could be that Travis hasn't updated .NET core, but it is showing as not installed, so it shouldn't be building and shouldn't be giving MSBuild errors. Possibly because I am using the newer DotNetCore cake tasks? Maybe we need to run --experimental on Travis?

build.ps1
jnm2: I ran into this as well. An alternative which I would recommend is to leave the bootstrapper alone and replace `nuget restore` with `msbuild /t:restore`.
See https://gitter.im/cake-build/cake?at=58d156017b3f37e7541e55b0 and http://blog.nuget.org/20170316/NuGet-now-fully-integrated-into-MSBuild.html

fir3pho3nixx:
@rprouse - is this any help? 

https://github.com/dotnet/docs/blob/master/docs/core/tools/dotnet-install-script.md

You can install the runtime by way of a bash script.

rprouse:
@fir3pho3nixx thanks for that, it might be helpful.

fir3pho3nixx:
Figured this out, the main battle here is figuring out how to get libunwind working on travis using ubuntu trusty. 

This is my travis yaml file that worked. 

```yaml

language: csharp

solution: fortress.sln

mono: none

dotnet: none

sudo: required

dist: trusty

before_install:  

    - sudo apt-get -qq update
    
    - sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
  
    - sudo apt-get install -y libunwind8

    - wget -O - https://raw.githubusercontent.com/dotnet/cli/rel/1.0.0/scripts/obtain/dotnet-install.sh | bash -s -- -i ./dotnetcli/

script:

    - ./dotnetcli/dotnet restore ./src/Fortress.Windsor/Fortress.Windsor.csproj
    
    - ./dotnetcli/dotnet restore ./src/Fortress.Facilities.Logging/Fortress.Facilities.Logging.csproj
    
    - ./dotnetcli/dotnet restore ./src/Fortress.Windsor.Tests/Fortress.Windsor.Tests.csproj
    
    - ./dotnetcli/dotnet test ./src/Fortress.Windsor.Tests/Fortress.Windsor.Tests.csproj

```

fir3pho3nixx:
Try enabling SUDO.

rprouse:
I am going to disable building .NET Core on Travis, but in the meantime, this is ready for review. It creates a .NET Standard version of the engine with the limitations mentioned above. It is working well in the Visual Studio adapter to run tests in Visual Studio and on the command line with `vstest` and `dotnet test`.

My thinking is to release an alpha version of the .NET Standard engine to NuGet and consume it in an alpha release of the adapter.

build.ps1
rprouse: Switching to MSBuild would require v15 which would force all team members and build machines to use Visual Studio 2017.

rprouse:
@nunit/engine-team I cannot seem to get Travis building for the life of me. It is failing in the NuGet restore of the main project for some Travis builds. The nuget restore hasn't changed and the main solution hasn't changed. The only difference that I can think of is that there are project.json files in some of the projects. That is no different than the nunit framework repo though.

I turned up the verbosity of the nuget restore and it is showing the error 

```
NuGet.CommandLine.CommandLineException: 
  at NuGet.CommandLine.MsBuildUtility+<GetProjectReferencesAsync>d__5.MoveNext () <0x40fa8830 + 0x0125b> in <filename unknown>:0 
```

The successful Travis build is finding MSBuild 15, but this is finding MSBuild 14. My Windows box finds 14 and succeeds though.

Any help or ideas would be appreciated.

rprouse:
I've managed to get this building on my Linux box. At this point, it doesn't look like it will require any code changes or changes to `build.cake`, it will only be changes to the Travis environment to ensure the correct tools are installed on Linux and Mac.

Based on that, this is ready to review. I would like feedback from the @nunit/core-team and @nunit/engine-team on the approach. I would also like to do an alpha release of just the .NET Core engine soon.

@CharliePoole, you also had concerns about calling this an engine. I would appreciate it if you could take a look and see if you still feel that way or possibly suggest better naming to prevent confusion.

One other major question is the fact that I copied and rewrote several files rather than use too many `#if` blocks. I worry that decision might cause maintenance problems in the future. Most of the files are shared between the two solutions, but the following files are copies,

- TestEngine.cs
- DirectTestRunner.cs
- MasterTestRunner.cs
- DefaultTestRunnerFactory.cs
- DriverService.cs

rprouse:
CI is now passing on this PR. The Linux Mono 4.8.1 failure is unrelated, see #211.

This can be reviewed and merged when ready.

src/NUnitEngine/nunit.engine.netstandard/Services/DefaultTestRunnerFactory.cs
```diff
@@ -0,0 +1,72 @@
...
+                if (PathUtils.IsAssemblyFileType(testFile))
+                    assemblyCount++;
+            }
+
+            if ( assemblyCount > 0)
+                return new AggregatingTestRunner(package);
+
+            // TODO: What about bad extensions?
+            return new DirectTestRunner(package);
+        }
```
ChrisMaddock: This method seems a little odd now with all the project stuff stripped out. Return an AggregatingTestRunner if there are 'any' assemblies, and a DirectTestRunner only if there are zero? If extensions aren't supported, shouldn't everything be an assembly, and anything else invalid?

src/NUnitEngine/nunit.engine.tests.netstandard/API/TestPackageTests.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        [Test]
+        public void PackageIDsAreUnique()
+        {
+            var another = new TestPackage(@"c:\another.dll");
+            Assert.That(another.ID, Is.Not.EqualTo(package.ID));
+        }
+
+        [Test]
+        public void AssemblyPathIsUsedAsFilePath()
```
ChrisMaddock: Should this be AssemblyPathIsUsedAsFullName?

src/NUnitEngine/nunit.engine.netstandard/Services/DefaultTestRunnerFactory.cs
rprouse: Agreed, but I didn't want to strip out too much that we might end up re-adding. We don't need NUnit project loading or solution loading for the VS Adapter, but I thought that once we move to other runners like Xamarin or UWP, we might end up adding some of the extensions back in.

src/NUnitEngine/nunit.engine.tests.netstandard/API/TestPackageTests.cs
rprouse: Yes, fixing.

rprouse:
@ChrisMaddock thanks for the review, I have pushed an update for one of your comments.

@jnm2 I would appreciate a second code review on this before I merge if you have some time.

@nunit/core-team I plan on putting out an `alpha1` release of this once it is merged so that it can be put into the adapter and we can get an `alpha1` release of that out too. I was planning on calling this 3.7.0-alpha1 in prep for the final 3.7 release. I was planning on calling the adapter 4.0.0-alpha1 because it is a major change in functionality.

Does anyone see any problem with that plan or does anyone have suggestions?

src/NUnitEngine/nunit.engine.netstandard/nunit.engine.netstandard.csproj
```diff
@@ -0,0 +1,174 @@
...
+    <AssemblyName>nunit.engine.netstandard</AssemblyName>
+    <IntermediateOutputPath>obj\$(Configuration)\netstandard\</IntermediateOutputPath>
+    <PackageVersion Condition=" '$(PackageVersion)' == '' ">3.7.0-dev</PackageVersion>
+  </PropertyGroup>
+  <PropertyGroup>
+    <SignAssembly>true</SignAssembly>
+  </PropertyGroup>
+  <PropertyGroup>
+    <AssemblyOriginatorKeyFile>..\..\nunit.snk</AssemblyOriginatorKeyFile>
+    <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
```
jnm2: Same [Pretty sure this is the default and this line could be left out?]

src/NUnitEngine/nunit.engine.netstandard/nunit.engine.netstandard.csproj
```diff
@@ -0,0 +1,174 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>netstandard1.3</TargetFramework>
+    <RootNamespace>NUnit.Engine</RootNamespace>
+    <AssemblyName>nunit.engine.netstandard</AssemblyName>
+    <IntermediateOutputPath>obj\$(Configuration)\netstandard\</IntermediateOutputPath>
```
jnm2: Same [Pretty sure this is the default and this line could be left out?]

src/NUnitEngine/nunit.engine.netstandard/nunit.engine.netstandard.csproj
```diff
@@ -0,0 +1,174 @@
...
+    <Compile Include="..\nunit.engine\Runners\TestExecutionTask.cs">
+      <Link>Runners\TestExecutionTask.cs</Link>
+    </Compile>
+    <Compile Include="..\nunit.engine\Services\ResultService.cs" Link="Services\ResultService.cs" />
+    <Compile Include="..\nunit.engine\Services\ResultWriters\NUnit3XmlResultWriter.cs" Link="Services\ResultWriters\NUnit3XmlResultWriter.cs" />
+    <Compile Include="..\nunit.engine\Services\ResultWriters\TestCaseResultWriter.cs" Link="Services\ResultWriters\TestCaseResultWriter.cs" />
+    <Compile Include="..\nunit.engine\Services\TestFilterBuilder.cs" Link="Services\TestFilterBuilder.cs" />
+    <Compile Include="..\nunit.engine\Services\TestFilterService.cs" Link="Services\TestFilterService.cs" />
+    <Compile Include="..\nunit.engine\Services\TestSelectionParser.cs" Link="Services\TestSelectionParser.cs" />
+    <Compile Include="..\nunit.engine\Services\Tokenizer.cs" Link="Services\Tokenizer.cs" />
```
jnm2: Seems like a glob could take care of almost all these items linked from the `..\nunit.engine` folder. That's one less csproj to constantly be touching. :D (I can not wait until we do this with all the projects!)

src/NUnitEngine/nunit.engine/Services/ResultWriters/TestCaseResultWriter.cs
```diff
@@ -32,15 +32,16 @@ public class TestCaseResultWriter : IResultWriter
...
             {
                 // Opening is enough to check
             }
         }
 
         public void WriteResultFile(XmlNode resultNode, string outputPath)
         {
-            using (var writer = new StreamWriter(outputPath, false, Encoding.UTF8))
+            using (var stream = new FileStream(outputPath, FileMode.Create, FileAccess.Write))
+            using (var writer = new StreamWriter(stream))
```
jnm2: This is better because it no longer writes a byte order mark.

src/NUnitEngine/nunit.engine.tests.netstandard/nunit.engine.tests.netstandard.csproj
```diff
@@ -0,0 +1,58 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp1.1</TargetFramework>
+    <SignAssembly>True</SignAssembly>
+    <AssemblyOriginatorKeyFile>..\..\nunit.snk</AssemblyOriginatorKeyFile>
+    <DelaySign>False</DelaySign>
```
jnm2: Pretty sure this is the default and this line could be left out?

src/NUnitEngine/nunit.engine.tests.netstandard/nunit.engine.tests.netstandard.csproj
```diff
@@ -0,0 +1,58 @@
...
+    <GenerateAssemblyCompanyAttribute>false</GenerateAssemblyCompanyAttribute>
+    <GenerateAssemblyConfigurationAttribute>false</GenerateAssemblyConfigurationAttribute>
+    <GenerateAssemblyCopyrightAttribute>false</GenerateAssemblyCopyrightAttribute>
+    <GenerateAssemblyDescriptionAttribute>false</GenerateAssemblyDescriptionAttribute>
+    <GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>
+    <GenerateAssemblyInformationalVersionAttribute>false</GenerateAssemblyInformationalVersionAttribute>
+    <GenerateAssemblyProductAttribute>false</GenerateAssemblyProductAttribute>
+    <GenerateAssemblyTitleAttribute>false</GenerateAssemblyTitleAttribute>
+    <GenerateAssemblyVersionAttribute>false</GenerateAssemblyVersionAttribute>
```
jnm2: Same [These 9 lines can be replaced with `<GenerateAssemblyInfo>false</GenerateAssemblyInfo>`]

src/NUnitEngine/nunit.engine.tests.netstandard/nunit.engine.tests.netstandard.csproj
```diff
@@ -0,0 +1,58 @@
...
+    <Compile Include="..\nunit.engine.tests\Internal\DirectoryFinderTests.cs" Link="Internal\DirectoryFinderTests.cs" />
+    <Compile Include="..\nunit.engine.tests\Internal\PathUtilTests.cs" Link="Internal\PathUtilTests.cs" />
+    <Compile Include="..\nunit.engine.tests\Internal\SettingsGroupTests.cs" Link="Internal\SettingsGroupTests.cs" />
+    <Compile Include="..\nunit.engine.tests\Properties\AssemblyInfo.cs" Link="Properties\AssemblyInfo.cs" />
+    <Compile Include="..\nunit.engine.tests\Runners\MasterTestRunnerTests.cs" Link="Runners\MasterTestRunnerTests.cs" />
+    <Compile Include="..\nunit.engine.tests\Services\ResultServiceTests.cs" Link="Services\ResultServiceTests.cs" />
+    <Compile Include="..\nunit.engine.tests\Services\TestFilterBuilderTests.cs" Link="Services\TestFilterBuilderTests.cs" />
+    <Compile Include="..\nunit.engine.tests\Services\TestFilteringTests.cs" Link="Services\TestFilteringTests.cs" />
+    <Compile Include="..\nunit.engine.tests\Services\TestSelectionParserTests.cs" Link="Services\TestSelectionParserTests.cs" />
+    <Compile Include="..\nunit.engine.tests\Services\TokenizerTests.cs" Link="Services\TokenizerTests.cs" />
```
jnm2: Same [Seems like a glob could take care of almost all these items linked from the `..\nunit.engine` folder.]

src/NUnitEngine/nunit.engine.netstandard/nunit.engine.netstandard.csproj
jnm2: Oops, removing the comment on this line.

src/NUnitEngine/nunit.engine.netstandard/nunit.engine.netstandard.csproj
rprouse: Possibly, but this is just a fraction of the files in the Engine and API projects so I would have to add `#if !NETSTANDARD1_3` to all of those. That may happen when/if the projects ever merge, but for now I wanted to keep this as unobtrusive as possible.

src/NUnitEngine/nunit.engine.tests.netstandard/nunit.engine.tests.netstandard.csproj
rprouse: Same, there are 53 CS files for the main test suite, we are only pulling in 14.

src/NUnitEngine/mock-assembly/mock-assembly.netstandard.csproj
```diff
@@ -0,0 +1,39 @@
...
+    <GeneratePackageOnBuild>False</GeneratePackageOnBuild>
+    <GenerateAssemblyCompanyAttribute>false</GenerateAssemblyCompanyAttribute>
+    <GenerateAssemblyConfigurationAttribute>false</GenerateAssemblyConfigurationAttribute>
+    <GenerateAssemblyCopyrightAttribute>false</GenerateAssemblyCopyrightAttribute>
+    <GenerateAssemblyDescriptionAttribute>false</GenerateAssemblyDescriptionAttribute>
+    <GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>
+    <GenerateAssemblyInformationalVersionAttribute>false</GenerateAssemblyInformationalVersionAttribute>
+    <GenerateAssemblyProductAttribute>false</GenerateAssemblyProductAttribute>
+    <GenerateAssemblyTitleAttribute>false</GenerateAssemblyTitleAttribute>
+    <GenerateAssemblyVersionAttribute>false</GenerateAssemblyVersionAttribute>
```
rprouse: That cleans things up, thanks. Do you know where all of the new options are documented? I've just been looking through examples of Microsoft projects.

src/NUnitEngine/nunit.engine.api/Extensibility/IFrameworkDriver.cs
```diff
@@ -39,7 +40,7 @@ public interface IFrameworkDriver
         /// used to ensure that test ids are unique across drivers.
         /// </summary>
         string ID { get; set; }
-
+
```
jnm2: Today I also discovered a VS Code extension for this too (https://marketplace.visualstudio.com/items?itemName=shardulm94.trailing-spaces) which is totally awesome. I enabled [trim on save](https://marketplace.visualstudio.com/items?itemName=shardulm94.trailing-spaces#trim-on-save).

src/NUnitEngine/nunit.engine.netstandard/nunit.engine.netstandard.csproj
jnm2: Oh, good strategy.

jnm2:
@rprouse 

> That cleans things up, thanks. Do you know where all of the new options are documented? I've just been looking through examples of Microsoft projects.

[Tell me about it.](https://github.com/NuGet/Home/issues/4059#issuecomment-297854399) 😄 
I still can't locate documentation for `GenerateAssemblyInfo`. Got it off dotnet/project-system.

I've used these recently:

https://docs.microsoft.com/en-us/dotnet/articles/core/tools/csproj
https://docs.microsoft.com/en-us/dotnet/articles/core/tools/project-json-to-csproj
http://www.natemcmaster.com/blog/2017/03/09/vs2015-to-vs2017-upgrade/
https://docs.microsoft.com/en-us/dotnet/articles/core/migration/

PR #208, opened 2017-04-12
Explore flags test update

### LINKED ISSUE #139: Console runner - Explore flag does not apply where filter to output
@chriscore commented on [Thu Dec 17 2015](https://github.com/nunit/nunit/issues/1143)

Hi there

Running nunit3-console.exe with arguments similar to the following does not apply the filter specified in /where to the results which are seen on the output.

"C:\MyRepo\MyAssembly.dll" /explore /noheader /where="cat==Awesome"

All tests are returned by /explore, including those which do not satisfy cat==Awesome

This was found in NUnit Console Runner v3.0.5813

---

@CharliePoole commented on [Thu Dec 17 2015](https://github.com/nunit/nunit/issues/1143#issuecomment-165499679)

I was uncertain whether it made sense to use a filter when listing tests, but I suppose it does if you tried it!

---

@chriscore commented on [Thu Dec 17 2015](https://github.com/nunit/nunit/issues/1143#issuecomment-165501578)

Hey Charlie

Thanks, before a test run I am currently trying to call this to see how many tests my filter would return - it would certainly be very useful for me at this point :)

Great work with nunit 3 so far

Chris

---

@CharliePoole commented on [Thu Dec 17 2015](https://github.com/nunit/nunit/issues/1143#issuecomment-165502901)

@chriscore If you have a runner, just call CountTestCases on it with your filter!

If you are simply scripting nunit3-console, that's not an option, of course.

---

@chriscore commented on [Thu Dec 17 2015](https://github.com/nunit/nunit/issues/1143#issuecomment-165505549)

@CharliePoole I don't currently have a runner instance as I haven't been able to figure out how to get that going, like you say I have just scripted nunit3-console, but having that control in process would be ideal.

I did find it difficult to find how to host a runner as there seemed to be a lot of dependent objects which I found difficult to create, do you have a C# snippet anywhere which might give some clues how to start? Being able to load and run tests in process would be ideal for what I'm trying to do here.

---

@CharliePoole commented on [Thu Dec 17 2015](https://github.com/nunit/nunit/issues/1143#issuecomment-165509169)

Add a reference to nunit.engine.api.dll. Then (simplified)...

```
ITestEngine engine = TestEngineActivator.CreateInstance();
TestPackage package = new TestPackage(ASSEMBLY_PATH);
ITestRunner runner = engine.GetRunner(package);
```

This leaves out setting options, using a filter, etc. For the full story, start here:
    https://github.com/nunit/nunit/blob/master/src/NUnitConsole/nunit3-console/Program.cs#L111

---

@chriscore commented on [Thu Dec 17 2015](https://github.com/nunit/nunit/issues/1143#issuecomment-165511283)

Thanks so much for that, I didn't know about the TestEngineActivator!

---

@rprouse commented on [Thu Apr 07 2016](https://github.com/nunit/nunit/issues/1143#issuecomment-206954366)

When this is fixed, ensure it also works with NUnitLite. See #1213

---

@NCarter3 commented on [Tue Jun 21 2016](https://github.com/nunit/nunit/issues/1143#issuecomment-227305222)

I'd also love to see filters working with --explore!

Poimen:
Fixes #139 

Updated Console driver tests for explore functionality. This test requires the framework changes from nunit/nunit#2094 PR.

This change just updates the test case for `Explore` as the rest of the console application did not appear need any new changes.

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/Drivers/NUnit3FrameworkDriverTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Drivers/NUnit3FrameworkDriverTests.cs
@@ -105,6 +105,22 @@ namespace NUnit.Engine.Drivers.Tests
             Assert.That(ex, Is.TypeOf<InvalidOperationException>());
             Assert.That(ex.Message, Is.EqualTo(LOAD_MESSAGE));
         }
+
+        [TestCase("<filter><cat>MockCategory</cat></filter>")]
+        [TestCase("<filter><cat>FixtureCategory</cat></filter>")]
+        [TestCase("<filter><or><cat>MockCategory</cat><cat>FixtureCategory</cat></or></filter>")]
+        [TestCase("<filter><and><cat>MockCategory</cat><cat>FixtureCategory</cat></and></filter>")]
+        [TestCase("<filter><not><cat>MockCategory</cat></not></filter>")]
+        [TestCase("<filter><and><not><cat>MockCategory</cat></not><not><cat>FixtureCategory</cat></not></and></filter>")]
+        public void Explore_WithCatagoryFilter_ReturnsFilteredSuite(string categoryFilter)
+        {
+            _driver.Load(_mockAssemblyPath, _settings);
+            var result = XmlHelper.CreateXmlNode(_driver.Explore(categoryFilter));
+
+            var testCaseCount = result.SelectNodes("//test-case").Count;
+
+            Assert.That(_driver.CountTestCases(categoryFilter), Is.EqualTo(testCaseCount), "Explore did not return expected test count");
+        }
         #endregion
 
         #region CountTests


CharliePoole:
I'm not clear as to what your purpose is here. The addition is only of a test, with no behavior being changed. The driver has no control over what is returned to it, so what exactly are we testing?

Also... are you saying this will work differently (succeed vs fail) depending on the version of the framework used? If that's what you mean, then it violates the isolation of the engine from the framework and might be better done in an integration test. 

We have talked about, an integration test suite that would - among other things - mix and match various component versions. Unfortunately, we haven't created it yet. 😞

Poimen:
The intention was to add more coverage to the `Explore` function. As I saw it the `Explore` function only contained 2 tests that didn't actually test the `Explore` function expected results. Hence there was no test that I could see that would fail if `Explore` was changed. So when the user uses the `/explore` option, the intention was that the test mimics that operation to ensure `Explore` worked as expected.

The purpose was more test coverage.

If there is already

CharliePoole:
OK, but you say the test requires the framework changes. What do you mean by that?

Poimen:
By the framework changes I meant the changes that actually have the `Explore` functionality working correctly. The test tests for the `--explore` option to return a filtered list. Seeing that the filtering happens in the framework, the framework functionality is required in order to complete the test successfully.

rprouse:
The framework changes have been merged, so you should be able to update to use the CI build of the framework for this PR.

ChrisMaddock:
It would be good to get this resolved for the 3.7 release at the end of the week.

@CharliePoole - you had some concerns about whether this test belongs here, or if it should be in an integration test suite one day. What do you think, should this be resolved and merged, or is it not a good fit?

CharliePoole:
I don't see this as relevant to any near-term release. The tests all fail. @Poimen explains they will pass when we are using the 3.7 framework for the tests. This presents two problems:

1) We don't want tests of the console that require a certain version of the framework to work.
2) The tests do not test anything that the console or engine actually implement.

A valid __unit__ test of this functionality would simply be that the test returns whatever it gets from the framework - ANY framework. Probably a mock is needed.

ChrisMaddock:
> A valid unit test of this functionality would simply be that the test returns whatever it gets from the framework - ANY framework.

That's presumably unrelated to whether a filter is passed or not, and theoretically should already be tested for the normal `Explore()` functionality, right? We have a test that tests this method with `TestFilter.Empty` already. Perhaps the unit test would be the inverse - testing that the framework receives the filter passed into the driver. I guess a test could potentially test the filter is round tripped and output in the xml's results - although this seems to become more tenuous.

As we don't currently have this integration test assembly - should we close this?

CharliePoole:
I don't believe we ever round-trip filters to/from the framework. They only go one way.

Anyway, I think we should close this.

ChrisMaddock:
Ok, let's go with that then. Thanks anyway @Poimen.

PR #209, opened 2017-04-20
Fixed issue with sending extra new lines when writing to listeners

### LINKED ISSUE #72: TestContext.Progress.Write writes new line
@yoyos06 commented on [Mon Oct 03 2016](https://github.com/nunit/nunit/issues/1819)

Hi,

The TestContext.Progress.Write method now works exactly the same as TestContext.Progress.WriteLine because it writes a newline character at the end.

---

@rprouse commented on [Mon Oct 03 2016](https://github.com/nunit/nunit/issues/1819#issuecomment-251127409)

Thanks for the report, we will look into it.

---

@CharliePoole commented on [Mon Oct 03 2016](https://github.com/nunit/nunit/issues/1819#issuecomment-251230619)

@yoyos06 Actually, TestContext.Progress.Write doesn't write a newline after the text. It's done in NUnit console.

That might seem like an academic distinction, but it's actually important. The framework puts out events which are received and handled by runners. Each runner may handle them in a different way. For example, some gui might have a special text box that received all the "Progress" output.

In the case of the console runner, the output is handled in a line-oriented way. Every write currently gets a newline because of some earlier errors where lines intended to be separate were run together. We need to be a little smarter about this.

What needs to happen (I'm writing this for whomever implements the fix) is that no newline should be written until it's actually needed. That means the console's TestEventHandler class has to maintain a bit of state regarding whether the last bit of output had a newline attached. When output arrives from some other test it should supply the newline, if it wasn't there already, before writing the new output. It's potentially complicated, but doable.

Since this is a console issue, I'm copying this there. I'm leaving this issue open to deal with NUnitLite, which no doubt has the same problem.

MoFtZ:
Fixes #72

This pull request tracks the current test (by full name), and will automatically insert a new line when the test changes.

### DIFF
--- a/src/NUnitConsole/nunit3-console/TestEventHandler.cs
+++ b/src/NUnitConsole/nunit3-console/TestEventHandler.cs
@@ -42,6 +42,9 @@ namespace NUnit.ConsoleRunner
         private readonly bool _displayAfterTest;
         private readonly bool _displayBeforeOutput;
 
+        private string _lastTestOutput;
+        private bool _wantNewLine = false;
+
         public TestEventHandler(TextWriter outWriter, string labelsOption)
         {
             _outWriter = outWriter;
@@ -103,7 +106,8 @@ namespace NUnit.ConsoleRunner
                 if (_displayBeforeOutput)
                     WriteLabelLine(testName);
 
-                WriteOutputLine(outputNode.InnerText);
+                FlushNewLineIfNeeded();
+                WriteOutputLine(testName, outputNode.InnerText);
             }
 
             if (_displayAfterTest)
@@ -120,7 +124,8 @@ namespace NUnit.ConsoleRunner
                 if (_displayBeforeOutput)
                     WriteLabelLine(suiteName);
 
-                WriteOutputLine(outputNode.InnerText);
+                FlushNewLineIfNeeded();
+                WriteOutputLine(suiteName, outputNode.InnerText);
             }
         }
 
@@ -132,7 +137,7 @@ namespace NUnit.ConsoleRunner
             if (_displayBeforeOutput && testName != null)
                 WriteLabelLine(testName);
 
-            WriteOutputLine(outputNode.InnerText, stream == "Error" ? ColorStyle.Error : ColorStyle.Output);
+            WriteOutputLine(testName, outputNode.InnerText, stream == "Error" ? ColorStyle.Error : ColorStyle.Output);
         }
 
         private string _currentLabel;
@@ -141,6 +146,9 @@ namespace NUnit.ConsoleRunner
         {
             if (label != _currentLabel)
             {
+                FlushNewLineIfNeeded();
+                _lastTestOutput = label;
+
                 using (new ColorConsole(ColorStyle.SectionHeader))
                     _outWriter.WriteLine("=> {0}", label);
 
@@ -150,6 +158,9 @@ namespace NUnit.ConsoleRunner
 
         private void WriteLabelLineAfterTest(string label, string status)
         {
+            FlushNewLineIfNeeded();
+            _lastTestOutput = label;
+
             if (status != null)
                 using (new ColorConsole(GetColorForResultStatus(status)))
                     _outWriter.Write("{0} ", status);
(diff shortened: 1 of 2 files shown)


MoFtZ:
Tests have failed.

Should the tests or code be updated?

rprouse:
@MoFtZ how does this work when tests are run in parallel? Isn't their potentially a concurrently issue with some of the label flags that output test output continuously and therefore might jump between tests?

MoFtZ:
@rprouse it should deal with concurrency just fine.

My understanding is that the TestEventHandler is not re-entrant, and calls to the OnTestEvent() method are serialised by the EventPump. This has not changed in this PR, so there should be no change in behaviour here.

The main change in this PR is the introduction of calls to FlushNewLineIfNecessary(). When OnTestEvent() outputs some text, we track whether it had a trailing newline. If it does not, we set a flag to indicate that we MAY need a newline. On subsequent calls to OnTestEvent(), we check whether the output came from the same test, or a different test - if the same test, we do nothing. If a different test, we output a newline ONLY IF there was no previous newline. This logic occurs before all the logic for labels, so as not to interfere with that part of the code.

In addition, with the PR in the NUnit Framework (https://github.com/nunit/nunit/pull/2129), each call to Write/WriteLine should now generate a matching call to OnTestEvent(). Without that PR, a call to Write(char[]) would generate multiple calls to OnTestEvent(). That PR therefore reduces the likelihood of the output being a mixture from multiple tests.

CharliePoole:
I've held off reviewing this until the framework issue is resolved and I'm still holding off. :smile:

However, to respond to the above discussion, this code definitely has to work in a situation of parallel tests. While it is true that the framework events are serialized through the use of the event pump and listener, there may be multiple instances of the framework running tests in different assemblies and sending the output to the console runner simultaneously.

Let's focus on nunit plus nunitlite first.

src/NUnitConsole/nunit3-console/TestEventHandler.cs
```diff
@@ -88,6 +91,8 @@ private void TestStarted(XmlNode testResult)
         {
             var testName = testResult.Attributes["fullname"].Value;
 
+            FlushNewLineIfNecessary(testName);
```
CharliePoole: Why flush here if the new test that is starting isn't going to write anything?

src/NUnitConsole/nunit3-console/TestEventHandler.cs
```diff
@@ -98,6 +103,8 @@ private void TestFinished(XmlNode testResult)
             var status = testResult.GetAttribute("label") ?? testResult.GetAttribute("result");
             var outputNode = testResult.SelectSingleNode("output");
 
+            FlushNewLineIfNecessary(testName);
+
```
CharliePoole: Same question as above: in this case a test has ended. Depending on the settings, it may or may not be about to write anything. If it isn't going to write anything, then it's still possible that the test with a pending newline may finish the line before any other test produces output.

src/NUnitConsole/nunit3-console/TestEventHandler.cs
```diff
@@ -115,6 +122,8 @@ private void SuiteFinished(XmlNode testResult)
             var suiteName = testResult.Attributes["fullname"].Value;
             var outputNode = testResult.SelectSingleNode("output");
 
+            FlushNewLineIfNecessary(suiteName);
+
```
CharliePoole: Same as above

src/NUnitConsole/nunit3-console/TestEventHandler.cs
```diff
@@ -129,6 +138,8 @@ private void TestOutput(XmlNode outputNode)
             var testName = outputNode.GetAttribute("testname");
             var stream = outputNode.GetAttribute("stream");
 
+            FlushNewLineIfNecessary(testName);
+
```
CharliePoole: In this case, it seems like the right place. Because the new test is calling TestOutput we know it is about to write.

src/NUnitConsole/nunit3-console/TestEventHandler.cs
```diff
@@ -171,11 +184,25 @@ private void WriteOutputLine(string text, ColorStyle color)
...
+        {
+            if (_currentTestName != testName)
+            {
+                if (_wantNewLine)
+                {
                     _outWriter.WriteLine();
+                    _wantNewLine = false;
                 }
+
+                _currentTestName = testName;
```
CharliePoole: Can this be incorporated with or called from the methods that actually produce output?

src/NUnitConsole/nunit3-console/TestEventHandler.cs
```diff
@@ -42,6 +42,9 @@ public class TestEventHandler : MarshalByRefObject, ITestEventListener
         private readonly bool _displayAfterTest;
         private readonly bool _displayBeforeOutput;
 
+        private string _currentTestName;
```
CharliePoole: _lastTestOutput ? or something like that? IOW, we use "current" to mean the label under which we are producing output in _currentLabel, so _currentTestName seems a tad confusing.

CharliePoole:
@MoFtZ You may want to take a peek at the changes I included in PR nunit/nunit#2141.

src/NUnitConsole/nunit3-console/TestEventHandler.cs
MoFtZ: Ok, will change the name.

_currentTestName was chosen because it is often initialised with the same value as _currentLabel.

MoFtZ:
Unfortunately, I am not able to run the nunit-console unit tests in VS2017 - approx 35 test fail.

A quick look suggests that it is looking for the Mock-Assembly.DLL in the wrong folder path. I have instead run the tests from the command line.

MoFtZ:
Also, yes, the unit tests that failed are because they expected Write to include newlines.

I have modified the unit tests to match the desired behaviour.

CharliePoole:
Thanks. I'll review again after you update the PR.

CharliePoole:
Looks ready to merge as soon as we get a second committer to approve it.

MoFtZ:
Question about expected behaviour of when a suite is finished, with labels = AFTER.. we do not output the label or suite result?

CharliePoole:
Currently, we only output a suite label if the suite has produced output, your test shows that working so I feel like we are OK. Producing Before or After labels for suites would be a lot of output and we have traditionally focused on test cases.

MoFtZ:
Makes sense, thanks for the clarification.

rprouse:
@MoFtZ looks good, thanks for the help 👍

PR #210, opened 2017-04-23
Resolve Argument File processing differences with NUnitLite

### LINKED ISSUE #126: Resolve differences between NUnit Console and NUnitLite implementations of @filename
@CharliePoole commented on [Mon Oct 31 2016](https://github.com/nunit/nunit/issues/1877)

We (well actually I) mistakenly merged the nunitlite changes thinking that the differences between the console runner and nunitlite implementations had already been resolved between @NikolayPianikov and @oznetmaster 

We need to decide first what features to include / exclude then change one or both of them so that users do not get confused and features we have not decided to support do not become supported by default.

Eventually, we should also resolve code differences for the common features but that can await the point where we create a shared codebase for conmmand-line option processing.

I'm putting this in the discussion pipeline so that the NUnit team can work out what we want to support. Everyone else is welcome to participate in the discussion as well.

---

@NikolayPianikov commented on [Tue Nov 01 2016](https://github.com/nunit/nunit/issues/1877#issuecomment-257515218)

I've used Regex from @oznetmaster and I could add the nesting of any levels during 5 minutes if it is needed. I agree it would be better to have one code base and it doesn't matter for me which code will be preferable. I am just interested in implementation of this feature.

So I would like to highlight some questions:

About requirements:
- We should clarify a list of requirements.
- Should we implement minimum sufficient set of requirements? Or ...
    - Will we support the nesting?
    - Will we support symbol " within the argument?

About implementation:
- Should we duplicate a code or should we use some common assembly?
- Which code we will choose?
- Who will be responsible for implementation?
- Is it possible to use our Linq in the implementation (additional reference)?

---

@CharliePoole commented on [Tue Nov 01 2016](https://github.com/nunit/nunit/issues/1877#issuecomment-257523352)

I support @oznetmaster 's proposal in PR #1856:

"I guess I am recommending we leave at least one level of nesting (currently my code has two - it is only a constant), and remove the alternative delimiter. Then modify the console code to support the same syntax."

I think this is a minimum fix that will keep users from getting confused.

As already stated, I don't support trying to make the code common right now. There's a bigger project I would like to do that would include this: developing an entire shared codebase for option handling, including our own version of Options.cs and using a submodule to share the code. But we have many more high priority things to worry about right now, so I think this can wait. So long as the two bits of code actually __do__ the same thing, we will avoid complaints from our users.

---

@CharliePoole commented on [Tue Nov 01 2016](https://github.com/nunit/nunit/issues/1877#issuecomment-257524550)

I intended this to be in the Discussion pipeline when I created it but didn't put it there.

@NikolayPianikov Once we are done discussing and have decided on a course of action, we'll move it to the Backlog. Someone will decide to do it or we will pick someone. Then they will do it. :smile:

---

@CharliePoole commented on [Sat Nov 05 2016](https://github.com/nunit/nunit/issues/1877#issuecomment-258639412)

OK... I don't see any action on this folks, so I'll lay out a proposal...

Let's allow a single level of nesting and remove the alternative delimiter as @oznetmaster suggested.

If this is agreed upon, then we need to
1. Add a single level of nesting to NUnit3-console.
2. Remove extra nesting level from NUnitLite
3. Remove alternative delimiter from NUnitLite

@rprouse @oznetmaster @NikolayPianikov Do you agree?

---

@oznetmaster commented on [Sat Nov 05 2016](https://github.com/nunit/nunit/issues/1877#issuecomment-258641276)

The number of levels of nesting in my code is a constant.  It is only there to eliminate the possibility of infinite recursion.  It serves no other function.

Removing the alternative delimiter is easy.

---

@rprouse commented on [Sat Nov 05 2016](https://github.com/nunit/nunit/issues/1877#issuecomment-258647455)

Agreed to all points.

---

@CharliePoole commented on [Sat Nov 05 2016](https://github.com/nunit/nunit/issues/1877#issuecomment-258656991)

Replicating the issue for nunit-console

CharliePoole:
Fixes #126 

Made the following changes to the console runner options to bring it in line with NUnitLite's features:

1. Implemented nesting of argument files to a maximum depth of 3
2. Allowed arguments in quotes to extend to multiple lines in an argument file
3. Fixed the implementation of comments in argument files
4. To keep consistency with nunitlite, I ended up eliminating the separate ArgumentsFileParser class.
5. Renamed various things for consistency.


rprouse:
Looks good now, merging.

PR #212, opened 2017-04-25
Add .gitattributes entry so that bash scripts checked out in Windows have Unix line endings

jnm2:
As requested https://github.com/nunit/nunit/pull/2136#issuecomment-297108156


rprouse:
Thanks for doing this. The Travis build failures are yet more random failures, so I am going to go ahead and  merge.

PR #214, opened 2017-04-30
Remove naming of build based on Git tag and move versioning to Setup

### LINKED ISSUE #213: Master build failing after merging .NET Standard Engine
Not sure why the PR builds pass, but the master build is now failing, but the copying of the .NET Standard nuget package is failing.

The package that is created is,

```
src\NUnitEngine\nunit.engine.netstandard\bin\Release\nunit.engine.netstandard.3.7.0-alpha1.nupkg
```

The package that is being copied is,

```
src/NUnitEngine/nunit.engine.netstandard/bin/Release/nunit.engine.netstandard.3.7-alpha1.nupkg
```

Notice one is 3.7.0, but the other is just 3.7.

rprouse:
Fixes #213 

- Move the versioning code to `Setup` from `InitializeBuild` so that it happens no matter which dependencies are called
- Remove the code that sets the `packageVersion` based on a git tag because we are not using it.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -16,7 +16,7 @@ var ErrorDetail = new List<string>();
 //////////////////////////////////////////////////////////////////////
 
 var version = "3.7.0";
-var modifier = "-alpha1";
+var modifier = "";
 
 var isAppveyor = BuildSystem.IsRunningOnAppVeyor;
 var dbgSuffix = configuration == "Debug" ? "-dbg" : "";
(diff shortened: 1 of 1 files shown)


build.cake
```diff
@@ -56,6 +56,37 @@ bool IsDotNetCoreInstalled = false;
...
+            packageVersion = version + "-dev-" + buildNumber + dbgSuffix;
+        }
+        else
+        {
+            var suffix = "-ci-" + buildNumber + dbgSuffix;
+
+            if (isPullRequest)
+                suffix += "-pr-" + AppVeyor.Environment.PullRequest.Number;
+            else if (AppVeyor.Environment.Repository.Branch.StartsWith("release", StringComparison.OrdinalIgnoreCase))
+                suffix += "-pre-" + buildNumber;
```
CharliePoole: I've been tempted to remove this branch. Do you make much use of the -pre- tag?

rprouse: I never use the `-pre-` tags either. I only test the locally built versions as I am doing a release.

PR #221, opened 2017-05-07
Added missing nuget package info

### LINKED ISSUE #217: NUnit .NET Standard NuGet package missing some values
The .NET Standard nuget package is created from the CSPROJ file. Not all values that we set in the main nuget package are currently set.

rprouse:
Fixes #217 

I had thought that packaging nuget packages in .NET Core would pull data from the assembly info, but that was not correct, so adding it to the project file.

Here is the result,

image


jnm2:
Yes. Ideally both the assembly info and the nuspec are generated from the project properties.

jnm2:
One thing I have done is `<Import Project="..\SharedAssemblyInfo.props" />` like I used to do with `<Compile Include="..\SharedAssemblyInfo.cs" />`.

rprouse:
The failing travis-ci/pr was a NuGet restore timeout. I restarted and it happened again. Third time lucky?

CharliePoole:
@jnm2 Kindly check your email. 😄

rprouse:
> Ideally both the assembly info and the nuspec are generated from the project properties.

I am trying to make as few changes to the main solution as possible, including the shared assembly info files. I don't want to get into generating the files until we decide to switch to VS2017 for this repo and move to the new CSPROJ format.

src/NUnitEngine/nunit.engine.netstandard/nunit.engine.netstandard.csproj
```diff
@@ -12,6 +12,11 @@
     <AssemblyOriginatorKeyFile>..\..\nunit.snk</AssemblyOriginatorKeyFile>
     <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+    <PackageId>nunit.engine.netstandard</PackageId>
+    <Title>NUnit .NET Standard Engine</Title>
```
ChrisMaddock: Suggestion - we have problems with naming lots of similar NuGet packages. This is equivalent to the package we call "NUnit Test Engine" - correct? How about "NUnit Test Engine - .NET Standard" or something, to reflect that these are related?

src/NUnitEngine/nunit.engine.netstandard/nunit.engine.netstandard.csproj
```diff
@@ -12,6 +12,11 @@
...
+    <Authors>Charlie Poole</Authors>
+    <Description>Provides a common interface for loading, exploring and running NUnit tests in .NET Core and .NET Standard</Description>
```
ChrisMaddock: "Provides a common interface" makes me think this would be the nunit.engine.api assembly (which we also have a NuGet package for) - but this is the full engine, isn't it? Could the description be reworded - or possibly the same as the main package, with .NET Core and .NET Standard named, of course.

ChrisMaddock:
Just seen the below in VS 2015 - I think this should have an `<Owners>` tag too. Our other packages do.

image

CharliePoole:
I'm not in love with this approach of generating nuget packages from the project file. Do I understand that this is somehow required for .NET Core - i.e. we can't just do it with a nuspec file as we always have? IF we have to do this, what IDEs does it rule out, if any. Are we deciding (silently) that it's OK to not be able to create packages on certain OSs and IDEs? If so, I don't like that direction but could easily be out-voted. But it's the "silently" part that really bothers me.

jnm2:
It is completely optional to `msbuild /t:pack`. Nuspecs still work as they always have for any project, including .NET Framework and .NET Core. It's been a big convenience for me for my company's .NET Framework projects, but that may not be true for NUnit at this time.

It also requires VS2017 until they start shipping the SDK as a NuGet package.

PR #222, opened 2017-05-10
Improve missing agent error message

ChrisMaddock:
A user on Gitter today encountered this error message:

    NUnit components for version 4.0.30319 of the CLR are not installed Parametername: targetRuntime

This error is thrown in the specific case where the agent executable can't be found. This PR improves that error message to something more targeted, which a user may be able to resolve without looking at source.

### DIFF
--- a/src/NUnitEngine/nunit.engine/Services/TestAgency.cs
+++ b/src/NUnitEngine/nunit.engine/Services/TestAgency.cs
@@ -143,7 +143,7 @@ namespace NUnit.Engine.Services
             bool debugAgent = package.GetSetting(EnginePackageSettings.DebugAgent, false);
             string traceLevel = package.GetSetting(EnginePackageSettings.InternalTraceLevel, "Off");
             bool loadUserProfile = package.GetSetting(EnginePackageSettings.LoadUserProfile, false);
-            
+
             // Set options that need to be in effect before the package
             // is loaded by using the command line.
             string agentArgs = "--pid=" + Process.GetCurrentProcess().Id.ToString();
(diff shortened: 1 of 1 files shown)


src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -161,10 +161,9 @@ private Guid LaunchAgentProcess(TestPackage package)
 
             string agentExePath = GetTestAgentExePath(useX86Agent);
 
-            if (agentExePath == null)
-                throw new ArgumentException(
-                    string.Format("NUnit components for version {0} of the CLR are not installed",
-                    targetRuntime.ClrVersion.ToString()), "targetRuntime");
+            if (!File.Exists(agentExePath))
+                throw new FileNotFoundException(
+                    $"{Path.GetFileName(agentExePath)} for version {targetRuntime.ClrVersion.ToString()} of the CLR could not be found.", agentExePath);
```
CharliePoole: The message is a holdover from when we had an optional agent for running .NET 1.1 tests. The agent no longer varies by framework version. We should probably drop the version part of the message and just say we can't find the agent at the path used.

ChrisMaddock: Ahh, thanks. I did wonder - I wondered if this was future proofed, as we've talked about having different agents to run on different platforms.

I've removed it - we can add it back when we need it. 😄

ChrisMaddock:
I've edited through the web GUI because I'm lazy - will squash-merge once CI has passed. :smil

CharliePoole:
Looks good to merge

ChrisMaddock:
Shame Travis disagrees! 😢 I've restarted the CI builds.

PR #233, opened 2017-05-31
Changes for supporting mono 4+ on linux

### LINKED ISSUE #232: Some questions on compiling
I have some questions:
1) why <a href="https://github.com/nunit/nunit-console/blob/e39fc7fed5d122aff26eb7b22b565c35b34a6e2c/src/NUnitEngine/nunit.engine/XmlHelper.cs#L32">this line<a> is not under conditional directive ? I am trying to build this project for .NET framework 4.5, and I don't have v2.0 installed. During compilation it give me an error about Attribute duplication.
2) <a href="https://github.com/nunit/nunit-console/blob/1bea82f0c6511c496fa26e2511b1a31f49cc1f81/src/NUnitEngine/nunit.engine/Extensibility/ExtensionNode.cs#L134">this line</a> give an error
<code>Error CS0618: Warning as Error: `System.AppDomain.CreateInstanceFromAndUnwrap(string, string, bool, System.Reflection.BindingFlags, System.Reflection.Binder, object[], System.Globalization.CultureInfo, object[], System.Security.Policy.Evidence)' is obsolete: `Use an overload that does not take an Evidence parameter' (CS0618) (nunit.engine)</code>
in <a href="https://github.com/nunit/nunit-console/blob/1bea82f0c6511c496fa26e2511b1a31f49cc1f81/src/NUnitEngine/nunit.engine/Drivers/NUnit3FrameworkDriver.cs#L169">some other place</a> it was placed under NET_4_0, but not here.
3) Some more similar errors <a href="https://github.com/nunit/nunit-console/blob/1bea82f0c6511c496fa26e2511b1a31f49cc1f81/src/NUnitEngine/nunit.engine/Services/DomainManager.cs#L67-L75">here</a>

ArsenShnurkov:
Partially fixes #232

I am not sure yet, that I found all places, which should be changed

### DIFF
--- a/src/NUnitConsole/nunit3-console/SafeAttributeAccess.cs
+++ b/src/NUnitConsole/nunit3-console/SafeAttributeAccess.cs
@@ -25,11 +25,13 @@ using System;
 using System.Globalization;
 using System.Xml;
 
+#if NET_2_0
 namespace System.Runtime.CompilerServices
 {
     [AttributeUsage(AttributeTargets.Assembly | AttributeTargets.Class | AttributeTargets.Method)]
     sealed class ExtensionAttribute : Attribute { }
 }
+#endif
 
 namespace NUnit.ConsoleRunner
 {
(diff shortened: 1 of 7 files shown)


rprouse:
The fact that our continuous integrations pass is a good sign that you haven't broken the standard build. We can continue discussing these changes in the issue, but thanks for putting up a PR for everyone to look at.

ChrisMaddock:
I think we should close this. I wouldn't want to merge it unless we had CI that could enforce we don't break it in future, and there doesn't seem to be the demand to put that in place.

@rprouse - What do you think?

PR #234, opened 2017-06-01
Build choco package for the console runner

Roemer:
Since chocolate does not support things like BasePath and Cake does not support chocolateys Out parameter, there were a few "hacks" needed to bring it to work.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -1,9 +1,4 @@
-// Load the extensions to create the chocolatey package
-#tool "NUnit.Extension.VSProjectLoader"
-#tool "NUnit.Extension.NUnitProjectLoader"
-#tool "NUnit.Extension.NUnitV2ResultWriter"
-#tool "NUnit.Extension.NUnitV2Driver"
-#tool "NUnit.Extension.TeamCityEventListener"
+#addin "Cake.FileHelpers"
 
 //////////////////////////////////////////////////////////////////////
 // ARGUMENTS
@@ -418,20 +413,73 @@ Task("PackageChocolatey")
 		
 		EnsureDirectoryExists(PACKAGE_DIR);
 		
-		// Since cake does not yet support a working directory and separate output directory for chocolatey, the following copying and hacks are needed.
-		EnsureDirectoryExists(currentImageDir + "addins");
-		CopyFileToDirectory("choco/nunit-console.portable.nuspec", currentImageDir);
-		CopyFileToDirectory("tools/NUnit.Extension.NUnitProjectLoader/tools/nunit-project-loader.dll", currentImageDir + "addins");
-		CopyFileToDirectory("tools/NUnit.Extension.NUnitV2Driver/tools/nunit.v2.driver.dll", currentImageDir + "addins");
-		CopyFileToDirectory("tools/NUnit.Extension.NUnitV2Driver/tools/nunit.core.dll", currentImageDir + "addins");
-		CopyFileToDirectory("tools/NUnit.Extension.NUnitV2Driver/tools/nunit.core.interfaces.dll", currentImageDir + "addins");
-		CopyFileToDirectory("tools/NUnit.Extension.NUnitV2Driver/tools/nunit.v2.driver.addins", currentImageDir + "addins");
-		CopyFileToDirectory("tools/NUnit.Extension.NUnitV2ResultWriter/tools/nunit-v2-result-writer.dll", currentImageDir + "addins");
-		CopyFileToDirectory("tools/NUnit.Extension.TeamCityEventListener/tools/teamcity-event-listener.dll", currentImageDir + "addins");
-		CopyFileToDirectory("tools/NUnit.Extension.VSProjectLoader/tools/vs-project-loader.dll", currentImageDir + "addins");
+		// Note: Since cake does not yet support a working directory and separate output directory for chocolatey, the following copying and hacks are needed.
+		
+		// List with the addins (addin name, primary dll, additional files)
+		var addins = new Tuple<string, string, string[]>[] {
+			new Tuple<string, string, string[]>(
+				"NUnit.Extension.VSProjectLoader",
+				"vs-project-loader.dll",
+				null
+			),
+			new Tuple<string, string, string[]>(
+				"NUnit.Extension.NUnitProjectLoader",
+				"nunit-project-loader.dll",
+				null
+			),
+			new Tuple<string, string, string[]>(
+				"NUnit.Extension.NUnitV2ResultWriter",
+				"nunit-v2-result-writer.dll",
+				null
+			),
+			new Tuple<string, string, string[]>(
+				"NUnit.Extension.NUnitV2Driver",
+				"nunit.v2.driver.dll",
+				new [] {
+					"nunit.core.dll",
+					"nunit.core.interfaces.dll",
+					"nunit.v2.driver.addins"
+				}
+			),
+			new Tuple<string, string, string[]>(
+				"NUnit.Extension.TeamCityEventListener",
+				"teamcity-event-listener.dll",
+				null
+			)
+		};
+		
+		// Install and copy the addin files
+		var toolsDir = "tools";
+		var nugetInstallSettings = new NuGetInstallSettings { OutputDirectory = toolsDir, ExcludeVersion = true };
+		var addinsDir = System.IO.Path.Combine(currentImageDir, "addins");
+		EnsureDirectoryExists(addinsDir);
+		foreach (var addin in addins) {
+			// Install the extension
+			NuGetInstall(addin.Item1, nugetInstallSettings);
+			var addinToolsPath = System.IO.Path.Combine(toolsDir, addin.Item1, "tools");
+			// Copy primary dll
+			var primaryDllPath = System.IO.Path.Combine(addinToolsPath, addin.Item2);
+			CopyFileToDirectory(primaryDllPath, addinsDir);
+			// Copy additional files
+			if (addin.Item3 != null) {
+				foreach (var additionalItem in addin.Item3) {
+					var additionalItemPath = System.IO.Path.Combine(addinToolsPath, additionalItem);
+					CopyFileToDirectory(additionalItemPath, addinsDir);
+				}
+			}
+			// Write the primary dll to the addins file
+			FileAppendLines(System.IO.Path.Combine(currentImageDir, "nunit.engine.addins"), new[] {
+				System.IO.Path.Combine("addins", addin.Item2)
+			});
+		}	
+				
+		// Copy the nuspec file
+		CopyFileToDirectory("choco/nunit-console.nuspec", currentImageDir);
+		
+		// Set the working directory
 		Context.Environment.WorkingDirectory = currentImageDir;
 		
-		ChocolateyPack("nunit-console.portable.nuspec", 
+		ChocolateyPack("nunit-console.nuspec", 
 			new ChocolateyPackSettings()
 			{
 				Version = packageVersion,
--- a/build.cake
+++ b/build.cake
@@ -1,3 +1,10 @@
+// Load the extensions to create the chocolatey package
+#tool "NUnit.Extension.VSProjectLoader"
+#tool "NUnit.Extension.NUnitProjectLoader"
+#tool "NUnit.Extension.NUnitV2ResultWriter"
+#tool "NUnit.Extension.NUnitV2Driver"
+#tool "NUnit.Extension.TeamCityEventListener"
+
 //////////////////////////////////////////////////////////////////////
 // ARGUMENTS
 //////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


build.cake
```diff
@@ -1,3 +1,10 @@
+// Load the extensions to create the chocolatey package
+#tool "NUnit.Extension.VSProjectLoader"
+#tool "NUnit.Extension.NUnitProjectLoader"
+#tool "NUnit.Extension.NUnitV2ResultWriter"
+#tool "NUnit.Extension.NUnitV2Driver"
+#tool "NUnit.Extension.TeamCityEventListener"
+
```
CharliePoole: That's a clever way to get all the extensions in!

build.cake
```diff
@@ -402,6 +409,37 @@ Task("PackageConsole")
...
+		EnsureDirectoryExists(currentImageDir + "addins");
+		CopyFileToDirectory("choco/nunit-console.portable.nuspec", currentImageDir);
+		CopyFileToDirectory("tools/NUnit.Extension.NUnitProjectLoader/tools/nunit-project-loader.dll", currentImageDir + "addins");
+		CopyFileToDirectory("tools/NUnit.Extension.NUnitV2Driver/tools/nunit.v2.driver.dll", currentImageDir + "addins");
+		CopyFileToDirectory("tools/NUnit.Extension.NUnitV2Driver/tools/nunit.core.dll", currentImageDir + "addins");
+		CopyFileToDirectory("tools/NUnit.Extension.NUnitV2Driver/tools/nunit.core.interfaces.dll", currentImageDir + "addins");
+		CopyFileToDirectory("tools/NUnit.Extension.NUnitV2Driver/tools/nunit.v2.driver.addins", currentImageDir + "addins");
+		CopyFileToDirectory("tools/NUnit.Extension.NUnitV2ResultWriter/tools/nunit-v2-result-writer.dll", currentImageDir + "addins");
+		CopyFileToDirectory("tools/NUnit.Extension.TeamCityEventListener/tools/teamcity-event-listener.dll", currentImageDir + "addins");
+		CopyFileToDirectory("tools/NUnit.Extension.VSProjectLoader/tools/vs-project-loader.dll", currentImageDir + "addins");
```
CharliePoole: It would be nice if we didn't have to know what files are included in a given extension. AFAIK we want every dll in the tools directory. I'm not sure about that `.addins` file - have to check.

choco/nunit-console.portable.nuspec
```diff
@@ -0,0 +1,52 @@
...
+    <file src="..\..\nuget\engine\nunit.nuget.addins" target="tools"/>
+	<file src="addins\nunit-project-loader.dll" target="tools\addins" />
+	<file src="addins\nunit.v2.driver.dll" target="tools\addins" />
+	<file src="addins\nunit.core.dll" target="tools\addins" />
+	<file src="addins\nunit.core.interfaces.dll" target="tools\addins" />
+	<file src="addins\nunit.v2.driver.addins" target="tools\addins" />
+	<file src="addins\nunit-v2-result-writer.dll" target="tools\addins" />
+	<file src="addins\teamcity-event-listener.dll" target="tools\addins" />
+	<file src="addins\vs-project-loader.dll" target="tools\addins" />
+  </files>
```
CharliePoole: OK... now I see where you are going with all these addins files. Have you verified that each addin works when installed this way locally?

Roemer:
Regarding portable: I think this is if the application just sits in a folder and is not "installed. So i think we can keep this. But I can remove it if you wish.

build.cake
Roemer: I think I can optimize this a little more and only load them in the task where the chocolatey package is being build.

build.cake
Roemer: I can try to just use all files (*.*) or dlls in each extension. Not sure about the addins either, just tried to do it as in the "official" zip file.

choco/nunit-console.portable.nuspec
Roemer: I did how it is done in the "official" zip file so I guess it should work. I might try a few.

CharliePoole: It depends on the file contents and whether you are creating an equivalent directory structure to that used by the zip.

The `.addins` file in tools should have a single entry pointing to the `addins` subdirectory. The other `.addins` files, in the addins directory, should simply list the file name of the assemblies that contain extensions, with no path. If it looks like that once you install, then it should work.

Are you able to test that each extension works? One simple way is to run the console telling it to list the extensions with the --list-extensions option. Each of the extensions should be listed under the proper extension point.

Roemer: How is the zip file in the "Releases" generated? Manually or is there some script that does it? If so I could more or less just copy the logic from the script into cake. I will do those fixes and test it.

CharliePoole: I'm not up on the details, but it's done in the nunit-distribution project. However, I think it's all pretty simple if you inspect what you are getting from each nuget package you read. You can either create a directory structure that matches the structure implied by the `.addins` file, or not use them and create new files of your own that match the structure you are creating.

In your case, you have a simple structure with the engine under tools and the extensions in tools/addins. Therefore to match that, the .addins file in tools should contain `addins/` and the file or files in addins should contain entries like: `nunit-project-loader.dll`.

That's not what you have. The file `nunit.nuget.addins` that is part of the engine nuget install contains some complex paths that allow it to use extensions installed side-by-side with it in other nuget packages rather than the simple reference to `addins` that you need. Except for the v2 framework driver, the individual extensions don't have `.addins` files. When installed by themselves, they don't need one because they consist of a single file, all by itself in the tools directory. You need to create your own .addins file that contains all the assemblies you want examined for extensions. That list is easy: it's one dll out of each of the extensions you load. All the extensions except for the v2 framework driver have just one dll, so that's easy enough. For an initial implementation, you could just add the .addins file to the project. Eventually, we should generate it from a list of desired extensions.

CharliePoole:
I notice that you use hard-coded `<file>` elements to build the contents. In project-editor, I don't do that but build up the file contents entirely in the script. This seems like a better approach if we want the list of extensions we use to vary from time to time. Was there some reason you couldn't do it in the same way as project editor?

choco/nunit-console.portable.nuspec
ChrisMaddock: nunit-distribution currently builds all our packages which include extensions in addition to the main console runner. This is currently just the .zip and .msi. At some point, we want to merge nunit-distribution back into this repo, but for now, it all lives there. 

I wonder if the chocolatey package would be better built from that repo - this PR is duplicating a lot of the build logic (although in a nicer way, I'm tempted to say!). Of course, we want to merge everything into this repo at some point, but even then, I still see us keeping a separate cake script for the cross-package builds, and just referencing that from the main script.

On specifics, nunit-distribution just copies all dll's in the NuGet packages `tools` directory to the zip. ([link](https://github.com/nunit/nunit-distribution/blob/master/build.cake#L164)). Regarding the .addins file - it currently just includes a static file. ([link](https://github.com/nunit/nunit-distribution/blob/master/resources/nunit.engine.addins)). This does become a separate maintenance cost however - were I doing it again, I'd probably instead generate this from the cake file - and include a tuple-like list of package-name/primary-dll, to make sure this wasn't forgotten again.

ChrisMaddock: (And apologies that train-of-thought comes late on - there's been a lot of issues to keep up with recently!)

CharliePoole: @ChrisMaddock I had it here, but @Roemer volunteered to take it over for me. I'd prefer to see it here, where I can work on it too, especially since it's all about the console runner and its extensions. You did some stuff in the nunit-distribution project, which would present me with a learning curve if I were to work on it there. Also, there may not be as much duplication as you think, particularly if he switches to the approach in nunit-project-editor that leverages the Cake chocolatey feature.

CharliePoole:
@Roemer Regarding naming, this too was resolved with the chocolatey guys when I did nunit-project-editor. Let's just use names without a suffix.

CharliePoole:
Let me know when this is ready for re-reviewing.

Roemer:
@CharliePoole I did some adjustments to it.:
- Removed the .portable
- Made the addins more dynamic (see the tuple array)
- Hardcoded only the non-addin files in the nuspec and just add the whole addins folder
What's your opinion?

Something I am wondering now: The nupkg of the choco package goes also into the package folder. Won't that now be published to nuget as well instead of to chocolatey?

CharliePoole:
I restarted the travis build, which hung while installing packages.

CharliePoole:
@Roemer Unless we change our appveyor.yml file, all the nupkg files will go to our MyGet feed. I don't see any harm in that, actually. Publishing the releases to nuget is currently manual, so it won't be a problem for a while. Eventually, we will want to automate that.

That said, you just gave me an idea. We already have a nuget package that knows the proper extensions to install. Can't we somehow leverage that rather than duplicating the logic? I may pull down your branch myself and play with the idea a little.

build.cake
```diff
@@ -402,6 +404,90 @@ Task("PackageConsole")
...
+				"nunit-v2-result-writer.dll",
+				null
+			),
+			new Tuple<string, string, string[]>(
+				"NUnit.Extension.NUnitV2Driver",
+				"nunit.v2.driver.dll",
+				new [] {
+					"nunit.core.dll",
+					"nunit.core.interfaces.dll",
+					"nunit.v2.driver.addins"
```
CharliePoole: Why are you copying this file if you are creating your own file?

build.cake
```diff
@@ -402,6 +404,90 @@ Task("PackageConsole")
...
+		};
+		
+		// Install and copy the addin files
+		var toolsDir = "tools";
+		var nugetInstallSettings = new NuGetInstallSettings { OutputDirectory = toolsDir, ExcludeVersion = true };
+		var addinsDir = System.IO.Path.Combine(currentImageDir, "addins");
+		EnsureDirectoryExists(addinsDir);
+		foreach (var addin in addins) {
+			// Install the extension
+			NuGetInstall(addin.Item1, nugetInstallSettings);
```
CharliePoole: Let's at least add a comment here, that this installs the latest published version of the extension. It's OK right now, but I think we will want to control the version we install since we don't control the publication of all extensions. We will want to only publish with versions we have tested, at least at some point.

build.cake
```diff
@@ -402,6 +404,90 @@ Task("PackageConsole")
...
+			CopyFileToDirectory(primaryDllPath, addinsDir);
+			// Copy additional files
+			if (addin.Item3 != null) {
+				foreach (var additionalItem in addin.Item3) {
+					var additionalItemPath = System.IO.Path.Combine(addinToolsPath, additionalItem);
+					CopyFileToDirectory(additionalItemPath, addinsDir);
+				}
+			}
+			// Write the primary dll to the addins file
+			FileAppendLines(System.IO.Path.Combine(currentImageDir, "nunit.engine.addins"), new[] {
```
CharliePoole: You probably shouldn't use the same name we use for the addins file that is included with the nuget package but has entirely different content. The names of addins files are not used by the engine at all - it just loads all files of the type.

choco/nunit-console.nuspec
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Do not remove this test for UTF-8: if “Ω” doesn’t appear as greek uppercase omega letter enclosed in quotation marks, you should use an editor that supports UTF-8, not this one. -->
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>nunit-console</id>
```
CharliePoole: So we will have NUnit.Console on nuget.org and nunit-console on chocolatey.org. I'm OK with this but I want to call it to everyone's attention.

choco/nunit-console.nuspec
```diff
@@ -0,0 +1,46 @@
...
+	<releaseNotes>https://raw.githubusercontent.com/nunit/nunit-console/master/CHANGES.txt</releaseNotes>
+	<mailingListUrl>https://groups.google.com/forum/#!forum/nunit-discuss</mailingListUrl>
+    <licenseUrl>http://nunit.org/nuget/nunit3-license.txt</licenseUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <authors>nunit</authors>
+    <owners>nunit</owners>
+	<language>en-US</language>
+    <tags>nunit console runner test testing tdd</tags>
+    <copyright>Copyright (c) 2017 nunit</copyright>
+  </metadata>
```
CharliePoole: You still haven't fixed authors, owners and copyright from earlier comments.

CharliePoole:
@rprouse It seems to me that we can release a chocolatey package for 3.7.1 at some point __after__ the 3.7.1 release. Do you agree?

build.cake
Roemer: No idea, this is in the official zip file as well so I kept it. Can I remove it?

build.cake
Roemer: I used the same name as in the official package. Please suggest another name.

choco/nunit-console.nuspec
Roemer: According to chocos faq (https://github.com/chocolatey/choco/wiki/CreatePackages#naming-your-package) it says **Don’t use dots** so I followed that :)

build.cake
Roemer: I will add version numbers as well.

choco/nunit-console.nuspec
Roemer: Forgot that, sorry.

Roemer:
I did the changes I already could without more comments. Travis seems to have hung up again on restoring nuget packages...

CharliePoole:
I'm pulling this down locally  to test.

CharliePoole:
I pulled the latest down from master and can no longer build. 😞 Unless I can fix my build I can't help much more with this. I posted the problem on the developer list.

CharliePoole:
I installed and tested - it works!

You still seem to have extra `.addins` files...

* `nunit.nuget.addins` in the tools directory - this sends the engine searching for things uselessly, since the paths are relative to the choco install location.
* `nunit.v2.driver.addins` in the addins directory - this is never used, since your main addins file points directly to the assemblies.

Travis is still failing. @rprouse can you take a look at that?

build.cake
CharliePoole: I think the name is OK if it's the __only__  `.addins` file you use. It seems to mean "the .addins file for the engine." But you have been using an approach involving multiple `.addins`, which begs th e question "If this is for the engine, what are the others for.

Another approach might be to follow the approach we use for nuget and call it the `nunit.choco.addins` file. I'm OK with either name though, provided it's the sole file.

See my other comment about getting rid of the other names.

I suggest you don't rely completely on the zip file content, but try to think of the logic that the engine follows in finding its extensions.

build.cake
CharliePoole: It probably shouldn't be there either. What you need to do is have a complete path from the `.addins` file in the engine directory to every extension dll. As you have created that file, it goes to the dlls in a single step. There is no `addins/` entry that would cause `tools/addins` to be scanned, so no `tools/addins/*.addins` file would ever be examined. 

IOW, you make a choice about whether to just reference `addins/` or to reference each assembly in your primary `.addins` file. Once that choice is made, it dictates what else you need to do. At this stage of the game, either choice is good - we can change later if we need to - but you don't want to mix the two approaches. 

Hopefully that explains it better rather than adding to the confusion.

CharliePoole:
@Roemer I couldn't figure out why this was failing in Travis CI, so I pulled everything into a new branch, called issue-207. So far it's building in CI - I don't know what the difference was.

I'm currently  testing it locally with a few glitches found. I'll do a new PR on that branch when it's working fully. 

I'm also talking with folks at chocolatey about how to package the extensions separately from the runner and use references. If that seems easy, I'll switch to doing it that way right away. If it looks harder, then I'll wait for a subsequent release.

ChrisMaddock:
Thanks for the information about the externeous `.addins` files in the zip. I've created the below issues to resolve.

https://github.com/nunit/nunit-distribution/issues/17
https://github.com/nunit/nunit-v2-framework-driver/issues/14

Roemer:
I guess we can close this as it is replaced with #238 right?

CharliePoole:
I agree! Your work carries on in the other issue. 😄

PR #236, opened 2017-06-07
Create Chocolatey packages for console

### LINKED ISSUE #207: Create Chocolatey package(s) for the console
This arises from discussion in nunit/nunit#452

We need to decide whether we want separate packages with and without extensions and which extensions should be supported in the packaging.

CharliePoole:
Fixes #207 

This PR incorporates work from PR #234 by @Roemer and adds the following:
* There are now two packages, one for console only and one with extensions
* File names and package ids are changed
* Creation of shims for the agent exes is suppressed - we don't want users to be executing them directly.

I marked this as "Do Not Merge" because I'd really prefer to see the extensions referenced as packages, the way we do for NuGet, rather than actually copied into the package with extensions. I can see the way to do that with Chocolatey, but it will take some more experimentation. Alternatively, we could release what we have and improve it later.

Improving would mean first trying to use the existing nuget packages for the extensions. I know chocolatey will install them, but I don't know how whether it will do so in a way that is useful for us. The fallback would be to modify the nuget packages, creating chocolatey packages. In either case, there would be a file installed with the engine, similar to `nunit.nuget.addins`, i.e. `nunit.choco.addins`, which would allow the engine to find the extensions.

So the question is whether to wait for the better solution or just go with what we have now. @rprouse what do you prefer?

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -482,7 +482,9 @@ Task("PackageChocolatey")
 		// Copy the nuspec files
 		CopyFileToDirectory("choco/nunit-console-runner.nuspec", currentImageDir);
 		CopyFileToDirectory("choco/nunit-console-with-extensions.nuspec", currentImageDir);
-		
+		CopyFileToDirectory("choco/nunit-agent.exe.ignore", currentImageDir + "bin/");
+		CopyFileToDirectory("choco/nunit-agent-x86.exe.ignore", currentImageDir + "bin");
+				
 		// Set the working directory
 		Context.Environment.WorkingDirectory = currentImageDir;
 		
--- a/build.cake
+++ b/build.cake
@@ -31,6 +31,7 @@ var packageVersion = version + modifier + dbgSuffix;
 var PROJECT_DIR = Context.Environment.WorkingDirectory.FullPath + "/";
 var PACKAGE_DIR = PROJECT_DIR + "package/";
 var BIN_DIR = PROJECT_DIR + "bin/" + configuration + "/";
+var TOOLS_DIR = PROJECT_DIR + "tools/";
 var IMAGE_DIR = PROJECT_DIR + "images/";
 
 var SOLUTION_FILE = "NUnitConsole.sln";
@@ -405,8 +406,7 @@ Task("PackageConsole")
     });
 
 Task("PackageChocolatey")
-	.Description("Creates chocolate packages of the console runner")
-    .IsDependentOn("CreateImage")
+	.Description("Creates chocolatey packages of the console runner")
 	.Does(() =>
 	{
 		var currentImageDir = IMAGE_DIR + "NUnit-" + packageVersion + "/";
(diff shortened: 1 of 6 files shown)


CharliePoole:
Just to be clear... in my preferred solution, the user could type...

```
choco install nunit-console-runner nunit-teamcity-extension
```

and get the runner with that single extension installed.

CharliePoole:
Hey @ferventcoder we'd love to have your comments on this as well.

build.cake
```diff
@@ -402,6 +404,105 @@ Task("PackageConsole")
...
+			// Write the primary dll to the addins file
+			FileAppendLines(System.IO.Path.Combine(currentImageDir, "nunit.engine.addins"), new[] {
+				System.IO.Path.Combine("addins", addin.Item3)
+			});
+		}	
+				
+		// Copy the nuspec files
+		CopyFileToDirectory("choco/nunit-console-runner.nuspec", currentImageDir);
+		CopyFileToDirectory("choco/nunit-console-with-extensions.nuspec", currentImageDir);
+		CopyFileToDirectory("choco/nunit-agent.exe.ignore", currentImageDir + "bin/");
```
Roemer: Wouldn't it be better to just create those empty files here in cake instead of creating them in the solution and copying them here? Or at least just copy all files from the choco directory.

choco/nunit-console-runner.nuspec
```diff
@@ -0,0 +1,45 @@
...
+    <mailingListUrl>https://groups.google.com/forum/#!forum/nunit-discuss</mailingListUrl>
+    <licenseUrl>http://nunit.org/nuget/nunit3-license.txt</licenseUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <authors>Charlie Poole</authors>
+    <owners>Charlie Poole</owners>
+    <language>en-US</language>
+    <tags>nunit console runner test testing tdd</tags>
+    <copyright>Copyright (c) 2017 Charlie Poole</copyright>
+  </metadata>
+  <files>
```
Roemer: The more I look at it the more i think it will be more maintainable if we have the entire content of the package build by the cake script. Now we have a mix of solution items, lines in nuspec, things copied and generated by cake...

CharliePoole: That's the approach I favor too. 😄 Will do!

build.cake
CharliePoole: I think it's clearer if they are in the same place as the agent exe and config files. If we do it all from the cake script, then I would favor what you are suggesting.

choco/nunit-console-runner.nuspec
CharliePoole: Part of the difficulty here is the use of the "Image" directory to build everything. This is a hold-over from when the framework and console were in the same repo and we built a complicated structure with various frameworks for different runtime versions. It serves little use here and was never removed purely out of laziness on my part. I may take a look at that, although it's probably better as a separate PR.

CharliePoole:
@rprouse I know it's easy to miss which issues require attention, so... THIS DOES. Your comment to  question above is the only thing that is blocking it right now. 😸 

Based on your input, I will either merge what we have or work toward the solution I prefer - described above. I also asked you a while back to tell me how you wanted this to relate to the release you are working on - i.e. does it need to be part of it or can it be done a bit later. The two questions are obviously related.

choco/nunit-console-with-extensions.nuspec
```diff
@@ -0,0 +1,56 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Do not remove this test for UTF-8: if “Ω” doesn’t appear as greek uppercase omega letter enclosed in quotation marks, you should use an editor that supports UTF-8, not this one. -->
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <id>nunit-console-with-extensions</id>
+    <version>0.0.0</version>
```
rprouse: Doesn't the version need to be a variable here?

rprouse: I see it gets injected in the Cake file, just thought it should be `$version$` like NuGet does it. I could very easily be wrong though 😄

CharliePoole: This is how chocolatey generates templates, so it's at least their preferred approach. Don't know if they accept `$version$` as well.

CharliePoole:
I'm taking "keep it simple" to mean "don't make changes" - what we have suffers precisely from not being simple. However, it's no worse than what we do in the distribution project. I'll take care of the review comments and merge.

In the longer (even medium) term, I'm convinced that this is __not__ the right way to use chocolatey with self-contained packages. My experience, their docs and my discussions with @ferventcoder have convinced me that use of package references is a better path. Hopefully, whoever does that next iteration will bear it in mind.

Still waiting to here how you want this to fit in with the release. It makes a difference in how this is eventually merged.

rprouse:
I think we can merge this and we can try to work it into the release process, or at least right after the release. I will likely need your help uploading. We do NUnit distribution just after the initial release and amend the GitHub release, we could do the same with this. Let's see how it goes, document the process and then decide if it needs improvements.

CharliePoole:
Rebased on master

build.cake
ChrisMaddock: Personally, I'd agree it would be nice to create them in cake - purely because I'm not too familiar with choco, and just went to look them up, and find out what was in the `.ignore` files. 😆 That would be more explicit if they were created right here - I assumed files being copied had content.

build.cake
```diff
@@ -402,6 +404,143 @@ Task("PackageConsole")
...
+			if (addin.Item4 != null) {
+				foreach (var additionalItem in addin.Item4) {
+					var additionalItemPath = System.IO.Path.Combine(addinToolsPath, additionalItem);
+					CopyFileToDirectory(additionalItemPath, addinsDir);
+				}
+			}
+			// Write the primary dll to the addins file
+			FileAppendLines(System.IO.Path.Combine(currentImageDir, "nunit.engine.addins"), new[] {
+				System.IO.Path.Combine("addins", addin.Item3)
+			});
```
ChrisMaddock: This is much better than the nunit-distribution handling. 😄

build.cake
```diff
@@ -402,6 +404,143 @@ Task("PackageConsole")
         });
     });
 
+Task("PackageChocolatey")
+	.Description("Creates chocolate packages of the console runner")
```
ChrisMaddock: chocolate -> Chocolatey. Although chocolate packages would be nice! 😄

build.cake
```diff
@@ -402,6 +404,143 @@ Task("PackageConsole")
...
+    .IsDependentOn("CreateImage")
+	.Does(() =>
+	{
+		var currentImageDir = IMAGE_DIR + "NUnit-" + packageVersion + "/";
+		
+		EnsureDirectoryExists(PACKAGE_DIR);
+		
+		// Note: Since cake does not yet support a working directory and separate output directory for chocolatey, the following copying and hacks are needed.
+
```
ChrisMaddock: Should this comment be attached to lines 480ish? I'm not sure what it's referring to.

CharliePoole: @Roemer This came from your branch and I just kept it. I actually don't understand it either.

build.cake
```diff
@@ -402,6 +404,143 @@ Task("PackageConsole")
...
+					"nunit.core.dll",
+					"nunit.core.interfaces.dll"
+				}
+			),
+			new Tuple<string, string, string, string[]>(
+				"NUnit.Extension.TeamCityEventListener",
+				"1.0.2",
+				"teamcity-event-listener.dll",
+				null
+			)
```
CharliePoole: @ChrisMaddock This came out of my directions to @Roemer. To the extent that we might build and distribute extensions written by other people, I thought we needed to be able to specify the version we have tested. RIght now, the only such extension is teamcity, which we don't test in any version. Now that I write that, it sounds bad, doesn't it. 😟

build.cake
CharliePoole: I'll look at that change and see if I can fit it in relatively smoothly.

build.cake
CharliePoole: FYI - and not particularly to do with these lines of code - Because I can no longer build nunit-console without commenting out the line that detects .NET Core, I'm actually using CI to test and making changes more incrementally than I would normally do. Each new commit gets in a bit more to fix the comments that have been made.

CharliePoole: @Roemer 's concept!

FYI - and not particularly to do with these lines of code - Because I can no longer build nunit-console without commenting out the line that detects .NET Core, I'm actually using CI as my primary test bed and making changes more incrementally than I would normally do. Each new commit gets in a bit more to fix the comments that have been made.

build.cake
```diff
@@ -402,6 +404,143 @@ Task("PackageConsole")
...
+			// Set the version
+			nugetInstallSettings.Version = addin.Item2;
+			// Install the extension
+			NuGetInstall(addin.Item1, nugetInstallSettings);
+			var addinToolsPath = System.IO.Path.Combine(toolsDir, addin.Item1, "tools");
+			// Copy primary dll
+			var primaryDllPath = System.IO.Path.Combine(addinToolsPath, addin.Item3);
+			CopyFileToDirectory(primaryDllPath, addinsDir);
+			// Copy additional files
+			if (addin.Item4 != null) {
```
CharliePoole: That would guard against `addin` being null, not `Item4`. Currently, the code assumes that addin is not null, which is valid since we create the list here.

ChrisMaddock: Sorry - you're right.

CharliePoole:
@Roemer @rprouse I think this is as good as it gets without doing some major reoranization on the build script. I'd actually like to do that, but not in this PR.

I'd also like to see a version with extensions that simply referenced them, but I think we should get this one out first.

There are some comments in the code that point out places where I left in duplication that should eventually be removed. That's because I think the script reorganization is needed first.

Please re-review for merging.

choco/nunit-console-with-extensions.nuspec
ferventcoder: Not yet, but we do accept `choco pack --version=1.2.3`. See https://github.com/chocolatey/choco/issues/1313

ferventcoder:
Late to the conversation but I will reiterate what I mentioned to @CharliePoole in an email here.

When you implement plugins/exensions, you could do it in one of two ways. Either within Chocolatey's lib directory, or gather it together under `$env:ChocolateyToolsLocation\nunit` - this allows users to determine where that tools location is (defaults to `c:\tools`). Using the Tools Location does have some drawbacks and considerations, as you are responsible for setting the path and removing that, etc - much of that is currently documented at https://github.com/chocolatey/choco/issues/1303#issuecomment-303804561

I do prefer each plugin as a separate package, but as mentioned here that could be a major restructuring.

build.cake
```diff
@@ -554,7 +669,8 @@ Task("Package")
     .IsDependentOn("CheckForError")
     .IsDependentOn("PackageEngine")
     .IsDependentOn("PackageConsole")
-    .IsDependentOn("PackageNetStandardEngine");
+    .IsDependentOn("PackageNetStandardEngine")
+	.IsDependentOn("PackageChocolatey");
```
ferventcoder: Small formatting issue.

choco/nunit-console-with-extensions.nuspec
ferventcoder: > This is how chocolatey generates templates, so it's at least their preferred approach.

I'm not sure what version of Chocolatey you are on, but this hasn't been the way that Chocolatey generates templates for over a year. Current version is 0.10.7. and it separates package specifics from software specifics now in generation.

CharliePoole: I'm on the latest... however I copied this from another project. 😊 Anyway, I'm running tests using this package, so we're good for now.

CharliePoole:
Thanks @ferventcoder for chiming in. I'm going to continue on this and we'll do it in the preferred way when possible. Meanwhile, there may be a first release using what we have so far. Key question is can we rely on relative paths from our engine to any extensions remaining stable over time as Chocolatey develops. We use such paths extensively.

ferventcoder:
> Key question is can we rely on relative paths from our engine to any extensions remaining stable over time as Chocolatey develops. We use such paths extensively.

You can't go wrong with `$env:ChocolateyInstall\lib\<id>` - we have no plans to change it currently. If we ever did it would be a VERY long transition and deprecation before the change was made (with plenty of warnings). So very unlikely.

CharliePoole:
We can adapt. 😄 Here's how we handle both nuget 2 and 3 for example in our addins file:

```
../../NUnit.Extension.*/**/tools/     # nuget v2 layout
../../../NUnit.Extension.*/**/tools/  # nuget v3 layout
```

This works well because we actually don't want a copy of the NUnit engine installed with NuGet to be aware of any extensions except those also installed in the project with NuGet.

Similarly, I think an app installed via Chocolatey should only use extensions installed the same way. Eventually, I'd hope to see all of the following separate...

* Console runner
* Gui Runner
* Engine
* Each extension to the engine

That would give the two (or more) runners access to the same engine with that engine automatically using all the extensions. It's not something that's so easy to do otherwise.

PR #257, opened 2017-07-09
Improve error catching around OutputSpecifications

### LINKED ISSUE #252: Console crashes when specifying both format and transform for result
See below:

```
nunit3-console Tests.exe --result:results.xml;format=nunit2;transform=a.xslt

Unhandled Exception: System.ArgumentException: Conflicting format options: results.xml;format=nunit2;transform=a.xslt
   at NUnit.Common.OutputSpecification..ctor(String spec)
   at NUnit.Common.CommandLineOptions.<ConfigureOptions>b__124_13(String v)
   at NUnit.Options.Option.Invoke(OptionContext c)
   at NUnit.Options.OptionSet.Parse(String argument, OptionContext c)
   at NUnit.Options.OptionSet.Parse(IEnumerable`1 arguments)
```

It appears transforms must only be used with the default format - but we should catch this exception and print a nice error message, like we do with invalid arguments. (It would also be helpful to document this more explicitly.)

ChrisMaddock:
Fixes #252.

As well as fixing the original crash, this also generally improves error catching in this area.

1. Fixes crash when an invalid `--result` format is passed.
2. Fixes NUnit Engine returning misleading error message when transform can not be loaded.
3. ConsoleOptions now validates for existence of transform, so we can provide a neat error message, rather than stack trace from the engine.

### DIFF
--- a/src/NUnitConsole/nunit3-console.tests/CommandLineTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/CommandLineTests.cs
@@ -475,7 +475,7 @@ namespace NUnit.ConsoleRunner.Tests
         [Test]
         public void ResultOptionWithFilePathAndTransform()
         {
-            ConsoleOptions options = new ConsoleOptions("tests.dll", "-result:results.xml;transform=transform.xslt");
+            ConsoleOptions options = new ConsoleOptions("tests.dll", "-result:results.xml;transform=TextSummary.xslt");
             Assert.True(options.Validate());
             Assert.AreEqual(1, options.InputFiles.Count, "assembly should be set");
             Assert.AreEqual("tests.dll", options.InputFiles[0]);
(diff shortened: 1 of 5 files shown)


oznetmaster:
Do any of these need to be applied to the nunitlite runner as well?

CharliePoole:
Has WorkDirectory been used in the past for the location of the transform? That seems wrong to me, since we define WorkDirectory as the place where we create files, not find them. When using the command-line, it seems like the most natural thing would be to expand non-absolute paths using the CurrentDirectory.

ChrisMaddock:
> Do any of these need to be applied to the nunitlite runner as well?

Looks like NUnitLite reads the transform option, but never uses it? But yes, the same crash looks like it would happen there.

> Has WorkDirectory been used in the past for the location of the transform?

I'm getting WorkDirectory and 'Working Directory' mixed up again, aren't I? 😞 Will rejig.

ChrisMaddock:
See https://github.com/nunit/nunit/pull/2312 for related NUnitLite changes.

ChrisMaddock:
This wasn't ready to merge - @CharliePoole's comment was right, that it checks in the wrong place for transform file existence. Sorry, I should have made that clearer, I didn't manage to get back to this last week. 😞 

The issue with this at the moment, is that if the user has a different 'working directory' and 'work directory' - then the console will refuse to load the xslt. Hopefully that won't affect too many users. I've made https://github.com/nunit/nunit-console/issues/262 for it - I'm not sure if I'll be able to get to this for a few days, but will try and get something in soon.

CharliePoole:
Note that we can only trust the current directory before we start running tests, since user code could change it. It's not a problem here, but worth remembering that we need to do anything that involves current directory before the tests start running.

PR #263, opened 2017-07-16
Use current directory instead of WorkDirectory

### LINKED ISSUE #262: Transform file existence check should check current directory instead of WorkDirectory
See discussion on https://github.com/nunit/nunit-console/pull/257

mikkelbu:
Fixes #262

### DIFF
--- a/src/NUnitConsole/nunit3-console/CommandLineOptions.cs
+++ b/src/NUnitConsole/nunit3-console/CommandLineOptions.cs
@@ -39,17 +39,27 @@ namespace NUnit.Common
     /// </summary>
     public class CommandLineOptions : OptionSet
     {
+        private static readonly string CURRENT_DIRECTORY_ON_ENTRY = Environment.CurrentDirectory;
+
         private bool validated;
         private bool noresult;
 
+        /// <summary>
+        /// An abstraction of the file system
+        /// </summary>
+        protected readonly IFileSystem _fileSystem;
+
 #region Constructor
 
-        internal CommandLineOptions(IDefaultOptionsProvider defaultOptionsProvider, params string[] args)
+        internal CommandLineOptions(IDefaultOptionsProvider defaultOptionsProvider, IFileSystem fileSystem, params string[] args)
         {
             // Apply default options
-            if (defaultOptionsProvider == null) throw new ArgumentNullException("defaultOptionsProvider");
+            if (defaultOptionsProvider == null) throw new ArgumentNullException(nameof(defaultOptionsProvider));
             TeamCity = defaultOptionsProvider.TeamCity;
-            
+
+            if (fileSystem == null) throw new ArgumentNullException(nameof(fileSystem));
+            _fileSystem = fileSystem;
+
             ConfigureOptions();            
             if (args != null)
                 Parse(args);
@@ -127,7 +137,7 @@ namespace NUnit.Common
         private string workDirectory = null;
         public string WorkDirectory 
         {
-            get { return workDirectory ?? Environment.CurrentDirectory; }
+            get { return workDirectory ?? CURRENT_DIRECTORY_ON_ENTRY; }
         }
         public bool WorkDirectorySpecified { get { return workDirectory != null; } }
 
@@ -143,7 +153,8 @@ namespace NUnit.Common
                     return new OutputSpecification[0];
 
                 if (resultOutputSpecifications.Count == 0)
-                    resultOutputSpecifications.Add(new OutputSpecification("TestResult.xml"));
+                    resultOutputSpecifications.Add(
+                        new OutputSpecification("TestResult.xml", CURRENT_DIRECTORY_ON_ENTRY, _fileSystem));
 
                 return resultOutputSpecifications;
             }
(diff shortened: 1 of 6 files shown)


src/NUnitConsole/nunit3-console/CommandLineOptions.cs
```diff
@@ -129,7 +127,7 @@ public CommandLineOptions(params string[] args)
         private string workDirectory = null;
         public string WorkDirectory 
         {
-            get { return workDirectory ?? DEFAULT_WORK_DIRECTORY; }
+            get { return workDirectory ?? Environment.CurrentDirectory; }
```
CharliePoole: The idea of using the static string was to get the field initialized before any other code had a chance to run. It probably doesn't matter so long as we don't do anything else with the CurrentDirectory in the options class.

src/NUnitConsole/nunit3-console/CommandLineOptions.cs
```diff
@@ -402,7 +400,7 @@ private void ResolveOutputSpecification(string value, IList<OutputSpecification>
 
             if (spec.Transform != null)
             {
-                var transformPath = Path.Combine(WorkDirectory, spec.Transform);
+                var transformPath = Path.Combine(Environment.CurrentDirectory, spec.Transform);
```
CharliePoole: Same concept applies here. We have to be sure that `Environment.CurrentDirectory` is not changed before we execute this. Since NUnit should __never__ change it, that essentially means that no user code can be allowed to execute before evaluating this line. Odds are, we are safe but "belt and suspenders" may apply.

mikkelbu: @CharliePoole I was actually more worried about the opposite situation that we would instantiate `CommandLineOptions` from several locations and then then `DEFAULT_WORK_DIRECTORY` would be fixed to something that was not correct anymore. So I only checked that `CommandLineOptions` was only constructed once. I will reintroduce the static. 

When looking more into this I have some questions regarding `WorkDirectory`. In https://github.com/nunit/nunit-console/blob/0500ce8c3216157d320b6072ae2b57544e5d9362/src/NUnitConsole/nunit3-console/ConsoleRunner.cs#L77 and https://github.com/nunit/nunit-console/blob/0500ce8c3216157d320b6072ae2b57544e5d9362/src/NUnitConsole/nunit3-console/ConsoleRunner.cs#L414 we check whether `WorkDirectory` is null and then use `Environment.CurrentDirectory` as a substitute. I cannot see that `WorkDirectory` can be null so these "guards" make little sense, but the last guard was introduced a year ago, so perhaps I've overlooked something?

mikkelbu:
@CharliePoole I was actually more worried about the opposite situation that we would instantiate `CommandLineOptions` from several locations and then then `DEFAULT_WORK_DIRECTORY` would be fixed to something that was not correct anymore. So I only checked that `CommandLineOptions` was only constructed once. I will reintroduce the static. 

When looking more into this I have some questions regarding `WorkDirectory`. In https://github.com/nunit/nunit-console/blob/0500ce8c3216157d320b6072ae2b57544e5d9362/src/NUnitConsole/nunit3-console/ConsoleRunner.cs#L77 and https://github.com/nunit/nunit-console/blob/0500ce8c3216157d320b6072ae2b57544e5d9362/src/NUnitConsole/nunit3-console/ConsoleRunner.cs#L414 we check whether `WorkDirectory` is null and then use `Environment.CurrentDirectory` as a substitute. I cannot see that `WorkDirectory` can be null so these "guards" make little sense, but the last guard was introduced a year ago, so perhaps I've overlooked something?

CharliePoole:
I think it's simply that the value from options was once null when not set by the user so console runner had to allow for it. Now options is taking care of the policy decision to use current directory instead.

ChrisMaddock:
Thanks for picking this one up, @mikkelbu!  I wanted to get it done quickly, but a bit limited for time at the moment. Appreciated! 🙂

ChrisMaddock:
One further thought, while we're talking about when `Environment.CurrentDirectory` may be changed - the transform is currently loaded from a path relative to the currentdirectory, much later in the process.

Maybe it would be good to create a fully rooted path at the point where we check for file existence, in case some future modification does alter the current directory.

CharliePoole:
@ChrisMaddock Good idea. If it happened it would be "the user's fault" but it could take a bunch of time to figure out what was going on.

mikkelbu:
Following @ChrisMaddock's idea I've recorded the full path at the point where we check for file existence. I've also moved the check and information about the full path into `OutputSpecification`, since the responsibility should be here. As a consequence of this I've changed the constructor of `OutputSpecification` such that it also accepts and possible folder for the transformation and an abstraction of the file system (i.e. `IFileSystem`).

ChrisMaddock:
@mikkelbu - apologies, I think we've mis-understood each other! My suggestion was to replace the existing transform path with a fullpath, rather than a relative one - not to record the full path additionally.

My concern is that the xslt file is loaded from `spec.Transform`, which is a relative path, much later in the process,  [here](https://github.com/nunit/nunit-console/blob/403740be90978c87fd8d85102cd301f87f0eca2d/src/NUnitConsole/nunit3-console/ConsoleRunner.cs#L161). The path that's loaded there, should match the one we checked existed earlier when parsing the options ([here](https://github.com/nunit/nunit-console/blob/403740be90978c87fd8d85102cd301f87f0eca2d/src/NUnitConsole/nunit3-console/CommandLineOptions.cs#L407)) - the best way to ensure that is probably to make spec.Transform a fully rooted path, rather than the relative path it currently is.

I like that you've added the IFileSystem however, a much better solution than the lazy tests that I did! I'd suggest the file existence check should stay in the CommandLineOptions class however - that should be the class that's responsible for validating options - the OutputSpecification class shouldn't know about the filesystem in my opinion, and should just be a store of pre-validated filepaths.

mikkelbu:
Wrt. why I put the existence check in OutputSpecification was to enforce that we only could create OutputSpecification instances that were valid.

Wrt. to why I kept both Transform and FullPathTransform was that I did not know to what extend Transform was used, and I decided to keep both just to be safe.

I'll try to move the validation and see how the result looks 😄

mikkelbu:
@ChrisMaddock Your proposed solution is simpler, so I've moved the validation to CommandLineOptions and removed FullPathTransform.

PR #267, opened 2017-07-19
Fix possible NRE

alexxn:
If for some reason NUnit.Engine.Runners.ProcessRunner failed to run task, in the resultNode there is no "result" attribute. Thus, OverallResult stays null and WriteSummaryReport() fails with NullReferenceException on line 112,  leaving user frustrated about reasons of fail and results of the test run.

Albeit main cause of this error is rethrowing exception in ProcessRunner.cs (See Dispose method, line 255) user should not see NullReference exception and must be infromed with a more meaningfull information.

### DIFF
--- a/src/NUnitConsole/nunit3-console/ResultReporter.cs
+++ b/src/NUnitConsole/nunit3-console/ResultReporter.cs
@@ -42,6 +42,8 @@ namespace NUnit.ConsoleRunner
             OverallResult = resultNode.GetAttribute("result");
             if (OverallResult == "Skipped")
                 OverallResult = "Warning";
+            if (OverallResult == null)
+                OverallResult = "Unknown";
 
             Summary = new ResultSummary(resultNode);
         }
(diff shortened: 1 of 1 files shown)


src/NUnitConsole/nunit3-console/ResultReporter.cs
```diff
@@ -42,6 +42,8 @@ public ResultReporter(XmlNode resultNode, ExtendedTextWriter writer, ConsoleOpti
             OverallResult = resultNode.GetAttribute("result");
             if (OverallResult == "Skipped")
                 OverallResult = "Warning";
+            if (OverallResult == null)
+                OverallResult = "Unknown";
```
CharliePoole: Seems simple enough. Does it get colored red, however? Does any error show up in the console report?

alexxn: Not really. Exceptions in other threads must be handled properly. Right now even  logging

string stopError = string.Format("Failed to stop the remote agent. {0}", e.Message);

in the ProcessRunner.cs does not write into log anything here.

In my second commit I've added output coloring as Error though.

CharliePoole:
Looks ready for merging.

ChrisMaddock:
Restarted Travis again.

mikkelbu:
I've restarted it 4-5 times already without any success :(. Should we be worried about the `CommandLineException` that happens on every build (even when the build is successful)

ChrisMaddock:
Hadn't noticed that exception - think it's new. One to keep an eye on perhaps...

Issue #271, opened 2017-07-25
Run Nunit test from .netcoreapp 1.1 dll via console runner


pavzaj:
I have some nunit tests in .netcoreapp 1.1 project. Tests are written using Nunit framework 3.7.1 and they test logic from a referenced dll compiled against .netStandard 1.0.
These tests are properly detected and executed by NUnit3TestAdapter 3.8.0 in VS 2017, but when I try to run in via Nunit CLI 3.7.0 I get the following error:

D:\NUnit.Console-3.7.0>nunit3-console.exe D:\github\MyProject\_tests\TestData\SampleLogicNunitTests\bin\Debug\netcoreapp1.1\SampleLogicNunitTests.dll
NUnit Console Runner 3.7.0
Copyright (c) 2017 Charlie Poole, Rob Prouse

Runtime Environment
   OS Version: Microsoft Windows NT 10.0.15063.0
  CLR Version: 4.0.30319.42000

Test Files
    D:\github\MyProject\_tests\TestData\SampleLogicNunitTests\bin\Debug\netcoreapp1.1\SampleLogicNunitTests.dll

Errors, Failures and Warnings

1) Error :
An exception occurred in the driver while loading tests.

Server stack trace:
   at NUnit.Engine.Runners.DirectTestRunner.LoadDriver(IFrameworkDriver driver, String testFile, TestPackage subPackage)
   at NUnit.Engine.Runners.DirectTestRunner.LoadPackage()
   at NUnit.Engine.Runners.DirectTestRunner.EnsurePackageIsLoaded()
   at NUnit.Engine.Runners.DirectTestRunner.RunTests(ITestEventListener listener, TestFilter filter)
   at System.Runtime.Remoting.Messaging.StackBuilderSink._PrivateProcessMessage(IntPtr md, Object[] args, Object server, Object[]& outArgs)
   at System.Runtime.Remoting.Messaging.StackBuilderSink.SyncProcessMessage(IMessage msg)

Exception rethrown at [0]:
   at System.Runtime.Remoting.Proxies.RealProxy.HandleReturnMessage(IMessage reqMsg, IMessage retMsg)
   at System.Runtime.Remoting.Proxies.RealProxy.PrivateInvoke(MessageData& msgData, Int32 type)
   at NUnit.Engine.ITestEngineRunner.Run(ITestEventListener listener, TestFilter filter)
   at NUnit.Engine.Runners.ProcessRunner.RunTests(ITestEventListener listener, TestFilter filter)

[...]

PR #278, opened 2017-08-07
Improve error message running .NET Standard assembly

CharliePoole:
This is intended to mitigate issue #271, although it doesn't resolve it. People keep being confused by the error message that indicates the .NET standard tests are portable, when they are not. Turns out this is a problem with VS, which I try to compensate for in the message.

### DIFF
--- a/src/NUnitEngine/nunit.engine/Services/DriverService.cs
+++ b/src/NUnitEngine/nunit.engine/Services/DriverService.cs
@@ -60,8 +60,13 @@ namespace NUnit.Engine.Services
 
             if (targetFramework != null)
             {
-                var platform = targetFramework.Split(new char[] { ',' })[0];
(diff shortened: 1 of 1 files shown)


build.cake
```diff
@@ -92,7 +92,7 @@ Setup(context =>
 
     // Executed BEFORE the first task.
     Information("Building version {0} of NUnit.", packageVersion);
-    IsDotNetCoreInstalled = CheckIfDotNetCoreInstalled();
+    //IsDotNetCoreInstalled = CheckIfDotNetCoreInstalled();
```
rprouse: Was this checked in by mistake because you can't get the .NET Core stuff to build?

CharliePoole: Exactly what happened. I'll re-push.

PR #283, opened 2017-08-19
Remove the ability to load a newer engine from the NUnit install directory

### LINKED ISSUE #6: TypeLoadException in nunit3-console 3.0.1
@yaakov-h commented on [Sun Mar 13 2016](https://github.com/nunit/nunit/issues/1348)

When using NUnit.Console 3.0.1 from the NuGet gallery, I'm getting an exception when starting the runner.

I'm only seeing this behaviour on some PCs.

```
NUnit Console Runner 3.0.5813
Copyright (C) 2015 Charlie Poole

System.TypeLoadException: Could not load type 'NUnit.Engine.IAvailableRuntimes' from assembly 'nunit.engine.api, Version=3.0.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb'.
   at NUnit.Engine.TestEngine.Initialize()
   at NUnit.Engine.TestEngine.NUnit.Engine.ITestEngine.get_Services()
   at NUnit.ConsoleRunner.ConsoleRunner..ctor(ITestEngine engine, ConsoleOptions options, ExtendedTextWriter writer)
```

This issue only occurs when trying to run NUnit.Console on a PC that has NUnit 3.2 installed globally by MSI.

---

@rprouse commented on [Mon Mar 14 2016](https://github.com/nunit/nunit/issues/1348#issuecomment-196372402)

@CharliePoole looks like an error with our engine resolution that we didn't think of. The console is finding the newer engine in the install directory and using that. Any ideas what we can do about it?

---

@CharliePoole commented on [Mon Mar 14 2016](https://github.com/nunit/nunit/issues/1348#issuecomment-196493221)

Ah ha! I didn't think of this. The new engine is getting loaded. It references the api assembly that is found in its own directory. But the old api assembly is already in memory.

@yaakov-h A workaround is to upgrade the console nuget package to 3.2.  You don't have to upgrade the framework if you don't want to.

A temporary fix we could make would be to only use the local engine.

A longer term fix that allows dropping in new engine releases will take some thinking.

---

@yaakov-h commented on [Tue Mar 15 2016](https://github.com/nunit/nunit/issues/1348#issuecomment-196731196)

Why is the old runner loading the new engine?

---

@CharliePoole commented on [Tue Mar 15 2016](https://github.com/nunit/nunit/issues/1348#issuecomment-196921026)

Because nunit.engine.api is designed to find the newest engine available and use it, within limits set by the client program - nunit3-console in this case. Clearly, we have a glitch, as described in my note above.

---

@rprouse commented on [Tue Apr 12 2016](https://github.com/nunit/nunit/issues/1348#issuecomment-209122846)

@CharliePoole, should we do a temporary fix for this for 3.2.1?

---

@CharliePoole commented on [Tue Apr 12 2016](https://github.com/nunit/nunit/issues/1348#issuecomment-209148218)

Good idea. It's just a matter of having the console runner specify `true` for privateCopy when it calls TestEngineActivator.CreateInstance. That's what Ive just done for the gui.

---

@rprouse commented on [Wed Apr 13 2016](https://github.com/nunit/nunit/issues/1348#issuecomment-209439174)

@CharliePoole, sounds good. I have put in the 3.2.1 milestone and assigned it to me for the temporary fix.

---

@knocte commented on [Wed Apr 13 2016](https://github.com/nunit/nunit/issues/1348#issuecomment-209744451)

>  A workaround is to upgrade the console nuget package to 3.2.

Are you sure about this? This is affecting me when simply running the test within XamarinStudio (not the console runner), and I'm already using version 3.2.

More info: https://bugzilla.xamarin.com/show_bug.cgi?id=40035

---

@rprouse commented on [Thu Apr 14 2016](https://github.com/nunit/nunit/issues/1348#issuecomment-209931711)

@knocte, I am pretty sure your problem is a different issue. It looks to be a problem with the XamarinStudio test runner which we did not write. I actually didn't even know it worked for NUnit 3 :smile: 

Looking at the [source for the MonoDevelop NUnit 3 runner](https://github.com/mono/monodevelop/tree/master/main/src/addins/MonoDevelop.UnitTesting.NUnit/NUnit3Runner), they have not updated it or the engine to 3.2 yet, so it is not you that needs to update to 3.2, it is Xamarin Studio. 

[Looking at their runner code](https://github.com/mono/monodevelop/blob/master/main/src/addins/MonoDevelop.UnitTesting.NUnit/NUnit3Runner/NUnitTestRunner.cs), they are loading v3.0.1 of the nunit.framework.dll into their test runner and directly linking to v3.0.1 of the nunit.engine. This locks Xamarin users into v3.0.1 of NUnit. Their runner code is totally incorrect. They really should have contacted us for guidance.

---

@rprouse commented on [Fri Apr 15 2016](https://github.com/nunit/nunit/issues/1348#issuecomment-210484904)
#1413 provides a short term fix for this by using a local engine. I am moving out of the 3.2.1 milestone.

---

@rprouse commented on [Sat Jun 25 2016](https://github.com/nunit/nunit/issues/1348#issuecomment-228538528)

I am going to leave this `On Deck`, but move out of the 3.4 milestone. The temporary fix is still in place and works. I think we need time to discuss possible fixes and work through the potential consequences of each. Until we do that, I think sticking with a local engine is safest.

---

@CharliePoole commented on [Sat Jun 25 2016](https://github.com/nunit/nunit/issues/1348#issuecomment-228550715)

I agree. This is really a symptom of the overall issue addressed by #1132. Should that be an Epic?

### LINKED ISSUE #255: NUnit console fails with error code -100

rprouse:
See #255 
Fixes #6 

See #255 for a full discussion and my analysis. For history, NUnit 3 has always had the ability to load a newer version of the engine from the NUnit install directory if it is available. The idea was that if 3rd Party vendors don't update the engine, users could do so by installing a newer NUnit. Once NUnit was released though, 3rd Party vendors preferred to disable that ability and only allow users to run with the engine they ship with so the only runner that ever shipped with that ability was our console runner.

As you can see from this PR, there were never any unit tests for this functionality as it is more an integration level test which we have none. At some point remoting stopped working when running in this mode and we never noticed. I walked through the code and it is running as intended and only loading the newer engine, but the remoting calls fail.

Since the intended use of this feature was 3rd Party runners and nobody uses it, it is easier to remove it than to fix remoting. NuGet is also our primary means of distribution now and it does not work with that and I expect it breaks NuGet addins. 

I have removed the functionality. I kept the parameter to maintain API compatibility, but renamed it to unused. 

I thought we had documentation on using the engine, but I cannot find it. @CharliePoole am I mistaken or can you point me to it? I assume it will need to be updated if this is merged.

### DIFF
--- a/src/NUnitConsole/nunit3-console/Program.cs
+++ b/src/NUnitConsole/nunit3-console/Program.cs
@@ -105,7 +105,7 @@ namespace NUnit.ConsoleRunner
                     return ConsoleRunner.INVALID_ARG;
                 }
 
-                using (ITestEngine engine = TestEngineActivator.CreateInstance(false))
+                using (ITestEngine engine = TestEngineActivator.CreateInstance())
                 {
                     if (Options.WorkDirectory != null)
                         engine.WorkDirectory = Options.WorkDirectory;
--- a/src/NUnitEngine/nunit.engine.api/TestEngineActivator.cs
+++ b/src/NUnitEngine/nunit.engine.api/TestEngineActivator.cs
@@ -8,10 +8,10 @@
 // distribute, sublicense, and/or sell copies of the Software, and to
 // permit persons to whom the Software is furnished to do so, subject to
 // the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be
 // included in all copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
 // EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
 // MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
@@ -39,40 +39,31 @@ namespace NUnit.Engine
         private const string DefaultAssemblyName = "nunit.engine.dll";
         internal const string DefaultTypeName = "NUnit.Engine.TestEngine";
 
-        private const string NunitInstallRegKey = @"SOFTWARE\Nunit.org\Engine";
-        private const string NunitInstallRegKeyWow64 = @"SOFTWARE\Wow6432Node\Nunit.org\Engine";
-
         #region Public Methods
 
         /// <summary>
         /// Create an instance of the test engine.
         /// </summary>
-        /// <remarks>If private copy is false, the search order is the NUnit install directory for the current user, then
-        /// the install directory for the local machine and finally the current AppDomain's ApplicationBase.</remarks>
-        /// <param name="privateCopy">if set to <c>true</c> loads the engine found in the application base directory, 
-        /// otherwise searches for the test engine with the highest version installed. Defaults to <c>true</c>.</param>
+        /// <param name="unused">This parameter is no longer used but has not been removed to ensure API compatibility.</param>
         /// <exception cref="NUnitEngineNotFoundException">Thrown when a test engine of the required minimum version is not found</exception>
         /// <returns>An <see cref="NUnit.Engine.ITestEngine"/></returns>
-        public static ITestEngine CreateInstance(bool privateCopy = false)
+        public static ITestEngine CreateInstance(bool unused = false)
         {
-            return CreateInstance(DefaultMinimumVersion, privateCopy);
+            return CreateInstance(DefaultMinimumVersion, unused);
         }
 
         /// <summary>
         /// Create an instance of the test engine with a minimum version.
         /// </summary>
-        /// <remarks>If private copy is false, the search order is the NUnit install directory for the current user, then
-        /// the install directory for the local machine and finally the current AppDomain's ApplicationBase.</remarks>
         /// <param name="minVersion">The minimum version of the engine to return inclusive.</param>
-        /// <param name="privateCopy">if set to <c>true</c> loads the engine found in the application base directory, 
-        /// otherwise searches for the test engine with the highest version installed. Defaults to <c>true</c>.</param>
-        /// <exception cref="NUnitEngineNotFoundException">Thrown when a test engine of the given minimum version is not found</exception>
+        /// <param name="unused">This parameter is no longer used but has not been removed to ensure API compatibility.</param>
+        /// <exception cref="NUnitEngineNotFoundException">Thrown when a test engine of the required minimum version is not found</exception>
         /// <returns>An <see cref="ITestEngine"/></returns>
-        public static ITestEngine CreateInstance(Version minVersion, bool privateCopy = false)
+        public static ITestEngine CreateInstance(Version minVersion, bool unused = false)
         {
             try
             {
-                Assembly engine = FindNewestEngine(minVersion, privateCopy);
+                Assembly engine = FindNewestEngine(minVersion);
                 if (engine == null)
                 {
                     throw new NUnitEngineNotFoundException(minVersion);
@@ -93,7 +84,7 @@ namespace NUnit.Engine
 
         #region Private Methods
 
-        private static Assembly FindNewestEngine(Version minVersion, bool privateCopy)
+        private static Assembly FindNewestEngine(Version minVersion)
         {
             var newestVersionFound = new Version();
 
@@ -116,18 +107,18 @@ namespace NUnit.Engine
                 }
             }
 
-            // Check for an engine that has been installed as part of a 
-            // development project using NuGet. 
+            // Check for an engine that has been installed as part of a
+            // development project using NuGet.
             //
             // NOTE: This code assumes that we are working in
             // bin/Debug or bin/Release and includes a lot of
             // knowledge of how we handle nuget packages. It
             // is only intended for use while a new runner is
-            // under development. 
+            // under development.
             if (newestAssemblyFound == null)
             {
                 var packages = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "../../packages");
-     
+
                 if (Directory.Exists(packages))
                 {
                     var packagesDir = new DirectoryInfo(packages);
@@ -145,25 +136,6 @@ namespace NUnit.Engine
                             break;
                     }
                 }
-            }
-
-            if (!privateCopy)
-            {
-                // Check the install for the current user, 32 bit process
-                path = FindEngineInRegistry(Registry.CurrentUser, NunitInstallRegKey);
-                newestAssemblyFound = CheckPathForEngine(path, minVersion, ref newestVersionFound, newestAssemblyFound);
-
-                // Check the install for the current user, 64 bit process
-                path = FindEngineInRegistry(Registry.CurrentUser, NunitInstallRegKeyWow64);
-                newestAssemblyFound = CheckPathForEngine(path, minVersion, ref newestVersionFound, newestAssemblyFound);
-
-                // check the install for the local machine, 32 bit process
-                path = FindEngineInRegistry(Registry.LocalMachine, NunitInstallRegKey);
-                newestAssemblyFound = CheckPathForEngine(path, minVersion, ref newestVersionFound, newestAssemblyFound);
-
-                // check the install for the local machine, 64 bit process
-                path = FindEngineInRegistry(Registry.LocalMachine, NunitInstallRegKeyWow64);
-                newestAssemblyFound = CheckPathForEngine(path, minVersion, ref newestVersionFound, newestAssemblyFound);
             }
 
             return newestAssemblyFound;


rprouse:
Interestingly, when we were testing earlier attempts to fix #255, we asked users to download dev builds from MyGet. This were always newer than any MSI that they would have had installed, so the local copy would have been used, thus appearing to fix the problem 😈

rprouse:
Travis failure is Mono Latest. It looks like Mono 5.2 has been released. The failure looks like a failure of our integration test, but is followed by a mono crash callstack so I am unsure if it is a real failure or not.

```
1) Failed : NUnit.Engine.Tests.Integration.RemoteAgentTests.Explore_does_not_throw_SocketException
  Expected: no item property IsError True
  But was:  < [stdout] NUnit Console Runner 3.8.0 
Copyright (c) 2017 Charlie Poole, Rob Prouse
Runtime Environment
   OS Version: Linux 4.4.0.83 
  CLR Version: 4.0.30319.42000
Test Files
    /tmp/iv9k7fgv.41f/mock-assembly.dll
* Assertion at mono-threads.c:544, condition `info' not met
, [stderr] /proc/self/maps:
00400000-007a2000 r-xp 00000000 08:01 2644                               /usr/bin/mono-sgen
009a1000-009a2000 r--p 003a1000 08:01 2644                               /usr/bin/mono-sgen
009a2000-009a5000 rw-p 003a2000 08:01 2644                               /usr/bin/mono-sgen
009a5000-009d9000 rw-p 00000000 00:00 0 
01b37000-01ea6000 rw-p 00000000 00:00 0                                  [heap]
40ce1000-40d41000 rwxp 00000000 00:00 0 
41791000-417a1000 rwxp 00000000 00:00 0 
41e55000-41f48000 rwxp 00000000 00:00 0 

SNIP...
```

rprouse:
Looking again at the Mono failure, it is definately a Mono crash and not our test failing. The test launches the console which crashes on Mono.

CharliePoole:
The only question I have is whether we should keep the old method signature and throw a not implemented exception of true is passed. As you know, we have been maintaining a stable API up to now, with only additive changes and keeping the assembly version fixed at 3.0.0.

I'll check for the docs when I'm back on my computer.

CharliePoole:
Engine API docs: https://github.com/nunit/docs/wiki/Test-Engine-API

CharliePoole:
I dislike seeing integration tests mixed in with unit tests.

ChrisMaddock:
The changes look good. 

I understand how this fixes #6 - but I'm missing the link to #255. Why do we think this is a cause of #255 - is it the same stack trace? 

#255 so far has been about a flakey race-condition crash caused by the remoting protocol. This fixes #6 - which seems deterministic, and a separate issue caused by engine location. Have I missed something that links the two?

If we are closing #255 - @jnm2 should have a look - he's put a lot of time into diagnosing #255 over the last few months.

CharliePoole:
Even though @ChrisMaddock has approved this for merge, I think we have to first address the compatibility issue I raised. Keeping the API backward-compatible has been pretty fundamental up to now.

rprouse:
The error message and callstack looked the same for #255. My guess is that @jnm2 did fix the intermittent failures as reported, but we reopened when people reported the same error was still happening. I might be wrong though.

As for API compatibility, my thinking is that throwing an exception is breaking the API. If we are going to do that, then I would prefer to see it at compile time. I would be okay with dropping the parameter since it's an easy fix, but if we do, we should probably increment the engine version.

rprouse:
I mean the engine API version...

CharliePoole:
Compile time only works if you are compiling. We currently support dropping in a new API and/or engine assembly in place. If we won't continue to support it, then we should announce this as a breaking change.

I don't think this would affect a lot of people, but it bothers me that the ones most likely to be affected are those who have been paying attention to our previous statements of direction. It's of course up to you.

jnm2:
I'm happy about this PR. It vastly simplifies complexity when diagnosing engine bugs.
After the dust settles from this PR, is it even possible that people's remaining SocketExceptions will be gone?
Either way I've already decided not to put any more time into diagnosing the SocketException but rather provide a non-remoting implementation where we control the transport.

jnm2:
I prefer compile-time breaks (via principle of failing as soon and loud as possible) but only if we put that at the top of the release notes and increment the major version. I could also begrudgingly support incrementing just the minor version, given sufficiently clear messaging.

ChrisMaddock:
I guess the other resolution to this problem would be to just have the console require a local engine, and leave the functionality in there - purely for API compatibility.

rprouse:
> Compile time only works if you are compiling. We currently support dropping in a new API and/or engine assembly in place. If we won't continue to support it, then we should announce this as a breaking change.

As this PR stands, if they drop the current version in over any other engine, it will continue to work. If we throw an exception, then we are breaking them just as much as if they hit a method not found exception. That is why I coded it this way. I think I would prefer the code as it is as it works no matter what. We can update the documentation accordingly. Any objections to leaving it as is?

jnm2:
No objections. Wish we could label this issue to track for engine engine vNext.

CharliePoole:
If it continued to work, I would have no objection. However, if the code is calling the method with the second boolean argument and that entry no longer exists in the new API, then it would surprise me a lot if it continued to work. 😄

CharliePoole:
Or are you saying we can change the engine so long as we don't change the API assembly?

rprouse:
> If it continued to work, I would have no objection. However, if the code is calling the method with the second boolean argument and that entry no longer exists in the new API, then it would surprise me a lot if it continued to work.

@CharliePoole by continues to work, I mean that it will continue to run if you pass in a boolean parameter, either true or false. The parameter is now `unused` so it will not have an effect, but it will not crash. If we changed it to throw an `ArgumentException`, a user that dropped a new engine in a runner that passes true would start crashing with that exception. Personally, I think continues to run is the better option.

> Or are you saying we can change the engine so long as we don't change the API assembly?

I don't think we should change the signature of the API assembly, but that we can change the underlying behavior of the engine, especially in this case where the promised behavior didn't work anyway.

That said, I actually don't mind changing the API assembly either if we increment the version and force 3rd Party runners to do a recompile when they update. Personally, I think the number of people that upgrade the engine by dropping individual copies of DLL's in a bin directory probably approaches zero.

As I see it, we are discussing 3 options,

1. Make the `privateCopy` parameter do nothing. Advantage, does not require recompile and works. Disadvantage, does not inform users that it no longer works to pass true even though it never worked.
2. Throw `ArgumentException` if `privateCopy=true`. Advantage, users know at runtime that the behavior has changed. Disadvantage, crashes and they need to rollback the engine and vendors need to update.
3. Change the API forcing 3rdparty vendors to update. Advantage, 3rdparty vendors know the option no longer supported. Disadvantage, 3rdparty vendors need to recompile when they update and any user that drops files into a 3rd party install hoping to update will crash if the vendor didn't rely on the optional argument.

Personally, I favor the third option as I think it balances letting the vendors know of the change along with a minor obvious update for the vendors. I don't think anyone updates the engine by overwriting files, so I think that disadvantage is minor.

I did code this PR to be safe though and didn't change the API.

@CharliePoole I think you are advocating the second option, but you said "Keeping the API backward-compatible has been pretty fundamental up to now." which seems to conflict with the second option.

CharliePoole:
Seems like I misunderstood what you are doing. I understood you favored option 3 and was offering option 2 as a compromise. If we are doing option 1, I'm completely happy.

As far as option 2 breaking the API, it doesn't if you're an old COM guy, where interfaces had to be satisfied but implementations frequently threw not implemented. 😸

Frudy:
Just upgraded to NUnit 3 (from 2.6) and we are having trouble with this issue. Will the fix be out soon? And what can I did meanwhile to get around it?

rprouse:
@Frudy we will hopefully get a new release out soon. In the meantime, just uninstall the MSI from your developer machines and use the copy of the console in the NuGet packages directory.

Frudy:
Problem is running tests on CI with Team City 10. Forgot to mention. NUnit not explicitly installed on build agent. Any temporal workaround for this?

rprouse:
@Frudy if NUnit is not installed on the build agent, then the issue you are having is unrelated to this issue. Please enter a new issue with a description of the errors you are seeing and any other information that might help us diagnose your problem.

Issue #273, opened 2017-07-28
Insufficient error handling message in ProcessRunner -> RunTests method


ChrisMaddock:
@Fernell commented on [Fri Jul 28 2017](https://github.com/nunit/nunit/issues/2334)

log.Error("Failed to run remote tests {0}", e.Message); is insufficeint. It lacks details required to diagnose problem. Here is sample of my log:
> An exception occurred in the driver while loading tests.
> Server stack trace: 
>    at NUnit.Engine.Runners.DirectTestRunner.LoadDriver(IFrameworkDriver driver, String testFile, TestPackage subPackage)
>    at NUnit.Engine.Runners.DirectTestRunner.LoadPackage()
>    at NUnit.Engine.Runners.TestDomainRunner.LoadPackage()
>    at NUnit.Engine.Runners.DirectTestRunner.EnsurePackageIsLoaded()
>    at NUnit.Engine.Runners.DirectTestRunner.RunTests(ITestEventListener listener, TestFilter filter)
>    at NUnit.Engine.Runners.AbstractTestRunner.Run(ITestEventListener listener, TestFilter filter)
>    at NUnit.Engine.Runners.AggregatingTestRunner.RunTestsSequentially(ITestEventListener listener, TestFilter filter, List`1 results, Boolean disposeRunners)
>    at NUnit.Engine.Runners.AggregatingTestRunner.RunTests(ITestEventListener listener, TestFilter filter)
>    at NUnit.Engine.Runners.AbstractTestRunner.Run(ITestEventListener listener, TestFilter filter)
>    at NUnit.Engine.Agents.RemoteTestAgent.Run(ITestEventListener listener, TestFilter filter)
>    at System.Runtime.Remoting.Messaging.StackBuilderSink._PrivateProcessMessage(IntPtr md, Object[] args, Object server, Object[]& outArgs)
>    at System.Runtime.Remoting.Messaging.StackBuilderSink.PrivateProcessMessage(RuntimeMethodHandle md, Object[] args, Object server, Object[]& outArgs)
>    at System.Runtime.Remoting.Messaging.StackBuilderSink.SyncProcessMessage(IMessage msg)

I was unable to recognize error only by this stack trace. I was forced to download nunit3-console binaries and debug it myself. I found out that useful error message was in ex.InnerException which is not returned in error message at all. Please extend error message by adding ex.InnerException message.

---

[...]

PR #301, opened 2017-10-19
Improve Exception information on Console/Engine Crash

### LINKED ISSUE #284: NUnit3: An exception occured in the driver while loading tests...   bei NUnit.Engine.Runners.ProcessRunner.RunTests(ITestEventListener listener, TestFilter filter)
@KlausNUnit commented on [Tue Aug 22 2017](https://github.com/nunit/nunit/issues/2377)

[TestResult.zip](https://github.com/nunit/nunit/files/1241797/TestResult.zip)

Hi guys!
When I try to run tests, I get alway an exception very early.
I went through the internet and also here and saw similiar "issues". 
I tried to add --inprocess, --trace:Debug 
but didnt get more information about the root issue.

The dll is working fine with NUnit Version 2.6.x, but unfortunately I have to pass parameters to the tests and found no suitable solution with 2.6.

Pls. find attached the console output as png, furthermore the ResultFile as Zip.
If there is something missing to analyse this issue, I would be happy to add more information.

Thanks for help in advance.

BR
 Klaus

console output
packages_config

---

@mikkelbu commented on [Tue Aug 22 2017](https://github.com/nunit/nunit/issues/2377#issuecomment-324003020)

Hi @KlausNUnit 
Can you also attach the trace files. They are named InternalTrace.<NUM>.*.dll and should be in the current folder unless specified otherwise.

---

@KlausNUnit commented on [Tue Aug 22 2017](https://github.com/nunit/nunit/issues/2377#issuecomment-324005702)

@mikkelbu: Thanks for pointing to the trace file.

I just ran a new session and X-ed the sensitive information.

[InternalTrace.14872.zip](https://github.com/nunit/nunit/files/1241922/InternalTrace.14872.zip)

---

@KlausNUnit commented on [Tue Aug 22 2017](https://github.com/nunit/nunit/issues/2377#issuecomment-324008783)

When inspecting the InternalTrace file, there is a cross reference for the logfile of the nunit-agent.

I add this log here for further investigation.
[nunit-agent_8308.zip](https://github.com/nunit/nunit/files/1241950/nunit-agent_8308.zip)

---

@rprouse commented on [Tue Aug 22 2017](https://github.com/nunit/nunit/issues/2377#issuecomment-324066697)

I've looked at the trace files and don't see anything helpful, just the same error as reported in the test run,

```
13:57:25.857 Error [ 1] ProcessRunner: Failed to run remote tests An exception occurred in the driver while loading tests.
```

Can you try running your tests with the `--inprocess` flag?

Also, please take a look at the NUnit reference in Visual Studio. Is it pointing to the `net45` directory in the NuGet package?

---

@KlausNUnit commented on [Tue Aug 22 2017](https://github.com/nunit/nunit/issues/2377#issuecomment-324078589)

I gave --inprocess already a try (see top of the screenshot and thread start). Same issue.

I have to check the NUnit reference, I will do this tomorrow and report.

---

@CharliePoole commented on [Tue Aug 22 2017](https://github.com/nunit/nunit/issues/2377#issuecomment-324081305)

An error while loading tests typically points to some problem in a `TestCaseSource` method. It usually gives a better error message, however. Does that suggest anything to you, considering what's in your tests? 

---

@KlausNUnit commented on [Tue Aug 22 2017](https://github.com/nunit/nunit/issues/2377#issuecomment-324217877)

NUnit reference is ok. It is pointing to net45
path to nunit framework dll in packages

---

@KlausNUnit commented on [Wed Aug 23 2017](https://github.com/nunit/nunit/issues/2377#issuecomment-324249978)

@ CharliePoole: TestCaseSource was not used in any code.

For considering what's in my tests, I have now disentangled the code and excluded lot of code, so that only my "fokus" will be executed later on.
It was the same behavior.

Then I did the following crazy thing:
I changed ONLY the r e f e r e n c e   for the nunit.framework.dll from 3.**7** to **2.5** and compiled the code again.
Therefore the content of packages.config was not touched (it was still pointing to 3.7 framework and testadapter).
Then I ran the tests against the NUnit**3**-console, only with parameter --trace:Debug, with **success!**
The test was executed properly, the internal-trace shows:
Info  [ 1] ProcessRunner: Done running  

Afterwards I tried to find out, if a "downgrade" of NUnit-framework version can help to avoid my issue.
I tried: 3.6.1 and 3.0.0. (theNuGet PM has resolved a lot of dependencies!)
Same error!

Now I have deinstalled all NuGet packages to have a clean area and went back to NUnit 3.7.1 and TestAdapter.

More ideas?

---

@CharliePoole commented on [Wed Aug 23 2017](https://github.com/nunit/nunit/issues/2377#issuecomment-324391697)

Why is the test adapter installed? Although unlikely, it's conceivable that it is pulling in a version of the NUnit engine that conflicts with what the console runner uses.

Now that I think about it, where does your console runner come from? That is, how did you install it?

---

@KlausNUnit commented on [Wed Aug 23 2017](https://github.com/nunit/nunit/issues/2377#issuecomment-324410440)

Maybe the test adapter was installed by a mistake.
I am not absolutely sure, but as I remember I picked the zip up from here:
[https://github.com/nunit/nunit-console/releases/tag/3.7](url)

The "installation" of the zip was done by extracting in a certain folder. Is this wrong?
My intention is to use the commandline for a POC.

---

@KlausNUnit commented on [Wed Aug 23 2017](https://github.com/nunit/nunit/issues/2377#issuecomment-324412972)

Additional information: the "driver" will be compiled on system A with VS2017 Enterprise.
The output (dll) will bei tested on system A and when verified copied to system B, which is a different system (VirtServer). 

---

@CharliePoole commented on [Wed Aug 23 2017](https://github.com/nunit/nunit/issues/2377#issuecomment-324412980)

Nothing wrong with your approach of installing the zip. If you look in the directory where you placed it, you'll see a number of assemblies, which also exist in the directory under your project where nuget placed the vs adapter. That's the conflict I am wondering about. In theory, so long as none of the assemblies are copied o your output directory, nothing should go wrong. 😸 But perhaps you should remove the adapter package just to be sure.

Another thought. Have you installed the nunit msi at any time? If so, uninstall it! That too may be causing a conflict due to a bug we have fixed but not yet released.

---

@KlausNUnit commented on [Thu Aug 24 2017](https://github.com/nunit/nunit/issues/2377#issuecomment-324611305)

I am ashamed!
The issue was sitting in front of the keyboard.

I forgot to copy the nunit.framework.dll v3.7.1 to the folder where the binaries are executed. 

I am wondering, why there came out no information that the assembly doesnt fit. 

---

@CharliePoole commented on [Thu Aug 24 2017](https://github.com/nunit/nunit/issues/2377#issuecomment-324712271)

The exception you got is an NUnitEngineException with a nested exception, which probably provides a more meaningful message. The console runner assumes that an NUnitEngineException is self-explanatory and so just displays the message. However, we are throwing such exceptions with a generic message and an Inner exception in some cases.

Somebody needs to examine all our use of this exception type and either make the messages fully explanatory or change the console to display any inner exception. I'm moving this to the nunit-console repository, which has the code for both the engine and the console.

ChrisMaddock:
Fixes #284, and covers the error-reporting side of #273.

The NUnit Console hasn't recently had the best error reporting for unexpected crashes. Exceptions generally are wrapped up inside an NUnitEngineException - which obscures the original exception and makes debugging problematic for the user.

This PR brings the framework's Exception hierachy flattening logic over to the console/engine. It assumes that any unexpected NUnitEngineException which would return a -100 from the console should display as much information as possible.

The problem is tackled in multiple places, as inprocess tests and out of process tests report the exceptions differently. I'll comment on the code diff as to what comes through where.

Looking in particular at the case in #273, the changes look as such:

**Old Message running in process**
```
An exception occured in the driver while loading tests.
```

**New Message running in process**
image

**Old Message running out of process**
image

**New Message running out of process**
image

### DIFF
--- a/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
+++ b/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
@@ -243,7 +243,7 @@ namespace NUnit.ConsoleRunner
 
                 // Since we got a result, we display any engine exception as a warning
                 if (engineException != null)
-                    writer.WriteLine(ColorStyle.Warning, Environment.NewLine + engineException.Message);
+                    writer.WriteLine(ColorStyle.Warning, Environment.NewLine + ExceptionHelper.BuildMessage(engineException));
 
                 if (reporter.Summary.UnexpectedError)
                     return ConsoleRunner.UNEXPECTED_ERROR;
@@ -258,7 +258,11 @@ namespace NUnit.ConsoleRunner
 
             // If we got here, it's because we had an exception, but check anyway
             if (engineException != null)
-                writer.WriteLine(ColorStyle.Error, engineException.Message);
+            {
+                writer.WriteLine(ColorStyle.Error, ExceptionHelper.BuildMessage(engineException));
+                writer.WriteLine();
+                writer.WriteLine(ColorStyle.Error, ExceptionHelper.BuildStackTrace(engineException));
+            }
 
             return ConsoleRunner.UNEXPECTED_ERROR;
         }
--- a/src/NUnitConsole/nunit3-console/Program.cs
+++ b/src/NUnitConsole/nunit3-console/Program.cs
@@ -134,7 +134,9 @@ namespace NUnit.ConsoleRunner
                     }
                     catch (Exception ex)
                     {
-                        OutWriter.WriteLine(ColorStyle.Error, ex.ToString());
+                        OutWriter.WriteLine(ColorStyle.Error, ExceptionHelper.BuildMessage(ex));
+                        OutWriter.WriteLine();
+                        OutWriter.WriteLine(ColorStyle.Error, ExceptionHelper.BuildStackTrace(ex));
                         return ConsoleRunner.UNEXPECTED_ERROR;
                     }
                     finally
@@ -147,8 +149,6 @@ namespace NUnit.ConsoleRunner
                                 Console.ReadKey(true);
                             }
                         }
-
-                        //    log.Info( "NUnit3-console.exe terminating" );
                     }
                 }
             }
--- a/src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
+++ b/src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
@@ -22,6 +22,7 @@
 // ***********************************************************************
 
 using System;
+using NUnit.Common;
 using NUnit.Engine.Internal;
 using NUnit.Engine.Services;
 
@@ -308,8 +309,8 @@ namespace NUnit.Engine.Runners
             XmlHelper.AddAttribute(suite, "asserts", "0");
 
             var failure = suite.AddElement("failure");
-            failure.AddElementWithCDataSection("message", e.Message);
-            failure.AddElementWithCDataSection("stack-trace", e.StackTrace);
+            failure.AddElementWithCDataSection("message", ExceptionHelper.BuildMessage(e));
+            failure.AddElementWithCDataSection("stack-trace", ExceptionHelper.BuildStackTrace(e));
 
             return new TestEngineResult(suite);
         }
--- a/src/NUnitConsole/nunit3-console.tests/ExceptionHelperTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/ExceptionHelperTests.cs
@@ -1,0 +1,85 @@
+﻿// ***********************************************************************
+// Copyright (c) 2017 Charlie Poole, Rob Prouse
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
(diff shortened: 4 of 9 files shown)


src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -258,7 +259,11 @@ private int RunTests(TestPackage package, TestFilter filter)
 
             // If we got here, it's because we had an exception, but check anyway
             if (engineException != null)
-                writer.WriteLine(ColorStyle.Error, engineException.Message);
+            {
+                writer.WriteLine(ColorStyle.Error, ExceptionHelper.BuildMessage(engineException));
+                writer.WriteLine();
+                writer.WriteLine(ColorStyle.Error, ExceptionHelper.BuildStackTrace(engineException));
+            }
```
ChrisMaddock: In-process exceptions are reported from here

src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
```diff
@@ -308,8 +308,8 @@ TestEngineResult CreateFailedResult(Exception e)
             XmlHelper.AddAttribute(suite, "asserts", "0");
 
             var failure = suite.AddElement("failure");
-            failure.AddElementWithCDataSection("message", e.Message);
-            failure.AddElementWithCDataSection("stack-trace", e.StackTrace);
+            failure.AddElementWithCDataSection("message", ExceptionHelper.BuildMessage(e));
+            failure.AddElementWithCDataSection("stack-trace", ExceptionHelper.BuildStackTrace(e));
```
ChrisMaddock: Out of process exceptions are reported via xml, and logged here

src/NUnitConsole/nunit3-console/Program.cs
```diff
@@ -134,7 +135,9 @@ public static int Main(string[] args)
                     }
                     catch (Exception ex)
                     {
-                        OutWriter.WriteLine(ColorStyle.Error, ex.ToString());
+                        OutWriter.WriteLine(ColorStyle.Error, ExceptionHelper.BuildMessage(ex));
+                        OutWriter.WriteLine();
+                        OutWriter.WriteLine(ColorStyle.Error, ExceptionHelper.BuildStackTrace(ex));
```
ChrisMaddock: This is a very top-level try catch, which would presumably handle exceptions which occur within the console, rather than engine. Seemed sensible to additionally add the functionality here.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -31,6 +31,7 @@
 using NUnit.Engine.Extensibility;
 using System.Runtime.InteropServices;
 using System.Text;
+using NUnit.Engine.Internal;
```
CharliePoole: Since the console does not reference the engine, it seems a bit confusing to have the utility in the `NUnit.Internal.Namespace`. Can it go somewhere else that signals it is used in common?

ChrisMaddock: Looks like we still have an NUnit.Common namespace. I can switch it to that. 🙂

CharliePoole:
Sounds good!

ChrisMaddock:
@CharliePoole - made the suggested change - but I can't merge - it needs one approved review. 🙂

CharliePoole:
Re-approved it. This GitHub review feature is more trouble than it's worth if you ask me.

PR #304, opened 2017-10-24
Catch agent debugger launch exceptions, and improve agent crash handling

### LINKED ISSUE #228: System.Reflection.TargetInvocationException with nunit3-console --debug on Mono
@PJB3005 commented on [Fri May 26 2017](https://github.com/nunit/nunit/issues/2191)

Attempting to run `nunit3-console --debug` 3.6.1 on Mono causes it to break due to the following error:
```
$ mono nunit3-console.exe --debug --inprocess ../../../SS14.UnitTesting/bin/Debug/SS14.UnitTesting.dll
NUnit Console Runner 3.6.1
Copyright (C) 2017 Charlie Poole

Runtime Environment
   OS Version: MacOSX 16.5.0.0
  CLR Version: 4.0.30319.42000

Test Files
    ../../../SS14.UnitTesting/bin/Debug/SS14.UnitTesting.dll

System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.NotImplementedException: The method or operation is not implemented.
  at System.Diagnostics.Debugger.Launch () [0x00000] in <ffb99659fc1c47faa4d5f883014d08aa>:0
  at NUnit.Framework.Api.NUnitTestAssemblyRunner.StartRun (NUnit.Framework.Interfaces.ITestListener listener) [0x000c1] in <a136915bdafc4f7db46d0b0a906d5367>:0
  at NUnit.Framework.Api.NUnitTestAssemblyRunner.RunAsync (NUnit.Framework.Interfaces.ITestListener listener, NUnit.Framework.Interfaces.ITestFilter filter) [0x0006f] in <a136915bdafc4f7db46d0b0a906d5367>:0
  at NUnit.Framework.Api.NUnitTestAssemblyRunner.Run (NUnit.Framework.Interfaces.ITestListener listener, NUnit.Framework.Interfaces.ITestFilter filter) [0x00000] in <a136915bdafc4f7db46d0b0a906d5367>:0
  at NUnit.Framework.Api.FrameworkController.RunTests (System.Web.UI.ICallbackEventHandler handler, System.String filter) [0x0001d] in <a136915bdafc4f7db46d0b0a906d5367>:0
  at (wrapper remoting-invoke-with-check) NUnit.Framework.Api.FrameworkController:RunTests (System.Web.UI.ICallbackEventHandler,string)
  at NUnit.Framework.Api.FrameworkController+RunTestsAction..ctor (NUnit.Framework.Api.FrameworkController controller, System.String filter, System.Object handler) [0x00006] in <a136915bdafc4f7db46d0b0a906d5367>:0
  at (wrapper managed-to-native) System.Reflection.MonoCMethod:InternalInvoke (System.Reflection.MonoCMethod,object,object[],System.Exception&)
  at System.Reflection.MonoCMethod.InternalInvoke (System.Object obj, System.Object[] parameters) [0x00002] in <ffb99659fc1c47faa4d5f883014d08aa>:0


Server stack trace:
  at System.Reflection.MonoCMethod.InternalInvoke (System.Object obj, System.Object[] parameters) [0x00014] in <ffb99659fc1c47faa4d5f883014d08aa>:0
  at System.Reflection.MonoCMethod.DoInvoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0007a] in <ffb99659fc1c47faa4d5f883014d08aa>:0
  at System.Reflection.MonoCMethod.Invoke (System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00000] in <ffb99659fc1c47faa4d5f883014d08aa>:0
  at System.RuntimeType.CreateInstanceImpl (System.Reflection.BindingFlags bindingAttr, System.Reflection.Binder binder, System.Object[] args, System.Globalization.CultureInfo culture, System.Object[] activationAttributes, System.Threading.StackCrawlMark& stackMark) [0x00213] in <ffb99659fc1c47faa4d5f883014d08aa>:0
  at System.Activator.CreateInstance (System.Type type, System.Reflection.BindingFlags bindingAttr, System.Reflection.Binder binder, System.Object[] args, System.Globalization.CultureInfo culture, System.Object[] activationAttributes) [0x00095] in <ffb99659fc1c47faa4d5f883014d08aa>:0
  at System.Activator.CreateInstance (System.String assemblyString, System.String typeName, System.Boolean ignoreCase, System.Reflection.BindingFlags bindingAttr, System.Reflection.Binder binder, System.Object[] args, System.Globalization.CultureInfo culture, System.Object[] activationAttributes, System.Security.Policy.Evidence securityInfo, System.Threading.StackCrawlMark& stackMark) [0x00077] in <ffb99659fc1c47faa4d5f883014d08aa>:0
  at System.Activator.CreateInstance (System.String assemblyName, System.String typeName, System.Boolean ignoreCase, System.Reflection.BindingFlags bindingAttr, System.Reflection.Binder binder, System.Object[] args, System.Globalization.CultureInfo culture, System.Object[] activationAttributes, System.Security.Policy.Evidence securityInfo) [0x00016] in <ffb99659fc1c47faa4d5f883014d08aa>:0
  at System.AppDomain.CreateInstance (System.String assemblyName, System.String typeName, System.Boolean ignoreCase, System.Reflection.BindingFlags bindingAttr, System.Reflection.Binder binder, System.Object[] args, System.Globalization.CultureInfo culture, System.Object[] activationAttributes, System.Security.Policy.Evidence securityAttributes) [0x0000e] in <ffb99659fc1c47faa4d5f883014d08aa>:0
  at (wrapper remoting-invoke-with-check) System.AppDomain:CreateInstance (string,string,bool,System.Reflection.BindingFlags,System.Reflection.Binder,object[],System.Globalization.CultureInfo,object[],System.Security.Policy.Evidence)
  at System.AppDomain.CreateInstanceAndUnwrap (System.String assemblyName, System.String typeName, System.Boolean ignoreCase, System.Reflection.BindingFlags bindingAttr, System.Reflection.Binder binder, System.Object[] args, System.Globalization.CultureInfo culture, System.Object[] activationAttributes, System.Security.Policy.Evidence securityAttributes) [0x00000] in <ffb99659fc1c47faa4d5f883014d08aa>:0
  at (wrapper remoting-invoke-with-check) System.AppDomain:CreateInstanceAndUnwrap (string,string,bool,System.Reflection.BindingFlags,System.Reflection.Binder,object[],System.Globalization.CultureInfo,object[],System.Security.Policy.Evidence)
  at (wrapper xdomain-dispatch) System.AppDomain:CreateInstanceAndUnwrap (object,byte[]&,byte[]&,string,string,bool)

Exception rethrown at [0]:
  at (wrapper xdomain-invoke) System.AppDomain:CreateInstanceAndUnwrap (string,string,bool,System.Reflection.BindingFlags,System.Reflection.Binder,object[],System.Globalization.CultureInfo,object[],System.Security.Policy.Evidence)
  at (wrapper remoting-invoke-with-check) System.AppDomain:CreateInstanceAndUnwrap (string,string,bool,System.Reflection.BindingFlags,System.Reflection.Binder,object[],System.Globalization.CultureInfo,object[],System.Security.Policy.Evidence)
  at NUnit.Engine.Drivers.NUnit3FrameworkDriver.CreateObject (System.String typeName, System.Object[] args) [0x00000] in <7b0953727751470ab10f0e0b547b85ae>:0
  at NUnit.Engine.Drivers.NUnit3FrameworkDriver.Run (NUnit.Engine.ITestEventListener listener, System.String filter) [0x00030] in <7b0953727751470ab10f0e0b547b85ae>:0
  at NUnit.Engine.Runners.DirectTestRunner.RunTests (NUnit.Engine.ITestEventListener listener, NUnit.Engine.TestFilter filter) [0x0002b] in <7b0953727751470ab10f0e0b547b85ae>:0
  at NUnit.Engine.Runners.AbstractTestRunner.Run (NUnit.Engine.ITestEventListener listener, NUnit.Engine.TestFilter filter) [0x00000] in <7b0953727751470ab10f0e0b547b85ae>:0
  at NUnit.Engine.Runners.MasterTestRunner.RunTests (NUnit.Engine.ITestEventListener listener, NUnit.Engine.TestFilter filter) [0x00085] in <7b0953727751470ab10f0e0b547b85ae>:0
  at NUnit.Engine.Runners.MasterTestRunner.Run (NUnit.Engine.ITestEventListener listener, NUnit.Engine.TestFilter filter) [0x00000] in <7b0953727751470ab10f0e0b547b85ae>:0
  at NUnit.ConsoleRunner.ConsoleRunner.RunTests (NUnit.Engine.TestPackage package, NUnit.Engine.TestFilter filter) [0x000d2] in <b0ef9cac03db4431890fe739906a4c7b>:0
  at NUnit.ConsoleRunner.ConsoleRunner.Execute () [0x000b6] in <b0ef9cac03db4431890fe739906a4c7b>:0
  at NUnit.ConsoleRunner.Program.Main (System.String[] args) [0x001bf] in <b0ef9cac03db4431890fe739906a4c7b>:0
```

Remove the `--debug` flag and everything works.

@ChrisMaddock determined on Gitter that it's caused by a bug in Mono: https://bugzilla.xamarin.com/show_bug.cgi?id=30489

---

@ChrisMaddock commented on [Fri May 26 2017](https://github.com/nunit/nunit/issues/2191#issuecomment-304319675)

The issue is caused by the fact that System.Diagnostics.Debugger.Launch() isn't implemented on mono. I'm not sure how we could work around that at all to make `--debug` actually work.

@PJB3005 had to add `--inprocess` before he could see this exception - although I believe that will be already be handled better by https://github.com/nunit/nunit-console/issues/11.

Regarding the NotImplementedException, should we catch this, and throw an NUnitFrameworkException in its place? I'm not totally sure how the architecture is designed to cope with such things?
@nunit/framework-team

Edit: Marked this low priorty, as the next version of the console should improve the error message shown to the user, and make it easier to diagnose, at least!

---

@jnm2 commented on [Fri May 26 2017](https://github.com/nunit/nunit/issues/2191#issuecomment-304324694)

I wonder what barriers there are to getting it implemented on Mono?

---

@CharliePoole commented on [Fri May 26 2017](https://github.com/nunit/nunit/issues/2191#issuecomment-304335718)

Amazing that nobody has reported this before! This is one of those system-level things for which we don't actually test.

Clearly we should handle this gracefully, whatever that means. If you type `--debug` I imagine that you would prefer to see an error rather than just running the tests without the debugger. So, what kind of error... ?

1. Since the code is in the framework, the framework should detect this in addition to any upstream check. After all the framework can be called directly. I think it should simply mark the assembly non-runnable in that case, with an appropriate message. 

2. If we do that, I would expect any runner to give an intelligent report.

3. In additon, the console could detect that it is on mono and not allow the option in the first place. That's definitely cleaner than waiting for a failure.

ChrisMaddock:
This primarily fixes #228 - by catching exceptions thrown by launching the debugger in mono.

The solution is possibly a little over-architected, however I've attempted to use this problem as a way to improve our error handling for situations where the agent crashes.

I've introduced a number of agent exit codes, which can be read by the engine process and converted to an NUnitEngineException with suitable Debug information. The drawback is that this is triggered by the agent process `Exited` event.

This causes two issues that I can see.
1. We crash the engine. That said, once the agent has crashed, I think we're already at worst case scenario - so in this situation I think we can only help by providing additional debug information? In the console runner this terminates the application with a stack trace - I don't know about other runners, but I imagine it would be possible to hook into `AppDomain.UnhandledException`? @OsirisTerje - would this affect the adapter, or does it already handle unhandledexceptions?
2. There seems to be something of a slight race condition, in that exceptions that happen at the end of the agent process won't raise the event in time, and therefore won't be displayed. I don't think this is anything new, but by putting this additional functionality in, we should probably acknowledge it doesn't cover all scenarios.

### DIFF
--- a/src/NUnitEngine/nunit-agent/AgentExitCodes.cs
+++ b/src/NUnitEngine/nunit-agent/AgentExitCodes.cs
@@ -1,0 +1,36 @@
+﻿// ***********************************************************************
+// Copyright (c) 2017 Charlie Poole, Rob Prouse
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+namespace NUnit.Common
+{
+    internal static class AgentExitCodes
+    {
+        public const int OK = 0;
+        public const int PARENT_PROCESS_TERMINATED = -1;
+        public const int UNEXPECTED_EXCEPTION = -2;
+        public const int FAILED_TO_START_AGENT = -3;
+        public const int NO_DEBUGGER_SECURITY = -4;
+        public const int NO_DEBUGGER_NOT_IMPLEMENTED = -5;
+
+    }
+}
--- a/src/NUnitEngine/nunit-agent/AgentExitCodes.cs
+++ b/src/NUnitEngine/nunit-agent/AgentExitCodes.cs
@@ -28,9 +28,9 @@ namespace NUnit.Common
         public const int OK = 0;
         public const int PARENT_PROCESS_TERMINATED = -1;
         public const int UNEXPECTED_EXCEPTION = -2;
-        public const int FAILED_TO_START_AGENT = -3;
-        public const int NO_DEBUGGER_SECURITY = -4;
-        public const int NO_DEBUGGER_NOT_IMPLEMENTED = -5;
+        public const int FAILED_TO_START_REMOTE_AGENT = -3;
+        public const int DEBUGGER_SECURITY_VIOLATION = -4;
+        public const int DEBUGGER_NOT_IMPLEMENTED = -5;
 
     }
 }
--- a/src/NUnitEngine/nunit-agent/AgentExitCodes.cs
+++ b/src/NUnitEngine/nunit-agent/AgentExitCodes.cs
@@ -27,10 +27,9 @@ namespace NUnit.Common
     {
         public const int OK = 0;
         public const int PARENT_PROCESS_TERMINATED = -1;
-        public const int UNEXPECTED_EXCEPTION = -2;
-        public const int FAILED_TO_START_REMOTE_AGENT = -3;
-        public const int DEBUGGER_SECURITY_VIOLATION = -4;
-        public const int DEBUGGER_NOT_IMPLEMENTED = -5;
-
+        public const int FAILED_TO_START_REMOTE_AGENT = -2;
+        public const int DEBUGGER_SECURITY_VIOLATION = -3;
+        public const int DEBUGGER_NOT_IMPLEMENTED = -4;
+        public const int UNEXPECTED_EXCEPTION = -100;
     }
 }
--- a/src/NUnitEngine/nunit-agent/Program.cs
+++ b/src/NUnitEngine/nunit-agent/Program.cs
@@ -23,6 +23,8 @@
 
 using System;
 using System.Diagnostics;
+using System.Security;
+using NUnit.Common;
 using NUnit.Engine;
 using NUnit.Engine.Agents;
 using NUnit.Engine.Internal;
@@ -51,6 +53,7 @@ namespace NUnit.Agent
             AgencyUrl = args[1];
 
             InternalTraceLevel traceLevel = InternalTraceLevel.Off;
+            int pid = Process.GetCurrentProcess().Id;
 
             for (int i = 2; i < args.Length; i++)
             {
@@ -61,11 +64,15 @@ namespace NUnit.Agent
                 if (arg == "--debug-agent")
                 {
                     if (!Debugger.IsAttached)
-                        Debugger.Launch();
+                        TryLaunchDebugger();
                 }
                 else if (arg.StartsWith("--trace:"))
                 {
                     traceLevel = (InternalTraceLevel)Enum.Parse(typeof(InternalTraceLevel), arg.Substring(8));
+
+                    // Initialize trace so we can see what's happening
+                    string logname = string.Format(LOG_FILE_FORMAT, pid);
+                    InternalTrace.Initialize(logname, traceLevel);
                 }
                 else if (arg.StartsWith("--pid="))
                 {
@@ -73,11 +80,6 @@ namespace NUnit.Agent
                     AgencyProcess = Process.GetProcessById(agencyProcessId);
                 }
             }
-
-            // Initialize trace so we can see what's happening
-            int pid = Process.GetCurrentProcess().Id;
-            string logname = string.Format(LOG_FILE_FORMAT, pid);
-            InternalTrace.Initialize(logname, traceLevel);
 
             log.Info("Agent process {0} starting", pid);
             log.Info("Running under version {0}, {1}",
@@ -107,7 +109,7 @@ namespace NUnit.Agent
 
             // TODO: We need to get this from somewhere. Argument?
             engine.InternalTraceLevel = InternalTraceLevel.Debug;
-            
+
             // Custom Service Initialization
             //log.Info("Adding Services");
             engine.Services.Add(new SettingsService(false));
@@ -129,16 +131,19 @@ namespace NUnit.Agent
                 if (Agent.Start())
                     WaitForStop();
                 else
+                {
                     log.Error("Failed to start RemoteTestAgent");
+                    return AgentExitCodes.FAILED_TO_START_AGENT;
+                }
             }
             catch (Exception ex)
             {
                 log.Error("Exception in RemoteTestAgent", ex);
+                return AgentExitCodes.UNEXPECTED_EXCEPTION;
             }
+            log.Info("Agent process {0} exiting cleanly", Process.GetCurrentProcess().Id);
 
-            log.Info("Agent process {0} exiting", Process.GetCurrentProcess().Id);
-
-            return 0;
+            return AgentExitCodes.OK;
         }
 
         private static void WaitForStop()
@@ -150,11 +155,35 @@ namespace NUnit.Agent
                 if (AgencyProcess.HasExited)
                 {
                     log.Error("Parent process has been terminated.");
-                    Environment.Exit(-1);
+                    Environment.Exit(AgentExitCodes.PARENT_PROCESS_TERMINATED);
                 }
             }
 
             log.Debug("Stop signal received");
+        }
+
+        private static void TryLaunchDebugger()
+        {
+            try
+            {
+                Debugger.Launch();
+            }
+            catch (SecurityException se)
+            {
+                if (InternalTrace.Initialized)
+                {
+                    log.Error($"System.Security.Permissions.UIPermission is not set to start the debugger. {se} {se.StackTrace}");
+                }
+                Environment.Exit(AgentExitCodes.NO_DEBUGGER_SECURITY);
+            }
+            catch (NotImplementedException nie) //Debugger is not implemented on mono
+            {
+                if (InternalTrace.Initialized)
+                {
+                    log.Error($"Debugger is not available on all platforms. {nie} {nie.StackTrace}");
+                }
+                Environment.Exit(AgentExitCodes.NO_DEBUGGER_NOT_IMPLEMENTED);
+            }
         }
     }
 }
--- a/src/NUnitEngine/nunit-agent/Program.cs
+++ b/src/NUnitEngine/nunit-agent/Program.cs
@@ -41,8 +41,6 @@ namespace NUnit.Agent
         static Process AgencyProcess;
         static RemoteTestAgent Agent;
 
-        private const string LOG_FILE_FORMAT = "nunit-agent_{0}.log";
-
         /// <summary>
         /// The main entry point for the application.
         /// </summary>
@@ -54,6 +52,7 @@ namespace NUnit.Agent
 
             InternalTraceLevel traceLevel = InternalTraceLevel.Off;
             int pid = Process.GetCurrentProcess().Id;
+            bool debugArgPassed = false;
 
             for (int i = 2; i < args.Length; i++)
             {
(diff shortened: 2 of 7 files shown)


src/NUnitEngine/nunit-agent/AgentExitCodes.cs
```diff
@@ -0,0 +1,36 @@
...
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+namespace NUnit.Common
+{
+    internal static class AgentExitCodes
+    {
+        public const int OK = 0;
+        public const int PARENT_PROCESS_TERMINATED = -1;
+        public const int UNEXPECTED_EXCEPTION = -2;
```
CharliePoole: It's a nit, but maybe this should be -100 as for the console runner. That would allow us to eventually consolidate the return codes.

src/NUnitEngine/nunit-agent/AgentExitCodes.cs
```diff
@@ -0,0 +1,36 @@
...
+        public const int FAILED_TO_START_AGENT = -3;
```
CharliePoole: FAILED_TO_START_REMOTE_AGENT ? 

That is, it's not the agent process that failed to start, but the RemoteAgent class.

src/NUnitEngine/nunit-agent/AgentExitCodes.cs
```diff
@@ -0,0 +1,36 @@
...
+        public const int NO_DEBUGGER_SECURITY = -4;
+        public const int NO_DEBUGGER_NOT_IMPLEMENTED = -5;
```
CharliePoole: I didn't know about these problems. Are there issues for them? Of course, it's unlikely users will ever see these codes, since `--debug-agent` is for our debugging, but it could save time if we get a sensible message.

How about DEBUGGER_SECURITY_VIOLATION and DEBUGGER_NOT_IMPLEMENTED.

Also, we should have a page in the wiki that lists these codes.

src/NUnitEngine/nunit-agent/Program.cs
```diff
@@ -61,11 +64,15 @@ public static int Main(string[] args)
                 if (arg == "--debug-agent")
                 {
                     if (!Debugger.IsAttached)
-                        Debugger.Launch();
+                        TryLaunchDebugger();
                 }
```
CharliePoole: Rather than depending on the order of args in the call in TestAgency, what if you merely set flags in this switch statement and pushed all the stuff that may generate an error down after the switch?

src/NUnitEngine/nunit-agent/Program.cs
```diff
@@ -129,16 +131,19 @@ public static int Main(string[] args)
...
                     log.Error("Failed to start RemoteTestAgent");
+                    return AgentExitCodes.FAILED_TO_START_AGENT;
+                }
             }
             catch (Exception ex)
             {
                 log.Error("Exception in RemoteTestAgent", ex);
+                return AgentExitCodes.UNEXPECTED_EXCEPTION;
             }
+            log.Info("Agent process {0} exiting cleanly", Process.GetCurrentProcess().Id);
```
CharliePoole: Not something you introduced, but we could use `pid` here.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -154,10 +155,10 @@ private Guid LaunchAgentProcess(TestPackage package)
             // Set options that need to be in effect before the package
             // is loaded by using the command line.
             string agentArgs = "--pid=" + Process.GetCurrentProcess().Id.ToString();
+            if (traceLevel != "Off")                    //Pass --trace as first non-optional arg as it intialises log
+                agentArgs += " --trace:" + traceLevel;
```
CharliePoole: This order-dependency seems fragile. See my comment in Program.cs.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -253,6 +256,42 @@ private static string GetTestAgentExePath(bool requires32Bit)
...
+                return;
+
+            string errorMsg;
+
+            switch (process.ExitCode)
+            {
+                case AgentExitCodes.OK:
+                    return;
+                case AgentExitCodes.PARENT_PROCESS_TERMINATED:
+                    errorMsg = "Remote test agent believes agency process has exited.";
```
CharliePoole: 😸 Obviously, we don't ever expect to see this message - belt and suspenders though.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -253,6 +256,42 @@ private static string GetTestAgentExePath(bool requires32Bit)
...
+                    break;
+                case AgentExitCodes.FAILED_TO_START_AGENT:
+                    errorMsg = "Failed to start remote test agent.";
+                    break;
+                case AgentExitCodes.NO_DEBUGGER_SECURITY:
+                    errorMsg = "Debugger could not be started on remote agent due to System.Security.Permissions.UIPermission not being set.";
+                    break;
+                case AgentExitCodes.NO_DEBUGGER_NOT_IMPLEMENTED:
+                    errorMsg = "Debugger could not be started on remote agent as not available on platform.";
+                    break;
```
CharliePoole: This brings to mind that the same errors can occur when the user passes the `--debug` option to the console runner. What happens in that case? What happens in general if there is an exception in the engine? We may not want to deal with it in the same PR, of course.

CharliePoole:
@ChrisMaddock The adapter makes no use of the agent in that it doesn't pass settings to the engine that would cause tests to run out of process. That's because VS handles all the process stuff.

There is a sort of a race condiditon. The `TestAgency` waits for the agent process to terminate and the agent checks that the agency process has not terminated. I've never been quite sure how to deal with that, since either one could terminate first.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
ChrisMaddock: Yeah - I protected against a crash by checking the log was initialised - but this was mainly me being lazy and making the minimal required changes. I can do this bit better.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
ChrisMaddock: `--debug` launches the debugger in the framework - I PR'd that side a few weeks back.

> What happens in general if there is an exception in the engine?

Not sure what you mean by this?

src/NUnitEngine/nunit-agent/AgentExitCodes.cs
ChrisMaddock: This one annoyed me - I wasn't quite sure what to call it. It's not really equivalent to the console's -100 'unexpected exception anywhere' - as this code only wraps a specific part of the agent (post-initialisation.). That was about as much info as was in the original log message though, and I wasn't sure what else to call it!

I don't mind changing it to -100 - would you envisage these ever being consolidated though? The list of reasons doesn't seem to have much in common.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
ChrisMaddock: I included all the existing exit points into this functionality. This one seemed odd, but I knew if I didn't write a custom message, then it would be the first to come up! 😆

src/NUnitEngine/nunit-agent/AgentExitCodes.cs
CharliePoole: Yes, the --debug option allows you to debug tests, even if you are running on a separate agent. The setting is passed in the package. The --debug-agent option is just for us to debug the agent.

ChrisMaddock:
Thanks @CharliePoole - I've added comments and a couple to discuss, I'll get back to the changes tomorrow.

ChrisMaddock:
@CharliePoole - I've made the requested changes. There was a couple of points we didn't finish discussing - will you take a second look and merge if you're happy? Thanks!

src/NUnitEngine/nunit-agent/AgentExitCodes.cs
CharliePoole: In the console UNEXPECTED_EXCEPTION generally means it's our error, even if it's due to a user error we neglected to handle. No user exceptions are "expected" to escape. I think it's the same here.

Yes, I think this needs a better name but I don't know what it might be.

ChrisMaddock:
Ok @CharliePoole, you've convinced me. :-p

Made the change, and will merge after CI.

ChrisMaddock:
Oh - this needs an 'approve review' before it can be merged. @CharliePoole, mind revisiting?

CharliePoole:
Tied up in a conference for the next few days. I'll take a look after if nobody beats me to it.

ChrisMaddock:
Thanks. 😄

CharliePoole:
Looks good to me too.

Issue #111, opened 2016-10-24
Provide better info when AppDomain won't unload


CharliePoole:
We fixed the problem of the agent failing to terminate when this happens but the user is not getting much information to allow debugging the reason the AppDomain failed to close. We should deal with this to the extent possible either in a console message or in the log.

octoberclub:
Not sure if there's any way of detecting or diagnosing why an Appdomain has failed to close - likely it's a native call originating from the test case that has locked some resource?  If anyone has any ideas, I'm happy to add something later.

In the meantime, which of these would be useful additions to the exception messages? 

- The full message from System.CannotUnloadAppDomainException (or other such exception)
- Information on the timeout that occured when attempting to unload. e.g. Unloading Appdomain xxx did not respond in xxx secs
- AppDomain friendly name 
- ApplicationBase
- A list of loaded assembly modules for the Appdomain (Using ReflectionOnlyGetAssemblies)

CharliePoole:
Sorry, I thought I had already added this to our Backlog. Hopefully somebody will pick it up now. Good suggestions @octoberclub - any chance you want to do a PR?

octoberclub:
Yes please. Will do.

ChrisMaddock:
@octoberclub - I've removed you assignment to open this issue up to others - if you're able to come back to it, the contribution would still be welcome! 😄

PR #320, opened 2017-12-02
Return error code -5 when AppDomain fails to unload

### LINKED ISSUE #256: Rewrite ConsoleRunnerTests.ThrowsNUnitEngineExceptionWhenTestResultsAreNotWriteable()
This test was previously dependent on engine services which were available to the engine, and shouldn't necessairily have been. It was ignored as part of a fix to the way engine services are initialised in https://github.com/nunit/nunit-console/pull/251.

At some point, the necessary fake services should be added to allow this test to be reimplemented.

ChrisMaddock:
This is the first of a few PRs to improve the experience when the appdomain fails to unload. I was hoping to get further that this - unfortunately mocking out the console took a few more hours than expected. 😬 

This implements part 4 of the solution proposed [here](https://github.com/nunit/nunit-console/pull/308#issuecomment-341934320) i.e. returning `-5` when the app domain fails to unload. Note this is a 'breaking change' - previously we issued a warning in the console, and then silently exited with code zero. I'm of the opinion that we shouldn't exit silently when these errors come up. With a specific error code, user's can chosoe to accept either 0 or -5 as a pass in their CI if they do wish to ignore the problem, but it will never silently slip by without the user taking action.

The reason there's been a few people cropping up over this issue is that, when using a `.nunit` project (i.e. everyone on TeamCity), behaviour seems to be different, and the process does crash mid-result if the appdomain fails to unload. I imagine that could be related to #116 - and will be looking at that as a separate issue.

Other than that, the remaining issue to do here is #111 - but I'll put these in as separate PRs, to be easier to review!

Also fixes #256

### DIFF
--- a/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
+++ b/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
@@ -47,6 +47,7 @@ namespace NUnit.ConsoleRunner
         public static readonly int INVALID_ASSEMBLY = -2;
         //public static readonly int FIXTURE_NOT_FOUND = -3;    //No longer in use
         public static readonly int INVALID_TEST_FIXTURE = -4;
+        public static readonly int UNLOAD_ERROR = -5;
         public static readonly int UNEXPECTED_ERROR = -100;
 
         #endregion
@@ -204,6 +205,7 @@ namespace NUnit.ConsoleRunner
                 : "ON";
 
             XmlNode result = null;
+            NUnitEngineUnloadException unloadException = null;
             NUnitEngineException engineException = null;
 
             try
@@ -217,6 +219,10 @@ namespace NUnit.ConsoleRunner
 
                     result = runner.Run(eventHandler, filter);
                 }
+            }
+            catch (NUnitEngineUnloadException ex)
+            {
+                unloadException = ex;
             }
             catch (NUnitEngineException ex)
             {
@@ -238,12 +244,20 @@ namespace NUnit.ConsoleRunner
                 {
                     var outputPath = Path.Combine(_workDirectory, spec.OutputPath);
                     GetResultWriter(spec).WriteResultFile(result, outputPath);
-                    _outWriter.WriteLine("Results ({0}) saved as {1}", spec.Format, spec.OutputPath);
+                    writer.WriteLine("Results ({0}) saved as {1}", spec.Format, spec.OutputPath);
                 }
 
-                // Since we got a result, we display any engine exception as a warning
                 if (engineException != null)
-                    writer.WriteLine(ColorStyle.Warning, Environment.NewLine + ExceptionHelper.BuildMessage(engineException));
+                {
+                    writer.WriteLine(ColorStyle.Error, Environment.NewLine + ExceptionHelper.BuildMessage(engineException));
+                    return ConsoleRunner.UNEXPECTED_ERROR;
+                }
+
+                if (unloadException != null)
+                {
+                    writer.WriteLine(ColorStyle.Warning, Environment.NewLine + ExceptionHelper.BuildMessage(unloadException));
+                    return ConsoleRunner.UNLOAD_ERROR;
+                }
 
                 if (reporter.Summary.UnexpectedError)
                     return ConsoleRunner.UNEXPECTED_ERROR;
--- a/src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
+++ b/src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
@@ -224,7 +224,7 @@ namespace NUnit.Engine.Runners
             {
                 _disposed = true;
 
-                string unloadError = null;
+                Exception unloadException = null;
 
                 try
                 {
@@ -233,8 +233,9 @@ namespace NUnit.Engine.Runners
                 catch(Exception ex)
                 {
                     // Save and log the unload error
-                    unloadError = ex.Message;
-                    log.Error(unloadError);
+                    unloadException = ex;
+                    log.Error(ExceptionHelper.BuildMessage(ex));
+                    log.Error(ExceptionHelper.BuildStackTrace(ex));
                 }
 
                 if (_agent != null && _agency.IsAgentRunning(_agent.Id))
@@ -252,16 +253,16 @@ namespace NUnit.Engine.Runners
                         _agent = null;
 
                         // Stop error with no unload error, just rethrow
-                        if (unloadError == null)
+                        if (unloadException == null)
                             throw;
 
                         // Both kinds of errors, throw exception with combined message
-                        throw new NUnitEngineException(unloadError + Environment.NewLine + stopError);
+                        throw new NUnitEngineUnloadException(unloadException.Message + Environment.NewLine + stopError);
                     }
                 }
 
-                if (unloadError != null) // Add message line indicating we managed to stop agent anyway
-                    throw new NUnitEngineException(unloadError + Environment.NewLine + "Agent Process was terminated successfully after error.");
+                if (unloadException != null) // Add message line indicating we managed to stop agent anyway
+                    throw new NUnitEngineUnloadException("Agent Process was terminated successfully after error.", unloadException);
             }
         }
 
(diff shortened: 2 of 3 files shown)


src/NUnitConsole/nunit3-console.tests/ConsoleRunnerTests.cs
```diff
@@ -1,32 +1,58 @@
...
 using NUnit.Engine.Extensibility;
 using NUnit.Engine.Services;
 using NUnit.Framework;
+using NUnit.Framework.Internal;
 
 namespace NUnit.ConsoleRunner.Tests
 {
     class ConsoleRunnerTests
     {
-        [Test, Ignore("Needs to be rewritten")]
```
ChrisMaddock: This is unrelated, but I wanted to add NSubstitute for my tests, so put this one back while I was here.

src/NUnitConsole/nunit3-console.tests/nunit3-console.tests.csproj
```diff
@@ -8,13 +8,14 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>NUnit.ConsoleRunner.Tests</RootNamespace>
     <AssemblyName>nunit3-console.tests</AssemblyName>
-    <TargetFrameworkVersion>v2.0</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v3.5</TargetFrameworkVersion>
```
ChrisMaddock: Required for NSubstitute. nunit.engine.tests is already the same way - a 3.5 test dll referencing a 2.0 product.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -227,11 +228,9 @@ private int RunTests(TestPackage package, TestFilter filter)
                 RestoreErrorOutput();
             }
 
-            var writer = new ColorConsoleWriter(!_options.NoColor);
```
ChrisMaddock: Does anyone know why we recreated the ConsoleWriter here? This meant my mock wasn't being used - and the two halves of console output were coming through two different writers.

src/NUnitEngine/nunit.engine/Services/ServiceManager.cs
```diff
@@ -126,6 +126,14 @@ public void StopServices()
...
         }
 
+        public void SubstituteService<T>(T newService) where T : IService
+        {
+            var currService = GetService(typeof(T));
+            _services.Remove(currService);
+            _serviceIndex.Remove(typeof(T));
+            AddService(newService);
+        }
+
```
ChrisMaddock: For test purposes. I'd have ideally made this internal and used InternalsVisibleTo - but as nunit.engine is signed, it wasn't possible without additionally signing nunit-console, and I didn't want to start that cycle off!

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
CharliePoole: The reason is that output from the test itself, which is handled by the first writer, may be redirected. This code goes back to when only a few people worked on NUnit but should have been explained in a comment anyway.

ChrisMaddock: Thanks. I wonder if it's possible to re-redirect the original ConsoleWriter instead? That would allow other writers to be passed in to the class. It feels a little odd that the class takes a Writer in the constructor, but then doesn't use that writer to write all the output. 

I'll have another look.

ChrisMaddock:
Hmm - I've just moved on to the next stage which is a problem with the AggregatingTestRunner

CharliePoole:
Returning an error code when one assembly encounters a problem is a little problematic. We moved away from this so as to give the error on the assembly result in many other cases, as when the assembly can't be loaded.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
CharliePoole: I can't see it all very well browsing on the phone but I do remember that adding color made it more complicated and o had to work to ensure that two writers still made one smooth output flow. It may have been changed as well since then.

There are two ways to redirect, --output parameter, which affects only test output and not the final report, and command-line she'll redirection, which affects anything that might go to the console. Taken together, that's four possibilities.

ChrisMaddock:
@CharliePoole - let's discuss this in more details on #321. 😄

ChrisMaddock:
This is ready for re-review.

The Console will now return -5 in cases of app domain unload failure, meaning unload failure will return a non-zero code, but is distingushable from other errors, so can be chosen to be ignored, if they user wishes.

I removed the test in the end, with the TextWriter being recreated and other remoting issues, it just seemed too awkward to test at this level. I guess this is another one for our integration test wishlist. 😞

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -238,12 +244,20 @@ private int RunTests(TestPackage package, TestFilter filter)
                 {
                     var outputPath = Path.Combine(_workDirectory, spec.OutputPath);
                     GetResultWriter(spec).WriteResultFile(result, outputPath);
-                    _outWriter.WriteLine("Results ({0}) saved as {1}", spec.Format, spec.OutputPath);
+                    writer.WriteLine("Results ({0}) saved as {1}", spec.Format, spec.OutputPath);
```
ChrisMaddock: This change is not related to the main PR, but believe this should use the 'new' writer instead. See discussion here: https://github.com/nunit/nunit-console/pull/320#discussion_r154501112 - came across this after my own confusion!

ChrisMaddock:
Sample output:

image

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -238,12 +244,20 @@ private int RunTests(TestPackage package, TestFilter filter)
...
                 }
 
-                // Since we got a result, we display any engine exception as a warning
                 if (engineException != null)
-                    writer.WriteLine(ColorStyle.Warning, Environment.NewLine + ExceptionHelper.BuildMessage(engineException));
+                {
+                    writer.WriteLine(ColorStyle.Error, Environment.NewLine + ExceptionHelper.BuildMessage(engineException));
+                    return ConsoleRunner.UNEXPECTED_ERROR;
```
CharliePoole: I set this up as a way for the engine to issue warnings and used it for a while. Since I stopped working on the engine I'm not sure if it is still used that way, but if it is, you would break the behavior here. It may be that you have replaced the last such usage and could then remove this code. In retrospect I should have used a different exception to make it clearer.

ChrisMaddock: My thought process was we now separate out the unload exceptions to a separate path. Therefore, any other crash at this point is 'unexpected'.

That said, this still returns the same -100 error code at this point - the only actual change is the colour of the text from Yellow -> Red - unless there's something else I'm missing here?

ChrisMaddock:
> One question... if running multiple assemblies and one of them fails to unload, do we get to see which one caused the problem?

Not as a 'direct' feature, but you will see the app domain which crashed, and I believe are app domains are named by the assembly they're running, right? 🤔

Left a reply above just to be sure we're thinking along the same lines before merging.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
CharliePoole: That makes sense.

CharliePoole:
The domain name will point to the proper assembly a lot of the time, since it uses the "package name". It will not directly indicate the assembly when running multiple assemblies in the same domain. In that case, you'll see either the name of the project file that contains the assemblies or the name of the first assembly specified on the command-line. My general sense is that this is not something people do a lot of the time.

I think this could be refined so that the unload exception always included the name of the assembly, but I'll leave it up to you whether that should be part of this PR or not.

ChrisMaddock:
I agree that would be nice - but I want to get this into the next release, and don’t really have the time to revisit this before Robs planned date next weekend. Can we call that a separate enhancement? 🙂

PR #321, opened 2017-12-02
AggregatingTestRunner to handle app domain unload failure in same fashion as single assembly runner

### LINKED ISSUE #191: Exception encountered unloading AppDomain

ChrisMaddock:
Fixes #191. This still needs tests, but sounds like it might be worth discussion before I go much further.

To clarify current behaviour:

1. When running a single assembly, if the AppDomain fails to unload as `ProcessRunner` is disposed, the exception is caught - and the console runner continues to write the already completed TestResults. A warning is then issued, and the console exits with 0. (To be changed in https://github.com/nunit/nunit-console/pull/320)

2. When running multiple assemblies via AggregatingTestRunner, if `EnginePackageSettings.DisposeRunners = false`, the same behaviour is maintained. 

3. If `EnginePackageSettings.DisposeRunners=true` however (default for the console), then the test runners are disposed at the end of each test run, and there is no functionality to handle CannotUnloadAppDomain exceptions, as there is at the other points where TestRunners are disposed. 

This is less of a problem when running via AggregatingTestRunner in **parallel**, as the CannotUnloadAppDomain exception is thrown on a separate thread, after the result has already been set. This means (I think, I haven't dug into this) that the exception just crashes the test running thread right at the end of the run - but the error is silently ignored, and the result can still be retrieved.

However, when AggregatingTestRunner is run sequentially (As the user may choose, or as #116 is currently causing all .nunit project users to see) then the exception is thrown on the main thread, with no handing, and the run is terminated immediately with no results (or only the results achieved so far.)

My opinion is that this situation should be handled the same way regardless of the user's choice of execution method. This PR fixes the early termination for running the AggregatingTestRunner sequentially, and catches the exceptions not being handled properly when running AggregatingTestRunner in parallel. This should standardise the handling of CannotUnloadAppDomain, regardless of the number of assemblies, or parallelisation choices.

WIP as I need to add tests, but it sounded like it would be sensible to put this up for feedback before putting much more work into it. 😄 

(Edit: looks like I broke netstandard. Will revisit when I come back to this!)


CharliePoole:
Based solely on the description, this makes sense to me. I'll be flying home tomorrow, so if the jet lag lets me I can look at code the next day.

ChrisMaddock:
Thanks Charlie. No rush, I won't be back to this for a few days now. 🙂

src/NUnitEngine/nunit.engine/Runners/AggregatingTestRunner.cs
```diff
@@ -210,16 +227,40 @@ protected override void Dispose(bool disposing)
...
+                    runner.Dispose();
+                }
+                catch (NUnitEngineException e)
+                {
+                    _thrownExceptions.Add(e);
+                }
+            }
 
             Runners.Clear();
+            throw new NUnitEngineException(_thrownExceptions);
```
ChrisMaddock: Just come across this - D'oh! Of course, that should only throw when there are exceptions to be thrown - will fix this before merge.

src/NUnitEngine/nunit.engine.api/NUnitEngineException.cs
```diff
@@ -38,23 +39,45 @@ namespace NUnit.Engine
...
+        }
+
+        /// <summary>
+        /// Construct with a message and collection of inner exceptions
+        /// </summary>
+        public NUnitEngineException(ICollection<Exception> aggregatedExceptions) 
+            : base(AggregatedExceptionsMsg, new NUnitEngineException(AggregatedExceptionsMsg))
+        {
+            AggregatedExceptions = aggregatedExceptions;
+        }
```
CharliePoole: Not clear what you are doing here. Seems like you end up with an `NUnitEngineException` with an `InnerException` that's also an `NUnitEngineException`, both having the same message. The outer also gets a collection of aggregated exceptions. What's the point of the `InnerException`?

src/NUnitEngine/nunit.engine.api/NUnitEngineException.cs
```diff
@@ -38,23 +39,45 @@ namespace NUnit.Engine
...
 
 #if !NETSTANDARD1_3
         /// <summary>
         /// Serialization constructor
         /// </summary>
         public NUnitEngineException(SerializationInfo info, StreamingContext context) : base(info, context) { }
 #endif
+
+        /// <summary>
+        /// Gets the Exception instance that caused the current exception.
```
CharliePoole: Comment is wrong.

src/NUnitEngine/nunit.engine/Runners/AggregatingTestRunner.cs
CharliePoole: So, throwing is the whole point of this change. I want to go back and question why we are  throwing in the case of failure to unload. I mentioned this elsewhere (on the issue I think) and I wonder why we can't just report this as an error (or warning) associated with the individual assembly as we do in other cases. The difference here, of course, is that only the engine can detect the error as opposed to the framework. Worth discussing?

src/NUnitEngine/nunit.engine.api/NUnitEngineException.cs
ChrisMaddock: I think I was maybe thinking of `NUnitException` in the framework, where I thought we'd always expect an inner exception? I'll get rid of it.

src/NUnitEngine/nunit.engine/Runners/AggregatingTestRunner.cs
ChrisMaddock: For the avoidance of doubt, to be clear of the things I'm aiming to do here:

1. Bring the AggregatingTestRunner inline with the other test runners.
2. Have the engine throw an exception on domain unload failure, caught by the runner.

The current issue is that the AggregatingTestRunner is throwing the exception mid-run, where as the other TestRunners would all throw the exception at the end of the test run, as the runner is disposed, and once the full results have already been returned to the runner.

I think that design currently works. A few reasons:
1. I don't think we should silently swallow the exception, this exception indicates an unclean shut down in the user's code. By returning a distinct unload-failure-error-code such as in #320 - we still give the user the _choice_ to ignore this if they wish to.
2. Re: associating with xml - by the time the domain is unloaded and the engine cleaned up, the results have already been returned to the runner. Would we want to alter this behaviour in all situations - e.g. with a GUI, unload wouldn't happen after each run, correct? I don't see the advantage of holding on to the xml, incase we wish to add an error that's not strictly related to the test results.
3. I feel like app domain unload belongs semantically as an engine exception, rather than a 'test result' in the xml. i.e. I could chose to run in the main domain, in which case I would get no exception.

I _don't_ think that this exception should escape the runner - the current behaviour of handling this within the runner seems correct to me.

ChrisMaddock: (The term 'runner' is overloaded in this conversation! I've edited to be clear when I mean ITestEngineRunner - and when I mean runner as in 'console runner'/GUI etc.)

CharliePoole: I see your point about the separation of running from unloading. In that case, I suppose it's up to the runner to decide what to do with the exceptions - error, warn, etc.  That would mean that unload exceptions need to be distinguishable from others of course.

ChrisMaddock: How about an NUnitUnloadException which inherits from NUnitEngineExceptions, to not break any existing runners by throwning a new Exception type? 🙂

CharliePoole:
Still "do not merge"?

ChrisMaddock:
Was just waiting for CI to run last night, and didn't get a chance to come back to it. 😄 Ready for a second review.

ChrisMaddock:
Thanks Charlie. 😄

PR #322, opened 2017-12-03
Use Environment Exit in Agent to ensure all threads are terminated

### LINKED ISSUE #259: NUnit3 agent hangs after encountering an "CannotUnloadAppDomainException"
I do believe that the nunit-agent process hangs after it run into a problem with unloading app domains after all tests have run from a certain test assembly. Unfortunately we suffer from a small number of "shaky" tests which can lead to the problem in unloading app domains (thread which cannot be aborted or something like this). Of course we are trying to fix these problems in our tests, but nevertheless I think that the test runner never must break because of errors in the tests assemblies.

We're using Nunit3Console version 3.6.1

### Stack trace of the monitored app domain issue taken from TeamCity (version 2017.1) :
_Error : C:\TcTemp\buildTmp\EUSULw5pVgYsE72Ff9xROS7C6Rb9b4jB.nunit
Exception encountered unloading AppDomain

Server stack trace: 
bei NUnit.Engine.Services.DomainManager.DomainUnloader.Unload()
bei NUnit.Engine.Runners.TestDomainRunner.UnloadPackage()
bei NUnit.Engine.Runners.AbstractTestRunner.Dispose(Boolean disposing)
bei NUnit.Engine.Runners.AbstractTestRunner.Dispose()
bei NUnit.Engine.Runners.AggregatingTestRunner.RunTests(ITestEventListener listener, TestFilter filter)
bei System.Runtime.Remoting.Messaging.StackBuilderSink._PrivateProcessMessage(IntPtr md, Object[] args, Object server, Object[]& outArgs)
bei System.Runtime.Remoting.Messaging.StackBuilderSink.SyncProcessMessage(IMessage msg)

After this issue has been encountered, the nunit-agent.exe process keeps the test assembly locked. And with it all system resources allocated by the last executed test such as tcp ports. With the effect, that future tests cannot be run successful on the same TeamCity agent again.

STEPS NEEDED

1. Write a test which launches a thread which refuses to stop and abort (Trick here: catch tread abort exception and keep thread busy in a finally block)
2. Run test with TeamCity, NUnit3 with nunit3.console runner 3.6 (Probably TeamCity is not really required, but I didn't check this out so far)

STATUS
The test will fail - as expected. But you will find that the nunit-agent.exe process is still running (check with windows task manager), locking the test assembly. You won't be able to delete it.

ChrisMaddock:
Fixes #259 (In combination with #320 and #321 - although this PR can be merged independently.)

At the point where the agent exits the main method, we've tried everything possible to shut down nicely. If the AppDomain does fail to unload, after the specified timeout, we may reach this point in the code with some threads still running. Using `Environment.Exit()` as opposed to `return` ensures these will be cleaned up, and the process will definitely terminate.

The additional noise is just because the logging in Program.cs wasn't working - the log was declared in a static variable, and then a new log re-initialised later.

### DIFF
--- a/src/NUnitEngine/nunit-agent/Program.cs
+++ b/src/NUnitEngine/nunit-agent/Program.cs
@@ -34,18 +34,17 @@ namespace NUnit.Agent
 {
     public class NUnitTestAgent
     {
-        static Logger log = InternalTrace.GetLogger(typeof(NUnitTestAgent));
-
         static Guid AgentId;
         static string AgencyUrl;
         static Process AgencyProcess;
         static RemoteTestAgent Agent;
+        private static Logger log;
 
         /// <summary>
         /// The main entry point for the application.
         /// </summary>
         [STAThread]
-        public static int Main(string[] args)
+        public static void Main(string[] args)
         {
             AgentId = new Guid(args[0]);
             AgencyUrl = args[1];
@@ -76,6 +75,7 @@ namespace NUnit.Agent
             }
 
             InternalTrace.Initialize($"nunit-agent_{pid}.log", traceLevel);
+            log = InternalTrace.GetLogger(typeof(NUnitTestAgent));
 
             if (debugArgPassed)
                 TryLaunchDebugger();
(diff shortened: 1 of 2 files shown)


mikkelbu:
LGTM, but I did not get the part about `log`. Where was it  re-initialised later before this change?

Ps. The build on travis failed with (I've restarted the job)
```
1) Failed : NUnit.Engine.Services.Tests.AgentDataBaseTests.AddData_Parallel(3)
  Expected and actual are both <System.Collections.Generic.List`1[System.Guid]> with 3 elements
  Values differ at index [1]
  Expected: c16ce2d9-f989-4f57-b5d5-29354389c25a
  But was:  90194d77-629a-4fd7-9542-d3fe7d2b9b15
at NUnit.Engine.Services.Tests.AgentDataBaseTests.AddData_Parallel (System.Int32 count) [0x00000] in <33c50ed7acbf44b985fba7ae2ae9d2fa>:0
```

src/NUnitEngine/nunit-agent/Program.cs
```diff
@@ -76,6 +75,7 @@ public static int Main(string[] args)
             }
 
             InternalTrace.Initialize($"nunit-agent_{pid}.log", traceLevel);
+            log = InternalTrace.GetLogger(typeof(NUnitTestAgent));
```
ChrisMaddock: @mikkelbu - These are the relvant two lines. Previously, line 78 was a static class var - so ran as soon as the class was created. When the log was initialised on line 77, I think the state changed, meaning all later messages using this log variable were lost. Running this after initialisation instead of before means we don't loose the log messages. 😄

ChrisMaddock:
Thanks Mikkel. 😄 Have also fixed the CI failure in #323.

PR #326, opened 2017-12-09
Display warnings even when there are only warnings

### LINKED ISSUE #289: Warnings not displayed
The console displays a report of failures, errors and warnings only if the overall result is failure. If there are only warnings it is not displayed, but it should be.

mikkelbu:
The "Errors, Failures and Warnings" section is now also written,
if there are tests with warnings, but no tests failed.

Note that we now only create ConsoleTestResult if the site is
different from `null` as warnings do not set the site.
Otherwise we will report warnings "One or more child tests had warnings"
for "test-suite"-elements above tests with warnings.

Fixes #289

### DIFF
--- a/src/NUnitConsole/nunit3-console.tests/ResultReporterTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/ResultReporterTests.cs
@@ -127,7 +127,7 @@ namespace NUnit.ConsoleRunner.Tests
                 "Warning Message",
                 "6) Invalid : NUnit.Tests.BadFixture" + nl +
                 "No suitable constructor was found"
-        };
+            };
             
             var actualErrorFailuresReport = GetReport(_reporter.WriteErrorsFailuresAndWarningsReport);
 
--- a/src/NUnitConsole/nunit3-console/ResultReporter.cs
+++ b/src/NUnitConsole/nunit3-console/ResultReporter.cs
@@ -191,7 +191,7 @@ namespace NUnit.ConsoleRunner
                         else
                         {
                             var site = resultNode.GetAttribute("site");
-                            if (site != null && site != "Parent" && site != "Child")
+                            if (site == "SetUp" || site == "TearDown")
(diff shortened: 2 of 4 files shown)


CharliePoole:
Confused about the null business. It sounds as if we are not setting site correctly and should fix that rather than compensating for it. Am I misunderstanding?

mikkelbu:
As far as I can tell from the code the "site" is only used for failures (the enum is also called `FailureSite` which again indicates that it is only for failures). In the NUnitLite runner we "handle" that site can be `null` by inverting the check (we know for the context that site cannot be `FailureSite.Test`, hence it must be either be `null` or one of the other 4 possibilities: `SetUp`, `TearDown`, `Parent`, or `Child`. 

https://github.com/nunit/nunit/blob/3c028034e717c79bf00661c5852b358f9097d2e1/src/NUnitFramework/nunitlite/TextUI.cs#L467
```csharp
var suite = result.Test as TestSuite;
var site = result.ResultState.Site;
if (suite.TestType == "Theory" || site == FailureSite.SetUp || site == FailureSite.TearDown)
    DisplayTestResult(result);
```
Ignoring `suite.TestType == "Theory"`, we only write the result if the site is either `FailureSite.SetUp` or `FailureSite.TearDown`. 
So I can also change the check from `site != null && site != "Parent" && site != "Child"` to `site == "SetUp" || site == "TearDown"`.

CharliePoole:
@mikkelbu Part of the problem, of course, is that we are working with the XML, which exposes the internals of `ResultState` as three fields. As you point out, not having a site attribute means that the `Site` was `FailureSite.Test`. Maybe as a longer-term solution, we should create a client-side model that puts everything back together as a proper `ResultState`.

If memory serves, we also use Site it for Ignored and Inconclusive tests, if the result came from an Assert in the setup.

I like the positive tests better.

CharliePoole:
Looks good to me

PR #328, opened 2017-12-11
Provide additional debug information when AppDomain fails to unload

### LINKED ISSUE #111: Provide better info when AppDomain won't unload

ChrisMaddock:
Fixes #111. Debug information is displayed on the console, and intended for the user to better be able to debug the issue in their code.

It's fair to say that what's available at this point is pretty limited - open to any more useful ideas, if anyone has any.

### DIFF
--- a/src/NUnitEngine/nunit.engine/Internal/DomainDetailsBuilder.cs
+++ b/src/NUnitEngine/nunit.engine/Internal/DomainDetailsBuilder.cs
@@ -1,0 +1,81 @@
+﻿// ***********************************************************************
+// Copyright (c) 2017 Charlie Poole, Rob Prouse
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Text;
+
+namespace NUnit.Engine.Internal
+{
+    /// <summary>
+    /// DomainDetailsBuilder provides human readable information on 
+    /// an AppDomain, to assist with debugging.
+    /// </summary>
+    internal static class DomainDetailsBuilder
+    {
+        /// <summary>
+        /// Get human readable string containing details of AppDomain.
+        /// </summary>
+        /// <param name="domain">AppDomain to get details on.</param>
+        /// <param name="errMsg">An optional overall error message e.g. The messaged from an AppDomainUnload exception.</param>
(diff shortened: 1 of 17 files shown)


src/NUnitEngine/nunit.engine/Internal/DomainDetailsBuilder.cs
```diff
@@ -0,0 +1,81 @@
...
+    /// DomainDetailsBuilder provides human readable information on 
+    /// an AppDomain, to assist with debugging.
+    /// </summary>
+    internal static class DomainDetailsBuilder
+    {
+        /// <summary>
+        /// Get human readable string containing details of AppDomain.
+        /// </summary>
+        /// <param name="domain">AppDomain to get details on.</param>
+        /// <param name="errMsg">An optional overall error message e.g. The messaged from an AppDomainUnload exception.</param>
```
mikkelbu: "The" => "the" and "messaged" => "message".

src/NUnitEngine/nunit.engine/Internal/DomainDetailsBuilder.cs
```diff
@@ -0,0 +1,81 @@
...
+                if (reflectionLoadedAssemblies.Count != 0)
+                {
+                    sb.AppendLine("--- Assemblies loaded in current AppDomain via reflection ---");
+                    foreach (var assembly in reflectionLoadedAssemblies)
+                        WriteAssemblyInformation(sb, assembly);
+                }
+            }
+            catch (AppDomainUnloadedException)
+            {
+                sb.AppendLine("App Domain was unloaded before all details could be read.");
```
mikkelbu: In all other places we use the term "AppDomain" (i.e. in one word)

src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
```diff
@@ -261,7 +261,7 @@ protected override void Dispose(bool disposing)
                 }
 
                 if (unloadError != null) // Add message line indicating we managed to stop agent anyway
-                    throw (new NUnitEngineException(unloadError + "\nAgent Process was terminated successfully after error."));
+                    throw (new NUnitEngineException(unloadError + Environment.NewLine + "Agent Process was terminated successfully after error."));
```
mikkelbu: Perhaps we should also remove the parentheses surrounding `new NUnitEngineException(....)`

src/NUnitEngine/nunit.engine/Services/DomainManager.cs
```diff
@@ -194,12 +192,13 @@ private void UnloadOnThread()
...
                 {
-                    _unloadException = ex;
-
                     // We assume that the tests did something bad and just leave
-                    // the orphaned AppDomain "out there". 
-                    // TODO: Something useful.
-                    log.Error("Unable to unload AppDomain " + domainName, ex);
+                    // the orphaned AppDomain "out there".
+                    var msg = DomainDetailsBuilder.DetailsFor(_domain,
+                    $"Exception encountered unloading AppDomain: {ex.Message}");
```
mikkelbu: Either indent this line or extract `$"Exception encountered unloading AppDomain: {ex.Message}"` to a variable and keep `var msg = DomainDetailsBuilder.DetailsFor(_domain, errMsg);` on one line

src/NUnitEngine/nunit.engine/Internal/DomainDetailsBuilder.cs
ChrisMaddock: Thanks for this comment. I decided to refer to what MS do in their docs - they write out 'application domain' in full, when referring to the concept. Based on that, I went through and changed all the public string to that convention instead. 🙂

ChrisMaddock:
@rprouse - I want to merge this after making trivial changes, but I can neither merge without an approved review, nor review my own PR. Is there a setting to change one of these things, to save someone else needing to come back to it?

@ anyone-else - any chance of a review. 😄 Mikkel has reviewed the content - the final commit is just string changes.

src/NUnitEngine/nunit.engine/Internal/DomainDetailsBuilder.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            var sb = new StringBuilder();
+            if (errMsg != null) sb.AppendLine(errMsg);
+
+            try
+            {
+                sb.AppendLine($"Application domain name: {domain.FriendlyName}");
+                sb.AppendLine($"Application domain BaseDirectory: {domain.BaseDirectory}");
+
+                var reflectionLoadedAssemblies = new List<Assembly>(domain.ReflectionOnlyGetAssemblies());
```
CharliePoole: Why only ReflectionOnly assemblies?

ChrisMaddock: Because `domain.GetAssemblies()` threw at this point. Perhaps because appdomain unload has already been attempted?

PR #329, opened 2017-12-13
Bug fixes for #309

### LINKED ISSUE #309: No driver found if framework assembly reference has uppercase characters
.NET assembly references are not case sensitive, but the engine fails to find a driver unless the assembly reference is precisely the lowercase string `"nunit.framework"`.

https://github.com/nunit/nunit-console/blob/416b54d2a215466a861b3fbea82adeb3c820ec62/src/NUnitEngine/nunit.engine/Drivers/NUnit3DriverFactory.cs#L32-L42
We should be doing something similar to:

```cs
return NUNIT_FRAMEWORK.Equals(reference.Name, StringComparison.OrdinalIgnoreCase) && reference.Version.Major == 3;
```

Besides the correctness issue, this prevents us from ever deciding to ship an `NUnit.Framework.dll` (using the conventional namespace styling) in any packaging form.

lasthero:
Fixes #309

No driver found if framework assembly reference has uppercase characters

### DIFF
--- a/src/NUnitEngine/nunit.engine.netstandard/Drivers/NUnitNetStandardDriver.cs
+++ b/src/NUnitEngine/nunit.engine.netstandard/Drivers/NUnitNetStandardDriver.cs
@@ -79,7 +79,7 @@ namespace NUnit.Engine.Drivers
             if(_testAssembly == null)
                 throw new NUnitEngineException(string.Format(FAILED_TO_LOAD_TEST_ASSEMBLY, assemblyRef.FullName));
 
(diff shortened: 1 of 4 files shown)


CharliePoole:
Looks good to me.

jnm2:
Thanks for taking the time to do this! I added "Fixes #309" to your text so that GitHub links and autocloses the issue when the PR is merged. (https://help.github.com/articles/closing-issues-using-keywords/).

lasthero:
👍  my pleasure. @jnm2 - thanks for the tips.
On to the next one...

PR #330, opened 2017-12-14
Updating the TeamCity extension to 1.0.3

### LINKED ISSUE #318: Update NUnit.Extension.TeamCityEventListener to 1.0.3
The extension has been updated, we should update the NuGet dependency.

phillipsj:
This updates the extension.  Fixes #318

### DIFF
--- a/choco/nunit-console-with-extensions.nuspec
+++ b/choco/nunit-console-with-extensions.nuspec
@@ -40,7 +40,7 @@
           <dependency id="nunit-extension-vs-project-loader" version="3.7.0" />
           <dependency id="nunit-extension-nunit-v2-result-writer" version="3.5.0" />
           <dependency id="nunit-extension-nunit-v2-driver" version="3.6.0" />
-          <dependency id="nunit-extension-teamcity-event-listener" version="1.0.2" />
+          <dependency id="nunit-extension-teamcity-event-listener" version="1.0.3" />
       </group>
     </dependencies>
   </metadata>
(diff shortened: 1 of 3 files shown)


CharliePoole:
@rprouse Are we still releasing NUnit.Runners?

ChrisMaddock:
I'll merge this with a single review as it's a minor change. Thanks for making it @phillipsj!

Charlie - I think we should make any required changes to NUnit.Runners in a separate PR, but I think we should keep releasing it for now. It still has a significant number of downloads. 🙂

CharliePoole:
@ChrisMaddock I'm wondering if we should deprecate it for real.

ChrisMaddock:
Is that a feature NuGet has yet? My memory was that there wasn't much option previously other than just to label it up, like we have done, and stop releasing it - which I'm not keen on!

CharliePoole:
There's no feature. I figured that what we have already done was just the warning we were about to obsolete the package and that we could follow use one of the fairly well established conventional ways of doing it eventually...

1. Change the display name from NUnit.Runners to Obsolete Package or something similar
2. Remove all references to dependent projects OR
3. Change the reference to NUnit.Console, the replacement project

If we do #2, folks will be forced to change their references. If we do 3 it continues to be updated but without any need to modify it on our part.

Or, we can keep it in the state it is indefinitely but that's just one more little thing to remember to update.

ChrisMaddock:
1 and 3 sound good to me! I don't think we should do anything which would prevent users getting automatic updates.

rprouse:
I just changed the title of the NuGet package to just `Deprecated: Use NUnit.Console`, but the current version still has 100K downloads, so I hate to drop it outright.

CharliePoole:
That should help limit any new downloads. Problem is folks can find many writeups online that say to use this package. Next time we have a chance, I suggest we change it to depend on nunit.console so we will then only have two places to update our list of dependencies.

PR #346, opened 2018-01-11
Ensure we don't try to unload agent if the process already terminated

### LINKED ISSUE #343: Superfluous unload error shown in console
We are repeatedly seeing stack traces that show two things:

1. The agent crashes in process runner.
2. We subsequently get an error when trying to unload the crashed runner.

This is confusing to users. We should not be trying to unload a runner that we already know has crashed. Fixing this to eliminate the second error will help people to focus on finding the real problem and will also allow us to isolate __real__ problems with unload when they do occur.

CharliePoole:
Fixes #343 

My original thought had been that the extra unload error didn't cause any problem but it turns out to be confusing to users who report that as the error rather than the original error. 

In order to suppress the attempt to unload the agent and runner when the agent process has terminated, the `ProcessRunner` needed to ask about the status of the process. In order to look that up, it had to use the agent id.

Since `RemoteTestAgent` is... remote 😄 ... accessing the Id property was causing an error in itself! Consequently, I created a `RemoteTestAgentProxy` that wraps the `RemoteTestAgent` and remembers its id. It doesn't do anything else, forwarding all calls to the remote agent, but it could be extended in the future.

I mention this, because I suspect that the access to the agent id may have caused some problems in the past and may still cause it if done by `TestAgency` directly on the `RemoteTestRunner`.

I tested this by forcing the agent to terminate and observing that only the initial termination message appeared in the console.

### DIFF
--- a/src/NUnitEngine/nunit.engine/Services/TestAgency.cs
+++ b/src/NUnitEngine/nunit.engine/Services/TestAgency.cs
@@ -27,6 +27,7 @@ using System.Threading;
 using System.Diagnostics;
 using System.Collections.Generic;
 using NUnit.Common;
+using NUnit.Engine.Agents;
 using NUnit.Engine.Internal;
 
 namespace NUnit.Engine.Services
@@ -40,7 +41,8 @@ namespace NUnit.Engine.Services
         Starting,
         Ready,
         Busy,
-        Stopping
+        Stopping,
+        Terminated
     }
 
     /// <summary>
@@ -106,7 +108,7 @@ namespace NUnit.Engine.Services
 
         #endregion
 
-        #region Public Methods - Called by Clients
+        #region Method Called by Clients
 
         public ITestAgent GetAgent(TestPackage package, int waitTime)
         {
@@ -124,6 +126,12 @@ namespace NUnit.Engine.Services
                 r.Status = AgentStatus.Ready;
                 log.Debug("Releasing agent " + agent.Id.ToString());
             }
+        }
+
+        internal bool IsAgentRunning(Guid id)
+        {
+            var agentRecord = _agentData[id];
+            return agentRecord != null && agentRecord.Status != AgentStatus.Terminated;
         }
 
         #endregion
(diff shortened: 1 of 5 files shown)


src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -256,15 +265,18 @@ private static string GetTestAgentExePath(bool requires32Bit)
                 ? "nunit-agent-x86.exe"
                 : "nunit-agent.exe";
 
-            return Path.Combine(engineDir, agentName);
+            return Path.Combine(engineDir, agentName);//
```
mikkelbu: The `//` can be removed

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -291,6 +303,8 @@ private static void OnAgentExit(object sender, EventArgs e)
                     errorMsg = $"Remote test agent exited with non-zero exit code {process.ExitCode}";
                     break;
             }
+
+            //_agentData[agentId]
```
mikkelbu: Please remove this

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -236,11 +244,12 @@ private ITestAgent CreateRemoteAgent(TestPackage package, int waitTime)
             {
                 Thread.Sleep( pollTime );
                 if ( !infinite ) waitTime -= pollTime;
-                ITestAgent agent = _agentData[agentId].Agent;
-                if ( agent != null )
+                var agentRecord = _agentData[agentId];
+                
+                if ( agentRecord.Agent != null )
```
mikkelbu: The spacing around `(` and `)` is wrong, but then again we use the same kind of spacing in the rest of the method.

CharliePoole: ...which is why I continued to use it. 😄 

Not relevant here, but I really don't see why we should have a spacing standard. I don't think the presence or absence of a space makes the code more difficult to understand.

mikkelbu: But you forgot to do it 3 lines below 😄 . Personally I don't feel too strongly about a space or not, but I prefer that the code is consistent.

CharliePoole:
We've been waiting for the apple builds to start for about an hour. I'm going to go ahead and merge since everything else is passing.

PR #348, opened 2018-01-12
Add command-line option to specify configuration file

### LINKED ISSUE #246: No way to specify app.config with console runner
@cyanite commented on [Tue Jun 27 2017](https://github.com/nunit/nunit/issues/2284)

Using nunit-console runner on individual dll's, there is no way to specify the app.config file to use. To do that you have to create an .nunit file and use configfile="app.config". There should be a command line option to do this.

We have a huge number of dll's which require assembly redirects and we don't have config files written for each of them, but only for our executables.

CharliePoole:
Fixes #246 

Adds the option to the console runner. The underlying capability has been there for some time and is already used by the NUnit 3 VS adapter.

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/Services/DomainManagerStaticTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Services/DomainManagerStaticTests.cs
@@ -35,6 +35,9 @@ namespace NUnit.Engine.Services.Tests
         static string path2 = TestPath("/test/bin/debug/test2.dll");
         static string path3 = TestPath("/test/utils/test3.dll");
 
+        const string STANDARD_CONFIG_FILE = "nunit.engine.tests.dll.config";
+        const string ALTERNATE_CONFIG_FILE = "alt.config";
+
         [Test]
         public static void GetPrivateBinPath()
         {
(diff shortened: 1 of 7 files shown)


src/NUnitEngine/nunit.engine.tests/Services/DomainManagerStaticTests.cs
```diff
@@ -88,15 +91,18 @@ public static void GetCommonAppBase_ThreeElements_DifferentDirectories()
         [Test]
         public static void ProperConfigFileIsUsed()
         {
-            var configFileName = "nunit.engine.tests.dll.config";
-            var expectedPath = Path.Combine(TestContext.CurrentContext.TestDirectory, configFileName);
-            Assert.That(AppDomain.CurrentDomain.SetupInformation.ConfigurationFile, Is.SamePath(expectedPath));
+            var expectedPath = Path.Combine(TestContext.CurrentContext.TestDirectory, STANDARD_CONFIG_FILE);
+            var alternatePath = Path.Combine(TestContext.CurrentContext.TestDirectory, ALTERNATE_CONFIG_FILE);
+            Assert.That(AppDomain.CurrentDomain.SetupInformation.ConfigurationFile, Is.SamePath(expectedPath).Or.SamePath(alternatePath));
```
ChrisMaddock: Would you mind explaining the changes in this file to me? I'm a bit confused where alt.config has come from?

CharliePoole: Good catch... I should add `alt.config` in GitHub rather than directly in my bin directory where it is, of course, being ignored.

Basically, `alt.config` is a standard way of manually testing that the command-line option works. By running the tests with `--configfile=alt.config` I am able to verify the complete chain of events that cause the correct file to be read. The changes here are so that existing tests do not get broken when I do that.

This is really not absolutely necessary because we have unit tests that verify each individual link in the chain but it made me more confident to be able to add it.

I think it will be clearer once I actually add the file.

ChrisMaddock: Thanks - that made sense. 😄 Could I suggest a comment to cover that - that the alternatePath part of these two tests is only in the case where a --configFile option has been used manually?

mikkelbu: I completely overlooked the tests when I reviewed (so much for multitasking between tasks).

CharliePoole: @ChrisMaddock I'll do that. Things like this are always "obvious" to the guy who writes them!

CharliePoole:
Still having problem with Apple agents not starting. Overriding to merge.

Issue #347, opened 2018-01-12
TestEngine: 34 failed tests in NUnitAdapter - none via build


alski:
![image](https://user-images.githubusercontent.com/991624/34889122-cecc765c-f7c4-11e7-8a15-13cdc47def5d.png)

When I run the tests via NUnitAdapter (v3) I get 34 failures, but when running via build all pass. It appears to be due to tests being unable to find the mock assemblies that the tests expect to run against, possibly due to the incorrect working directory. Still checking, just wanted to raise this for now.

`<test-suite type="Assembly" id="0-1" name="mock-assembly.dll" fullname="C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\mock-assembly.dll" testcasecount="0" runstate="NotRunnable"><properties><property name="_SKIPREASON" value="File not found: C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\mock-assembly.dll" /></properties></test-suite>`

alski:
Ok, looks like this has been remedied in #if NETCOREAPP1_1 so just needs to cover these cases too. Pull Request coming shortly.

alski:
Hmmm.. Some of the failures are because MockAssembly.dll is being loaded in and run.
I know I'm due an opticians appointment, but I couldn't see a way to exclude via .RunSettings

https://github.com/nunit/nunit3-vs-adapter/blob/master/src/NUnitTestAdapter/NUnitTestAdapter.cs#L147

ChrisMaddock:
My understanding is that we've never formally supported running the NUnit tests themselves in the adapter. My opinion would be that if there's no good reason for that, then a PR to improve it would be welcome - although other team members may disagree!

--------

> Hmmm.. Some of the failures are because MockAssembly.dll is being loaded in and run.
I know I'm due an opticians appointment, but I couldn't see a way to exclude via .RunSettings

You're correct - there currently is no method to exclude the assembly. Take a look at the tail end of this PR, from when @jnm2 proposed the same idea for the framework tests. https://github.com/nunit/nunit/pull/2297#issuecomment-334939925 My personal preference is allowing the test adapter to provide the `--skipnontestassemblies` option (or more, have it on by default) - but that hasn't yet been discussed or implemented.

CharliePoole:
[...]

PR #349, opened 2018-01-12
Get all TestEngine tests running under NUnitAdapter apart from those …

alski:
…in the MockAssembly which will still need excluding.

The dotnetCore build seems to work via the build script but someone with more knowledge should probably check the conditional being correct.


src/NUnitConsole/nunit3-console.tests/CommandLineTests.cs
```diff
@@ -513,7 +520,13 @@ public void ResultOptionWithoutFileNameIsInvalid()
         [Test]
         public void ResultOptionMayBeRepeated()
         {
-            const string transformFile = "TextSummary.xslt";
+#if !NETCOREAPP1_1
+            var dir = Path.GetDirectoryName(typeof(CommandLineTests).Assembly.Location);
+#else
+            var dir = Path.GetDirectoryName(typeof(CommandLineTests).GetTypeInfo().Assembly.Location);
+#endif
```
CharliePoole: This and the other conditional places should use `TestContext.TestDirectory`, which I believe already has the necessary conditional code. If it doesn't it should be added there rather than in multiple places.

ChrisMaddock:
I'd be glad to accept these changes, but instead using `TestContext.TestDirectory` as @CharliePoole has said. In my opinion it makes the tests more robust and runner agnostic, which is always an advantage.

In fact, I'm suspicious the adapter is a red herring in this situation - what's actually causing the issue is the work**ing** directory the test process is launched from. For example, I just saw the same 27 failures @alski reported in the nunit console, using the following command line:

```
nunit3-console nunit-console/bin/Debug/nunit.engine.tests.dll
```

ChrisMaddock:
@alski - regardless of what conclusion we come to - thank you for taking the time to dig in to this. 🙂

alski:
No problem Chris, I simply need a stable base to try out a spike.

ChrisMaddock:
@alski - We have a console release coming up towards the end of this week - and I think it would be great if we can get this in. Would you have time to revist this during this week, and make the change @CharliePoole suggested to use `TestContext.TestDirectory` instead of locating the directory manually? If not - perhaps one of the team could get that change in.

Charlie - based on my comment above, are you happy this is a worthwhile change to merge in?

alski:
Chris,
No I don't mind and can put in a bit of time. 

However, I've also noticed that `TestContext` is a fairly low level `nUnit.Framework` object that is not usually instantiated at the time we are looking at creating a test for nunit-console.

I also noticed that there exists

https://github.com/nunit/nunit-console/blob/master/src/NUnitEngine/mock-assembly/MockAssembly.cs#L105
`public class MockAssembly`
`{`
` #if !NETSTANDARD1_6`
`            public static readonly string AssemblyPath = AssemblyHelper.GetAssemblyPath(typeof(MockAssembly).Assembly);`
 ` #endif `

and this could be used to cover at least two of the cases ( https://github.com/nunit/nunit-console/pull/349/commits/e779307ae7254a27b2364ee845822b2e09c65a83#diff-c6456741755862ce820b5efbe37974d0L75 and https://github.com/nunit/nunit-console/pull/349/commits/e779307ae7254a27b2364ee845822b2e09c65a83#diff-650c001d1dda89ada4fb5df0c43fd7b3L47).

This makes me wonder if we should be using a similar pattern for the remainder of the cases as this is really the level that the tests are looking at. For example most of the remainder are simply looking for a TestSummary.xslt so we could create a similar `public class TestSummaryXslt { public string TestPath => AssemblyHelper.GetAssemblyPath(…); }`

This would then only leave 

https://github.com/nunit/nunit-console/pull/349/commits/e779307ae7254a27b2364ee845822b2e09c65a83#diff-1f357fc76a1c83c019ccddb28679d21aL61 which currently takes multiple args 
`  [TestCase("mock-assembly.dll", false)]`
`  [TestCase("nunit-agent.exe", false)]`
`  [TestCase("nunit-agent-x86.exe", true)]`



If @CharliePoole 's basic objection is the proliferation of the `#if` conditional, could we simply introduce a `TestHelper.GetDirectoryFromType(typeof(XXXXX))` (for example) pattern, that we reuse in the `MockAssembly.AssemblyPath` and `TestSummartXslt.Path` implementations?

Let me know your thoughts.

CharliePoole:
Comments on the last comment...

* `TestContext` is not low-level at all. It's available to all tests in the `NUnit.Framework` namespace and serves as a facade to the internal, low-level classes. As you probably noticed in the code an instance is created each time you reference `TestContext.CurrentContext`. However, `TestContext.TestDirectory` is static and causes no construction. All our advice to users for the past few years has been to use `TestContext.TestDirectory` in this situation, so we should follow it too!

* `MockAssembly` exists for some very specific uses: loading and running an entire test assembly and verifying that everything adds up as needed. The tests that use it are by definition more like feature or integration tests than unit tests. In the past folks have added extra methods to this assembly and tried to create tests using them. We have had to spend extra time to remove those methods and put them in the right place, which is generally the test-data assembly.

* Some of our tests were written before `TestContext.TestDirectory` existed. We didn't generally go back and "clean up" things like this just for consistency in the past. Now that there are more new people looking at the code, we may want to consider it. Of course, there are situations where we are __testing__ how `TestContext.TestDirectory` works. In those cases, we obviously can't use it in the test.

Can you state briefly why you don't want to use `TestContext.TestDirectory` in your tests? Obviously you can't use it in the engine or console code, but the tests are what it's for.

alski:
Sorry maybe I'm looking at the wrong code but that is not what I see. Firstly I haven't pulled down the nunit repo, only nunit-console, so my view is from Metadata via VS, but I see only an instance method...

```
#region Assembly nunit.framework, Version=3.6.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb
// C:\Dev\nunit-console\packages\NUnit.3.6.0\lib\net20\nunit.framework.dll
#endregion

using System;
using System.IO;
using NUnit.Framework.Constraints;
using NUnit.Framework.Interfaces;
using NUnit.Framework.Internal;

namespace NUnit.Framework
{
    //
    // Summary:
    //     Provide the context information of the current test. This is an adapter for the
    //     internal ExecutionContext class, hiding the internals from the user test.
    public class TestContext
    {
….
        //
        // Summary:
        //     Gets the directory containing the current test assembly.
        public string TestDirectory { get; }
```
image

alski:
Seems to be the same in GitHub too

https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/TestContext.cs#L131 

````
        /// <summary>
        /// Gets the directory containing the current test assembly.
        /// </summary>
        public string TestDirectory
        {
            get
            {
````

**Comment by CharliePoole** [comment]
You're quite right. I wasn't able to look at the code when I posted that and remembered it incorrectly. 

However, static or instance property aside, I still don't understand your not wanting to use it. Are you trying to get the information outside of a test method execution?

**Comment by ChrisMaddock** [comment]
Just to be clear - the way to access the TestDirectory property should be `TestContext.CurrentContext.TestDirectory`. 

Apologies - I forgot about the CurrentContext accessor as well!

**Comment by alski** [comment]
No problem. No, I would be happy to use it, if it was available at the call site. 
I've updated the pull request with an "as close as I can see how to get it" for now.  Ignore the TestContextHelper being in the MockAssembly, I would assume this code is taken back to NUnit.Framework.TestContext as soon as it can be.
Would this be acceptable, once the static/instance accessibility issue is resolved?

**Comment by CharliePoole** [comment]
I hope I'm not the only one to not understand what you're saying. 😄 Please help me out anyway! What is the call site `TestContext.CurrentContext.TestDirectory` is not available? It should be available anywhere you are actually __running__ a test: in SetUp, TearDown or the test method itself and in any helper methods they call. It __should__ also(*) be available in a `TestCaseSource` method, if you have one, although there are bugs around that in earlier framework versions.

IOW I am not understanding why you aren't able to use it or why we would need to change the framework. We are probably misconnecting over some bit of semantics.

* `TestCaseSource` methods are not invoked during the running of the test, but earlier.

**Comment by alski** [comment]
No it's not you, I think I'm being a little dopey there. 

Using `TestContext.CurrentContext.TestDirectory` fixes all the problems. Just checking all tests run and rolling back all other code changes, and it will be pushed.

File: src/NUnitEngine/mock-assembly/MockAssembly.cs [hunk 2]
```diff
@@ -103,6 +103,7 @@ public class MockAssembly
             public const int Inconclusive = MockTestFixture.Inconclusive;
 
 #if !NETSTANDARD1_6
+            //Replace with TestContext.CurrentContext.TestDirectory ?
```
CharliePoole: You can't do that since it's initialized in static constructor. This needs to be as it is because the field is accessed externally, by our tests, eons after the test was run. You can just drop the comment or add one that says we can't use TestContext.TestDirectory here.

src/NUnitEngine/mock-assembly/TestContextHelper.cs
```diff
@@ -0,0 +1,56 @@
+﻿// ***********************************************************************
```
CharliePoole: This file is no longer needed, right?

alski:
In terms of "what will we have accomplished", it fixes all tests that currently fail when running tests from nunit-console under the nunit-adapter ins VS, _apart_ from the ones that are designed to fail because they are effectively data for other tests.
If I remember correctly, 34 fails becomes 5.
image
These five can further be fixed by implementing the ability to exclude tests via the .runsettings available in VS. See https://github.com/nunit/nunit-console/issues/347#issuecomment-357329467

CharliePoole:
@ChrisMaddock Ah! I didn't catch that comment. Yes that would explain the adapter failures as a side issue. I'm fine with making things run under the adapter where it just requires a fix we should otherwise make. I'm not so much into pushing any changes into nunit itself for that purpose, that's all. Some of our work with Microsoft has pushed us in that direction and I was mixing this one up with those changes.

Thanks @alski for your work!

PR #351, opened 2018-01-18
Display subnodes of settings nodes in Console Run Summary

### LINKED ISSUE #285: ColorConsoleWriter.WriteLabel causes NullReferenceException
I've gotten this 7 times out of the last 9 builds on the build server at work. <s>The only thing that should have changed from last night is that there's a lot more parallelization going on.</s> Oh, also I updated to NUnit Framework 3.8.0, duh...

```c#
Run Settings
    DisposeRunners: True
    WorkDirectory: [redacted]
    TestParametersDictionary: System.NullReferenceException: Object reference not set to an instance of an object.
   at NUnit.ConsoleRunner.ColorConsoleWriter.WriteLabel(String label, Object option, ColorStyle valueStyle)
   at NUnit.ConsoleRunner.ColorConsoleWriter.WriteLabelLine(String label, Object option, ColorStyle valueStyle)
   at NUnit.ConsoleRunner.ResultReporter.WriteRunSettingsReport()
   at NUnit.ConsoleRunner.ResultReporter.ReportResults()
```

Fails with error -100.

ChrisMaddock:
Fixes #285 - now reads/displays the subnodes of the TestParamsDictionary settingnode from the framework, as opposed to reading the `value` attribute we added for legacy runners. 

The new output looks like this. Do we approve of the arrows? I wanted to avoid characters users may be likely to use in their own params.

image

This involved a couple of other changes - I suggest reviewing the two commits separately. I'll line-comment some of the unexpected bits.

Also involved:
- Adding the Guard.cs utility class to the console, to guard the method which previously crashed. Involved moving it to the NUnit.Common namespace
- Updating the NUnit framework version referenced by the tests, to remove a reference to the NUnit.Framework.Internal namespace. (Commented further inline.)

### DIFF
--- a/NUnitConsole.sln.DotSettings
+++ b/NUnitConsole.sln.DotSettings
@@ -7,6 +7,7 @@
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/ALIGN_MULTIPLE_DECLARATION/@EntryValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/ALIGN_MULTLINE_TYPE_PARAMETER_LIST/@EntryValue">True</s:Boolean>
 	<s:String x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/EMPTY_BLOCK_STYLE/@EntryValue">TOGETHER_SAME_LINE</s:String>
+	<s:String x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/PLACE_ACCESSORHOLDER_ATTRIBUTE_ON_SAME_LINE_EX/@EntryValue">NEVER</s:String>
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/SPACE_AFTER_TYPECAST_PARENTHESES/@EntryValue">False</s:Boolean>
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/SPACE_AROUND_MULTIPLICATIVE_OP/@EntryValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/SPACE_WITHING_EMPTY_BRACES/@EntryValue">True</s:Boolean>
@@ -36,7 +37,12 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.&#xD;
 </s:String>
 	<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/PredefinedNamingRules/=PrivateInstanceFields/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="_" Suffix="" Style="aaBb"&gt;&lt;ExtraRule Prefix="" Suffix="" Style="aaBb" /&gt;&lt;/Policy&gt;</s:String>
 	<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/PredefinedNamingRules/=PrivateStaticFields/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="_" Suffix="" Style="aaBb"&gt;&lt;ExtraRule Prefix="" Suffix="" Style="aaBb" /&gt;&lt;/Policy&gt;</s:String>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpAttributeForSingleLineMethodUpgrade/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpKeepExistingMigration/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpPlaceEmbeddedOnSameLineMigration/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpRenamePlacementToArrangementMigration/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EAddAccessorOwnerDeclarationBracesMigration/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002ECSharpPlaceAttributeOnSameLineMigration/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EMigrateBlankLinesAroundFieldToBlankLinesAroundProperty/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EMigrateThisQualifierSettings/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/PatternsAndTemplates/LiveTemplates/Template/=1256596694514F41A4D79DB082E50E51/@KeyIndexDefined">True</s:Boolean>
--- a/src/NUnitConsole/nunit3-console.tests/ResultReporterTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/ResultReporterTests.cs
@@ -30,7 +30,6 @@ using NUnit.Common;
 using NUnit.Engine;
 using NUnit.Framework;
 using NUnit.Framework.Api;
-using NUnit.Framework.Internal;
 using NUnit.Tests.Assemblies;
 using NUnit.Engine.Internal;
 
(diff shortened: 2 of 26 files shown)


src/NUnitConsole/nunit3-console.tests/ResultReporterTests.cs
```diff
@@ -46,12 +45,21 @@ public class ResultReporterTests
...
+                }}
+            };
 
-            var runner = new NUnitTestAssemblyRunner(new DefaultTestAssemblyBuilder());
-            runner.Load(mockAssembly, emptySettings);
-            var xmlText = runner.Run(TestListener.NULL, Framework.Internal.TestFilter.Empty).ToXml(true).OuterXml;
+            var controller = new FrameworkController(mockAssembly, "id", emptySettings);
+
+            controller.LoadTests();
+            var xmlText = controller.RunTests(null);
```
ChrisMaddock: Turns out the settings node is inserted at the FrameworkController level, and not the NUnitTestAssemblyRunner. To test this, I instead needed to move a level up in the framework API.

While doing that, I also removed the `Framework.Internal.TestFilter.Empty` reference - as we recently started supporting passing null or string.empty in it's place. Turns out that happened in a later version of the framework than the tests were currently using - hence the framework update.

src/NUnitConsole/nunit3-console.tests/ResultReporterTests.cs
```diff
@@ -169,11 +177,23 @@ public void TestsNotRunTest()
...
+                "        1 -> d",
+                "        2 -> c",
+                "    NumberOfTestWorkers: 8"
+            };
 
-        private TestEngineResult AddMetadata(TestEngineResult input)
+            var report = GetReportLines(_reporter.WriteRunSettingsReport);
+            Assert.That(report, Is.SupersetOf(expected));
+        }
+
```
ChrisMaddock: Tests both forms of the `TestParameters` setting, to ensure ongoing support, until such a time as we chose not to. `Is.SubsetOf` is an attempt to make the test more maintainable than expecting an exact string.

ChrisMaddock:
Whoops - accidentally included the number of cores line in my settings tests. Worked on my machine! 😆

NUnitConsole.sln.DotSettings
```diff
@@ -7,6 +7,7 @@
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/ALIGN_MULTIPLE_DECLARATION/@EntryValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/ALIGN_MULTLINE_TYPE_PARAMETER_LIST/@EntryValue">True</s:Boolean>
 	<s:String x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/EMPTY_BLOCK_STYLE/@EntryValue">TOGETHER_SAME_LINE</s:String>
+	<s:String x:Key="/Default/CodeStyle/CodeFormatting/CSharpFormat/PLACE_ACCESSORHOLDER_ATTRIBUTE_ON_SAME_LINE_EX/@EntryValue">NEVER</s:String>
```
mikkelbu: I tried to figure out what `PLACE_ACCESSORHOLDER_ATTRIBUTE_ON_SAME_LINE_EX` does, but I only found examples from other github projects (and all has it configured to `NEVER`). So probably ok 😄 .

NUnitConsole.sln.DotSettings
```diff
@@ -36,7 +37,12 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.&#xD;
...
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpKeepExistingMigration/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpPlaceEmbeddedOnSameLineMigration/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ECSharpRenamePlacementToArrangementMigration/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002EAddAccessorOwnerDeclarationBracesMigration/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EReSharper_002EPsi_002ECSharp_002ECodeStyle_002ESettingsUpgrade_002ECSharpPlaceAttributeOnSameLineMigration/@EntryIndexedValue">True</s:Boolean>
```
mikkelbu: I've not looked into the settings above.

src/NUnitConsole/nunit3-console.tests/ResultReporterTests.cs
mikkelbu: 👍 (but you probably meant `Is.SupersetOf` and not `Is.SubsetOf` 😄)

src/NUnitConsole/nunit3-console.tests/ResultReporterTests.cs
```diff
@@ -169,11 +177,23 @@ public void TestsNotRunTest()
...
+        private static TestEngineResult AddMetadata(TestEngineResult input)
         {
-            input.Add("<settings><setting name=\"Setting1Name\" value=\"Setting1Value\"></setting><setting name=\"Setting2Name\" value=\"Setting2Value\"></setting></settings>");
```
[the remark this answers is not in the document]
mikkelbu: At least if we can trust https://github.com/nunit/docs/wiki/Test-Result-XML-Format

NUnitConsole.sln.DotSettings
CharliePoole: Maybe such changes should really have been in a separate PR. I wouldn't hold this one up on account of it, however.

src/NUnitEngine/nunit.engine/Guard.cs
```diff
@@ -23,7 +23,7 @@
 
 using System;
 
-namespace NUnit.Engine
+namespace NUnit.Common
```
CharliePoole: We might want to come back and gather the various common classes into a separate directory somewhere.

src/NUnitConsole/nunit3-console.tests/ResultReporterTests.cs
```diff
@@ -46,12 +45,21 @@ public class ResultReporterTests
         [OneTimeSetUp]
         public void CreateResult()
         {
-            var mockAssembly = typeof (MockAssembly).Assembly;
-            var emptySettings = new Dictionary<string, object>();
+            var mockAssembly = typeof(MockAssembly).Assembly;
+            var emptySettings = new Dictionary<string, object>
```
CharliePoole: Since they are no longer empty, perhaps a different name would fit better. 😄

src/NUnitConsole/nunit3-console.tests/ResultReporterTests.cs
CharliePoole: In fact, we should really stop using the framework API here. The console actually only deals with the engine but we ended up with code like this because I couldn't get a secondary engine to work alongside the "real" engine. That's due to the duplicate URI that arises when the secondary engine tries to initialize the `TestAgency` service.

In a different PR, I recently came up with a way to run a secondary `MasterTestRunner`, which is almost as good as a secondary engine. Alternatively, we could create code to configure a secondary engine that lacks the `TestAgency` and possibly some other services that are not needed.

I don't think that should be part of this PR, however, but something to look at in the future.

src/NUnitConsole/nunit3-console.tests/ResultReporterTests.cs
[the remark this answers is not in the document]
CharliePoole: Settings can appear in different places because some settings apply only to an assembly. The ones you are concerned with should actually be created at the engine level, at least to my memory, so I'm a bit surprised here. This is a more or less obscure part of the architecture (not well documented) so it could be that some things have moved around over time.

NUnitConsole.sln.DotSettings
jnm2: As far as I could figure out, it means you have ReSharper 2017.3 and nothing more. 😄

NUnitConsole.sln.DotSettings
ChrisMaddock: Sorry - I should have commented this. I made no purposeful changes to the resharper settings - these are just applied by default with the latest version of ReSharper. It sometimes changes the settings under `IsMigratorApplied` with a new update. I could have reverted, but I think it would have insisted on applying them again every time I opened the solution. 😄

NUnitConsole.sln.DotSettings
ChrisMaddock: I agree Joseph - see comment below. Sorry!

src/NUnitConsole/nunit3-console.tests/ResultReporterTests.cs
ChrisMaddock: Whoops! 😊 I'll fix that one!

src/NUnitConsole/nunit3-console.tests/ResultReporterTests.cs
ChrisMaddock: That's a good point. I didn't think that far through.

src/NUnitConsole/nunit3-console.tests/ResultReporterTests.cs
ChrisMaddock: I think what I have is correct? I'm expecting the actual report to contain more stuff than the `expected` array. Or am I missing something? 🙂

src/NUnitConsole/nunit3-console.tests/ResultReporterTests.cs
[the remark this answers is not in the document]
ChrisMaddock: I _think_ I had a look to see if there was any code to create `settings` at test-run level, and couldn't find any - but it was a while ago I did that bit now, so can't be sure!

src/NUnitConsole/nunit3-console.tests/ResultReporterTests.cs
ChrisMaddock: OH - I explained it wrong in my line comment! Apologies! 😅

src/NUnitConsole/nunit3-console.tests/ResultReporterTests.cs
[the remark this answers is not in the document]
CharliePoole: It looks as if all settings have moved to the assembly level, including those that function at a higher level than the assembly. So no problem. 😄

CharliePoole:
@ChrisMaddock The WIP in title is still indicating you don't want this merged.

ChrisMaddock:
That's correct, I'm coming back to fix your settings comment at some point this week, Charlie. 🙂 

> Since they are no longer empty, perhaps a different name would fit better.

ChrisMaddock:
Final fix made, this is ready for merge once CI passes. 👍

rprouse:
LGTM. Travis Mono Latest is failing on `apt-get install mono` and the OSX builds are backed up like traffic in Toronto. Other builds are fine, so overriding and merging.

PR #353, opened 2018-01-19
Ensure that assembly-level errors are actually displayed by console runner

### LINKED ISSUE #350: Invalid assemblies no longer give an error message
In the NUnit3 console runner 3.7, the argument `x.dll` gives an error message indicating the file was not found. Typing `x.junk` gives a message saying that the file type is not known.

In the current master build, the run fails without specifying a message in both cases.

CharliePoole:
Fixes #350 

This is the console part of the fix. I'll do the framework fix next.

As I envision this, the console should not need any changes once the framework fix is in. However, the console still has to deal with older framework versions and it is for those versions that this PR is targeted.

The problem is with failures, errors or warnings that come from child tests of a suite, but which do not have the `site` attribute set to `Child`. That means the runner assumes the suite itself is the cause of the error. A previous fix solved the problem by ignoring any suites without a `site` attribute but that causes the problem described in the issue. Assembly not found and similar errors don't use the site at all.

After experimenting with various approaches, including looking at the `type` attribute, I decided that the best way, although ugly, was to examine the message used. For all our releases, we have used constant messages for suites with a failed or warning status due to problems with the child tests. I'm using those messages to decide whether a suite with no site specified should actually have the `Child` value instead of the default `Test` value.

This is pretty adhoc, but fixes on top of existing framework releases often are. Please review with that in mind and try to see if you can think of any situations where this will not work.

### DIFF
--- a/src/NUnitConsole/nunit3-console.tests/BadFileTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/BadFileTests.cs
@@ -34,8 +34,6 @@ namespace NUnit.ConsoleRunner.Tests
 {
     public class BadFileTests : ITestEventListener
     {
-        static string NL = Environment.NewLine;
-
         [TestCase("junk.dll", "File not found")]
         [TestCase("EngineTests.nunit", "File type is not supported")]
         public void MissingFileTest(string filename, string message)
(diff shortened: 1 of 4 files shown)


CharliePoole:
@rprouse If this looks good to you, let's merge without waiting for an IOS build, which may never come.

jnm2:
We need to do something about Travis. Maybe the .NET Foundation can get us a different service that does OSX.

CharliePoole:
I can force the merge at any time, of course.

This problem has been going on for a while. Should we make the iOS build optional? Disable entirely?

CharliePoole:
Yes, we previously had these builds set up to allow failure. If you set it up right, the build is flagged as complete even if the jobs that allow failure have not started. We changed to requiring it because we wanted to be sure that it worked on iOS.

PR #363, opened 2018-01-27
3.8 Release Changes

rprouse:
This merges the 3.8 release changes back into master. It includes,

- Update to Changes.txt
- Update the copyright year to 2018
- Switch the version to 3.9 for ongoing development

Still outstanding in the release,

- [x] Create and upload the installer to the release
- [x] Update the website
- [ ] Upload the two Chocolatey packages
- [x] Notify users

### DIFF
--- a/nuget/runners/nunit.runners.nuspec
+++ b/nuget/runners/nunit.runners.nuspec
@@ -28,7 +28,7 @@
     <releaseNotes>This release uses the latest version of the TeamCityEventListener extension.</releaseNotes>
     <language>en-US</language>
     <tags>nunit test testing tdd runner</tags>
-    <copyright>Copyright (c) 2017 Charlie Poole</copyright>
+    <copyright>Copyright (c) 2018 Charlie Poole</copyright>
     <dependencies>
       <group>
         <dependency id="NUnit.ConsoleRunner" version="$version$" />
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,4 @@
-﻿NUnit 3.8 - February 27, 2018
+﻿NUnit 3.8 - January 27, 2018
 
 This release includes several fixes when unloading AppDomains and better error reporting. The
 aggregate NuGet packages also include updated versions of several extensions.
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,40 @@
-﻿NUnit 3.7 - July 13, 2017
+﻿NUnit 3.8 - February 27, 2018
+
+This release includes several fixes when unloading AppDomains and better error reporting. The
+aggregate NuGet packages also include updated versions of several extensions.
+
+Issues Resolved
+
+ * 6 TypeLoadException in nunit3-console 3.0.1
+ * 93 Update Readme with information about the NuGet packages
+ * 111 Provide better info when AppDomain won't unload
+ * 116 NUnit 3.5.0 defaults to single agent process when using an nunit project file
+ * 191 Exception encountered unloading AppDomain
+ * 228 System.Reflection.TargetInvocationException with nunit3-console --debug on Mono
+ * 246 No way to specify app.config with console runner
+ * 256 Rewrite ConsoleRunnerTests.ThrowsNUnitEngineExceptionWhenTestResultsAreNotWriteable()
+ * 259 NUnit3 agent hangs after encountering an "CannotUnloadAppDomainException"
+ * 262 Transform file existence check should check current directory instead of WorkDirectory
+ * 267 Fix possible NRE
+ * 273 Insufficient error handling message in ProcessRunner -> RunTests method
+ * 275 Integrate chocolatey packages with build script
+ * 284 NUnit3: An exception occured in the driver while loading tests...   bei NUnit.Engine.Runners.ProcessRunner.RunTests(ITestEventListener listener, TestFilter filter)
+ * 285 ColorConsoleWriter.WriteLabel causes NullReferenceException
+ * 289 Warnings not displayed
+ * 298 Invalid --framework option throws exception
+ * 300 Agents do not respect the Console WORK parameter when writing log file
+ * 304 Catch agent debugger launch exceptions, and improve agent crash handling
+ * 309 No driver found if framework assembly reference has uppercase characters
+ * 314 Update NUnit.Extension.VSProjectLoader to 3.7.0
+ * 318 Update NUnit.Extension.TeamCityEventListener to 1.0.3
+ * 320 Return error code -5 when AppDomain fails to unload
+ * 323 Assertion should not be ordered in AgentDatabaseTests
+ * 343 Superfluous unload error shown in console
+ * 349 Get all TestEngine tests running under NUnitAdapter apart from those .
+ * 350 Invalid assemblies no longer give an error message
+ * 355 NuGet package links to outdated license
+
+NUnit 3.7 - July 13, 2017
 
 Engine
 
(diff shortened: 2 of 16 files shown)


rprouse:
@CharliePoole can you upload the Chocolatey packages? I don't have permissions. You can grab them from the release. They are the first two packages at https://github.com/nunit/nunit-console/releases/tag/3.8

nuget/runners/nunit.runners.nuspec
```diff
@@ -28,7 +28,7 @@
     <releaseNotes>This release uses the latest version of the TeamCityEventListener extension.</releaseNotes>
     <language>en-US</language>
     <tags>nunit test testing tdd runner</tags>
-    <copyright>Copyright (c) 2017 Charlie Poole</copyright>
+    <copyright>Copyright (c) 2018 Charlie Poole</copyright>
```
mikkelbu: Is it on purpose that this file as the only .nuspec file that only contains Charlie's name?

CharliePoole: I think I left it as is because I wasn't going to update it any longer, just have it depend on NUnit.Console latest version. Rob decided to keep releasing it, so it should be updated like the others.

CharliePoole:
@rprouse I'll update the releases on Chocolatey.

CharliePoole:
@rprouse The chocolatey packages are in the queue. Do you have a chocolatey user id so I can add you as a package maintainer?

Also... are there any new extension versions that have not been uploaded to chocolatey yet?

rprouse:
@CharliePoole I am rprouse on Chocolatey.

Looking at the history, the only extensions that have been updated in the last year are the VS Project Loader which is at the latest version and TeamCity which is one version behind. It should be updated to the 1.0.3 version. You are the owner for that too, https://chocolatey.org/packages/nunit-extension-teamcity-event-listener, but I don't see a chocolatey package created for the release at https://github.com/nunit/teamcity-event-listener/releases/tag/1.0.3. Did you just rename the existing NuGet package?

CharliePoole:
@rprouse IIRC @ChrisMaddock had me update chocolatey for the VS project loader.

I created an adhoc 1.0.2 package for the teamcity extension using code that read the nuget file and transformed it to create a nuspec following chocolatey rules. IIRC I offered to work on the actual extension build script so that a chocolatey file would be created whenever it was built. @NikolayPianikov do you remember discussing that?

At any rate, AFAIK it has never been done. It's really up to @NikolayPianikov as to whether the teamcity extension can be distributed on chocolatey but I'm happy to look at it if need be.

CharliePoole:
@rprouse I added you as a maintainer of both packages. I just realized that the one including extensions is being uploaded for the first time. I wanted to note that this was not my intention for chocolatey. Since you can install using `choco <list of names>` I figured we didn't need a package with extensions.

Apparently we are creating one, so maybe you and I already discussed this, leading to creation of the combined package.

CharliePoole:
@rprouse @NikolayPianikov The package for the runner with extensions is failing verification because it depends on the non-existent teamcity package.

For what it's worth, here's what I thought the process for integrating packages should be:

1. Extension maintainer makes changes and tests against the current release (also probably against the latest master,, but that doesn't impact users.

2. Extension maintainer releases, including on chocolatey. Presumably, the maintainer tests the package itself by installing locally first.

3. Users who need the new extension install it separately and provide feedback.

4. When the dependent package (like Console) is released, the person doing it makes sure the old extensions are working with the new release. In the case of chocolatey, that means installing the console runner using the chocolatey file locally and then installing the extensions from chocolatey.org.

There's no doubt that this is a very involved and time-consuming process. It's one of the prices of unbundling the various components, I'm afraid.

CharliePoole:
@rprouse I'm glad to help out once you decide which way you want to go to fix this.

rprouse:
@CharliePoole I don't remember a discussion about the console package with extensions. I had assumed that you created it, so included it with the release. I'm fine if you want to stick with the packages that are available.

As for the TeamCity extension, I think it is up to @NikolayPianikov to release and maintain it if he feels it is necessary. It is another good reason not to maintain the Console with Extensions package on Chocolatey. What do you think?

As for testing each of the extensions with each release of the console, we should do that with the NuGet package too, but to be honest I don't. I install the version with extensions and run projects using it, but I don't test every combination of extensions. Once again, we need some sort of integration tests 😄

CharliePoole:
@NikolayPianikov Do you recall our discussion about releasing a chocolatey package? Chocolatey users who have the console runner installed could then install an update of the teamcity extension, which is then automatically used. They can also install the runner plus extension or extensions by listing their names on a single command-line if they like. I think this is a great advantage for distribution.

The existing package on chocolatey (1.0.2) is one I created on an adhoc basis. If you want to continue to support chocolatey, we need to integrate creation of that package alongside your nuget package. Although both systems use nuspecs, they are organized differently and chocolatey has extensive validation rules that must be followed.

I'm available to help if you tell me what you want to do.

CharliePoole:
@rprouse I'll cancel the upload for the package with extensions. It can always be re-created out of band if we decide we want it.

NikolayPianikov:
@CharliePoole Honestly, I do not remember, but I would like to do all required changes to make it. I'll be grateful if you give a reference to solution which I could use to follow it.
@rprouse  About integration tests: I have a set of integration tests for 3 extensions in base scenarios for different version of nunit console/frameworks. They could be used to tests combination of extensions as well.

CharliePoole:
@NikolayPianikov Please see discussion on PR nunit/teamcity-event-listener#43 and issue nunit/teamcity-event-listener#41

I had added a chocolatey package to the script and you had me merge the PR. IIRC, the coding of the script was a bit different from all the other extensions because the teamcity extension itself uses a somewhat different build script with a lot of integration tests. Consequently, I wasn't 100% confident that it would work as you needed it to work.

I assume a chocolatey package was created when you did the 1.0.3 release. It isn't one of the files in the release on GitHub, however. Here are the steps that I think are needed...

1. Verify that a chocolatey package is being created.
2. Test the package
3. If it's OK, add the package to the release on github. If there are problems, we should work to resolve them first.
4. Release the package on chocolatey. This presents a problem in terms of who has permission to upload. Our chocolatey packages (like the nuget ones) include the actual file for the extension rather than referencing a download location. For that reason, they may not be accepted if the person uploading is not a copyright holder. I can upload it this time but we need to work out that problem for the future.

Maybe we need an issue on your project to keep discussing this. 😄

NikolayPianikov:
@CharliePoole thank you, I will go through all points but little bit later - I have critical bugs to fix

NikolayPianikov:
@CharliePoole I've been through points 1-3. 
I've checked - everything is ok. But I am planning to create a set of tests for it on our CI.
Do you have an idea how we will publish it in the future?

PS Why the package has id which is not equal to NuGet package or just to use single package for both repositories?

CharliePoole:
@NikolayPianikov The package id is different for two reasons...

1. Chocolatey has specific naming standards, which we have to follow.

2. The content is different. Creating two different packages with the same id could lead to a lot of confusion, even if we don't intend to upload them both to the same place.

If you feel that the package works, I suggest you build it matching the 1.0.3 release, test it and post on GitHub. I'll then upload the same thing to chocolatey and we can deal with future releases at more leisure.

The key problem is that my uploads to chocolatey are treated differently because my name is listed as the holder of the copyright on the software. If that isn't the case, then other rules apply and it's not normal to include actual binaries in the package. Instead, they point to a download site. You and your company had not wanted me to add your name as a copyright holder, although I'm still willing to do that.

We actually have a similar issue with the other extensions and the console runner itself. I'll try to find out from my "guy" at chocolatey what our options are.

NikolayPianikov:
@CharliePoole ok, thank you

PR #366, opened 2018-01-31
Redirect error output to file when specified

### LINKED ISSUE #360: CommandLineOption --err does not write error input to ErrorFile
Hi guys,

I am trying this:
"nunit3-console.exe MyTestAssembly_uTest.dll --err=ErrorOutput.txt --out=StandardOutputfile.txt"

The assembly to test has only one test:
```C#
public void SampleTest()
       {
            Console.Error.WriteLine("I am a standard error output.");
            TestContext.Error.WriteLine("I am a standard error output.");
        }
```
After execution of the test case above, the file which should contain the error output is always empty. 
I realized that the strings are written to the StandardOutputfile.txt.

Can you help me guys?

rprouse:
Fixes #360 

This was never hooked up, if `--err` was specified, the file was created, but no output was ever written to the file. Error output always went to whatever `--out` went to or to the console. NUnit has been out for years and nobody noticed 😳 

Labels complicate things a bit and don't work very well with error output. The events for these come in as they happen making our various Before, After, All, etc. difficult. Since this is seldom used and to keep it simple, I add the test name label before error output in all cases except when `labels == None`.

### DIFF
--- a/src/NUnitConsole/nunit3-console.tests/TestEventHandlerTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/TestEventHandlerTests.cs
@@ -32,14 +32,19 @@ namespace NUnit.ConsoleRunner.Tests
     {
         private StringBuilder _output;
         private TextWriter _writer;
+        private StringBuilder _errOutput;
+        private TextWriter _errWriter;
 
         private string Output {  get { return _output.ToString(); } }
+        private string ErrorOutput { get { return _errOutput.ToString(); } }
 
         [SetUp]
         public void CreateWriter()
         {
             _output = new StringBuilder();
             _writer = new StringWriter(_output);
+            _errOutput = new StringBuilder();
+            _errWriter = new StringWriter(_errOutput);
         }
 
         [TestCaseSource("SingleEventData")]
@@ -76,7 +81,7 @@ namespace NUnit.ConsoleRunner.Tests
         {
             var handler = new TestEventHandler(_writer, _writer, labels);
 
-            handler.OnTestEvent(ErrorOutput);
+            handler.OnTestEvent(ErrorOutputXml);
 
             Assert.That(Output, Is.EqualTo(expected));
         }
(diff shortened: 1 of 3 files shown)


CharliePoole:
I would have expected that separately directing the error output would imply no labels being used - just a file with the text itself. That's how it was done before we integrated error output with the labels.
 
That said, I don't think it matters much. There's little reason to be sending the two outputs to two different files in most cases, since the single combined output is well labeled.

In fact, I could imagine just dropping the --err option, since it has apparently not worked without anyone noticing for so long!

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -354,14 +347,16 @@ private void DisplayTestFilters()
...
         {
             if (_options.ErrFileSpecified)
             {
                 var errorStreamWriter = new StreamWriter(Path.Combine(_workDirectory, _options.ErrFile));
                 errorStreamWriter.AutoFlush = true;
-                _errorWriter = errorStreamWriter;
+                return new ExtendedTextWrapper(errorStreamWriter);
             }
+
+            return _outWriter;
```
ChrisMaddock: Won't this change the default output for errors from `stderr` to `stdout`? Meaning if I'm redirecting the two streams separately (in cmd, outside of NUnit) - it'll be a change in behaviour?

src/NUnitConsole/nunit3-console/TestEventHandler.cs
```diff
@@ -133,7 +137,44 @@ private void TestOutput(XmlNode outputNode)
         {
             var testName = outputNode.GetAttribute("testname");
             var stream = outputNode.GetAttribute("stream");
+            if (stream == "Error" && _outWriter != _errWriter)
```
ChrisMaddock: If my above comment is correct - then here, could we instead pass a TextWriter into `TestErrorOutput`, and simplify this check?

src/NUnitConsole/nunit3-console/TestEventHandler.cs
```diff
@@ -133,7 +137,44 @@ private void TestOutput(XmlNode outputNode)
...
+                TestErrorOutput(outputNode, testName);
+            else
+                TestOutput(outputNode, testName, stream);
+        }
+
+        private string _currentErrorLabel;
```
ChrisMaddock: Nitpick - could this field move to live with the other similar ones at the top of the class?

src/NUnitConsole/nunit3-console.tests/TestEventHandlerTests.cs
```diff
@@ -71,6 +76,83 @@ public void MultipleEvents(string[] reports, string labels, string expected)
...
+            Assert.That(ErrorOutput, Is.EqualTo($"=> ErrorTest{Environment.NewLine}OneTwo"));
+        }
+
+        [Test]
+        public void ErrorWriteLineDoesIncludeNewlinesInSameTest()
+        {
+            var handler = new TestEventHandler(_writer, _errWriter, "Before");
+
+            handler.OnTestEvent(string.Format(ErrorOutputFormat, "ErrorTest", "One\r\n"));
+            handler.OnTestEvent(string.Format(ErrorOutputFormat, "ErrorTest", "Two\r\n"));
```
ChrisMaddock: Why are these two explicit `\r\n`s - and everything else uses Environment.NewLine? Is that to do with what the framework writes?

ChrisMaddock: I'm glad this stuff's explicitly tested thought - there's some subtle functionality here. 😄

src/NUnitConsole/nunit3-console.tests/TestEventHandlerTests.cs
```diff
@@ -71,6 +76,83 @@ public void MultipleEvents(string[] reports, string labels, string expected)
...
+            handler.OnTestEvent(string.Format(ErrorOutputFormat, "ErrorTest1", "One\r\n"));
+            handler.OnTestEvent(string.Format(ErrorOutputFormat, "ErrorTest2", "Two\r\n"));
+
+            Assert.That(ErrorOutput, Is.EqualTo($"=> ErrorTest1{Environment.NewLine}One{Environment.NewLine}=> ErrorTest2{Environment.NewLine}Two{Environment.NewLine}"));
+        }
+
+        const string ErrorOutputFormat = "<test-output stream=\"Error\" testname=\"{0}\"><![CDATA[{1}]]></test-output>";
+        static readonly string ErrorOutputXml = string.Format(ErrorOutputFormat, "ErrorOutput", "This is an error\r\n");
+
+        static TestCaseData[] ErrorData = new TestCaseData[]
```
ChrisMaddock: This needs to be inside the pragma I think - it's failing Travis.

CharliePoole:
This seems to have become a new feature rather than a bug fix. And I'm not sure it's a feature we want.

Originally, NUnit had the ability to send standard and error output to two separate files. Then we created labels in standard output. Next, we made sure that error output fell within the same labels with standard output when it was not redirected. Finally, we discovered that the error output was not being redirected even if requested.

I think we should fix the bug. I don't think anyone asked for a separate labeled output file for errors only. If one redirects error output, it's likely that they just want the errors for some particular purpose. IMO, if they do that, they simply lose the combined report provided when the errors go to the same place as standard output.

rprouse:
Closing in favor of #381

PR #368, opened 2018-02-01
Fix #367

### LINKED ISSUE #367: nunit3-console loads nunit.framework with partial name
This prevents running tests when there are non-trivial assembly codebase entries in the app.config file, since redirection polices are not applied when loading with partial name. In general, you should never load with partial name.

The bug is in the NUnit3 driver and the fix is not difficult, since the full assembly information is already available for a load with full name. I'll create a pull request for it ASAP.

A similar issue exists in NUnit 2 and its driver for NUnit 3. I'll create a separate issue for it.

cyanite:


src/NUnitEngine/nunit.engine.tests/Drivers/NUnit3FrameworkDriverTests.cs
```diff
@@ -35,6 +35,7 @@ namespace NUnit.Engine.Drivers.Tests
     public class NUnit3FrameworkDriverTests
     {
         private const string MOCK_ASSEMBLY = "mock-assembly.dll";
+        private const string NUNIT_FRAMEWORK_NAME = "nunit.framework, Version=3.9.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb";
```
CharliePoole: We have no way of knowing that the version of the framework used. That is, not unless we require installation of an entirely new engine for each framework version.

cyanite: This is the version used by mock-assembly which is part of the same project and which has the version number hardcoded into its csproj:

    <Reference Include="nunit.framework, Version=3.9.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb, processorArchitecture=MSIL">
      <HintPath>..\..\..\packages\NUnit.3.9.0\lib\net20\nunit.framework.dll</HintPath>
    </Reference>

CharliePoole: The test assemblies naturally all reference a particular version of the NUnit framework. That's the way NUnit is used and in our tests we are users of NUnit. We usually update the engine tests to the latest release of the framework when we do a new engine release, but there's no requirement to do so unless we are using new features.

There are, however, no references to the framework from the code that we package. Users may use any version of the framework starting with 3.0. This has always been a design goal of NUnit since it's beginning and NUnit 3 continued it. Many users have different projects using different NUnit framework versions but all running under the latest release of some runner.

I think you will do better if you go back to the issue and help us confirm that there is a problem.

cyanite: >There are, however, no references to the framework from the code that we package. Users may use any version of the framework starting with 3.0.

That's not relevant here since my reference is specifically to the nunit.framework used in mock-assembly which can most certainly not be "any version", since it's hard coded in the csproj.

CharliePoole:
OK, I see you are doing more than I initially saw. I'll do a new review shortly.

cyanite:
Thanks 👍

src/NUnitEngine/nunit.engine.tests/Drivers/NUnit3FrameworkDriverTests.cs
CharliePoole: This would still be good to avoid, however, since it gives us an extra place to change the framework version when we update it, rather than just changing the reference.

CharliePoole: This could be fixed by pulling the assembly name out of mock-assembly. Code can be brief since it's guaranteed to have the reference.

CharliePoole:
Restarted one travis job.

cyanite:
For 2, I really don't think we should get rid of the hardcoded framework because:

1. The mock-assembly is part of the nunit3-console project, and is only used for test. It's not user provided, nor in a separate repository.
2. My hard coded string is in the test where "mock-assembly" is also hard coded.
3. The framework version mock-assembly uses is hard coded in its csproj-file.

In normal operation, Mono.Cecil would have reflected the real assembly reference to use at this point, but in this unit test we start at this point instead. So if we don't hard code the name, we would need to e.g. use Mono.Cecil to pull it out.

CharliePoole:
Let's imagine what happens when the maintainers decide it's time to update the console/engine tests to use a new framework version...

CURRENTLY:
Use the VS menu to upgrade the nunit package

WITH THIS CHANGE:
Use the VS menu to upgrade the nunit packge
Edit the code of this particular test to change the version

The second step can easily be forgotten and is actually unnecessary. All we have to do is leave the old constructor in place. Since this test assembly doesn't use any binding redirect, it shouldn't cause a problem. Alternatively, the test could get the proper reference from the mock-assembly and use it.

src/NUnitEngine/nunit.engine.tests/Drivers/NUnit3FrameworkDriverTests.cs
jnm2: Let's make this `static readonly` and pull the reference from the current assembly, since it's easy.

src/NUnitEngine/nunit.engine.tests/Services/TestFilteringTests.cs
```diff
@@ -34,6 +35,7 @@ namespace NUnit.Engine.Services.Tests
     public class TestFilteringTests
     {
         private const string MOCK_ASSEMBLY = "mock-assembly.dll";
+        private const string NUNIT_FRAMEWORK_NAME = "nunit.framework, Version=3.9.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb";
```
jnm2: Same

src/NUnitEngine/nunit.engine.tests/Drivers/NUnit3FrameworkDriverTests.cs
jnm2: Also, HintPath doesn't meet my definition of 'hardcoding', nor do NuGet-restore-generated references. What's really hardcoded from my point of view is the NuGet package version 3.9.0.

CharliePoole:
@jnm2 What kind of new test were you thinking of?

jnm2:
So long as it fails before the change and passes afterwards, I don't care if it's more unit-ish or integration-ish.

CharliePoole:
Well, it's pretty much an integration test. We'll need to crate a new assembly with a special config in order to duplicate the use case as outlined in the issue. Personally, that's more than I would ask in order to approve this.

jnm2:
Isn't there any way to make a test double assembly loader so we can make sure the requested assembly name contains a version?

CharliePoole:
What would you be mocking? The assembly loader is basically .NET. You could certainly write a test that verifies we construct the framework driver with an AssemblyName argument, but that doesn't need testing with the PR because you can't call it any other way.

This is system stuff and very hard to test other than in a system-level test IMO.

That said, I think we need to decide whether we want to fix the problem first. That's why I haven't done a review yet.

CharliePoole:
Restarted failing CI job again. Appears that the PPA for mono latest is still out of service.

src/NUnitEngine/nunit.engine.tests/Services/TestFilteringTests.cs
cyanite: > Also, HintPath doesn't meet my definition of 'hardcoding'

It's not the hint path that's hardcoded, it's this part: `<Reference Include="nunit.framework, Version=3.9.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb, processorArchitecture=MSIL">` combined with the fact that there is no SpecificVersion false tag (for good reason).

Pulling the reference out of the assembly (using Mono.Cecil, I assume you mean?) sounds fine to me, I can update the PR.

cyanite:
> We should add a test that protects this new behavior.

Well, it's not so much new behavior as it's a bug fix: loading by partial name is, in my opinion, a bug, since you take away all foreign appdomain control (via app.config) of the assembly load process.

cyanite:
> I think we need to decide whether we want to fix the problem first.

I'm not sure I understand why you wouldn't? The fix is basically two lines and it will behave the same in all existing situations, plus it will avoid loading by partial name. Microsoft has obsoleted the method to do this (although it can be done via Assembly.Load, which is essentially what's happening in NUnit), and warns against it here: https://docs.microsoft.com/en-us/dotnet/framework/deployment/best-practices-for-assembly-loading ("Avoid Binding on Partial Assembly Names"), listing several problems.

This fix only makes NUnit more capable, without taking anything away. It even removes a hard coded string "nunit.framework" in the code, since we *already* reflected the complete assembly name for the framework used.

CharliePoole:
@cyanite Frankly, the need to confirm comes from how you originally presented the issue. You described a best practice ("You should never... ") rather than giving a use case. It's best to present issues in terms of the effect they have rather than just giving opinions. Different programmers have different opinions.

You also cited Microsoft guidelines, which is basically an appeal to higher authority. I'm surprised that you believe it would matter to me what those guys think. 😄 I'm a little offended that you didn't realize we were are violating the usual rules of assembly loading in full knowledge of what those rules are.

See the issue for reasons for and against fixing this - the PR isn't the place to discuss it.

CharliePoole:
@cyanite I'm not clear how you can remove the hard-coded name of the assembly. Each test assembly has a list of direct references, which may or may not be test frameworks. There is nothing in the assemblyname to tell us that a particular reference is a test framework unless we have the names of valid test assemblies somewhere.

Imagine that we created `charlie.framework` as a new test framework. How will we select its driver in the current design without using that string?

cyanite:
>  I'm surprised that you believe it would matter to me what those guys think.

Maybe because it's not what they *think* that matters (to me either), but what actually happens when you load by partial name. They have a check point list about that. The main thing that happens is that the configuration of the target domain is disregarded. Anyway, let's move discussion to the issue, then.

cyanite:
> I'm not clear how you can remove the hard-coded name of the assembly. Each test assembly has a list of direct references, which may or may not be test frameworks. There is nothing in the assemblyname to tell us that a particular reference is a test framework unless we have the names of valid test assemblies somewhere.

I don't understand this comment. The only hardcoded assembly name in my PR is in a test, refering to a mock-assembly which also has a hardcoded assembly name in its csproj. You're right, if "nunit.framework" is no longer the only valid test framework, I can't remove it. But I am also against removing it, because this is a unit test, not an integration test.

jnm2:
@cyanite 

> > We should add a test that protects this new behavior.
> 
> Well, it's not so much new behavior as it's a bug fix:

We should add a test that protects this behavior we didn't have before due to this bug fix.

Looks like it would need to be an integration test. 

@CharliePoole 

> Personally, that's more than I would ask in order to approve this.

Okay. :-) I wouldn't mind @cyanite wanting to contribute their minimal repro as an integration test to ensure that the fix lasts. I guess it comes down to how important this change is?

> I think we need to decide whether we want to fix the problem first.

Absolutely right. I'm going to respond to https://github.com/nunit/nunit-console/issues/367.

cyanite:
> Thank you for correcting me.

By the way, in case this was sarcastic, I apologise for putting it that way. I didn't mean to correct you, just pointing out that the PR doesn't change the behaviour for anything existing (except when it didn't work before).

CharliePoole:
@cyanite Our team members are never sarcastic...

Unless that was sarcasm. 😄

jnm2:
@cyanite Lol, yes! I was making a point since it seemed you hadn't noticed what I had actually _said_ and were only clarifying a point which I already understood and agreed with. No hard feelings. I've been on the other side myself!

jnm2:
@cyanite Lol, yes! I was making a point since it seemed you hadn't noticed what I had actually _said_ (that we should add a test to go with the change) and that you were only clarifying a point which I already understood and agreed with. No hard feelings. I've been on the other side myself!

cyanite:
> I'm OK with this provided the test gets changed so that we don't hard code the AssemblyName definition.

@CharliePoole ok, but what do you suggest instead? That I pull it out of the dll using Mono.Cecil or that I invoke other parts of NUnit to do that?

Is there anything else I need to do? @rprouse mentioned he had some comments, but I don't see them here yet.

cyanite:
@jnm2 here you go: https://github.com/cyanite/nunit3-issue367-repro

Building this and running nunit3-console.exe on the produced assembly will reproduce the issue.

jnm2:
@cyanite For the hardcoded name, why not `typeof(NUnit.Framework.TestAttribute).Assembly.GetName().FullName`?

Thanks for the repro. Unless you want to, I'll see if I can work it into an integration test so we can validate this change.

CharliePoole:
Exactly, Cecil doesn't enter into it at all.

cyanite:
> @cyanite For the hardcoded name, why not typeof(NUnit.Framework.TestAttribute).Assembly.GetName().FullName?

@jnm2, @CharliePoole  Because that will give us the full name of whatever nunit.framework is loaded into the test, not necessarily the one that our testing assembly (i.e. mock-assembly.dll) uses. Are those guaranteed to be the same? The whole point of my PR is to improve the "always load the correct assembly" logic.

jnm2:
@cyanite You're right that mock-assembly is set up to use whatever version the tests are using. That's what the integration test is needed for.

cyanite:
@jnm2 alright, then I'll change it and update the PR :)

CharliePoole:
@cyanite I'm glad you explained! I didn't understand why you were making an issue of it.

Yes. As @jnm2 says, our __unit tests__ all use the same version of the framework and are designed that way. Back before the framework was separated, they used the version we were building. Now that the framework and engine are separate, the engine tests use one common version of the framework installed across the solution using nuget.

We definitely need separate integration tests for all of this, which would include mixing different framework versions, etc. But so far we have not been able to put the time into creating it.

Just FYI, we use Mono.Cecil for checking assemblies that we do __not__ necessarily want to load into the domain where the checking is done. Once we have made the check and we know we are loading it, then there is no problem using reflection.

cyanite:
@CharliePoole I believe I addressed your review comment with my updated PR :)
@jnm2 Do I need to add further tests or something? I also recall that @rprouse said he had some comments.

PR #393, opened 2018-03-29
Remove duplicate nunit.engine.api packaging

ChrisMaddock:
Noticed nunit.engine.api.dll is packaged in two places in the package.

@CharliePoole - you added this package - this reference has been in twice since the nuspec was added to the repo. Am I missing a subtlety, or is it just a copy and paste error?

### DIFF
--- a/nuget/engine/nunit.engine.tool.nuspec
+++ b/nuget/engine/nunit.engine.tool.nuspec
@@ -21,7 +21,6 @@
     <file src="LICENSE.txt" />
     <file src="NOTICES.txt" />
     <file src="CHANGES.txt" />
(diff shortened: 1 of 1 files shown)


CharliePoole:
This is an experimental package, which we don't distribute. The app assembly is deliberately in both the lib and the tools directory so it gets copied to output from lib and is also in tools where the engine references it. The idea was that runners could start using the engine as a package dependency rather than bundling it.

Rob decided on a different direction for the engine so I stopped working on this but I left it until it was actually decided to delete it.

ChrisMaddock:
Thanks for the history Charlie. I've added a new PR at https://github.com/nunit/nunit-console/pull/396. I think we should remove to avoid future confusion!

PR #396, opened 2018-03-30
Remove experimental nunit.engine.tool packaging

ChrisMaddock:
See discussion on https://github.com/nunit/nunit-console/pull/393.

Here's the package on NuGet - it hasn't been updated since v3.6.0. It would be good to delist it, and add to the description that's its no longer released.

https://www.nuget.org/packages/NUnit.Engine.Tool/

In my opinion, the downloads are sufficiently low, and it's been forgotten about for long enough that this is the best approach - we shouldn't put extra effort into releasing an additional package.


ChrisMaddock:
@rprouse - I believe you'll need to handle the de-listing, if that's ok! 😄

PR #397, opened 2018-03-30
Merge nunit-distribution into nunit-console repo

### LINKED ISSUE #218: Move Distribution back into the nunit-console project.
As discussed in nunit/governance#1

ChrisMaddock:
Fixes #218. This brings the current msi and zip packaging functionality into the main repo - keeping functionality pretty much as is.

**What's changed:**
- The zip file now additionally includes test files. This reflects the framework zip, it didn't seem sensible to filter the files out. The structure of the zip is otherwise as before, to avoid breaking changes.
- The 'COMPONENTS.TXT' file that used to be packaged with the zip has caused us an amount of pain recently, due to it's construction relying on a Cake addin - and various incompatibilities there. It didn't play nice with the current Cake version in this repo, and given the high cost of it recently, I've cut it.

Other comments inline. 🙂 

@rprouse - once this is merged, would you please delete the old repo? (Seems preferable to archiving, given we've kept all the content and history.)


build.cake
```diff
@@ -1,15 +1,12 @@
 //////////////////////////////////////////////////////////////////////
-// ARGUMENTS
+// ARGUMENTS & INITIALISATION
 //////////////////////////////////////////////////////////////////////
 
 var target = Argument("target", "Default");
-var configuration = Argument("configuration", "Debug");
```
ChrisMaddock: The msi uses the configuration to build - and it felt like it might slip people minds to package under a Release configuration! Making this 'Release' by default also makes it consistent with the framework repo

build.cake
```diff
@@ -375,41 +383,35 @@ Task("PackageConsole")
...
             NoPackageAnalysis = true
         });
     });
 
-// Unlike other tasks, this one does not depend on creation of the
-// image directory. I think this is the direction we want to go
-// for all the different packages, but it's for a separate change.
 Task("PackageChocolatey")
 	.Description("Creates chocolatey packages of the console runner")
-	.IsDependentOn("Build")
```
ChrisMaddock: This initially confused me - by running the 'Package' task, you ended up with different binaries being packaged for NuGet than for Chocolatey. That seemed dangerous - so I removed it for now. I think if we want this dependency, we need to add it for all packaging - but we should look at that as a separate issue.

build.cake
```diff
@@ -473,6 +475,70 @@ Task("PackageNetStandardEngine")
...
+Task("PackageZip")
+.IsDependentOn("CreateCombinedImage")
+.Does(() =>
+{
+    CleanDirectory(ZIP_IMG);
+
+    //Flatten for zip
+    CopyDirectory(CURRENT_IMG_DIR, ZIP_IMG);
+    CopyDirectory(ZIP_IMG + "bin/", ZIP_IMG);
+    DeleteDirectory(ZIP_IMG + "bin/", new DeleteDirectorySettings { Recursive = true });
```
ChrisMaddock: Flattening is to ensure no change in current zip structure.

rprouse:
Great to finally get this done. Thanks

rprouse:
I will archive the distribution repo for now. We can delete after the next release?

ChrisMaddock:
Sounds good, thanks Rob!

build.cake
CharliePoole: If you verified that, it maybe a Cake bug. Things depended on more than once should only run once. That's how every build system since make has worked. :-)

ChrisMaddock: I don't think so in this case, it's a based on the order of dependencies of the Package task. So, packaging currently runs
```
PackageNuGet ->PackageChoco -> Package Msi
```

Because only Choco depended on the build task, that evaluated to this:
```
PackageNuGet -> Build -> PackageChoco -> Package Msi
```

That meant NuGet packaging was done with the old binaries, then they were rebuilt, then choco packaging was done.

CharliePoole: I misunderstood. I thought the other packages also depended on build directly or indirectly. If we want packaging to __not__ build but use the latest build, then there should be no dependencies. If we want __any__ packaging to trigger a build, then each individual package should depend on build. If we want only the full package target to trigger a build, then that should have the dependency.

It's possible I created the inconsistency when I was experimenting with different ways to do this. I'm afraid you may find various half-done things from the point where I stopped working on the build because the only way to test some of them in CI was to commit them. 😞

PR #399, opened 2018-04-02
Fix minor doccoment issues

ChrisMaddock:
Minor fixes. Resolving these as they're flagged by ReSharper - and I like working with a clean sheet. 😄


src/NUnitEngine/nunit.engine/IDriverService.cs
```diff
@@ -38,6 +38,7 @@ public interface IDriverService
         /// <param name="domain">The application domain in which to run the tests</param>
         /// <param name="assemblyPath">The path to the test assembly</param>
         /// <param name="targetFramework">The value of any TargetFrameworkAttribute on the assembly, or null</param>
+        /// <param name="skipNonTestAssemblies">True if non-test assemblies should simply be skipped rather than reporting an error</param>
```
CharliePoole: IMO we should clean this up at some time. The first three args are necessary to select and create a driver. The fourth should really not be there since it has to do with how we handle the situation when we can't get a driver. Not relevant for this PR, however.

ChrisMaddock:
Thanks Charlie.

PR #401, opened 2018-04-05
Remove -5 UNLOAD_ERROR exit code

### LINKED ISSUE #371: Remove -5 exit code on app domain unload failures
Hi,

After updating from 3.7.0 to 3.8.0, NUnit gives an error exit code.
Below is the error output from both 3.7.0 and 3.8.0.
The way the NUnitEngineUnloadException is handled has changed, but I do believe it hasn't been for the better...
In 3.7.0 I didn't get an error exit code, but in 3.8.0 I do.

In 3.7 I get this as engine errors:
`Unhandled Exception: NUnit.Engine.NUnitEngineException: Exception encountered unloading AppDomain
Agent Process was terminated successfully after error.
   at NUnit.Engine.Runners.TestExecutionTask.Execute()
   at NUnit.Engine.Runners.ParallelTaskWorkerPool.ProcessTasksProc()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object
state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ThreadHelper.ThreadStart()`
powershell gives True as output for this process

In 3.8 I get the following:
`NUnit.Engine.NUnitEngineUnloadException : Multiple exceptions encountered. Retrieve AggregatedExceptions property for more information
  ----> NUnit.Engine.NUnitEngineUnloadException : Agent Process was terminated successfully after error.
  ----> NUnit.Engine.NUnitEngineUnloadException : Exception encountered unloading application domain
  ----> NUnit.Engine.NUnitEngineException : Exception encountered unloading application domain: Error while unloading ap
pdomain. (Exception from HRESULT: 0x80131015)
Application domain name: domain-f7378c0-*.dll
Application domain BaseDirectory: Path`
powershell gives False as output for this process

ChrisMaddock:
Looking at the code for a recent email discussion, I realised how trivial it was to fix #371 

This reverts to the previous-intended-functionality regarding app domain unloads (ignoring the bugs causing -100's up till 3.7!). Behaviour should now be that the engine continues to throw a NUnitEngineUnloadException - but the console just writes a warning to screen and exists with the normal code for the test run.

In future, we may wish to instead attach the warning to the test assembly at engine level, but that's a bigger implementation!

### DIFF
--- a/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
+++ b/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
@@ -47,7 +47,7 @@ namespace NUnit.ConsoleRunner
         public static readonly int INVALID_ASSEMBLY = -2;
         //public static readonly int FIXTURE_NOT_FOUND = -3;    //No longer in use
         public static readonly int INVALID_TEST_FIXTURE = -4;
-        public static readonly int UNLOAD_ERROR = -5;
+        //public static readonly int UNLOAD_ERROR = -5;         //No longer in use
         public static readonly int UNEXPECTED_ERROR = -100;
 
         #endregion
@@ -248,7 +248,6 @@ namespace NUnit.ConsoleRunner
                 if (unloadException != null)
                 {
                     writer.WriteLine(ColorStyle.Warning, Environment.NewLine + ExceptionHelper.BuildMessage(unloadException));
-                    return ConsoleRunner.UNLOAD_ERROR;
                 }
 
                 if (reporter.Summary.UnexpectedError)


src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -47,7 +47,7 @@ public class ConsoleRunner
         public static readonly int INVALID_ASSEMBLY = -2;
         //public static readonly int FIXTURE_NOT_FOUND = -3;    //No longer in use
         public static readonly int INVALID_TEST_FIXTURE = -4;
-        public static readonly int UNLOAD_ERROR = -5;
+        //public static readonly int UNLOAD_ERROR = -5;         //No longer in use
```
rprouse: LGTM, but we should just remove it if we are removing it.

ChrisMaddock: I wanted to leave it there to make sure we don't end up reusing -5 for something else in a couple of releases time. I can remove if  you disagree - what do you think?

rprouse: Good enough reason 😄

PR #408, opened 2018-04-22
Relocate DotNetFrameworkLocator methods

ChrisMaddock:
Relates to #173 

Another small one with no intended change in functionality - moving .NET Framework location code out of RuntimeFramework into a static helper class. I plan to eventually do the same with Mono location code, and then use both of these from RuntimeFrameworkService.

Note that RuntimeFramework class is public -  but users should only be referencing via nunit.engine.api anyway...

### DIFF
--- a/src/NUnitEngine/nunit.engine/Internal/RuntimeFrameworks/DotNetFrameworkLocator.cs
+++ b/src/NUnitEngine/nunit.engine/Internal/RuntimeFrameworks/DotNetFrameworkLocator.cs
@@ -1,0 +1,104 @@
+// ***********************************************************************
+// Copyright (c) 2007 Charlie Poole, Rob Prouse
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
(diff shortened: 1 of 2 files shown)


src/NUnitEngine/nunit.engine/Internal/RuntimeFrameworks/DotNetFrameworkLocator.cs
```diff
@@ -0,0 +1,81 @@
+using System;
```
mikkelbu: We are missing the copyright

src/NUnitEngine/nunit.engine/nunit.engine.csproj
```diff
@@ -64,6 +64,7 @@
       <Link>Agents\AgentExitCodes.cs</Link>
     </Compile>
     <Compile Include="Agents\RemoteTestAgentProxy.cs" />
+    <Compile Include="Internal\RuntimeFrameworks\DotNetFrameworkLocator.cs" />
```
mikkelbu: I was actually going to say that we should move this down to the other Internal files, but when @rprouse merges in the new csproj format the includes are going to disappear. So let us just keep it as it is.

src/NUnitEngine/nunit.engine/Internal/RuntimeFrameworks/DotNetFrameworkLocator.cs
ChrisMaddock: D’oh! Thanks, will get back to it!

src/NUnitEngine/nunit.engine/Internal/RuntimeFrameworks/DotNetFrameworkLocator.cs
```diff
@@ -1,3 +1,26 @@
+// ***********************************************************************
+// Copyright (c) 2007 Charlie Poole, Rob Prouse
```
mikkelbu: Shouldn't this be 2018, since this is a new file, although the logic is moved from an existing file.

ChrisMaddock: I thought we kept the original date when code was just moved? I am absolutely not an authority on this stuff though!

rprouse: Could be either, I don't think we have a hard and fast rule. At my company, if we make substantial changes, we update the copyright to the current year.

ChrisMaddock:
Thanks Mikkel - I just need to rebase. Sorry - been pretty busy lately. 😅

ChrisMaddock:
@mikkelbu - I finally got round to rebasing this! It just needs a 'Github review' to merge, which I can't do myself...would you mind? 🙂 Nothing's changed since the last time you looked!

ChrisMaddock:
Thanks!

Issue #412, opened 2018-05-01
Extensions not dectected for version 3.9.0-dev-03997


MJLHThomassen-Eurocom:
As discussed in #390 , I grabbed the latest 3.9.0-dev-03997 version of the nuget3-console runner from myget and i have the nunit-extension-vs-project-loader installed. When executing nunit3-console with a .sln, it comes up with an error:

```
nunit3-console .\MySolution.sln --config=Release
NUnit Console Runner 3.9.0
Copyright (c) 2018 Charlie Poole, Rob Prouse

Runtime Environment
   OS Version: Microsoft Windows NT 10.0.16299.0
  CLR Version: 4.0.30319.42000

Test Files
    .\MySolution.sln


Errors, Failures and Warnings

1) Invalid : E:\Path\To\MySolution\MySolution.sln
File type is not supported

Test Run Summary
  Overall result: Failed
  Test Count: 0, Passed: 0, Failed: 0, Warnings: 0, Inconclusive: 0, Skipped: 0
  Start time: 2018-05-01 13:39:52Z
    End time: 2018-05-01 13:39:53Z
    Duration: 0.930 seconds

Results (nunit3) saved as TestResult.xml
```

I installed both the console and the extension using Chocolatey v0.10.10:

```
choco install NUnit.ConsoleRunner --version 3.9.0-dev-03997 --source https://www.myget.org/F/nunit/api/v2
choco install nunit-extension-vs-project-loader
```

The extension is detected and works when installing the nunit console runner 3.8.0 via the regular Chocolatey stream:

```
choco install nunit-console-runner
```

I suspect this has something to do with the nunit.choco.addins/nunit.nuget.addins files (since they appear to be pointing to the extension directory).

Installation (of 3.9.0-dev-03997) via MyGet resulted in a `C:\ProgramData\chocolatey\lib\NUnit.ConsoleRunner\tools\nunit.nuget.addins` file with contents

```
../../NUnit.Extension.*/**/tools/     # nuget v2 layout
../../../NUnit.Extension.*/**/tools/  # nuget v3 layout
```

Installation (of 3.8.0) via regular feed rulted in a `C:\ProgramData\chocolatey\lib\nunit-console-runner\tools\nunit.choco.addins` file with contents

```
../../nunit-extension-*/tools/     # find extensions installed under chocolatey
```

however copying either the `nunit.choco.addins` file or its contents to the `nunit.nuget.addins` location/file results in another error when calling nunit3-console:

```
[...]

PR #419, opened 2018-05-14
Use API version for nunit.engine.api

ChrisMaddock:
Fixes https://github.com/nunit/nunit-console/issues/412. Just a simple typo in the PR converting to the new csproj - it shows up a gap in our integration testing - but that's a gap I believe we're already well aware of...


rprouse:
Good catch @ChrisMaddock, thanks for tracking it down.

Issue #433, opened 2018-06-27
All messages from EventListenerTextWriter goes to console output independent on stream name


nvborisenko:
I want to put some hidden message from test to `ITestEventListener` extensions.
```csharp
public void Test()
{
  (new NUnit.Framework.Internal.Execution.EventListenerTextWriter("DiagnosticStream", System.IO.TextWriter.Null)).WriteLine("This is message from custom stream");
}
```
But this message also goes to console output. Is there some way to accomplish my needs?

CharliePoole:
This seems like something that could be needed in general: a way to send messages from the framework to the engine and/or runner.

The example code creates an event in XML with element name textout, which conceptually represents text output from the text. The event has a stream attribute which is used by runners to decide how to output the text. A special stream name would mean "this is not really text output, ignore it."

Piggybacking on top of text output like this seems a bit messy. It's equally easy to create a new type of event element for the purpose of communicating with the engine. Let's accept this and label for design. I have some ideas which I'll post later, when I'm actually awake!

nvborisenko:
Right, I need communication between test domain and engine domain. This communication should be dedicated only for me and no others participants should react on these messages. Thanks @CharliePoole

CharliePoole:
Currently, our convention is that unknown event types (XML elements) are ignored. Unknown text streams ate displayed to avoid information loss.

CharliePoole:
Some thoughts for the @nunit/framework-and-engine-team on design...

1. If this is to be handled as an event, i.e. going through the framework event pump, then it will be completely asynchronous. If a synchronous message is wanted, that's a couple of orders of magnitude more in re-design.

2. At a minimum, I think a new message element is needed, rather than piggybacking on text output. It's reasonable to think that all text output events will result in the output of text!

3. In designing a new element, we should consider carefully whether it will be targeted at a particular extension, at all extensions, at the engine or at the engine and runners. In particular, I'm wondering if we should only handle some messages within the engine and its extensions and avoid forwarding them to runners. Should there be some attribute in the new element that indicates how far to distribute?

4. I think some refactoring is called for so that we actually have an identifiable messaging facility in the engine. The current messages would then make use of it rather than creating their `EventListenerTextWriters` directly. Use of a `EventListenerTextWriter` would then be an implementation detail within the messaging facility.

jnm2:
[...]

PR #434, opened 2018-06-27
Don't output message to console if the stream name is "Diagnostic"

nvborisenko:


CharliePoole:
Let's first discuss design approach on issue #433.

ChrisMaddock:
Thanks for this @nvborisenko. I think we've decided on #433 that this isn't the approach we want to take, so I'll close this PR.

Issue #439, opened 2018-07-13
.nunit projects do not work when running under --process=Separate


ChrisMaddock:
This situation shows the "File type not supported" message - as an inappropriate TestRunner is being selected. Related to #XXX

ChrisMaddock:
Closing in favour of https://github.com/nunit/nunit-console/issues/546

CharliePoole:
Ah! I searched but didn't find this one!

Issue #382, opened 2018-03-05
Only 1 agent running with multiple projects and --process=Multiple


Matteoz991:
Using a .nunit file to run multiple projects and --process=Multiple will only run one agent process per time while with version 3.6.1 i was able to run one agent per test project.

The command line i'm using is:
nunit3-console InsertAnyPathHere\MyNunitProjFile.nunit --x86 --dispose-runners --process=Multiple

My .nunit file is like this: 
```
<NUnitProject>
  <Settings activeconfig="Release" processModel="Multiple" domainUsage="Default" />
  <Config name="Release" runtimeFramework="net-4.0" binpathtype="Manual">
    <assembly path="folder/folder/folder/MyTestDll.dll" />
    <assembly path="folder/folder/folder/MyTestDll1.dll" />
    <assembly path="folder/folder/folder/MyTestDll2.dll" />
    <assembly path="folder/folder/folder/MyTestDll3.dll" />
    <assembly path="folder/folder/folder/MyTestDll4.dll" />
    <assembly path="folder/folder/folder/MyTestDll5.dll" />
    <assembly path="folder/folder/folder/MyTestDll6.dll" />
    <assembly path="folder/folder/folder/MyTestDll7.dll" />
    <assembly path="folder/folder/folder/MyTestDll8.dll" />
  </Config>
</NUnitProject>
```

I couldnt find any way to run again, as per version 3.6.1, more than one agent per time.

ChrisMaddock:
This should be fixed in NUnit Console 3.8 ,as per https://github.com/nunit/nunit-console/issues/116. Please try updating, let us know if this is still a problem.

ChrisMaddock:
Actually, this could be a different issue - `--process=multiple` would previously have forced multiple processes.

Can you try: a) updating the the latest console, and b) removing the `--x86` flag, and see if the results are different in either case?

Matteoz991:
So.. i've try to update to the latest official release 3.8, and downloading the latest code from git and with both i have only 1 agent running per time.. even without --x86 flag.

I took a look at the source code and i think the problem is because the number of agents running is based on the SubPackages count in TestPackage class but the moment this is verified this count is equal to 1 and the SubPackages from the .nunit file are loaded only after inside the agent itself. I could be wrong but i hope it helps.

rprouse:
Thanks for the help @Matteoz991. You are probably right in your analysis, we've seen similar issues before.

CharliePoole:
[...]

PR #440, opened 2018-07-13
Fix --inprocess NUnit Project loader

### LINKED ISSUE #386: nUnit project loader does not work when --inprocess is set
When starting nunit3-console 3.8.0 with an nUnit project (requiring nUnit project loader) and either the --inprocess or --process=Single parameter, I get an "File type is not supported" error message. If I use exactly the same parameters but leave out the --inprocess, the project file is recognized correctly.

(Tested with nunit3-console msi package on Windows 10.)

This seems to be a regression in 3.8.0; 3.7.0 worked fine.

ChrisMaddock:
Fixes #386 

As far as I can see - any project must at some point be fed into an AggregatingTestRunner - as this is runner which currently has the knowledge to unpack the project.

In a multiple process situation (via ProcessModel.Default, or ProcessModel.Multiple), projects are fed either directly to an AggregatingProcessRunner, or to the MultipleProcessRunner (incidentally, also not the correct solution - logged as: https://github.com/nunit/nunit-console/issues/382)

This PR fixes the specific issue of running a project a single nunit project under ProcessModel.Single - by ensuring such a project would be passed to the MultipleDomainRunner. (Which inherits from AggregatingTestRunner, and therefore is able to unpack the project TestPackage.) I'd like to merge this as an incremental step - although given #382, and also #439, I think this area may need a more substantial refactor.

The majority of actual code changes here are additional lines of logging, which I thought helpful to leave  in, as this code is also hit in the agent process. The only production code change is in InProcessTestRunnerFactory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ build/
 .vscode/
 tools/
 !tools/packages.config
+launchSettings.json
 
 # Enable "build/" folder in the NuGet Packages folder since NuGet packages use it for MSBuild targets
 !packages/*/build/
--- a/src/NUnitEngine/nunit.engine.tests/Services/DefaultTestRunnerFactoryTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Services/DefaultTestRunnerFactoryTests.cs
@@ -56,7 +56,7 @@ namespace NUnit.Engine.Services.Tests
         // Single file
         [TestCase("x.nunit",           null,        typeof(AggregatingTestRunner))]
         [TestCase("x.dll",             null,        typeof(ProcessRunner))]
-        [TestCase("x.nunit",           "Single",    typeof(TestDomainRunner))]
+        [TestCase("x.nunit",           "Single",    typeof(MultipleTestDomainRunner))]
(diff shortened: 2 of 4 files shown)


ChrisMaddock:
Sigh...missed that test file. It's late - I'll come back to this.

CharliePoole:
According to docs, assemblies should be run according to the `--process` and `--domain` settings no matter how they are specified (either directly on the command-line or indirectly within some kind of project file).

However, and this is a bit of a contradiction, some project formats (like `.nunit`) allow you to specify the equivalent process and domain settings internally.

The way we resolved this in the past is that a command-line setting should override the project settings. This means we need to know whether we are running Multiple processes (for example) by default or because it was on the command-line. I don't think we retain that info internally, which may be a problem.

ChrisMaddock:
Yes, I think that’s roughly the conclusion I came to, although I think that particular issue fits under #382, rather than here. This issue is simpler I think, as once we’ve selected a single process runner, then we can’t accidentally end up with a multiple one again I don’t think. 

There’s already some transfer of settings between the top level test package, and project test package. I guess we should maybe just update those, and ensure things are being overwritten correctly.

PR #452, opened 2018-08-18
Multi-Targeting of engine extensions (.NET Framework only)

### LINKED ISSUE #394: Multi-targeted Engine Extensions
It would be good to support multi-targeted engine addin's. Two motivations:

1. To allow extensions to target two different .NET Framework versions, where there is a significant advantage to build a version targeting a higher framework, but we don't wish to drop support for older frameworks. https://github.com/nunit/vs-project-loader/issues/25 is a great example of this.

1. To allow the packaging of extensions for potential separate .NET Standard and .NET Framework engines, which we may have in the future.

Here's a rough breakdown of what I think would be involved: 
- [ ] Allow engine to prioritise 'highest possible' available framework of addin.
- [ ] Permit engine to safely ignore extensions targeting a framework which can't be loaded. (May already exist?)
- [ ] Define .addins file format for these situations, closely tied to the recommended NuGet packaging method of extensions.

Here's some of my thoughts on how to go about this - comments welcome:

_Updated post-discussion 02/04/18_

**Extension precedence**
The engine already has code to select the extension of the highest assembly version, when multiple extensions of the same type are available. I propose we expand this to look at **Highest Assembly Version** and then **'Best' target framework**.

For the sake of consistency with the wider .NET ecosystem, I think our definition of 'Best' framework should match [NuGet's](https://docs.microsoft.com/en-us/nuget/reference/target-frameworks). We can limit the overhead by only initially implementing support for addins targetting .NET Framework or .NET Standard.

**NuGet packaging of Extensions**
~For consistency with the wider .NET ecosystem, I believe we should aim to make use of the existing multi-target packaging conventions introduced by NuGet - or the defaults created by `dotnet pack` on a multi-targeting project.~

~We currently seem to package our extensions under `/tools/` in their NuGet package. Is this the right location, given the main engine NuGet package, packages that under `/lib/`? The NuGet docs recommend `/tools/` for "Powershell scripts and programs accessible from the Package Manager Console".~

~Taking account of backwards compatibility - I'd like to look at repackaging our extension dll's with the dlls under the `/lib/` directory - as I believe is more standard. I could be wrong there - open to discussion on this one with people more knowledgable than me!~

We'll continue to package extensions under the `tools` subdir. We'll add in framework specific subdirectories, as below:

```
tools -> net20 -> extension.dll
      -> net45 -> extension.dll
```

**Addins file format/NuGet packaging**
The format of the addins file itself doesn't need to change - it can of course be pointed at a specific dll targeting a specific platform already. The question to look at here is the paths we put in the default .addins files to cater for NuGet and .msi installations.

I'd be tempted just to add an extra level of wildcard directory to the existing paths for msi/NuGet resolution, and have the engine attempt to load all extensions following the extension precedence logic described above. This is less efficient than attempting to read directory names and work out the , but would simplify the code, as the below functionality would already be required, given the current functionality to specify specific paths in the .addins file. Thoughts on this bit?

To ensure backwards compatibility between new extension packages, and old engine installations, and `.addins` file will be added to the `tools` directory of new extensions with subdirectory. This will allow the .addins files to 'chain' - and old engines to locate new addins, without the user needing to make any manual edits to the addins file.

This is spun out of https://github.com/nunit/nunit-console/issues/388. @rprouse - if you had particular plans for this part, feel free to boot me off and take the issue back! :smile:

ChrisMaddock:
This is some unfinished work from a few months back - excuse it being a bit random! Fixes #394 as far as is currently possible - if we're considering implementing extension functionality for the .NET Standard engine as a whole separate issue!

The engine will now prioritise the extension with the highest target framework - when two of the same type are made available. Preference order in determining which extension version to load is now:

> Assembly Version -> Target Framework -> Found via wildcard/path directly specified

In the short term, this unblocks https://github.com/nunit/vs-project-loader/issues/25 - something I might have a look at soon. In the longer-term - this is hopefully the foundations for resolving extensions made available for different platforms.

Again, I recommend review commit-by-commit - I'll add in a few line-comments shortly on some bits worth highlighting. The changes are less significant then they look - I did an amount of refactoring so we could test this better, as I imagine this functionality may become more complex over time!


src/NUnitEngine/nunit.engine.tests/Extensibility/ExtensionAssemblyTests.cs
```diff
@@ -22,96 +22,60 @@
 // ***********************************************************************
 
 using System;
-using System.Collections.Generic;
 using System.Reflection;
-using System.Text;
+using NUnit.Engine.Extensibility;
 using NUnit.Framework;
```
ChrisMaddock: Significant changes in this file are mainly due to tests here split out into ExtensionSelectorTests

src/NUnitEngine/nunit.engine/Extensibility/ExtensionAssembly.cs
```diff
@@ -0,0 +1,77 @@
+﻿// ***********************************************************************
```
ChrisMaddock: This file has been moved and isn't entirely new - I guess GitHub hasn't noticed as there's also some changes.

src/NUnitEngine/nunit.engine/Extensibility/ExtensionSelector.cs
```diff
@@ -0,0 +1,75 @@
+﻿// ***********************************************************************
```
ChrisMaddock: Functionality in this new class mostly extracted from ExtensionAssembly.cs - some changes to consider targetframework. Refactor was for test purposes.

src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
```diff
@@ -397,6 +391,11 @@ private void ProcessCandidateAssembly(string filePath, bool fromWildCard)
                     if (!fromWildCard)
                         throw new NUnitEngineException(String.Format("Specified extension {0} could not be read", filePath), e);
                 }
+                catch (NUnitEngineException)
+                {
+                    if (!fromWildCard)
+                        throw;
+                }
```
ChrisMaddock: Additional check as we now attempt to read target frameworks in this section, which can throw.

ChrisMaddock:
Tests are failing on mono-only, I'm going to guess some differences in assembly name reflection. Looking into it...

ChrisMaddock:
Issue is that our RuntimeFramework code currently believes Mono 5.10.0.160 can't support .NET 2.0 targeted things. I think I might have to deal with this as a separate issue...

ChrisMaddock:
@rprouse - this is a non-trivial change, and could probably be merged post-release, if you'd prefer? I don't mind either way. 🙂

src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
```diff
@@ -432,7 +446,7 @@ public void FindExtensionsInAssembly(ExtensionAssembly assembly)
                 if (versionArg != null && new Version((string)versionArg) > ENGINE_VERSION)
                     continue;
 
-                var node = new ExtensionNode(assembly.FilePath, type.FullName);
+                var node = new ExtensionNode(assembly.FilePath, type.FullName, assemblyTargetFramework);
```
CharliePoole: Once we decide that we can load the assembly in the current runtime, why is the assemblyTargetFramework still needed in the extension node? Is it only so that the console runner can display it?

ChrisMaddock: Yes, exactly. Info is only stored in `ExtensionNode` to be accessible to runners.

src/NUnitEngine/nunit.engine/Properties/AssemblyInfo.cs
```diff
@@ -27,4 +27,10 @@
     "fa6d1ea760e1ca6065cee41a1a279ca234933fe977a096222c0e14f9e5a17d5689305c6d7f1206"+
     "a85a53c48ca010080799d6eeef61c98abd18767827dc05daea6b6fbd2e868410d9bee5e972a004"+
     "ddd692dec8fa404ba4591e847a8cf35de21c2d3723bc8d775a66b594adeb967537729fe2a446b5"+
-    "48cd57a6")]
\ No newline at end of file
+    "48cd57a6")]
+
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2,PublicKey=002400000480000094" +
```
CharliePoole: Did I miss where this is explained? Maybe a comment here would help.

CharliePoole:
I had to go back and re-read all the discussion, including my own comments, to try to remember why we were doing this! Restating what I think you are trying to accomplish in my own terms so you can tell me if I get it correctly... 😄 

1. Currently, we can write extensions that target .NET 2.0. We ask that extension-writers do so in order to make them available to all users.

2. You'd like to write extensions that target higher framework versions and eventually .NET Core. This change takes care of the first and prepares for doing the second later on.

Where I am a bit confused is that I __thought__ we already handled the first case. If you write an extension that targets say 4.5, then we don't load it unless we are running under 4.5 or higher. The existing extension service is supposed to take care of that and I'm pretty sure it once did.

So, are we doing anything new here? Or are we just preparing for eventual .NET Core extensions? IOW, what is the use case that currently fails but will succeed under this PR?

If that case is, for example, loading a 4.6 extension when running under 3.5, I'd like to figure out when it stopped working correctly - that is, ignoring the extension. At least I think it used to work that way!

src/NUnitEngine/nunit.engine/Properties/AssemblyInfo.cs
ChrisMaddock: Sorry - this is to allow using NSubstitute with internal types. It should have a comment, yes!

CharliePoole:
My memory (which could easily be wrong) was that we caught the exception (from 4.6 for example) logged the error and ignored the assembly without crashing. I had to deal with it even though we didn't have extensions > 2.0 because there could still be assemblies in the same directory that targeted a higher version. In fact, at one point, we threw if there was a non-assembly in the same directory! However, this all may have been before we started using Mono.Cecil and you obviously have to deal with what we now do, not with what we used to do.

In any case, selecting the highest loadable version is a new thing that I hadn't thought about. 

BTW, I assume you are dealing with the identical assembly being found twice? It's possible to set up connected `.addins` files in a way that the same directory is scanned twice and it's also possible to have two copies of the same assembly in different directories. In fact, that's what the original "duplicate" logic was about.

ChrisMaddock:
Ah. So if there's an exception at the point of examining an assembly - then yes, we would catch and continue - adding that was my first ever contribution to NUnit, in fact! 😄 I had thought however that it would be possible to examine an assembly targeting a higher framework without crashing...but I'm not sure.

> BTW, I assume you are dealing with the identical assembly being found twice?

I didn't change existing functionality here - if there's no differentiating factors two assemblies, the engine chooses the first it encounters.

CharliePoole:
Then it all makes sense to me. Nice work!

src/NUnitEngine/nunit.engine/Properties/AssemblyInfo.cs
ChrisMaddock: Done!

PR #454, opened 2018-08-18
Misc improvements to ExtensionServiceTests

ChrisMaddock:
These tests caused me some debugging issues during #452, as they were failing on Mono only.

I've switched them over to use the newly-improved CollectionConstraint, which gives us more detailed error messages when they fail, in terms of what the collection actually looked like. I also tidied formatting/naming-conventions etc. while I was there.

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/Services/ExtensionServiceTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Services/ExtensionServiceTests.cs
@@ -22,7 +22,7 @@
 // ***********************************************************************
 
 using System;
-using System.Reflection;
+using System.Linq;
 using NUnit.Framework;
 using NUnit.Engine.Extensibility;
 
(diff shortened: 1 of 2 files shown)


src/NUnitEngine/nunit.engine.tests/Services/ExtensionServiceTests.cs
```diff
@@ -135,23 +108,18 @@ public void AllExtensionPointsAreKnown()
...
         {
-            foreach (IExtensionNode node in _serviceClass.Extensions)
-                if (node.TypeName == typeName)
-                {
-                    Assert.True(node.Enabled);
-                    return;
-                }
-            
-            Assert.Fail("Couldn't find known Extension {0}", typeName);
+            Assert.That(_serviceInterface.Extensions,
```
mikkelbu: Before we iterated over `_serviceClass.Extensions`. I don't know much about this area, so why this change?

mikkelbu: In all the refactorings you did in this PR you "translated" a foreach loop, e.g. `foreach (IExtensionNode node in _serviceInterface.Extensions)` into an assertion over the same collection `Assert.That(_serviceInterface.Extensions, ...`. 

But in this test the foreach loop was `foreach (IExtensionNode node in _serviceClass.Extensions)`, but this was translated into `Assert.That(_serviceInterface.Extensions, ...`, so I was wondering why the change from `_serviceClass.Extensions` to `_serviceInterface.Extensions`.

ChrisMaddock: https://github.com/nunit/nunit-console/pull/459

PR #459, opened 2018-08-21
Correct refactoring error

ChrisMaddock:
Correct a refactoring error @mikkelbu spotted in https://github.com/nunit/nunit-console/pull/454.

Thanks - good catch!


mikkelbu:
@ChrisMaddock Is it on purpose that this is merged into `ChrisMaddock-patch-3` and not `master`. It will probably end up on `master` eventually, but I just found it strange 😄.

ChrisMaddock:
Just ended up on this PR for unrelated reasons, and realised it never did get into master! 😱 Attempt no. 3 at: https://github.com/nunit/nunit-console/pull/518

PR #462, opened 2018-08-25
[wip] Replace Mono.Cecil

### LINKED ISSUE #389: Update Mono.Cecil to latest
Mono.Cecil 10.0 is no longer in preview. Update the .NET Standard build to use it. We cannot update the full framework engine because Mono.Cecil has dropped support for .NET 2.0.

jnm2:
Closes #389, nunit/nunit3-vs-adapter#366, and nunit/nunit3-vs-adapter#488.

Fully implemented and ready for review for .NET Framework. All tests are passing, but this seems almost too easy and I'd appreciate ideas for stress tests. For example, is there a way I could test the VSTest adapter against a PR build?

/cc @OsirisTerje because this will impact the VSTest adapter which currently bundles Mono.Cecil but only because the engine needed it.


msi/nunit/engine-files.wxi
```diff
@@ -32,10 +32,6 @@
                 <File Id="nunit_agent_x86.exe.config"
                       Source="$(var.InstallImage)bin/nunit-agent-x86.exe.config" />
             </Component>
-            <Component Id="MONO.CECIL" Location="local" Guid="D4A15752-6685-3494-9346-F623175CEFFA">
-                <File Id="Mono.Cecil.dll"
-                      Source="$(var.InstallImage)bin/Mono.Cecil.dll" />
-            </Component>
```
jnm2: I assume this is all I need to do here?

src/NUnitEngine/nunit.engine/Internal/MetadataFacts.cs
```diff
@@ -0,0 +1,53 @@
...
+            foreach (var reference in assembly.AssemblyReferences)
+            {
+                if (reference.Name == "mscorlib" && BitConverter.ToUInt64(reference.GetPublicKeyToken(), 0) == 0xac22333d05b89d96)
+                {
+                    // We assume 3.5, since that's all we are supporting
+                    // Could be extended to other versions if necessary
+                    // Format for FrameworkName is invented - it is not
+                    // known if any compilers supporting CF use the attribute
+                    return ".NETCompactFramework,Version=3.5";
+                }
```
jnm2: @ChrisMaddock Should this CF thing be removed?

src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
```diff
@@ -98,32 +97,20 @@ public void EnableExtension(string typeName, bool enabled)
...
-                if (ep.TypeName == type.FullName)
-                    return ep;
-
-            return null;
-        }
-
-        /// <summary>
-        /// Get an ExtensionPoint based on a Cecil TypeReference.
-        /// </summary>
-        public ExtensionPoint GetExtensionPoint(TypeReference type)
```
jnm2: Public API shape change for `nunit.engine`

src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
```diff
@@ -98,32 +97,20 @@ public void EnableExtension(string typeName, bool enabled)
...
         {
             return _pathIndex.ContainsKey(path) ? _pathIndex[path] : null;
         }
 
         /// <summary>
-        /// Get an ExtensionPoint based on the required Type for extensions.
+        /// Gets the extension point with the specified full type name.
         /// </summary>
-        public ExtensionPoint GetExtensionPoint(Type type)
+        public ExtensionPoint GetExtensionPointFromTypeName(string fullName)
```
jnm2: Public API shape change for `nunit.engine`
(`internal`? Do we care about public shape changes in this assembly?)

src/NUnitEngine/nunit.engine/Services/ExtensionAssembly.cs
```diff
@@ -36,43 +38,21 @@ public ExtensionAssembly(string filePath, bool fromWildCard)
...
-            }
-        }
-
-        public ModuleDefinition MainModule
+        internal IAssemblyMetadataProvider Metadata
         {
-            get { return Assembly.MainModule; }
-        }
-
-        public AssemblyNameDefinition AssemblyName
```
jnm2: Public API shape change for `nunit.engine`

src/NUnitEngine/nunit.engine/Services/ExtensionAssembly.cs
```diff
@@ -36,43 +38,21 @@ public ExtensionAssembly(string filePath, bool fromWildCard)
...
-
-                    _assemblyDefinition = AssemblyDefinition.ReadAssembly(FilePath, parameters);
-                }
+        private IAssemblyMetadataProvider _metadata;
 
-                return _assemblyDefinition;
-            }
-        }
-
-        public ModuleDefinition MainModule
```
jnm2: Public API shape change for `nunit.engine`

src/NUnitEngine/nunit.engine/Services/ExtensionAssembly.cs
```diff
@@ -36,43 +38,21 @@ public ExtensionAssembly(string filePath, bool fromWildCard)
         public string FilePath { get; private set; }
         public bool FromWildCard { get; private set; }
 
-        private AssemblyDefinition _assemblyDefinition;
-        public AssemblyDefinition Assembly
```
jnm2: Public API shape change for `nunit.engine`

bording:
FYI, this should close https://github.com/nunit/nunit3-vs-adapter/issues/366 too!

jnm2:
Removing Mono.Cecil from .NET Core could be a lot of work, so I want to do that separately. @nunit/engine-team I'll wait to hear from you.

@bording Thanks, I thought I missed one!

jnm2:
nunit/nunit3-vs-adapter#366
nunit/nunit3-vs-adapter#488

...interesting numbers there :D

bording:
> Removing Mono.Cecil from .NET Core could be a lot of work, so I want to do that separately. 

If you do it in a separate PR, then that PR would also need to be merged before you're really be able to consider those issues closed.

jnm2:
@bording I was thinking that everyone with an issue with Cecil only used the .NET Framework builds, but fair point.

I started taking the System.Reflection.Metadata path since I'm quite familiar with MetadataReader. Finishing the implementation is just a matter of time. The nice thing about this approach is that it's lightweight and it's built in to the .NET Core BCL. However, a .NET Standard target requires an additional package reference to System.Reflection.Metadata similar to the current dependency on Cecil.

@rprouse What do you think of multitargeting the engine as `net20;netcoreapp1.1`? We have so many hacks due to missing capabilities in netstandard1.3 and the question that arises in my mind is whether it's worth it: which platforms do we actually expect the engine to be operating on?

rprouse:
The VS adapter targets `netcoreapp1.0`, that is why I targeted `netstandard1.3`. I selected that target based on the target of the MSTest adapter. Before we up the version, we should probably switch the target of the adapter and test that it will still work.

bording:
> @bording I was thinking that everyone with an issue with Cecil only used the .NET Framework builds, but fair point.

I'm definitely hitting it on .NET Core as well. Since that version of the engine is referencing 0.10.0-beta6, the problem isn't as catastrophic, but it does still mean I'm not getting the version I'm referencing in my test projects.

>  However, a .NET Standard target requires an additional package reference to System.Reflection.Metadata much like the current dependency on Cecil (but I'd insist that we make it a package reference rather than bundling it).

As long as the package reference approach would actually work to let the engine properly resolve the dependency, it seems like that approach would be fine. Do we know for sure that the VS test adapter architecture would use the test project's dep.json file to be able to resolve it?

jnm2:
Given the above, I want to remove Cecil entirely.

@bording Thanks!

> As long as the package reference would actually work to let the engine properly resolve the dependency, it seems like that approach would be fine. Do we know for sure that the VS test adapter architecture would use the test project's dep.json file to be able to resolve it?

I don't know, good question.

@rprouse 

> I also have a big PR out merging the .NET Standard engine back into the main projects using multi-targeting. We're going to conflict, but feel free to move forward with your work, I will deal with it on my end 

I was looking for the PR last night to plan for .NET Core/Standard, actually hoping to make this easier for you and make sure my work was in the right direction. But I can't find it and got confused. Is it a branch?

> The VS adapter targets `netcoreapp1.0`, that is why I targeted `netstandard1.3`. I selected that target based on the target of the MSTest adapter. Before we up the version, we should probably switch the target of the adapter and test that it will still work.

I was thinking the adapter was 1.1 and that 1.0 was out of support, but neither of those is true so never mind. Can we do `net20;netcoreapp1.0` rather than `net20;netstandard1.3`? That way System.Reflection.Metadata is in the BCL rather than requiring a separate package reference, among other capabilities that .NET Standard doesn't have.

jnm2:
Rebased. @ChrisMaddock's PR made some of the exact same changes that this one did and answered some of my questions, so that's nice! 😆

jnm2:
As far as I know, there's nothing else for me to do in this PR unless we decide not to have a .NET Standard 1.x build that does assembly-parsing file I/O.
I'm doing stuff just in case it's useful, so don't feel bad turning down the PR. While I wait to know what happens next here, I'll try to learn if it's actually possible to load two versions of Mono.Cecil in the same process on .NET Core 1.0, and if not, if it's possible to ILMerge Mono.Cecil into nunit.engine.

src/NUnitEngine/nunit.engine/Internal/MetadataFacts.cs
ChrisMaddock: Not too sure what this is replacing at this level. The only CF functionality I left in, is the stuff that identifies an assembly which targets CF - in order to return an appropriate error message about that platform not being supported by the console. Would you mind following up the stack, to see where this is called?

I'm away at the moment, but will try and have a proper look at this at some point!

CharliePoole: When the engine tries to identify target platforms, it relies first of all on the target framework attribute. But the `TargetFrameworkAttribute` was only introduced in (relatively) recent compilers and NUnit has to respond sensibly even if the attribute is not present.

For CF, which was last supported by VS2008, there is no attribute. So I added this code to detect a CF assembly and return a fake value, using a format similar to what is used in the attribute itself. The return is supposed to be used precisely as you suggest: to allow us to give an intelligent error message.

jnm2:
As long as the VSTest host is able to resolve all direct assembly references as it reflects over the types in the adapter assembly searching for types implementing extensions, .NET Core is good.

jnm2:
Hey, guess what? I fixed it this morning. Only proved it now that I've finished writing package acceptance tests for CI. See you over at https://github.com/nunit/nunit3-vs-adapter/pull/541!

PR #479, opened 2018-10-07
Merge .NET Standard Engine code back into the main solution

rprouse:
This is the work @ChrisMaddock and I have been doing on merging the .NET Standard version of the engine back into the main solution. I am putting the pull request up because we are getting close to done and wanted to involve the team in the final steps.

I think that the tasks that we want to complete before a merge are;

- [x] Add the .NET Standard engine and engine.api into the NuGet packages
- [x] Test this version of the engine in the adapter in Visual Studio and with `dotnet test` on the command line.

Once this is complete, we can start exploring the other enhancements that @ChrisMaddock has entered recently like allowing the console runner to run Core tests, adding a Core console runner, allowing the engine to load Core extensions, adding back features to the .NET Standard engine like communicating to an agent, etc.


ChrisMaddock:
Thanks for putting this up Rob. (And fixing my NuGet packaging. 😩)

Never asked before - what was your thinking on adding a .NET Standard 1.3 engine - was it Xamarin support? I ask, as I'm not sure the dependencies are available on the Xamarin platforms (XmlDocument appears not to be available on iOS/Android - although I'd have to dig more deeply to check!) If we're only intending to support .NET Core with this engine however, then I'm not sure why you dropped the target level from 1.6?

rprouse:
I've tested this build in the adapter and everything seems to be working fine in my test projects, so I think we are pretty close to being able to merge this.

> what was your thinking on adding a .NET Standard 1.3 engine

To be honest, this has been on the back burner so long that I forget why I made that change 😦 The adapter needs to be compiled .NET Core 1.0 which is fine for 1.6 and the latest versions of Xamarin support it. It might have been for the original Windows 10 UWP, but even that could go to 1.4. We should probably bring it up to latest we can manage. Do you think that should be 1.4 for UWP support, or 1.6 for more features?

ChrisMaddock:
Hmm, we don’t have a runner for original UWP, but of course that would allow others to consume the engine.

I wonder if we should stick at 1.6 for now, until that demand is expressed? UWP is a very small part of our userbase - so much so, I’d be tempted to wonder if the portable agent could be sufficient.

rprouse:
I updated to .NET Standard 1.6 and the only new functionality that it enabled was getting the command line arguments. The two other big areas that I'd hoped to enable, extensions and running tests async, were still not possible because `Assembly.LoadFrom` and `Thread` are still not available in 1.6. We may as well leave these at 1.3 in the hope of adding full UWP support to the adapter.

I'm pushing some minor fixes that I found while going through trying to enable features for 1.6.

I think this is ready for review.

src/NUnitEngine/nunit.engine/Extensibility/ExtensionNode.cs
```diff
@@ -128,7 +134,17 @@ public object ExtensionObject
...
         public object CreateExtensionObject(params object[] args)
         {
+#if NETSTANDARD2_0
+            var assembly = Assembly.LoadFrom(AssemblyPath);
+            var typeinfo = assembly.DefinedTypes.FirstOrDefault(t => t.FullName == TypeName);
+            if (typeinfo == null)
+            {
+                return null;
+            }
+            return Activator.CreateInstance(typeinfo.AsType(), args);
```
ChrisMaddock: Ref comment on ExtensionService - I think we might want to try/catch from here. (Also, I don't think this method is safe to return null - I think perhaps we should throw instead?)

src/NUnitEngine/nunit.engine/Internal/Logging/Logger.cs
```diff
@@ -165,9 +170,12 @@ private void WriteLog(InternalTraceLevel level, string message)
             writer.WriteLine(TRACE_FMT,
                 DateTime.Now.ToString(TIME_FMT),
                 level == InternalTraceLevel.Verbose ? "Debug" : level.ToString(),
-                System.Threading.Thread.CurrentThread.ManagedThreadId, name, message);
+#if !NETSTANDARD1_3
+                System.Threading.Thread.CurrentThread.ManagedThreadId,
```
ChrisMaddock: We could use `Environment.CurrentManagedThreadId` for all .NET Standard platforms. Note TRACE_FMT above needs updating in sync

src/NUnitEngine/nunit.engine/Internal/RuntimeFrameworks/DotNetFrameworkLocator.cs
```diff
@@ -1,5 +1,5 @@
 // ***********************************************************************
-// Copyright (c) 2007 Charlie Poole, Rob Prouse
+// Copyright (c) 2018 Charlie Poole, Rob Prouse
```
ChrisMaddock: Any particular reason for this change? If the thinking was this being a 'new' file - I originally kept the 2007 date as I'd only moved code here, and not written anything new.

I don't know the specifics on what we _should_ be doing in this case, however! :smile:

src/NUnitEngine/nunit.engine/Internal/AssemblyHelper.cs
```diff
@@ -21,6 +21,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
+#if !NETSTANDARD1_3
 using System;
 using System.IO;
 using System.Reflection;
```
ChrisMaddock: There's a TODO in this file in terms of making `GetAssemblyPathFromCodeBase` xplat. Probably can be a future issue.

src/NUnitEngine/nunit.engine/Internal/TargetFrameworkHelper.cs
```diff
@@ -92,17 +92,6 @@ public string FrameworkName
...
-                    if (reference.Name == "mscorlib" &&
-                        BitConverter.ToUInt64(reference.PublicKeyToken, 0) == 0xac22333d05b89d96)
-                    {
-                        // We assume 3.5, since that's all we are supporting
-                        // Could be extended to other versions if necessary
-                        // Format for FrameworkName is invented - it is not
-                        // known if any compilers supporting CF use the attribute
-                        return ".NETCompactFramework,Version=3.5";
-                    }
-
```
ChrisMaddock: We'd previously kept this code, for the purpose of providing a sensible error message if someone tried to run .NET CF 3.5 tests in the console. I'm not strongly against removing it, but it doesn't seem to be causing problems to keep it.

src/NUnitEngine/nunit.engine/Runners/DirectTestRunner.cs
```diff
@@ -104,32 +105,40 @@ protected override TestEngineResult LoadPackage()
...
+#endif
                 driver.ID = TestPackage.ID;
                 result.Add(LoadDriver(driver, testFile, subPackage));
                 _drivers.Add(driver);
             }
-
             return result;
         }
 
+        private IDriverService GetDriverService()
```
ChrisMaddock: I'm guessing there's been some complexity removed here, and it might make sense to re-inline this 1-line method.

src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
```diff
@@ -38,8 +38,12 @@ public class MasterTestRunner : ITestRunner
         private const string TEST_RUN_ELEMENT = "test-run";
         private readonly ITestEngineRunner _engineRunner;
         private readonly IServiceLocator _services;
+#if NET20
```
ChrisMaddock: This is the only file which does `#if NET20` rather than `#if !NET_STANDARD`. We should switch and be consistent. (My bad, this one!)

src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
```diff
@@ -53,8 +57,12 @@ public MasterTestRunner(IServiceLocator services, TestPackage package)
 
             // Get references to the services we use
             _projectService = _services.GetService<IProjectService>();
+#if NET20
```
ChrisMaddock: As above.

src/NUnitEngine/nunit.engine/Services/ResultWriters/XmlTransformResultWriter.cs
```diff
@@ -21,6 +21,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
+#if !NETSTANDARD1_3 && !NETSTANDARD2_0
```
ChrisMaddock: I think this can be enabled in .NET Standard 2.0 with no changes?

src/NUnitEngine/nunit.engine/Services/ResultService.cs
```diff
@@ -27,17 +27,14 @@
...
-    public class ResultService :
-#if !NETSTANDARD1_3
-        Service, 
-#endif
-        IResultService
+    public class ResultService : Service,  IResultService
     {
-#if NETSTANDARD1_3
+#if NETSTANDARD1_3 || NETSTANDARD2_0
         private readonly string[] BUILT_IN_FORMATS = new string[] { "nunit3", "cases" };
```
ChrisMaddock: If we enable XmlTransforms for .NET Standard 2.0, we need to permit 'user' here as well.

src/NUnitEngine/nunit.engine/AsyncTestEngineResult.cs
```diff
@@ -31,7 +31,9 @@ namespace NUnit.Engine
     /// <summary>
     /// The TestRun class encapsulates an ongoing test run.
     /// </summary>
+#if !NETSTANDARD1_3 && !NETSTANDARD2_0
     [Serializable]
+#endif
```
ChrisMaddock: Any reason not to be serializable in NETSTANDARD2_0

src/NUnitEngine/nunit.engine/RuntimeFramework.cs
```diff
@@ -686,5 +687,82 @@ private static void AddMonoFramework(Version frameworkVersion, string profile)
         }
 
         #endregion
+
+        #region Helper Methods - .NET
+
```
ChrisMaddock: These methods moved to DotNetFrameworkLocator - is putting them back maybe a merge error?

src/CommonAssemblyInfo.cs
```diff
@@ -31,26 +31,28 @@
...
-[assembly: AssemblyConfiguration(".NET 4.0")]
-#elif NET_2_0
+#if NET35
+[assembly: AssemblyConfiguration(".NET 3.5")]
+#elif NET20
 [assembly: AssemblyConfiguration(".NET 2.0")]
-#elif NETSTANDARD1_3 || NETSTANDARD1_6 || NETCOREAPP1_0
-[assembly: AssemblyConfiguration(".NET Standard")]
+#elif PORTABLE
+[assembly: AssemblyConfiguration("Portable")]
```
ChrisMaddock: Is adding back PORTABLE an error?

src/NUnitEngine/nunit.engine.tests/nunit.engine.tests.csproj
```diff
@@ -2,7 +2,9 @@
   <PropertyGroup>
     <RootNamespace>NUnit.Engine.Tests</RootNamespace>
     <AssemblyName>nunit.engine.tests</AssemblyName>
-    <TargetFrameworks>net35</TargetFrameworks>
+    <TargetFrameworks>net35;netcoreapp1.1;netcoreapp2.0</TargetFrameworks>
+    <OutputType Condition="'$(TargetFramework)'!='net35'">Exe</OutputType>
+    <!--<OutputType>Exe</OutputType>-->
```
ChrisMaddock: Remove comment, presumably!

src/NUnitEngine/nunit.engine.tests/Internal/PathUtilTests.cs
```diff
@@ -86,7 +86,7 @@ public void Canonicalize()
 				PathUtils.Canonicalize( @"folder1\folder2\..\..\..\file.tmp" ) );
 		}
 
-#if !NETCOREAPP1_1
+#if !NETCOREAPP1_1 && !NETCOREAPP2_0
```
ChrisMaddock: I suggest we `#if` out the actual method inline with the test being excluded here. Nothing actually uses this method in the .NET Standard builds.

ChrisMaddock:
> We may as well leave these at 1.3 in the hope of adding full UWP support to the adapter.

This sounds like a sensible enough justification to me! 👍

ChrisMaddock:
I'm hoping Travis on macOS is just timing out on a build step. I threw together an Azure DevOps just to see if it would 'just work' there - but it's going to require repo changes and such, so I think that should be work for a separate PR.

I've disabled the job on VSTS, so hopefully that won't show on any other CI runs/PRs, until I come back to it.

ChrisMaddock:
Wow...asking macOS Travis to wait for 30 minutes got it past the section it was 'stuck on', which suggests it is just a timeout issue, but still didn't get it to the end of package restore. 😱 

I suggest we disable the macOS .NET Core CI temporarily, and try out Azure DevOps as a separate issue. I can revist at some point.

src/NUnitEngine/nunit.engine.api/TestEngineActivator.cs
```diff
@@ -39,6 +39,19 @@ public static class TestEngineActivator
         private const string DefaultAssemblyName = "nunit.engine.dll";
         internal const string DefaultTypeName = "NUnit.Engine.TestEngine";
 
+#if NETSTANDARD1_3 || NETSTANDARD2_0
+        /// <summary>
+        /// Create an instance of the test engine.
+        /// </summary>
+        /// <returns>An <see cref="NUnit.Engine.ITestEngine"/></returns>
+        public static ITestEngine CreateInstance()
+        {
```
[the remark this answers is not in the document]
ChrisMaddock: Ah ok - we can't actually do anything useful with the `minVersion`. Will leave this as is. 👍

src/NUnitEngine/nunit.engine.tests/Internal/PathUtilTests.cs
ChrisMaddock: Ahh, it's the PlatormAttribute that's the issue. 

I'll try upgrading .NET Core 2.0 to 2.1 again. Last time I did that it caused a problem somewhere along the line, but I can't quite remember what.

ChrisMaddock:
Patched up a few of the minor things I spotted, and disabled iOS CI for now. 😞

jnm2:
@rprouse @ChrisMaddock 

> > We may as well leave these at 1.3 in the hope of adding full UWP support to the adapter.
> 
> This sounds like a sensible enough justification to me! 👍

The oldest UWP works with .NET Standard 1.4. Do you see any benefit to one versus the other? Just curious, no need to change anything.

src/NUnitEngine/nunit.engine/Drivers/NUnitNetStandardDriver.cs
```diff
@@ -21,6 +21,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
+#if NETSTANDARD
```
jnm2: Do you think we can get this similar enough to NUnit3FrameworkDriver to use the same name?

src/NUnitEngine/nunit.engine/Internal/Logging/Logger.cs
```diff
@@ -165,9 +159,14 @@ private void WriteLog(InternalTraceLevel level, string message)
             writer.WriteLine(TRACE_FMT,
                 DateTime.Now.ToString(TIME_FMT),
                 level == InternalTraceLevel.Verbose ? "Debug" : level.ToString(),
-                System.Threading.Thread.CurrentThread.ManagedThreadId, name, message);
+#if NET20
+                System.Threading.Thread.CurrentThread.ManagedThreadId,
+#elif NETSTANDARD
+                Environment.CurrentManagedThreadId,
+#endif
```
jnm2: Since `Environment.CurrentManagedThreadId` is preferred, can we special-case NET20 as the odd one out and just use `#else`?

src/NUnitEngine/nunit.engine/Services/ResultService.cs
```diff
@@ -27,17 +27,14 @@
 
 namespace NUnit.Engine.Services
 {
-    public class ResultService :
-#if !NETSTANDARD1_3
-        Service, 
-#endif
-        IResultService
+    public class ResultService : Service,  IResultService
```
jnm2: Double space

src/NUnitEngine/nunit.engine/Services/ServiceManager.cs
```diff
@@ -53,7 +54,11 @@ public IService GetService( Type serviceType )
             else
                 foreach( IService service in _services )
                 {
+#if NETSTANDARD1_3
+                    if (serviceType.GetTypeInfo().IsAssignableFrom(service.GetType().GetTypeInfo()))
+#else
                     if( serviceType.IsInstanceOfType( service ) )
+#endif
```
jnm2: This would enable unifying the code:

https://github.com/nunit/nunit/blob/6af1a1536e2ad86d3974b1167cb3c984b0c23f83/src/NUnitFramework/framework/nunit.framework.csproj#L22-L24

src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
```diff
@@ -434,6 +447,7 @@ internal void FindExtensionsInAssembly(ExtensionAssembly assembly)
                     return;
                 }
             }
+#endif
```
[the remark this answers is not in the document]
rprouse: I added a unit test for this where the .NET Core 2.0 targeted tests try to load extensions from the .NET 3.5 assembly and visa-versa. Interestingly, .NET 3.5 correctly throws an `NUnitEngineException` because it thinks the .NET Core 2.0 targets the 4.0 full framework. The .NET Core 2.0 seems to successfully load the .NET 3.5 extensions. Both of these behaviors seem wrong to me.

rprouse:
I am working through the comments and updating code where appropriate. Meanwhile,

> The oldest UWP works with .NET Standard 1.4. Do you see any benefit to one versus the other? Just curious, no need to change anything.

I looked at the [diff for .NET Standard 1.4 vs 1.3](https://github.com/dotnet/standard/blob/master/docs/versions/netstandard1.4_diff.md) and it only adds a couple of crypto related classes so gains us nothing for the work of updating...

src/NUnitEngine/nunit.engine/Drivers/NUnitNetStandardDriver.cs
rprouse: At this point, the two drivers are very different.

jnm2: Okay. Just checking. If we ship like this now, we're committing to keeping a driver named `NUnitNetStandardDriver` for folks using the package from .NET Standard, so if merging the two is on the roadmap, now's the best time.

rprouse:
@ChrisMaddock I'll push my changes based on feedback. The only thing I am not 100% happy with are my changes to the engine service. I'll tag those...

rprouse:
> nunit/docs#289 has just led me to this code as well, which will be broken by our repackaging of the engine. That's if it works currently - I'm not sure if any of our actual runners make use of this.

I am going to remove this code. As the comments read, it is only used when developing a runner, but that runner can simply add the NuGet package and take a reference.

src/NUnitEngine/nunit.engine.api/TestEngineActivator.cs
```diff
@@ -39,6 +39,19 @@ public static class TestEngineActivator
...
+            var assemblyName = new AssemblyName(DefaultAssemblyName);
+            var assembly = Assembly.Load(assemblyName);
```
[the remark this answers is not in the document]
ChrisMaddock: Update - this weekend was unexpectedly sunny, so no code was written. Sorry! 🌞 

image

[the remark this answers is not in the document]
rprouse: I came across this project that works around issues with assembly loading, https://natemcmaster.com/blog/2018/07/25/netcore-plugins/ with a GitHub repo at https://github.com/natemcmaster/DotNetCorePlugins. The code is under Apache license, so we can't use it directly, but maybe if we ask nicely, the project owner might allow us to use some of his ideas in our code under MIT? I've only had a cursory look, but if it pans out, I will reach out to him.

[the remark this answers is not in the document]
ChrisMaddock: Gah, @rprouse - that assembly loading blog/logic is all dependent on being [AssemblyLoadContext](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.loader.assemblyloadcontext?view=netcore-2.1#applies-to) - which .NET Core rather than .NET Standard. 

I'm not sure how we can achieve this targetting .NET Standard right now. Still digging - appreciate any thoughts. 😞

ChrisMaddock:
Latest commit adds XSLT functionality to .NET Standard 2.0, and fixes the load from outside of build dir issue highlighted above. 

I haven't been able to fix this issue in .NET Standard 1.3, as I'm not sure how to find the location of nunit.engine.api.dll. (Available via the `Assembly` type in .NET Standard 2.0.) I've added the tests in `NUnit.Engine.Tests.WorkingDirectoryTests` to demonstrate the issue (hence the failing CI).

ChrisMaddock:
.NET Standard 1.5 adds the required APIs. This is a problem we wouldn't have had for the previous .NET Standard engine, as the runner had to directly reference the engine, without the api assembly.

We could instead retarget 1.3 -> 1.6. This would be suitable for .NET Core 1.1, and still support UWP 10.0.16299. We'd lose potential support for UWP 10.0, although that's something which _would_ have been a new feature anyway.

What do you think, Rob?

src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
```diff
@@ -537,7 +517,37 @@ internal void FindExtensionsInAssembly(ExtensionAssembly assembly)
...
+            else if (runnerHelper.FrameworkName?.StartsWith(".NETCoreApp") == true)
+            {
+                if (extHelper.FrameworkName?.StartsWith(".NETStandard") != true && extHelper.FrameworkName?.StartsWith(".NETCoreApp") != true)
+                {
+                    throw new NUnitEngineException(".NET Core runners require .NET Core or .NET Standard extensions");
+                }
+            }
+            else if (extHelper.FrameworkName?.StartsWith(".NETCoreApp") == true || extHelper.FrameworkName?.StartsWith(".NETStandard") == true)
+            {
+                throw new NUnitEngineException(".NET Framework runners cannot load .NET Core extensions");
```
ChrisMaddock: This exception should maybe say .NET Core/.NET Standard extensions. I'd expect .NET Core extensions not to be loadable, it's .NET Standard extensions that are the 'unusual case' here in my opinion. 🙂

ChrisMaddock:
Todo's on this PR, to my mind:

- [ ] Decision on .NET Standard 1.3 support
- [ ] Fix CI

In subsequent PRs
- [ ] Add MacOS CI back, possible on Azure DevOps

Anything I'm missing? 🙂

ChrisMaddock:
Actually, I don't think there's anything to fix in CI, I forgot I broke it on purpose for https://github.com/nunit/nunit-console/pull/479#issuecomment-433178999 🙈 

@rprouse - what do you think upping the .NET Standard from 1.3 to 1.5. We lose potential support for UWP 10.0, but I'm not sure if we can reliably load the engine via reflection in .NET Standard 1.3, which is a must if we're using the API assembly.

rprouse:
Sorry, I've been really out of touch lately and will be away all weekend. As long as 1.5 works with the adapter I am fine with going with that. As you say, we don't support UWP 10.0 now anyway.

ChrisMaddock:
Great, thanks Rob - and no worries!

rprouse:
@ChrisMaddock I've resolved the merge conflicts, updated to .NET Standard 1.6 and got the load engine from another directory test passing. See the changes in `TestEngineActivator.CreateInstance()` for .NET Standard 1.6.

ChrisMaddock:
Awesome! Sorry, I haven't had time to get back to this recently! 

I started a PR here to update the adapter to the _current_ engine release, so that we can hopefully easily switch to this version once it's done. Unfortunately that seems to be non-trivial, but it's a WIP!

rprouse:
Looks like my recent changes or maybe the merge from master broke packaging in Cake. I am fixing it. @ChrisMaddock anything else we should do before merging this PR and moving on to the additional tasks? I started this branch back in May, so it would be nice to get it behind us 😄

rprouse:
Packaging has been fixed, now the new load engine from another directory is failing on Travis Linux on .NET 2.0

```
1) Failed : NUnit.Engine.Services.Tests.ExtensionServiceTests.FailsGracefullyLoadingOtherFrameworkExtensionAssembly
  Expected: <NUnit.Engine.NUnitEngineException>
  But was:  no exception thrown
  at NUnit.Engine.Services.Tests.ExtensionServiceTests.FailsGracefullyLoadingOtherFrameworkExtensionAssembly () [0x00000] in <839ae0be2830488ba956d1dfeb047f37>:0 
```

ChrisMaddock:
Yes - let’s get this merged! I’m pretty busy at the moment, but will try and give it a final look over at some point, unless you’re happy that the latest changes are simple?

Also - the failing test you’ve posted looks like a different test from the newly added one? Can’t see the appveyor logs on my phone myself.

rprouse:
> the failing test you’ve posted looks like a different test from the newly added one? Can’t see the appveyor logs on my phone myself.

You're right, I misread that. I didn't get much of a chance to look at it last night. I ended up struggling with getting my Linux environment instead. I'll probably be able to get back to it tomorrow.

The latest changes I made are fairly small. You can concentrate a review on the new .NET Standard 1.6 version of `TestEngineActivator.CreateInstance`. That was the only change of substance.

rprouse:
Further to that, we might be able to use the NETSTANDARD1_6 version of `CreateInstance` for the NETSTANDARD2_0 target. I was just playing it safe.

ChrisMaddock:
Assigned myself so I could apply PR 'code suggestions' (which were all trivial) - that was a super sweet experience! 😄

nuget/engine/nunit.engine.nuspec
```diff
@@ -19,7 +19,7 @@
     <copyright>Copyright (c) 2018 Charlie Poole, Rob Prouse</copyright>
     <dependencies>
       <group targetFramework="net20" />
-      <group targetFramework="netstandard1.3">
+      <group targetFramework="netstandard1.6">
```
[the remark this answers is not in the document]
ChrisMaddock: Update - looks like it can be removed, commit incoming. 😄

ChrisMaddock:
Last commit fixes the two outstanding issues - so to my mind, this is now ready to be merged! 😱 @rprouse - I'll leave it with you to merge when you're happy!

Last things I patched up:
1. `System.Reflection.TypeExtension` dependency needed under NS1.3 seemed unnecessary under NS1.6, so I removed it
2. The failing test was due to `Assembly.GetEntryAssembly()` returning null on mono, and thus the framework loading validation not being run. I just added the line below for now - although I assume trying to load a Net Core extension under mono would now lead to an noisier failure later in the process. Maybe something we should log for future...

```
//May be null on mono
Assume.That(Assembly.GetEntryAssembly(), Is.Not.Null, "Entry assembly is null, framework loading validation will be skipped.");
```

mikkelbu:
We also had some unexpected behaviour regarding `Assembly.GetEntryAssembly()` in https://github.com/nunit/nunit/issues/3084#issuecomment-436037830.

rprouse:
Looks good, I was just looking at making the same fixes tonight 😄 

If anyone is using it, I updated the docker image that `build-mono-docker.ps1` uses to the latest release of .NET Core 2.6.1. I was using that for testing locally.

PR #484, opened 2018-10-13
Add datetime of run to header

### LINKED ISSUE #178: Add date and time to console output
While working on NUnit, I repeatedly run the console tests. Sometimes, especially if I need to interrupt my work for a few minutes, I can't remember if the last run displayed uses the most recent compile or not. If the header display had a third line with the date and time, it would be easier to tell.

The output would use the local time and culture for display and could look like this...

```
NUnit Console Runner 3.7.0
Copyright (C) 2017 Charlie Poole
February 11, 2017 5:26PM
```

If we agree that this change should be made, I'll duplicate it for NUnitLite as well.

@nunit/core-team @nunit/engine-team @nunit/framework-team What do you think?

abborg:
Fixes #178

Adds localized time to the console runner header. The format of the datetime is using the current culture's full datetime pattern.

### DIFF
--- a/src/NUnitConsole/nunit3-console/Program.cs
+++ b/src/NUnitConsole/nunit3-console/Program.cs
@@ -23,6 +23,7 @@
 
 using System;
 using System.Diagnostics;
+using System.Globalization;
 using System.IO;
 using System.Reflection;
 using System.Text;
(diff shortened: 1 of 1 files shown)


CharliePoole:
Looks good to me as well. Thanks for your contribution!

PR #486, opened 2018-10-16
Format error message when invalidating integer-valued arguments

### LINKED ISSUE #485: Invalid integer arguments do not display properly in error message
When an invalid integer-valued arguments is passed to the nunit console, the error message does not display the option or value that is invalid. The code suggests the error message should be formatted  to contain that information. When validating string values, a similar error message is formatted to contain the invalid option and value. Example:
```
PS > .\bin\Debug\net35\nunit3-console.exe .\bin\Debug\net35\nunit3-console.tests.dll --workers j -noh
An int value was expected for option '{0}' but a value of '{1}' was used
PS > .\bin\Debug\net35\nunit3-console.exe .\bin\Debug\net35\nunit3-console.tests.dll --trace j -noh
The value 'j' is not valid for option '--trace'.
```

From CommandLineOptions.cs:

```
try
{
    result = int.Parse(val);
}
catch (Exception)
{
     ErrorMessages.Add("An int value was expected for option '{0}' but a value of '{1}' was used");
}
```

This should be updated to format the error message for integer-valued arguments so that the error contains the option name and value supplied.

abborg:
Fixes #485 

Provide more information in error message when invalid values are passed into integer-valued arguments.

```
PS > .\bin\Debug\net35\nunit3-console.exe .\bin\Debug\net35\nunit3-console.tests.dll --workers j -noh
An int value was expected for option '--workers' but a value of 'j' was used
```

### DIFF
--- a/src/NUnitConsole/nunit3-console/CommandLineOptions.cs
+++ b/src/NUnitConsole/nunit3-console/CommandLineOptions.cs
@@ -235,7 +235,7 @@ namespace NUnit.Common
                 }
                 catch (Exception)
                 {
-                    ErrorMessages.Add("An int value was expected for option '{0}' but a value of '{1}' was used");
+                    ErrorMessages.Add(String.Format("An int value was expected for option '{0}' but a value of '{1}' was used", option, val));
                 }
             }
 


ChrisMaddock:
Whoops! Thanks for the fix!

PR #489, opened 2018-10-19
Append user extensions

### LINKED ISSUE #488: Simplify locating of addins
Currently there is only one way how user is able to say where `nunit-console` should find extensions. This is via `*.addins` file. What if user will be able to install extension via nuget into tests project and that's all what is required from user.

I propose to extend nunit-console and try to search extensions in the directory where user's tests dll is located. It would be default behavior.

nvborisenko:
Trying to resolve #488 

I understand that this implementation is not good, hoping you will help. This solution works for me, but tests are failing.

### DIFF
--- a/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
+++ b/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
@@ -231,7 +231,7 @@ namespace NUnit.Engine.Runners
 
             var processModel = TestPackage.GetSetting(EnginePackageSettings.ProcessModel, "").ToLower();
 
-            if (IntPtr.Size == 8 && (processModel == "inprocess" || processModel == "single")  &&
+            if (IntPtr.Size == 8 && (processModel == "inprocess" || processModel == "single") &&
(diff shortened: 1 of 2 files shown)


ChrisMaddock:
Thanks for starting this one off @nvborisenko. I think we need to do a little more design work on this one first, before we can start writing code. I'll comment back on the issue with my thoughts there.

src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
```diff
@@ -351,6 +351,9 @@ private TestEngineResult RunTests(ITestEventListener listener, TestFilter filter
             var eventDispatcher = new TestEventDispatcher();
             if (listener != null)
                 eventDispatcher.Listeners.Add(listener);
+
+            FindExtensionsInTestPackage(this.TestPackage);
+
```
CharliePoole: I'm not sure we want to do this, but if we did I believe we would want the engine to do it by itself, without requiring the console to do quite so much work. If we make it a console option, then we would just pass a new setting in the test package itself.

nvborisenko:
Sure. Created a PR to force a discussion :)

PR #493, opened 2018-11-02
Correct order of params to Guard.ArgumentValid()

jwg4:
The arguments are in order: condition, error message, name of parameter. See https://github.com/nunit/nunit-console/blob/master/src/NUnitEngine/nunit.engine/Guard.cs#L76


ChrisMaddock:
Good spot - thanks for the fix!

PR #497, opened 2018-11-14
Add to path environment variable tip for console msi install

goroyal:


CharliePoole:
Are you suggesting that the program be added toothed path? Or do you have some other environment variable in mind?

goroyal:
> Are you suggesting that the program be added to the path? Or do you have some other environment variable in mind?

Yes, I am suggesting that the program be added to the path.

README.md
```diff
@@ -20,7 +20,7 @@ The latest stable release of the NUnit Console is [available on NuGet](https://w
 
 The Console/Engine are available in various packages:
 - [NUnit.ConsoleRunner](https://www.nuget.org/packages/NUnit.ConsoleRunner/): The NUnit Console, with no extensions.
-- [NUnit.Console](https://www.nuget.org/packages/NUnit.Console/): The NUnit Console, with key extensions additionally packaged. Also available as an [msi installer](https://github.com/nunit/nunit-console/releases).
+- [NUnit.Console](https://www.nuget.org/packages/NUnit.Console/): The NUnit Console, with key extensions additionally packaged. Also available as an [msi installer](https://github.com/nunit/nunit-console/releases), remember to add its install location to your environment variable after installing, such as `C:\Program Files (x86)\NUnit.org\nunit-console`.
```
CharliePoole: I think the original is trying to say to much in a single bullet-point and adding the tip demonstrates that. Nuget packages don't want to have anything added to the path, clearly, only the msi.

What if we change line 21 to say "various NuGet packages"? Then we could put in a new paragraph for the msi, like "An msi installer is also available if you want to install the runner in a central location. You may want to add the location to the path after installing..."

Should we also deal with the fact that the chocolatey packages have different names from the nuget packages?

goroyal: OK, I got it. Thanks for your reply

goroyal: So, I close this PR?

CharliePoole: Only if you don't want to improve it. For me, it should at least say which environment variable to change if you install the msi. The other changes would improve it IMO but aren't essential.

To update a PR you just push further changes to your branch on which it is based.

goroyal: Hi CharliePoole, I have updated the README, can you please take some time to review it?

Thanks.

mikkelbu:
I was thinking the same thing. I think it is quite normal for an installer to add the directory of the executable to the PATH.

PR #499, opened 2018-11-19
Add event handler to reset console colors on cancel keypress

### LINKED ISSUE #498: Reset console colors after Ctrl-C
When running nunit3-console.exe (version 3.9) the output in the console window gets colored.
On aborting the program with Ctrl-C the color of the console are not getting resetted.

Context:
I'm using a [solarized](https://ethanschoonover.com/solarized/) console window (MS Windows, cmd). After Ctrl-C the foreground color stays white on solarized dark and black on solarized light.

Proposal:
Use [CancelKeyPressed](https://docs.microsoft.com/en-us/dotnet/api/system.console.cancelkeypress?view=netframework-4.7.2) to catch this event and reset to original colors (`Console.ResetColor()`).

abborg:
Fixes #498. Use `Console.ResetColor()` to prevent the NUnit console runner from changing console colors when canceled.

### DIFF
--- a/src/NUnitConsole/nunit3-console/Program.cs
+++ b/src/NUnitConsole/nunit3-console/Program.cs
@@ -56,6 +56,8 @@ namespace NUnit.ConsoleRunner
         [STAThread]
         public static int Main(string[] args)
         {
+            Console.CancelKeyPress += new ConsoleCancelEventHandler(CancelHandler);
+
             try
             {
                 Options.Parse(Options.PreParse(args));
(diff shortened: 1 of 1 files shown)


mikkelbu:
The finally block is only run if if the app is still running :) (see e.g. https://stackoverflow.com/a/3216078)

mikkelbu:
Seems to work (at least in my simple tests 😄 ). Thanks for the contribution @abborg

PR #510, opened 2018-12-07
Remove unused TempResourceFile.cs

### LINKED ISSUE #8: TempResourceFile.Dispose causes run to hang
@CharliePoole commented on [Tue Sep 29 2015](https://github.com/nunit/nunit/issues/848)

This Dispose method contains an infinite loop, exited via a break statement. Sometimes, when all other threads are also waiting, the exit requirements are never met.

Someone needs to examine the reasons behind the logic that is there and modify it so it doesn't hang.

---

@CharliePoole commented on [Tue Sep 29 2015](https://github.com/nunit/nunit/issues/848#issuecomment-144281039)

I've noticed that this is often a problem when running a non-parallelizable STA test. Makes sense.

mikkelbu:
Fixes #8


mikkelbu:
No problem 😄.

PR #512, opened 2018-12-09
Support excluding tests in a testlist file

lanfeust69:
A use-case for this feature is for CI pipeline, when using results of previous
run to separate long running tests in their own process : we want to have a
reliable (and convenient) way to say "everything else".

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/Services/TestFilterBuilderTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Services/TestFilterBuilderTests.cs
@@ -77,6 +77,40 @@ namespace NUnit.Engine.Services.Tests
                 "<filter><or><test>My.First.Test</test><test>My.Second.Test</test><test>My.Third.Test</test></or></filter>"));
         }
 
+        [Theory]
+        public void OneTestExcluded(bool useBangSyntax)
+        {
+            builder.AddTest((useBangSyntax ? "!" : "-") + "My.Test.Name");
+            TestFilter filter = builder.GetFilter();
+
+            Assert.That(filter.Text, Is.EqualTo(
+                "<filter><not><test>My.Test.Name</test></not></filter>"));
+        }
+
+        [Theory]
+        public void ThreeTestsExcluded(bool useBangSyntax)
+        {
+            builder.AddTest((useBangSyntax ? "!" : "-") + "My.First.Test");
+            builder.AddTest((useBangSyntax ? "!" : "-") + "My.Second.Test");
+            builder.AddTest((useBangSyntax ? "!" : "-") + "My.Third.Test");
+            TestFilter filter = builder.GetFilter();
+
+            Assert.That(filter.Text, Is.EqualTo(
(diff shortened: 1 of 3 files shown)


CharliePoole:
Interesting idea. May have been better to discuss the feature first in an issue.

Note that you can already do this by putting `where` clause in a file and including it on the command-line using @filename. However, that only makes the feature available when using the console runner and it's a bit more verbose to use.

As implemented, the feature doesn't only apply to test lists but to any test specification. A user of the console runner could indicate `--test:-Some.Test.Name` right now and it would run a test named "-Some.Test.Name", presumably created using `SetTestName`. That's admittedly a contrived example but it's still potentially breaking.

I'd be happier if we could actually limit this syntax to test lists.

lanfeust69:
I actually have been using the feature in a customized NUnit 2.6.3 runner for years, before needing it with NUnit3, hence no issue created...

Indeed, I could have changed the driving part that generates the testlist, but it was much easier (and potentially useful for others) to port the feature.

I hadn't thought of the possibility to explicitly name tests, so that I thought it was impossible for a test to actually start with a "-" or "!". OTOH (especially assuming that an ambiguity wasn't possible), I considered having the feature also available through `--test` desirable (as a matter of fact, I did my first tests like that). That being said, is the (hopefully unlikely) break really worse for command-line parameters than within a testlist ?

ChrisMaddock:
Hi @lanfeust69, thanks for putting this together.

Cool feature - but I echo Charlie's concerns about how this PR would be adding new constraints on how tests can be named. I agree the potential breaking changes are niche, but do think we have a responsibility to keep our "API" clean and simple here, and I'm not totally comfortable that this change wouldn't lead to unexpected behaviour for someone unaware of this feature. To me, that's always been one fun thing about the popularity of NUnit, that people do tend to find these edge cases! 😅 

If we were going to add this functionality, I would prefer us to add e.g. an `--exclusionTestList` flag (actual name to be decided!). Given you've ported this for personal use - what would be your thoughts on that?

May be worth waiting for others comments from @nunit/engine-team too...busy time of year! 👍

ChrisMaddock:
Great - thanks! 😄 Yeeeah, we end up with a lot of command line options, but it always seems to be the safest option...

Just to be clear on what I was thinking - I was thinking of a second text file in the same format as the test list file, just passed into the parameter `--exclusiontestlist` rather than `--testlist`. There'd need to be some precedence I guess, so I guess that anything in exclusiontestlist, overrides what was specified in test list.

That would mean we wouldn't need to `!` and `-` syntax, so avoid the test naming problem. Opening that up to others on @nunit/engine-team for their opinions, however. 🙂

CharliePoole:
Another approach using the test list would be to require the `-` or `!` to be in the first character position.

Yet another approach would be to create some simplifications of the TSL syntax and use it in an @file. Currently, you could have a file like...

```
--where test=Something || test=SomethingElse || etc.
```
(Tests names with special characters have to be put into quotes)

We could create some shortcut syntax or an is-one-of sort of operator to reduce the overhead.

Issue #515, opened 2018-12-22
Appveyor CI failing on master


ChrisMaddock:
It failed on the last merge, and on #516. 😞 Taking a look...

mikkelbu:
Wonder if it is related to the following error? (from my phone so I cannot figure out how to make code blocks)


C:\projects\nunit-console\src\NUnitEngine\mock-assembly\mock-assembly.csproj : error NU3005: Package 'Microsoft.NETCore.Platforms 2.0.0' from source 'C:\Program Files\dotnet\sdk\NuGetFallbackFolder': The package signature file entry is invalid. The central directory header field 'compression method' has an invalid value (8). [C:\projects\nunit-console\NUnitConsole.sln]

ChrisMaddock:
Exactly my conclusions. 😄 Appveyor made a change on the 13th Dec from what I could find, which fits with when out builds started failing. 😒 

I'm just seeing if upgrading the tests to .NET Core 2.2 works...seeing as I'm not holding out much hope for a quick fix on that issue...

ChrisMaddock:
Siiigh, it doesn't.

As far as I can see, we can't currently build anything that cross-targets .NET Core 2.x and .NET Core 1.x on Appveyor. I wonder why the framework builds aren't failing in the same way. 🤔 

I'm out of time for today - anyone else is welcome to pick this up, if they fancy a dig. Feel free to work on the issue-515 branch, where I've upgraded all projects to .NET Core 2.2 in the hopes that that might fix this...

ChrisMaddock:
Switched to previous appveyor image. Ticket opened here: https://help.appveyor.com/discussions/problems/18758-nuget-package-restore-failing-for-net-core-20-since-dec-13-windows-image-update

Issue #514, opened 2018-12-22
Add higher-level unit tests for structure of TestRunners


ChrisMaddock:
We've had a number of issues recently which have related to the structure and order of TestRunners created. This is currently a relatively complex area of code, key concerns being:

- Multiple assembly test packages
- 'Project' test packages
- Test Packages which are a mix of the above
- Precedence of top level process/domain options over options specified at lower levels in the Test Package

Currently, we have the following open issues in or around this area: https://github.com/nunit/nunit-console/issues/495, https://github.com/nunit/nunit-console/issues/439, https://github.com/nunit/nunit-console/issues/428, https://github.com/nunit/nunit-console/issues/418, https://github.com/nunit/nunit-console/issues/386, https://github.com/nunit/nunit-console/issues/382, https://github.com/nunit/nunit-console/issues/357, https://github.com/nunit/nunit-console/issues/29, https://github.com/nunit/nunit-console/issues/615

One issue I've always found in addressing these problems, is that the only testing we currently have in this area is tests of `DefaultTestRunnerFactory`, which only test the top level runner returned. To address this, I'd like to add some higher-level tests, which take a constructed test package, and test the entire structure of TestRunners that will be built.

In the long run, I would plan for this suite of tests to replace `NUnit.Engine.Services.Tests.DefaultTestRunnerFactoryTests.CorrectRunnerIsUsed`. It's turning into a pretty big piece of work however, so I'll do this as multiple PRs instead.

Progress:

To-do's on this:

- [x] Tests for multiple projects
- [x] Tests for mix of projects and assemblies
- [x] Tests for both types of TestPackage ctor
- [x] Replace all `CorrectRunnerIsUsed` tests
- [x] Replace all `CheckInternalRunner` tests

CharliePoole:
This is a great idea. I think there is a lot to be done and that having a separate set of higher-level (integration?) tests in their own assembly might be better than combining them into the unit test assembly.

Maybe you could come up with a structure or framework that allowed other tests to be added, thereby sharing the workload.

ChrisMaddock:
To-do's on this:

- [ ] Tests for multiple projects
- [ ] Tests for mix of projects and assemblies
- [ ] Tests for using the TestPackage(string) ctor vs the TestPackage(List<string>) ctor. (These currently give different results, which I feel they should not! To be investigated further.)

[...]

PR #516, opened 2018-12-22
Initial Test Runner Selection Tests (1/4)

ChrisMaddock:
Contributes to #514.

This is the basic structure for the new tests I want to add to test the overall tree of `ITestRunners` which are built by the engine, in various situations. This currently just tests simple command-line-assembly usage - I started off trying to do everything in one go, but it gets more complicated once process/domain models and packages are added in, so I thought this would be easier for review.

I separated out the Full-Framework and NetStandard test cases, as the normal expected results are different anyway, and once process/domain model tests are added in, it becomes much more complex trying to maintain a single file.

This currently duplicates some test cases in the DefaultTestRunnerFactoryTests - I plan to finish off this test suite entirely first, before removing the old version.

This is ready for review and merge as is - I'll do the further work in separate PRs. 🙂

~Edit: Tracking the Appveyor failure at https://github.com/nunit/nunit-console/issues/515~


ChrisMaddock:
Thanks Charlie. 🙂 Given this is just test code, and I want to carry on with the next steps, I'll merge it now!

PR #520, opened 2018-12-23
Runner selection tests (2/4) - Add Process/Domain model tests

ChrisMaddock:
Next step of https://github.com/nunit/nunit-console/issues/514. 

N.B. This is branched off of #516 - that one should be reviewed and merged first.

This is where things start to get a little more wordy. This adds exhaustive tests for every combination of process/domain model which may be applied to a package, for single and multiple assembly packages. The tests added reflect _current behaviour_ rather than correct behaviour - I'm intending to use these tests later to start pinning down various bugs. We may also decide later down the line that some of these combinations of settings are nonsensical and choose to remove those tests, but again, I think that should be a future change.

I've written out each expected result explicitly here as there are so many subtleties that this seems like a situation where additional lines of codes is favourable in terms of explicitness. I've also removed the "three assemblies" tests in this PR, as I couldn't see any place where the behaviour of three assemblies differs to two assemblies. (That will however not be the case, once we start adding 'projects' in to the mix. 😄


src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20SingleAssemblyExpectedRunnerResults.cs
```diff
@@ -0,0 +1,104 @@
...
+                return new RunnerResult { TestRunner = typeof(MultipleTestProcessRunner) };
+            }
+            if (processModel == ProcessModel.Multiple && domainUsage == DomainUsage.Multiple)
+            {
+                return new RunnerResult { TestRunner = typeof(MultipleTestProcessRunner) };
+            }
+            if (processModel == ProcessModel.Multiple && domainUsage == DomainUsage.Single)
+            {
+                return new RunnerResult { TestRunner = typeof(MultipleTestProcessRunner) };
+            }
```
CharliePoole: Would nested `switch` statements be clearer? Also below.

ChrisMaddock: Yeah, I considered both. My thinking was that if we ever wanted to add a third option to this (e.g. framework), then this would be clearer than trebly-nested switch statements. I think that's probably overthinking things though...I think I'll rewrite it. Thanks! 😄

ChrisMaddock:
As review - now refactored to nested switch statements. Much easier to follow. 🙂

CharliePoole:
I agree!

PR #525, opened 2019-01-05
Separate NuGet Restore for Appveyor build

ChrisMaddock:
Following on from: https://github.com/nunit/nunit-console/pull/517, @IlyaFinkelshteyn from the @appveyor team found this solution, which will allow us to continue using the current appveyor image, and our builds not to fail when the "Previous" image next changes.

The core issue seems to be this one: https://github.com/NuGet/Home/issues/7414, which Appveyor are tracking via: https://github.com/appveyor/ci/issues/2787

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,9 +2,10 @@ branches:
   except:
     - /travis-.*/
 
-image: Previous Visual Studio 2017
+image: Visual Studio 2017
 
 build_script:
+  - nuget restore src\NUnitEngine\mock-assembly\mock-assembly.csproj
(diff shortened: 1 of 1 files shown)


IlyaFinkelshteyn:
@gep13, @devlead maybe this (switch to `nuget` instead of `msbuild` to restore packages) can be done somehow within Cake for it to look less hacky?

devlead:
@IlyaFinkelshteyn sure one could have a `Restore` task in the script that the build task depends on.

ChrisMaddock:
Thanks both! Will take a look. 🙂

Issue #46, opened 2016-08-30
ProcessRunner.RunTestsAsync doesn't work


CharliePoole:
The AsyncTestEngineResult returned from RunRestsAsync never shows as completed. This error doesn't actually affect how we currently work, since we don't use the RunAsync method in the console runner. However, we may use it in the future.

CharliePoole:
This is important if still correct but it's so old that it should be confirmed once again on the 4.0 codebase!

veleek:
I've verified that this is still a bug in V4.  

The problem is that the `RunAsync` call returns an `AsyncTestEngineResult` across the remoting proxy (from the nunit-agent-net462 process).  The `WaitHandle` it contains is properly marshaled across the boundary so it can be waiting from the caller and [triggered from the agent](https://github.com/nunit/nunit-console/blob/version4/src/NUnitEngine/nunit.engine.api/AsyncTestEngineResult.cs#L52-L53) after it sets the `_result` property to a non-null value.  However, the setting of the _result does not make it across the proxy boundary so it's still null.

I don't know much about the proxy boundary, is there an easy way to make something like this flow across?

CharliePoole:
We have a V4 issue to remove Windows Remoting as a comm transport, replacing it with straight TCP. So if this is purely a remoting issue, we can leave it for now. Is it possible to check this using the .NET 8.0 agent, which doesn't involve remoting?

veleek:
`ProcessRunner` is ifdef'd in for `NETFRAMEWORK` only, so it's a moot point in .NET Core.

CharliePoole:
'ProcessRunner' is on the runner side of things, but it launches agents for both .NET Framework and .NET Core. It's ifdef'd because the .NET Core _runner_ doesn't launch processes at all, but may in the future.

Windows Remoting is only used when communicating with agents that run the .NET Framework. Under #1660 we'll stop using Windows Remoting for that situation. But I agree that Won't Fix is probably the best solution.

veleek:
Ah right!  I missed that extra "agent" word in there.  I'll consider taking a look at #1660 next if I have a chance.

PR #526, opened 2019-01-06
Ensure Load and Reload incorporate test-run and prevent exception

### LINKED ISSUE #523: Reloading multiple files causes exception
When multiple files are loaded and `MasterTestRunner.Reload()` is called an InvalidOperationException is thrown with the message "May not use 'Xml' property on a result with multiple XmlNodes." This is thrown by the `TestEngineResult.Xml` property.

The `TestEngineResult` class was designed for efficiency in transmitting results over a comm channel, such as used by .NET remoting. It retains the XML representation in string form as long as possible. Since combining multiple results requires conversion to an `XmlNode`, combining results is delayed until the result is received by the master test runner. It's up to the master test runner to transform a result that combines multiple sources into a single XmlNode. It does that for load and run results but since Reload has never been exercised before, that seems to have slipped through the cracks.

CharliePoole:
Fixes #523 

Note that this changes  some engine behavior because it's basically the only way to fix the bug. While unlikely, somebody may have come to rely on the bug.

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
@@ -89,7 +89,7 @@ namespace NUnit.Engine.Runners.Tests
         {
             var prop = typeof(MasterTestRunner).GetField("_engineRunner", BindingFlags.NonPublic | BindingFlags.Instance);
             var runner = prop.GetValue(_runner);
-#if NET35
+#if !NETCOREAPP1_1 && !NETCOREAPP2_0
             if (_numAssemblies == 1)
                 Assert.That(runner, Is.TypeOf<TestDomainRunner>());
             else
@@ -185,7 +185,6 @@ namespace NUnit.Engine.Runners.Tests
 
             foreach (XmlNode suite in children)
             {
-                //Assert.NotNull("No suite found");
                 Assert.That(suite.GetAttribute("testcasecount", 0), Is.EqualTo(MockAssembly.Tests));
                 Assert.That(suite.GetAttribute("result"), Is.EqualTo("Failed"));
                 Assert.That(suite.GetAttribute("passed", 0), Is.EqualTo(MockAssembly.PassedInAttribute));
--- a/src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
@@ -32,18 +32,29 @@ using NUnit.Tests;
 
 namespace NUnit.Engine.Runners.Tests
 {
+    [TestFixture(1)]
+    [TestFixture(2)]
     public class MasterTestRunnerTests : ITestEventListener
     {
+        private List<string> _assemblies = new List<string>();
+        private int _numAssemblies;
         private TestPackage _package;
         private ServiceContext _services;
         private MasterTestRunner _runner;
         private List<XmlNode> _events;
 
+        public MasterTestRunnerTests(int numAssemblies)
+        {
+            _numAssemblies = numAssemblies;
+            var testDirectory = TestContext.CurrentContext.TestDirectory;
+            for (int i = 0; i < numAssemblies; i++)
+                _assemblies.Add(Path.Combine(testDirectory, "mock-assembly.dll"));
+        }
+
         [SetUp]
         public void Initialize()
         {
-            _package = new TestPackage(Path.Combine(TestContext.CurrentContext.TestDirectory, "mock-assembly.dll"));
-
+            _package = new TestPackage(_assemblies);
 
             // Add all services needed
             _services = new ServiceContext();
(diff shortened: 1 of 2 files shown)


src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
```diff
@@ -73,21 +84,57 @@ public void CleanUp()
                 _services.ServiceManager.Dispose();
         }
 
+        [Test]
+        public void CheckInternalRunner()
+        {
+            var prop = typeof(MasterTestRunner).GetField("_engineRunner", BindingFlags.NonPublic | BindingFlags.Instance);
+            var runner = prop.GetValue(_runner);
+#if NET35
```
mikkelbu: In most other places we use `!NETCOREAPP1_1 && !NETCOREAPP2_0`. Even though it will have the same behaviour now, it could make a difference in the future.

src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
```diff
@@ -105,7 +105,9 @@ protected bool IsPackageLoaded
         /// <returns>An XmlNode representing the loaded assembly.</returns>
         public XmlNode Load()
         {
-            LoadResult = _engineRunner.Load();
+            LoadResult = PrepareResult(_engineRunner.Load())
+                .Aggregate(TEST_RUN_ELEMENT, TestPackage.Name, TestPackage.FullName);
```
mikkelbu: Before the top-level xml-node was `test-suite`, now it is `test-run`. I guess that this could affect client code? (but looking at a resulting TestResult.xml from a execution it seems that this is already the case for the xml from a test execution, so I guess that this change will only affect the clients uses the engine directly, and not just rely on the console).

src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
CharliePoole: My usage reflected the fact that I actually was only using it for .NET 3.5, so that's what I did first. I later added the net core tests based on what I saw it doing... I'm glad to change it if that's the more standard way. Of course, any new platform could change all of this logic and possibly require a new runner type.

src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
CharliePoole: The master test runner never returns the direct result of Load unless you call Load directly. Most times in the past it has only been called indirectly through Explore or Run so the LoadResult was immediately updated.

Looked at that way, this can only affect someone who is directly calling Load or Reload on MasterTestRunner, in which case they are already broken, or someone who has extracted our internal TestEngineRunner classes for their own use, since we don't expose the internal logic in the API.

CharliePoole:
@mikkelbu Totally agree... that's why I asked for both of you! However, answering your questions, I think I have talked myself into this being an OK change that can't make things worse for anyone. :smile:

src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
mikkelbu: Personally, I like the consistency of using `!NETCOREAPP1_1 && !NETCOREAPP2_0` as we only have a few places where we switch on `NET35`, but I don't know if it is standard. I can remember that @jnm2 wrote something about it somewhere, but I cannot find it in this or the nunit repository or on the wiki, but it can be that my Google ability is not good enough. Let's see what Chris - or another - says.

mikkelbu:
> @mikkelbu Totally agree... that's why I asked for both of you! However, answering your questions, I think I have talked myself into this being an OK change that can't make things worse for anyone. 😄

I also agree after having read the comment above.

src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
CharliePoole: It would be tempting to add a define like SUPPORTS_APP_DOMAIN for the 3.5 build and use that for the places where that's the deciding factor.

ChrisMaddock:
I need to see pull this and see it in context to understand what's going on here - will hopefully do that tonight! 👍

src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
```diff
@@ -126,28 +173,33 @@ public void RunAsync()
...
+            Assert.That(result.GetAttribute("failed", 0), Is.EqualTo(MockAssembly.Failed * _numAssemblies));
+            Assert.That(result.GetAttribute("skipped", 0), Is.EqualTo(MockAssembly.Skipped * _numAssemblies));
+            Assert.That(result.GetAttribute("inconclusive", 0), Is.EqualTo(MockAssembly.Inconclusive * _numAssemblies));
+
+            var children = result.SelectNodes("test-suite");
+            Assert.That(children.Count, Is.EqualTo(_numAssemblies));
+
+            foreach (XmlNode suite in children)
+            {
+                //Assert.NotNull("No suite found");
```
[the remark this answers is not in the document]
CharliePoole: Cool! I wasn't aware of the suggeston feature.

src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
CharliePoole: I added the change because I couldn't interpret the results I was seeing, particularly because I'm not familiar with the .NET Core builds. It does probably belong elsewhere and could be moved.

src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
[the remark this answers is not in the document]
CharliePoole: I think it was probably added at one time when there was a worry of the children containing null entries. Not relevant any longer.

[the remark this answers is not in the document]
ChrisMaddock: I more meant that, as written, `Assert.NotNull("No suite found");` would never fail!

CharliePoole:
So I learned that Resolving a conversation kills off any batched committed suggestions! Had to unresolve and re-batch to commit them.

src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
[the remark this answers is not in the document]
CharliePoole: Oh, that! Good eye. :smiley:

ChrisMaddock:
By the way, did you have any thoughts on the RunAsync comment above? 😊

mikkelbu:
Perhaps we could make it part of #46, if that is the same `RunTestsAsync` method?

src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
jnm2: Sorry, I'm (trying to be) not actually dead!

> I can remember that @jnm2 wrote something about it somewhere

The bottom of https://github.com/nunit/nunit/blob/master/BUILDING.md#defined-constants?

My rule of thumb is to try to name the odd one out, the one that is the reason that the condition is necessary. It's ambiguous to me many times though. Is supporting appdomains unusual, or is lack of support unusual?

Also consider `#if !(NETCOREAPP1_1 || NETCOREAPP2_0)` over `#if !NETCOREAPP1_1 && !NETCOREAPP2_0`.

`SUPPORTS_APP_DOMAIN` is even better especially if the same `#if` condition would need to be synchronized across more than location in the code if a new target was added.

CharliePoole:
Ideally, I would use platform defines for features that we implement on all the platforms but with different code.

For a feature we can't implement on a platform with a specific capability, I'd use "platform capability" defines like SUPPORTS_APPDOMAIN.

For complex NUnit features, like timeout and parallel execution I'd use "feature support defines". Up to now we use PARALLEL for our support of the parallel feature. PARALLEL_EXECUTION or PARALLEL_TESTS might have been better.

As to whether which form of expression we use to combine platforms, I don't think it's important enough to standardize but reasonable to review on a case-by-case basis for clarity.

My 2 cents.

PR #534, opened 2019-01-15
Fix duplicate test ids when loading projects

### LINKED ISSUE #533: Duplicate ids when loading a project
When loading an NUnit solution in the TestCentric GUI, I get an exception that there is a duplicate id when attempting to add the test to a dictionary of all tests in the tree. The same happens with NUnit projects, which lead me to suspect the engine rather than the project loader.

Here's what I see happening when I load a solution file:

1. I create a package using `new TestPackage("testcentric-gui.sln")`. This gives me an anonymous package representing the command-line with one subpackage named for the solution.

2. I call `GetRunner(TestPackage package)`, which has the side effect of expanding the package. The second-level package (testcentric-gui.sln) now has 14 subpackages. The engine has correctly assigned unique ids 59 through 72 to them.

3. I call `Runner.Explore(TestFilter.Empty)`. The package remains unchanged. The tests that are returned make no use of the id prefixes that were created. That is, there are no tests 59-xxxx, etc. All the tests use the prefix "1-", which represents the Id of the solution project package rather than any of the assembly subpackages.

Researching, this appears to happen here: https://github.com/nunit/nunit-console/blob/master/src/NUnitEngine/nunit.engine/Runners/DirectTestRunner.cs#L130

The `DirectRunner` is called with a test package, which may or may not have subprojects. As each subproject is loaded, the Id of the __parent__ project is used for the driver. I think the subproject id has to be used.

We have never detected this because we don't actually rely on unique ids across multiple assemblies in either the console runner or the adapters. We don't have tests that guarantee
1. That each driver is given a unique prefix.
2. That every test has a unique test ID.
This could easily be done by building a dictionary of ids, just as the GUI does.

CharliePoole:
Fixes #533 

The first three commits merely serve to verify that no duplicates are generated when there are __no__ projects involved.I added cases with the notests-assembly because that's where I was seeing errors when they are included in projects. However, the tests didn't break anything.

Next step is to create some tests that use a project.

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
@@ -30,13 +30,12 @@ using System.Text;
 using System.Xml;
 using NUnit.Framework;
 using NUnit.Tests.Assemblies;
+using NUnit.Engine.Services;
+using NUnit.Engine.Services.Tests.Fakes;
 
 namespace NUnit.Engine.Runners.Tests
 {
-    using Services;
-    using Services.Tests.Fakes;
-
-    [TestFixtureSource("FixtureData")]
+    [TestFixtureSource(nameof(FixtureData))]
     public class MasterTestRunnerTests : ITestEventListener
     {
         private List<AssemblyData> _data = new List<AssemblyData>();
@@ -176,7 +175,7 @@ namespace NUnit.Engine.Runners.Tests
             var result = _runner.Load();
 
             Assert.That(result.Name, Is.EqualTo("test-run"));
-            Assert.That(result.GetAttribute("testcasecount", 0), Is.EqualTo(_totalTests));
+            Assert.That(result.GetAttribute("testcasecount", -1), Is.EqualTo(_totalTests));
 
             var suites = result.SelectNodes("test-suite");
             Assert.That(suites.Count, Is.EqualTo(_numAssemblies));
@@ -185,7 +184,7 @@ namespace NUnit.Engine.Runners.Tests
             foreach (XmlNode child in suites)
             {
                 var data = _data[i++];
-                Assert.That(child.GetAttribute("testcasecount", 0), Is.EqualTo(data.Tests));
+                Assert.That(child.GetAttribute("testcasecount", -1), Is.EqualTo(data.Tests));
                 Assert.That(child.GetAttribute("runstate"), Is.EqualTo(data.RunState));
             }
 
@@ -199,7 +198,7 @@ namespace NUnit.Engine.Runners.Tests
             var result = _runner.Reload();
 
             Assert.That(result.Name, Is.EqualTo("test-run"));
-            Assert.That(result.GetAttribute("testcasecount", 0), Is.EqualTo(_totalTests));
+            Assert.That(result.GetAttribute("testcasecount", -1), Is.EqualTo(_totalTests));
 
             var suites = result.SelectNodes("test-suite");
             Assert.That(suites.Count, Is.EqualTo(_numAssemblies));
@@ -208,7 +207,7 @@ namespace NUnit.Engine.Runners.Tests
             foreach (XmlNode child in result.SelectNodes("test-suite"))
             {
                 var data = _data[i++];
-                Assert.That(child.GetAttribute("testcasecount", 0), Is.EqualTo(data.Tests));
+                Assert.That(child.GetAttribute("testcasecount", -1), Is.EqualTo(data.Tests));
                 Assert.That(child.GetAttribute("runstate"), Is.EqualTo(data.RunState));
             }
 
@@ -228,7 +227,7 @@ namespace NUnit.Engine.Runners.Tests
             var result = _runner.Explore(TestFilter.Empty);
 
             Assert.That(result.Name, Is.EqualTo("test-run"));
-            Assert.That(result.GetAttribute("testcasecount", 0), Is.EqualTo(_totalTests));
+            Assert.That(result.GetAttribute("testcasecount", -1), Is.EqualTo(_totalTests));
 
             var suites = result.SelectNodes("test-suite");
             Assert.That(suites.Count, Is.EqualTo(_numAssemblies));
@@ -237,7 +236,7 @@ namespace NUnit.Engine.Runners.Tests
             foreach (XmlNode suite in suites)
             {
                 var data = _data[i++];
-                Assert.That(suite.GetAttribute("testcasecount", 0), Is.EqualTo(data.Tests));
+                Assert.That(suite.GetAttribute("testcasecount", -1), Is.EqualTo(data.Tests));
                 Assert.That(suite.GetAttribute("runstate"), Is.EqualTo(data.RunState));
             }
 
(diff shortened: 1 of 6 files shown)


CharliePoole:
Deliberately committing with a failing test so @ChrisMaddock and @mikkelbu can see where we are going wrong. Additionally, the various test cases for the three target platforms are instructive in how they differ.

The current failure is that the test returned representing "project.nunit" has all zero counts, even though it contains a subproject for mock-assembly. So no aggregation is being performed.

Also interesting is that the .NET Core 2.0 build uses an AggregatingTestRunner for a single project, whereas the .NET 2.0 build uses a TestDomainRunner. Of course, as stated already, neither of them is actually aggregating anything. :smile: 

I'm pretty sure that the difference in which runner is used was not something intended, with the exception that TestDomainRunner isn't used in .NET Core. Other than that, all the runner selection should work the same way on all platforms to the extent that a common feature is supported.

@ChrisMaddock Maybe we should figure out how to merge some of these tests with the ones you are working on... or at least reflect the problem with runner selection that I'm seeing for MasterTestRunner in your tests of the runner service.

src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
```diff
@@ -21,53 +21,130 @@
...
+            new TestRunData( typeof(MultipleTestDomainRunner), NoTestsAssemblyData, NoTestsAssemblyData ),
+            new TestRunData( typeof(MultipleTestDomainRunner), MockAssemblyData, NoTestsAssemblyData )
+#endif
+        };
+
         [SetUp]
         public void Initialize()
         {
             _package = new TestPackage(_assemblies);
+            _package.AddSetting(EnginePackageSettings.ProcessModel, "InProcess");
```
CharliePoole: This was using `InProcessTestRunnerFactory`, as many of our tests do, in order to avoid creating a process. I realized that using `InProcessTestRunnerFactory` directly bypasses an important bit of code in `DefaultTestRunnerFactory`, which is where we recognize whether there are any projects listed in the command-line. The result is that the tests actually behave differently from how the console app behaves.

In this case, I replaced `InProcessTestRunnerFactory` with `DefaultTestRunnerFactory` and set the `ProcessModel` to `InProcess`. This gives the correct behavior.

I think we should review and fix any other tests that use this technique but also review the way we use InProcessTestRunner in our agent process. We may need to refactor so that the examination of how many projects we have occurs either in both runners or separately.

CharliePoole:
The latest commit fixes the problem... just one line of new code to fix but tons of new tests. Change is that any projects in command-line cause the `AggregatingTestRunner` to be used.

Note that this eliminates one possibility... that of a project where several test assemblies require running in the same AppDomain. This isn't a great loss and was not working anyway! However, if we want it to be possible, then more work is required. However, I think it should wait until after some of the other, higher priority issues related to projects are resolved.

Starting tomorrow, I'll be traveling for a week. If it seems that changes are needed, someone else should feel free to make them.

src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
```diff
@@ -170,40 +263,106 @@ public void RunAsync()
...
+        {
+            AssertThatIdsAreUnique(test, new Dictionary<string, bool>());
+        }
+
+        private void AssertThatIdsAreUnique(XmlNode test, Dictionary<string, bool> dict)
+        {
+            Assert.That(dict, Does.Not.ContainKey(test.GetAttribute("id")));
+
+            foreach (XmlNode child in test.SelectNodes("test-suite"))
+                AssertThatIdsAreUnique(child, dict);
```
ChrisMaddock: Is this method missing the step that adds the ID's to the dict? 😄

src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
ChrisMaddock: Sounds good. Could you log an issue to look at the tests?

src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
```diff
@@ -21,53 +21,131 @@
...
-using System.Reflection;
-using System.IO;
-using NUnit.Tests;
 
 namespace NUnit.Engine.Runners.Tests
 {
-    [TestFixture(1)]
-    [TestFixture(2)]
+    using Services;
+    using Services.Tests.Fakes;
```
ChrisMaddock: Minor: Could all the `usings` in this file go together, rather than having half inside the namespace and half outside?

src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
```diff
@@ -21,53 +21,131 @@
...
+
+    [TestFixtureSource("FixtureData")]
```
ChrisMaddock: Only as I'm asking for other changes anyway...how about `nameof(FixtureData)` - for extra future-refactor-safety? 😄

src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
```diff
@@ -105,13 +176,20 @@ public void Load()
             var result = _runner.Load();
 
             Assert.That(result.Name, Is.EqualTo("test-run"));
-            Assert.That(result.GetAttribute("testcasecount", 0), Is.EqualTo(MockAssembly.Tests * _numAssemblies));
+            Assert.That(result.GetAttribute("testcasecount", 0), Is.EqualTo(_totalTests));
```
ChrisMaddock: Picky again: zero is a valid value here now we're testing no-tests-assembly, right? Would it be a little stronger to default to `-1` for `GetAttribute()`? Same occurs in several other places in this file.

src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
```diff
@@ -170,40 +263,106 @@ public void RunAsync()
...
+                ExpectedRunner = expectedRunner;
+                AssemblyData = assemblyData;
+            }
+
+            public override string ToString()
+            {
+                var sb = new StringBuilder(ExpectedRunner.Name);
+                foreach (var assembly in AssemblyData)
+                    sb.Append($", {assembly.Name}");
+                return sb.ToString();
```
ChrisMaddock: I like this `TestRunData` class, compared to the first iteration of these tests in the commits. Good thinking adding the `ToString()` as well! 😄

src/NUnitEngine/nunit.engine.tests/Services/Fakes/FakeProjectService.cs
```diff
@@ -22,21 +22,39 @@
...
 using NUnit.Engine.Extensibility;
 
 namespace NUnit.Engine.Services.Tests.Fakes
 {
     public class FakeProjectService : FakeService, IProjectService
     {
+        private string _supportedExtension;
+        private Dictionary<string, string[]> _projects = new Dictionary<string, string[]>();
+
+        public FakeProjectService(string supportedExtension = ".nunit")
```
ChrisMaddock: This constructor is nice too. 🙂

src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
```diff
@@ -296,11 +296,8 @@ private void ExpandProjects(TestPackage package)
         {
             if (package == null) throw new ArgumentNullException("package");
 
-            string packageName = package.FullName;
-            if (File.Exists(packageName) && _projectService.CanLoadFrom(packageName))
-            {
```
ChrisMaddock: Presume this change is for the tests. Did we check this that still behaves nicely if the `packageName` doesn't exist as a file in 'production'?

CharliePoole: It was a redundant check because the ProjectService simply ignores package file names that don't exist. The check in this location was forcing me to create any file names I want to pass to the faike project loader so I removed it.

ChrisMaddock: Great - I’m happy so long as it’s checked elsewhere! 😄

PR #539, opened 2019-02-01
Updated RunTests to create output directory

### LINKED ISSUE #501: Create result directory if it does not exist
Right now if the directory does not already exist for the result file path, you get this:

```
NUnit.Engine.NUnitEngineException : The path specified in --result C:/Users/Joseph/Source/Repos/nunit/test-results/net20/nunit.framework.tests.xml could not be written to
  ----> System.IO.DirectoryNotFoundException : Could not find a part of the path 'C:\Users\Joseph\Source\Repos\nunit\test-results\net20\nunit.framework.tests.xml'.

--NUnitEngineException
The path specified in --result C:/Users/Joseph/Source/Repos/nunit/test-results/net20/nunit.framework.tests.xml could not be written to
--
DirectoryNotFoundException
Could not find a part of the path 'C:\Users\Joseph\Source\Repos\nunit\test-results\net20\nunit.framework.tests.xml'.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access)
   at NUnit.Engine.Services.NUnit3XmlResultWriter.WriteResultFile(XmlNode resultNode, String outputPath)
```

It would be convenient if the directory was just created.

JerryBLi:
Updated RunTests to create output directory if the directory does not currently exist.

Fixes #501

### DIFF
--- a/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
+++ b/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
@@ -171,9 +171,10 @@ namespace NUnit.ConsoleRunner
             foreach (var spec in _options.ResultOutputSpecifications)
             {
                 var outputPath = Path.Combine(_workDirectory, spec.OutputPath);
+                var outputDirectory = Path.GetDirectoryName(outputPath);
 
                 IResultWriter resultWriter;
-
+                
                 try
                 {
                     resultWriter = GetResultWriter(spec);
@@ -182,7 +183,19 @@ namespace NUnit.ConsoleRunner
                 {
                     throw new NUnitEngineException($"Error encountered in resolving output specification: {spec}", ex);
                 }
-
+                
+                try
+                {
+                    Directory.CreateDirectory(outputDirectory);
+                }
+                catch (SystemException ex)
+                {
+                    throw new NUnitEngineException(
+                        String.Format(
+                            "The directory in --result {0} could not be created",
+                            spec.OutputPath), ex);
+                }
+                
                 try
                 {
                     resultWriter.CheckWritability(outputPath);
--- a/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
+++ b/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
@@ -168,11 +168,13 @@ namespace NUnit.ConsoleRunner
 
         private int RunTests(TestPackage package, TestFilter filter)
         {
+
+            var writer = new ColorConsoleWriter(!_options.NoColor);
+
             foreach (var spec in _options.ResultOutputSpecifications)
             {
                 var outputPath = Path.Combine(_workDirectory, spec.OutputPath);
-                var outputDirectory = Path.GetDirectoryName(outputPath);
-
+                
                 IResultWriter resultWriter;
                 
                 try
@@ -183,19 +185,21 @@ namespace NUnit.ConsoleRunner
                 {
                     throw new NUnitEngineException($"Error encountered in resolving output specification: {spec}", ex);
                 }
-                
+
                 try
                 {
+                    var outputDirectory = Path.GetDirectoryName(outputPath);
                     Directory.CreateDirectory(outputDirectory);
                 }
                 catch (SystemException ex)
                 {
-                    throw new NUnitEngineException(
-                        String.Format(
-                            "The directory in --result {0} could not be created",
-                            spec.OutputPath), ex);
+                    writer.WriteLine(ColorStyle.Error, String.Format(
+                        "The directory in --result {0} could not be created",
+                        spec.OutputPath));
+                    writer.WriteLine(ColorStyle.Error, ExceptionHelper.BuildMessage(ex));
+                    return ConsoleRunner.UNEXPECTED_ERROR;
                 }
-                
+
                 try
                 {
                     resultWriter.CheckWritability(outputPath);
@@ -207,6 +211,7 @@ namespace NUnit.ConsoleRunner
                             "The path specified in --result {0} could not be written to",
                             spec.OutputPath), ex);
                 }
+
             }
 
             var labels = _options.DisplayTestLabels != null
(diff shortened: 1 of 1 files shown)


src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -182,7 +183,19 @@ private int RunTests(TestPackage package, TestFilter filter)
...
                 }
-
+                
+                try
+                {
+                    Directory.CreateDirectory(outputDirectory);
+                }
+                catch (SystemException ex)
+                {
+                    throw new NUnitEngineException(
```
CharliePoole: IF we do want to wrap the exception, an `NUnitEngineException` seems inappropriate. This exception has nothing to do with the engine. The runner is trying to create the directory itself and gets an exception from .NET. Wrapping it as an engine exception is potentially misleading.

Reviewing the console runner code, I can see at least one other place where `NUnitEngineException` is used in this way. It's directly after this code and may be the reason why it was used here.

Originally, `NUnitEngineException` was an exception thrown __only__ by the engine. It provided a way to clarify where an exception originated. Later, it came to be used in this file to wrap __other__ kinds of exceptions thrown by the engine. IMO, that's unfortunate, since it means that other runners have to figure out for themselves whether or how to wrap such exceptions. However, that's probably water under the bridge at that point.

I do, however, think it would be a good idea to stop using `NUnitEngineException` for errors that have nothing to do with the engine.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -182,7 +183,19 @@ private int RunTests(TestPackage package, TestFilter filter)
                 {
                     throw new NUnitEngineException($"Error encountered in resolving output specification: {spec}", ex);
                 }
-
+                
+                try
+                {
+                    Directory.CreateDirectory(outputDirectory);
```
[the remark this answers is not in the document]
JerryBLi: I do like the idea of wrapping the exception. I think the wrapped string message gives the user a concise explanation of what's going wrong while the original error message can be used to debug if they need more information, . 

Also, I can see a few different exceptions being thrown by `CreateDirectory`, if the user doesn't have permission \(`UnauthorizedAccessException`\) or if the user inputs an invalid directory name \(`ArgumentNullException`, `DirectoryNotFoundException`, or `NotSupportedException`\).  

While taking a look at this, I also encountered another thing - an exception will be thrown by https://github.com/nunit/nunit-console/blob/f979cb049afa8cf809cdcd8e31a9ca54266a4c29/src/NUnitConsole/nunit3-console/ConsoleRunner.cs#L174 if the user path is too long \(`PathTooLongException`\) and it isn't currently wrapped.  

I suggest adding the line inside the try block before the `CreateDirectory`. This would make any issues arising from creating a new directory have a uniform error message.
https://github.com/nunit/nunit-console/blob/f979cb049afa8cf809cdcd8e31a9ca54266a4c29/src/NUnitConsole/nunit3-console/ConsoleRunner.cs#L187-L190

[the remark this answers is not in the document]
CharliePoole: The nested exception will give additional information, but only if the runner actually displays it. OTOH, the raw exception gets displayed with whatever message is there.

See also my other comment about the appropriateness of using an `NUnitEngineException` here. That's something the @nunit/engine-team will have to weigh in on.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
JerryBLi: That makes sense. I definitely used `NUnitEngineException` because it was used in the code directly below. As i said above in https://github.com/nunit/nunit-console/pull/539#discussion_r253116045 , I personally like the idea of wrapping the error with a high-level comment. After taking a look at the errors that can happen, I can think of three solutions:
1. Don't wrap the exceptions and let the original exception just bubble up  
2. Wrap it in a`SystemException` and throw that instead
3. Wrap it in a`OptionException` and throw that instead

Does wrapping it in an `OptionException` make sense? From what I can tell, it's thrown when the user incorrectly inputs options in the command line. This exception seems like it fits for this concern.

CharliePoole: `OptionException` belongs to the underlying Mono.Options software that we use to parse the command-line. IMO it would make sense to use it if the command-line parser were detecting the error. However, the parser is pretty much unaware of what the string arguments to options mean or how they are used by the application. Also, the CreateDirectory error can only be detected by trying to create the directory, which is definitely not the parser's job.

The other options seem valid to me. Another wrapper possibility is `InvalidOperationException`.  We would still have to ensure that the console displayed both the general wrapping message and the specific reason. Otherwise all the user knows is that some kind of problem happened when creating a directory.

Personally, for software that is used by developers, I have no problem with just throwing the original exception, but there's some logic to wrapping as well, so long as we are consistent. 

@nunit/engine-team Does somebody want to weigh in on this?

JerryBLi:
@ChrisMaddock I changed it so that when the exception is caught, a message is printed to the console and an error value is returned. One note is that I had to move the `ColorConsoleWriter` instantiation to the top of the method so that I could make use of it. I didn't completely follow the previous error handling logic because it seems cumbersome in this situation:

If I were to set a SystemException variable when an exception happens, I would have to check it immediately after the try-catch block to create the directory \(and before the try-catch block of `resultWriter.CheckWritability(outputPath);`\). This is because  if the directory can't be created, the `CheckWritability` will fail and that error will mask the failure from create directory.

I felt that it is just better to handle the error in the catch block and save the unnecessary need to write to a variable and immediately check it afterwards. What do you think?

PR #542, opened 2019-02-09
Update result message body to use output colour

### LINKED ISSUE #519: Break up multiple console error messages with colour
Working on the NUnit tests recently, and trying to separate out the lines in multiple test failures made my eyes go a little square! What would people think to altering out colour scheme, so that the actual test failure 'header' line itself was written in read, and then the information below it written back in white. This would help provide separation between tests.

Same would also need to apply for warnings, of course.

image

abborg:
Resolves #519 

The header of the result will still use the error/warning colour set for
the result writer.

Example of console result:
image


abborg:
From what I can see in the Travis CI build, it looks unrelated to my changes. The failure seems to be related to a connection being closed causing a remoting exception. Is it possible to re-run the Travis CI build?

CharliePoole:
I restarted the Travis build

ChrisMaddock:
This looks much better - thanks for picking this up @abborg! 😄

PR #552, opened 2019-02-21
Use LTS versions of netcore for building

nvborisenko:
`build.cake` requires specific versions of netcore SDK. Let's use LTS versions in this case.


nvborisenko:
Appveyor will fail because of these versions of sdk are not installed on the `vs2017` appveyor image. Closing this PR.

Issue #544, opened 2019-02-13
Deprecate nunit.netstandard.engine NuGet package


ChrisMaddock:
With the 3.10 release, the .NET Standard engine and .NET Framework engine will be packaged together in the main NuGet packages, and the nunit.engine.netstandard will become redundant. 

We should mark it as obsolete on nuget.org - adding this issue to the 3.10 milestone so that we remember to do so.

PR #553, opened 2019-02-22
Don't try to load extensions in case of incompatible target frameworks (fix #524)

### LINKED ISSUE #524: .NET Standard 2.0 engine crashes when .NET Framework extensions are in Global NuGet Cache
_Found by @Xambey via [Gitter](https://gitter.im/nunit/nunit?at=5c2e001233e089363b3a5581)_

The .NET Standard 2.0 Engine currently throws if it attempts to load a .NET Framework extension (by design.) The NuGet packaging for the engine however includes a `.addins` file which attempts to locate and load all other NuGet packages installed alongside the engine. As later versions of NuGet download all packages to a global cache (e.g. _C:\Users\Chris\.nuget\packages\_), this means the engine can throw due to packages from an entirely separate project.

The fix should be to only throw when netfx extensions are located _directly_, rather than found from directories or wildcards. This behaviour would be consistent with current behaviour when attempting to load an invalid engine extension.

cc @rprouse - I think this needs fixing before the new .NET Standard builds are released. Not sure I'll have time in the near future, but I'm happy to look at this one when I can get round to it. Anyone else is welcome to jump in first!

nvborisenko:
Just skip loading of extensions if they are not compatible with target framework.
Extensions should not affect nunit engine execution.
Trying to fix #524

### DIFF
--- a/src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
+++ b/src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
@@ -517,14 +517,6 @@ namespace NUnit.Engine.Services
                     ep.Install(node);
                 }
             }
-            else if (!assembly.FromWildCard)
-            {
-                throw new NUnitEngineException($"Cannot load {assembly.FilePath} as extension assembly due not supported target framework.");
-            }
-            else
-            {
-                log.Warning($"Skipping {assembly.FilePath} assembly as extension assembly due not supported target framework.");
-            }
         }
 
         /// <summary>
--- a/src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
+++ b/src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
@@ -431,7 +431,7 @@ namespace NUnit.Engine.Services
         {
             log.Info("Scanning {0} assembly for Extensions", assembly.FilePath);
 
-            if (ValidateTargetFramework(Assembly.GetEntryAssembly(), assembly))
+            if (CanLoadTargetFramework(Assembly.GetEntryAssembly(), assembly))
             {
 
                 IRuntimeFramework assemblyTargetFramework = null;
@@ -525,7 +525,7 @@ namespace NUnit.Engine.Services
         /// </summary>
         /// <param name="runnerAsm">The executing runner</param>
         /// <param name="extensionAsm">The extension we are attempting to load</param>
-        internal static bool ValidateTargetFramework(Assembly runnerAsm, ExtensionAssembly extensionAsm)
+        internal static bool CanLoadTargetFramework(Assembly runnerAsm, ExtensionAssembly extensionAsm)
         {
             if (runnerAsm == null)
                 return true;
@@ -534,20 +534,19 @@ namespace NUnit.Engine.Services
             var runnerHelper = new TargetFrameworkHelper(runnerAsm.Location);
             if (runnerHelper.FrameworkName?.StartsWith(".NETStandard") == true)
             {
-                log.Warning($"{runnerAsm.FullName} test runner must target .NET Core or .NET Framework, not .NET Standard");
-                return false;
+                throw new NUnitEngineException($"{runnerAsm.FullName} test runner must target .NET Core or .NET Framework, not .NET Standard");
             }
             else if (runnerHelper.FrameworkName?.StartsWith(".NETCoreApp") == true)
             {
                 if (extHelper.FrameworkName?.StartsWith(".NETStandard") != true && extHelper.FrameworkName?.StartsWith(".NETCoreApp") != true)
                 {
-                    log.Warning($".NET Core runners require .NET Core or .NET Standard extension for {extensionAsm.FilePath}");
+                    log.Info($".NET Core runners require .NET Core or .NET Standard extension for {extensionAsm.FilePath}");
                     return false;
                 }
             }
             else if (extHelper.FrameworkName?.StartsWith(".NETCoreApp") == true)
             {
-                log.Warning($".NET Framework runners cannot load .NET Core extension {extensionAsm.FilePath}");
+                log.Info($".NET Framework runners cannot load .NET Core extension {extensionAsm.FilePath}");
                 return false;
             }
 
--- a/src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
+++ b/src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
@@ -517,6 +517,14 @@ namespace NUnit.Engine.Services
                     ep.Install(node);
                 }
             }
+            else if (!assembly.FromWildCard)
+            {
+                throw new NUnitEngineException($"Cannot load {assembly.FilePath} as extension assembly due not supported target framework.");
+            }
+            else
+            {
+                log.Warning($"Skipping {assembly.FilePath} assembly as extension assembly due not supported target framework.");
+            }
         }
 
         /// <summary>
--- a/src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
+++ b/src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
@@ -64,7 +64,7 @@ namespace NUnit.Engine.Services
         /// </summary>
         public IEnumerable<IExtensionNode> Extensions
         {
-            get { return _extensions.ToArray();  }
+            get { return _extensions.ToArray(); }
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


ChrisMaddock:
Hi @nvborisenko - thanks a lot for picking this up to unblock the 3.10 release, really appreciate that! 😄 

We just need to work on the implementation approach here, this behaviour should be consistent with the existing behaviour for loading full framework extensions.

There's two ways an extension can be found, either within an assembly which has been directly specified, or from a "wildcard". 'Directly specified' assemblies are named to load specifically by the user (e.g. through an `.addins` files). 'Wildcard' assemblies are found by the engine looking in sensible places, or searching an entire directory from an addins file. In the former case, the user has specified to load a specific assembly, so if it can't be loaded, the engine should error. In the latter case, then engine should ignore any unloadable assemblies.

For each extension assembly that we're attempting to load, you can see whether it is found via a wildcard or not with the `assembly.FromWildcard` property - you can see some existing use of that in this method.

What the behaviour should be - the exception should be suppressed ONLY if `assembly.FromWildcard` is true. In that case, the engine should silently ignore the assembly, after attempting to load it. If `assembly.FromWildcard` is false, this means that the user has explicitly said to try and load this assembly, and we _should_ throw the exception and error.

Would you be happy to try and move things around, to meet that behaviour? It might be worth refactoring this method a little to get there. 🙂

ChrisMaddock:
@rprouse - @nvborisenko has picked up this issue as it's blocking the 3.10 release, and he's keen to get using it. I think the only other thing that couldn't be pushed back is #544 which I can look at - are you happy that we'd be good to release once those two are in? 🙂

rprouse:
@chrismaddock, I trust your judgement, so yes I think we are due for a release. I am just returning from vacation and will be able to look more closely this week.

CharliePoole:
So if non-wildcard extensions always throw when the target framework doesn't match, are we then saying that loading by name is no longer an option in the addins file that is distributed with an extension?

nvborisenko:
It is still an option with the following assumption: user is aware about compatibility of runner and extension. If runner and extension are not compatible (for non-wildcard extensions), then user gets an error. Further workflow might be to wait until the extension supports runtime framework, in other cases usage of extension is blocked.

Generally the question is whether runner should silently ignore unsupported extensions, or throw an error. In my opinion all `addins` files are managed by user (implicitly or explicitly), so he knows what he is doing.

CharliePoole:
Well, it's not entirely a matter of opinion. :smiley: The engine,  for example, is distributed with an addins file and so are some runners. Also, it's pretty normal for an extension to come with an addins file.

It's true that a user could modify the addins file, but I don't think that should be the main use case.

I'm not saying we shouldn't do this, just that we have to think about how it will impact existing and future addins files whether distributed by NUnit or third parties.

The v2 framework driver is a good test case. It has no reason to exist outside of .net framework and it specifies an assembly by name so that other assemblies with it ate not scanned for extensions.

Also, note that a differing addins file is used for different package types, e.g. nuget, chocolatey, msi.

ChrisMaddock:
> So if non-wildcard extensions always throw when the target framework doesn't match, are we then saying that loading by name is no longer an option in the addins file that is distributed with an extension?

Thanks Charlie, this is a good catch. I even wrote the code to make that bit work myself. 🤦‍♂️ 

I think we need to think about this in a bit more detail, I'm travelling with work at the moment and don't have the time to do that right now, I'm afraid. It does sound like your original solution might have been the best way however @nvborisenko, apologies for that.

If anyone else on @nunit/engine-team (or @CharliePoole) wants to jump in and work out a way this will work as intended please go ahead - otherwise, afraid I'll have to get back to this, possibly at the weekend.

ChrisMaddock:
Ok. Apologies @nvborisenko - but I think your first implementation was what we wanted. Would you mind reverting to that? My fault, sorry.

Logic:
- This part of the code will see all extension assemblies, so should never throw
- The engine needs to silently decide which assemblies can be loaded on the platform, otherwise we have no way to package extensions targeting multiple platforms, where we don't know in advance which platform the engine will be running on.
- In the long run, perhaps we should combine `ValidateTargetFramework` with the logic in `ExtensionSelector.IsBetterVersionOf`. That's outside of the scope of this PR however.

Thanks for your work on this Nikolay - apologies it's taken me a few days to get back to. I'm going to go through now and leave a couple of more general comments too - with those and the change above, hopefully we can then get this merged! 🙂

src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
```diff
@@ -431,89 +431,99 @@ internal void FindExtensionsInAssembly(ExtensionAssembly assembly)
         {
             log.Info("Scanning {0} assembly for Extensions", assembly.FilePath);
 
-            ValidateTargetFramework(Assembly.GetEntryAssembly(), assembly);
+            if (ValidateTargetFramework(Assembly.GetEntryAssembly(), assembly))
```
ChrisMaddock: Can we rename this method, now we're changing the use of it a little? Maybe something like `CanLoadTargetFramework`?

src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
```diff
@@ -431,89 +431,99 @@ internal void FindExtensionsInAssembly(ExtensionAssembly assembly)
...
+                var currentFramework = RuntimeFramework.CurrentFramework;
+                assemblyTargetFramework = assembly.TargetFramework;
+                if (!currentFramework.CanLoad(assemblyTargetFramework))
                 {
-                    log.Info($"Assembly {assembly.FilePath} targets {assemblyTargetFramework.DisplayName}, which is not available. Assembly found via wildcard.");
-                    return;
+                    if (!assembly.FromWildCard)
+                    {
+                        throw new NUnitEngineException($"Extension {assembly.FilePath} targets {assemblyTargetFramework.DisplayName}, which is not available.");
+                    }
```
ChrisMaddock: According to my more general comment - this if clause is the bit that should be removed. My fault, sorry!

src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
```diff
@@ -523,28 +533,33 @@ internal void FindExtensionsInAssembly(ExtensionAssembly assembly)
...
             if (runnerAsm == null)
-                return;
+                return true;
 
             var extHelper = new TargetFrameworkHelper(extensionAsm.FilePath);
             var runnerHelper = new TargetFrameworkHelper(runnerAsm.Location);
             if (runnerHelper.FrameworkName?.StartsWith(".NETStandard") == true)
             {
-                throw new NUnitEngineException("Test runners must target .NET Core or .NET Framework, not .NET Standard");
+                log.Warning($"{runnerAsm.FullName} test runner must target .NET Core or .NET Framework, not .NET Standard");
```
ChrisMaddock: This one should still be an exception I think - this is looking at the framework of the runner. It's probably more of a sanity check than anything else however.

src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
```diff
@@ -523,28 +533,33 @@ internal void FindExtensionsInAssembly(ExtensionAssembly assembly)
...
+                return false;
             }
             else if (runnerHelper.FrameworkName?.StartsWith(".NETCoreApp") == true)
             {
                 if (extHelper.FrameworkName?.StartsWith(".NETStandard") != true && extHelper.FrameworkName?.StartsWith(".NETCoreApp") != true)
                 {
-                    throw new NUnitEngineException(".NET Core runners require .NET Core or .NET Standard extensions");
+                    log.Warning($".NET Core runners require .NET Core or .NET Standard extension for {extensionAsm.FilePath}");
```
ChrisMaddock: Suggestion: should this and the log below be `log.Info` instead of `Warning`? I think these log lines will be hit quite regularly, and don't indicate that anything is 'wrong' - as I would expect with `log.Warning`.

nvborisenko:
Reworked it to throw exception if runner is `netstandard`, in other cases - just silently ignore loading of extensions with `Info` log message.

CharliePoole:
Why is it OK with net standard runner?

nvborisenko:
It is not OK, added new test to verify it.

nvborisenko:
All review comments are resolved. Please review.

ChrisMaddock:
Thanks @nvborisenko - apologies, I've been away. Hoping to get back to this and the other outstanding PRs some point this weekend. :+1:

Thanks for your help here!

PR #555, opened 2019-02-22
Use ExtendedTextWriter in TestEventHandler

### LINKED ISSUE #547: --labels=Before ignores --nocolor
When using --nocolor the output of --labels still is colorized.

mikkelbu:
Replaces all usages of `new ColorConsole` in TestEventHandler with
matching calls to ExtendedTextWriter.

Edit: I've not added any unit tests. One solution would be to mock `System.Console` to see whether we are writing the `ForegroundColor` in color or not, but that is more testing `ColorConsoleWriter`/`ColorConsole`. A more direct approach would be to mock the `ExtendedTextWrapper` passed to `TestEventHandler` and check that we are only calling methods that accept a `ColorStyle`.

Fixes #547

### DIFF
--- a/src/NUnitConsole/nunit3-console.tests/TestEventHandlerTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/TestEventHandlerTests.cs
@@ -31,7 +31,7 @@ namespace NUnit.ConsoleRunner.Tests
     public class TestEventHandlerTests
     {
         private StringBuilder _output;
-        private TextWriter _writer;
+        private ExtendedTextWriter _writer;
 
         private string Output {  get { return _output.ToString(); } }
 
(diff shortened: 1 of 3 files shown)


CharliePoole:
What happens if you use the `--out` option and the tests write to console? Does the proper text get written without error even though it's not actually an ExtendedTextWriter getting the output?

mikkelbu:
As far as I can tell the output still looks corrected with `--out`.

If `--out` is specified then we return an `ExtendedTextWrapper` from `CreateOutputWriter`, and `ExtendedTextWrapper` derives from `ExtendedTextWriter`, but ignores all style indications.
https://github.com/nunit/nunit-console/blob/3cd4a2bb90fbd647d4e3c9a7038bf349bb881f5b/src/NUnitConsole/nunit3-console/ConsoleRunner.cs#L366-L377

CharliePoole:
@mikkelbu That makes sense then. i wrote that part, but forgot it. :smile:

ChrisMaddock:
Re-running CI - this is good to merge after that as far as I'm concerned. 🙂

PR #557, opened 2019-02-22
Disable CliFallbackFolder as a nuget source

### LINKED ISSUE #556: Appveyor CI failing due to nuget restore
Builds are failing on AppVeyor in nuget restore, see e.g. https://github.com/nunit/nunit-console/pull/549#issuecomment-465684710.

```
...
Installing NUnit 3.11.0.
  OK https://ci.appveyor.com/nuget/nunit/FindPackagesById()?id='System.Reflection.TypeExtensions'&semVerLevel=2.0.0 472ms
NU3005: Package 'Microsoft.NETCore.Platforms 2.0.0' from source 'C:\Program Files\dotnet\sdk\NuGetFallbackFolder': The package signature file entry is invalid. The central directory header field 'compression method' has an invalid value (8).
  GET https://www.myget.org/F/nunit-analyzers/api/v3/flatcontainer/nunit/index.json
  NotFound https://www.myget.org/F/nunit-analyzers/api/v3/flatcontainer/nunit/index.json 194ms
...
```

The problem with nuget is the same as in #515 which should be solved by #525, but it seems like the solution is not enough. 

AppVeyor had new images uploaded 11 February (see e.g. [AppVeyor Windows images update](https://www.appveyor.com/updates/2019/02/11/). There was no changes to msbuild, but nuget was bumped, so nuget now probably calls to the same msbuild version that was failing in #515.

mikkelbu:
Fixes #556

### DIFF
--- a/src/NUnitEngine/mock-assembly/mock-assembly.csproj
+++ b/src/NUnitEngine/mock-assembly/mock-assembly.csproj
@@ -9,6 +9,7 @@
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\..\nunit.snk</AssemblyOriginatorKeyFile>
     <LangVersion>7</LangVersion>
+    <DisableImplicitNuGetFallbackFolder>true</DisableImplicitNuGetFallbackFolder>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="NUnit" Version="3.11.0" />


ChrisMaddock:
Thanks for digging in to this Mikkel. 🙂

mikkelbu:
After approx. 10 force-pushes tonight I found a solution. By setting `DisableImplicitNuGetFallbackFolder` we avoid that [Microsoft.NET.NuGetOfflineCache.targets](https://github.com/dotnet/sdk/blob/2eb6c546931b5bcb92cd3128b93932a980553ea1/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.NuGetOfflineCache.targets) defines the `NuGetFallbackFolder`.

See https://github.com/dotnet/sdk/blob/2eb6c546931b5bcb92cd3128b93932a980553ea1/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.NuGetOfflineCache.targets#L14-L24 for some more context.

ChrisMaddock:
Woohoo - nice one Mikkel - thanks for digging in to that! 🎉

Issue #551, opened 2019-02-21
Make engine easier to Debug


CharliePoole:
I'm phrasing this as a set of questions. Depending on the answers, it might become a task.

When using the engine as a component of another runner via nuget, debugging is rather tricky.

1. Is there a reason that the pdb files are not distributed with the nuget packages? That's done with the adapter, for example.

2. If I build for debug, the version suffix changes in the assemblyinfo file. I'm wondering if use of "-dbg" at the end of the suffix is something folks are attached to or which seems like it would break anyone if we did it differently. I see a couple of alternatives
 2.1 Add -dbg or (Debug) to the name component name, making it a different package
 2.2 Don't change the name at all but just create a separate feed for debug builds.

3. How do you feel about having separate build targets for the engine and console?

These questions could apply to the framework as well, but it's the engine that I'm trying to debug under the GUI right now.

mikkelbu:
I ran into the same issue as in 1. when I did some debugging at work. But I think that from 3.10 we will actually get PDBs in the packages (at least in the zip-files) - at least I did when executing the target with "package". I guess this is a sideeffect of #388, but have not examined it in any detail.

Regarding 2. I think we could do it differently without harming people, and I don't know if there is a standard way to do this (meaning having debug builds in nuget).

Regarding 3. I won't mind separate build targets, but I do not know much about the history in this area.

CharliePoole:
I'm using the current master, which includes #388 and I'm not seeing pdbs in the nuget packages.

mikkelbu:
You are right Charlie. I was sure that I checked that the output contained PDBs because I was about to create an issue for this a couple of weeks ago, but I did a final check and then I thought that it was already solved on master. But I must have made an mistake when I did the check.

CharliePoole:
Let me throw in another item that I think would make life easier:

4. Allow specifying the productVersion as an argument to the build script. This would make it easier to create debug or release test builds for use with a third-party runner, without modifying the script or stepping on the pre-release prefixes used by NUnit.

CharliePoole:
So far, @mikkelbu has responded. I'd rather see some others comments before I put any effort into a PR. At the moment, based on Mikkel's feedback, I would be inclined to put in a PR that covered 1, 2 and 4.

ChrisMaddock:
I’d like to hear @jnm2’s thoughts on pdbs, after all the work he did in the framework here recently, I remember we discussed several options. We absolutely should have something!

[...]

PR #559, opened 2019-02-25
Specify productVersion as an argument

CharliePoole:
Part of issue #551.

This merely allows a user running the script to specify an additional argument that gives the product Version to be used. The assembly version is parsed out of the product version and the default values remain the same if nothing is specified.

Note that this setting is intended for local use and will actually be overridden on AppVeyor. If it were desired to use it there, more modifications would be needed.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -6,6 +6,7 @@
 
 var target = Argument("target", "Default");
 var configuration = Argument("configuration", "Release");
+var productVersion = Argument("productVersion", "3.10.0");
 
 var ErrorDetail = new List<string>();
 bool IsDotNetCoreInstalled = false;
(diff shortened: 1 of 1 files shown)


rprouse:
Looks like the AppVeyor builds started failing 10 days ago with 

```
Errors in C:\projects\nunit-console\src\NUnitEngine\mock-assembly\mock-assembly.csproj
    NU3005: Package 'Microsoft.NETCore.Platforms 2.0.0' from source 'C:\Program Files\dotnet\sdk\NuGetFallbackFolder': The package signature file entry is invalid. The central directory header field 'compression method' has an invalid value (8).
```

@ChrisMaddock [reported this back in December](https://help.appveyor.com/discussions/problems/18758-nuget-package-restore-failing-for-net-core-20-since-dec-13-windows-image-update) but it is back again 😦 

I see that @mikkelbu also [reported this to NuGet](https://github.com/NuGet/Home/issues/7414#issuecomment-466695804) but it doesn't look like it will be fixed because .NET Core 1.x is no longer supported.

I am going to override and merge, but we need to decide what to do about this. I will copy this into an issue for discussion.

PR #560, opened 2019-02-26
Comments documenting purpose of various runners

CharliePoole:
@ChrisMaddock Here are some comments as you requested, giving the overall purpose and usage of each runner. I also added some TODOs for things I am aware of as not working.

Doing this was a good idea. Aside from its intrisic value, it gave me some ideas about how to handle the creation of project results. If nobody has started on it after my weekend trip, I'll grab it and take a shot.

Essentially - in case someone else wants to mess with it - my idea is that the aggregation of project results, which could happen in various places, really should be done at the lowest possible level, i.e. the first opportunity. Each runner can detect whether it is handling a top-level TestPackage that is a project and do the aggregation. Most of the actual aggregating methods are already in a utility class, so only the check would need to be duplicated.

The reasonig behind this is that a runner at certain level can check what the lower-level runners have done, but there is no way for it to know what higher-level runners __will__ do after it returns. Setting a rule that this happens at the lowest possible level gives us a standard for which tests can be written.

### DIFF
--- a/src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
+++ b/src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
@@ -31,10 +31,18 @@ using NUnit.Engine.Services;
 namespace NUnit.Engine.Runners
 {
     /// <summary>
-    /// Summary description for ProcessRunner.
+    /// ProcessRunner loads and runs a set of tests in a single age process.
     /// </summary>
     public class ProcessRunner : AbstractTestRunner
     {
+        // ProcessRunner is given a TestPackage containing a single assembly
+        // multiple assemblies, a project, multiple projects or a mix. It loads
+        // and runs all tests in a single remote agent process.
+        //
+        // If the input contains projects, which are not summarized at a lower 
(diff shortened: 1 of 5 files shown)


src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
```diff
@@ -31,10 +31,18 @@
 namespace NUnit.Engine.Runners
 {
     /// <summary>
-    /// Summary description for ProcessRunner.
+    /// ProcessRunner loads and runs a set of tests in a single age process.
```
ChrisMaddock: "Single age process"

Is this a typo, or just terminology I've never come across before? 😅 Google couldn't help me with the terminology, but I can't work out what it could be a typo for either!

CharliePoole: "agent"

ChrisMaddock: Ah! That one didnt cross my mind! I can correct it now, thanks!

PR #564, opened 2019-03-05
Eliminate use of ProjectService in agents and AbstractTestRunner

### LINKED ISSUE #563: ProjectService is incorrectly initialized in agents
I noticed this when examining all the services needed by an agent, hoping to build a version of the agent that doesn't depend on the engine itself.

In `Program.cs` we create and initialize the `ProjectService` for the engine. That service

1. Is not intended to exist in the agent, because all projects are supposed to have been resolved before the agent is called and passed a TestPackage.

2. Is never otherwise referenced in the engine.

While initializing it does no harm, it's confusing and makes it difficult to simplify the agent. There's also a risk that someone will start using it, perhaps re-expanding projects that are already expanded once.

CharliePoole:
Fixes #563

After eliminating the reference in Program.cs, I realized that the reference in AbstractTestRunner was unused. That makes sense because AbstractTestRunner is the base class for all ITestEngineRunners. If the reference were needed, then the service would be needed in the agent as well.

### DIFF
--- a/src/NUnitEngine/nunit-agent/Program.cs
+++ b/src/NUnitEngine/nunit-agent/Program.cs
@@ -123,7 +123,6 @@ namespace NUnit.Agent
             //log.Info("Adding Services");
             engine.Services.Add(new SettingsService(false));
             engine.Services.Add(new ExtensionService());
-            engine.Services.Add(new ProjectService());
             engine.Services.Add(new DomainManager());
             engine.Services.Add(new InProcessTestRunnerFactory());
             engine.Services.Add(new DriverService());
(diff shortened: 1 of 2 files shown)


CharliePoole:
Right.

CharliePoole:
Since the change is still pending, I also added a small change removing the SettingsService reference from the engine. This fixes #566 as well.

src/NUnitEngine/nunit-agent/Program.cs
```diff
@@ -121,9 +121,7 @@ public static void Main(string[] args)
 
             // Custom Service Initialization
             //log.Info("Adding Services");
-            engine.Services.Add(new SettingsService(false));
             engine.Services.Add(new ExtensionService());
-            engine.Services.Add(new ProjectService());
```
jnm2: Can the SettingsService and ProjectService classes also be deleted?

CharliePoole: No. They are used when the engine is used by the runner. It's only the agent that uses a "downgraded" version of the engine.

PR #565, opened 2019-03-05
Eliminate -dbg suffix from version

CharliePoole:
This PR is part of issue #551, point 2.

Simply eliminating it, as I have done here, seems to do the job for me. I no longer have to contend with unwanted changes to AssemblyInfo.cs simply because I build Debug from the command-line.

If we wanted to actually __publish__ debug__ builds through AppVeyor, that would require more work.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -21,7 +21,6 @@ var version = dash > 0
     : productVersion;
 
 var isAppveyor = BuildSystem.IsRunningOnAppVeyor;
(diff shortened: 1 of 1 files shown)


CharliePoole:
Yeah, I tried to keep those two changes separate, but I think you're right.

mikkelbu:
I'll merge this, as it only affects AppVeyor + Debug, which we currently do not do.

PR #572, opened 2019-03-15
Fixed support for space characters in the work directory path

### LINKED ISSUE #571: Space characters in the work directory path are not properly handled
When nunit3-console constructs the command line to run an agent, it does not properly double-quote the work directory path, that causes the runner to fail.

Here is a sample internal trace log:

```
14:40:58.747 Debug [ 1] TestAgency: Command line: "C:\Contoso\bin\Release With Spaces\TestPackage\nunit-agent.exe" f3ab647d-8aa4-4cb9-8032-a55f48aae46e tcp://127.0.0.1:60774/TestAgency --pid=16708 --trace:Debug --work=C:\Contoso\bin\Release With Spaces\TestPackage
14:40:58.748 Debug [ 1] TestAgency: Waiting for agent {f3ab647d-8aa4-4cb9-8032-a55f48aae46e} to register
14:40:59.350 Error [ 1] ProcessRunner: Failed to run remote tests Unable to acquire remote process agent
```

As you may see, the last part:
```
--work=C:\Contoso\bin\Release With Spaces\TestPackage
```

contains the path without quotes.

The problem is in the "nunit-console-3.9\src\NUnitEngine\nunit.engine\Services\TestAgency.cs" source file inside LaunchAgentProcess method. 

Current code looks like this:

```
agentArgs += " --work=" + workDirectory;
```

The fixed version:

```
agentArgs += " --work \"" + workDirectory + "\"";
```

For now, we'll be using our custom fixed nunit-console build, but it would be great if the issue is fixed in the next official release. Thank you!

Funbit:
Fixes #571 

<sup>Edit by Mikkel: Updated original message to auto close the issue when merged to master.<sup>

### DIFF
--- a/src/NUnitEngine/nunit.engine/Services/TestAgency.cs
+++ b/src/NUnitEngine/nunit.engine/Services/TestAgency.cs
@@ -186,7 +186,7 @@ namespace NUnit.Engine.Services
             if (debugAgent)
                 agentArgs += " --debug-agent";
             if (workDirectory != string.Empty)
-                agentArgs += " --work=" + workDirectory;
+                agentArgs += " --work \"" + workDirectory + "\"";
 
             log.Info("Getting {0} agent for use under {1}", useX86Agent ? "x86" : "standard", targetRuntime);
 


src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -186,7 +186,7 @@ private Guid LaunchAgentProcess(TestPackage package)
             if (debugAgent)
                 agentArgs += " --debug-agent";
             if (workDirectory != string.Empty)
-                agentArgs += " --work=" + workDirectory;
+                agentArgs += " --work \"" + workDirectory + "\"";
```
mikkelbu: Will this work? If I look at the agent then it looks like it expects a `=` immediately after `--work`.

https://github.com/nunit/nunit-console/blob/a64809c808b134693936c28df201c65a5bb230e2/src/NUnitEngine/nunit-agent/Program.cs#L78-L81

CharliePoole: Good catch!

stevenaw: Thanks for this @Funbit 
I don't know much about this part of the code, but how will this behave if someone tries to pass in a path already wrapped in double quotes?

CharliePoole: This isn't intended to be run by users. It's only executed from Test Agency.

stevenaw: Ah ok. Thanks for explaining!

Funbit: In the current version, even if user tries to pass a quoted working directory to the nunit3-console.exe, the console tool itself gets the correct path (value of the workDirectory variable), but when it will try to run the agent using the code above - the path will become invalid. My fix solves that issue.

CharliePoole: That's because quotes are stripped from the command-line before being passed to the program.

ChrisMaddock:
Hey @Funbit - thanks for doing this PR. I want to get this fix into the 3.10 release that we're hoping to do soon, so I'm going to take your commit and move it to a new PR, with the fix applied. We can then get it merged into master and the release on track, and the work will still be accredited to you. Hope that's ok!

ChrisMaddock:
Moved to https://github.com/nunit/nunit-console/pull/585

Funbit:
Sure, no problem, @ChrisMaddock ! The important part is that the project will become better anyway! :)

PR #575, opened 2019-03-17
Add --testparam to enable test parameters to contain semicolons

jnm2:
Fixes https://github.com/nunit/nunit-console/issues/23. Feel free to ignore till after the release is done (unless there's a reason you'd like to get it in 3.10).

Visual updates:

image

image

Let me know if there's anything you'd like to do differently.

### DIFF
--- a/src/NUnitConsole/nunit3-console/CommandLineOptions.cs
+++ b/src/NUnitConsole/nunit3-console/CommandLineOptions.cs
@@ -154,11 +154,13 @@ namespace NUnit.Common
 
         // Error Processing
 
+        public IList<string> WarningMessages { get; } = new List<string>();
+
         public IList<string> ErrorMessages { get; } = new List<string>();
 
-#endregion
+        #endregion
 
-#region Public Methods
+        #region Public Methods
 
         public bool Validate()
         {
--- a/src/NUnitConsole/nunit3-console/CommandLineOptions.cs
+++ b/src/NUnitConsole/nunit3-console/CommandLineOptions.cs
@@ -292,28 +292,19 @@ namespace NUnit.Common
             this.Add("where=", "Test selection {EXPRESSION} indicating what tests will be run. See description below.",
                 v => WhereClause = RequiredValue(v, "--where"));
 
-            this.Add("params|p=", "Define a test parameter.",
+            this.Add("params|p=", "Deprecated and will be removed in a future release. Please use --testparam instead.",
                 v =>
                 {
                     string parameters = RequiredValue( v, "--params");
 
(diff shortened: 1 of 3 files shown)


jnm2:
The docs will need to be updated after this is merged.

jnm2:
I also want to send a pull request to Cake which switches to --testparam if the right version is detected and throws an exception if it had to use --params and a semicolon was specified.

jnm2:
Don't know if this fits, but I sometimes use the word 'hesitations' for that. =)

Your points are so good that I don't really want to merge this till I correct those things. I like your phrase "key-value pair separated by an equals sign."

> I found it very strange to have parameters where the key has leading/trailing whitespaces - or even worse where the key only consist of whitespace

Me too. Since we're introducing a new parameter, this is the best time to add additional rules. Shall we error for these instead of the previous behavior? /cc @nunit/engine-team 

> Also the tests show that the `=` in `--testparam=VALUE` can also be a `:`.

This is true and this is how I had it in the help section at first, but I can't just write whatever I want to in the help section. There's some magic going on. If you specify `"testparam:`, the help shows `--testparam[=VALUE]`. This wasn't really intuitive to me, so I copied `--param`. Looks like we're locked into the same `=VALUE` that `--params` was for the time being. Maybe we can make the system more flexible in another PR.

mikkelbu:
> Looks like we're locked into the same =VALUE that --params was for the time being. Maybe we can make the system more flexible in another PR.

Sounds like a good compromise 😄.

jnm2:
I forgot I had the MSBuild commit in all three PRs. Removed it, ready for review.

ChrisMaddock:
Will come back to this after 3.10, if that's ok! 🙂

jnm2:
For sure!

src/NUnitConsole/nunit3-console/CommandLineOptions.cs
```diff
@@ -292,37 +285,33 @@ protected virtual void ConfigureOptions()
...
-                            string name = param.Substring(0, eq);
-                            string val = param.Substring(eq + 1);
-
-                            TestParameters[name] = val;
-                        }
+                        ApplyTestParameter(param);
                     }
                 });
 
+            this.Add("testparam=", "Followed by a key-value pair separated by an equals sign. Test code can access the value by name.",
```
rprouse: Should we add a short version for this like `tp`? `--testparam` is long to type if you have many of them.

jnm2: Sounds like a good idea.

Issue #266, opened 2017-07-19
Planning: replacing remoting with custom protocol


jnm2:
The SocketException thing continues to give us grief and it's gotten to the point where I feel I'm sinking too much time into something we don't ultimately even want to keep. I'd like to put some energy towards the custom protocol idea.

Goals:

 - Minimalism over extensibility
   Protocol compatibility is not an issue since the console and agent are released as a unit. Let's not overengineer this protocol. That includes performance, unless it's identified as a real-world bottleneck.

 - .NET Standard compatibility
   This feature should never become a reason we can't build the console runner for platforms other than .NET Framework.

 - Explore the ability to run the agent on remote machines
   Lowest priority; this is something we want to do eventually. Let's just try not to come up with a design that we'd have to scrap as soon as we want to run on remote machines.

I don't expect there will actually be much coding to do, mostly design. I'm fine doing all the coding though of course the more the merrier, but it would be great to have everyone's thoughts on the design.

jnm2:
Marking as pri:high because I haven't been able to kick those SocketExceptions entirely and there is no workaround for folks.

jnm2:
@nunit/engine-team Marking as pri:high because I haven't been able to kick those SocketExceptions entirely and there is no workaround for folks.

jnm2:
### Discovery

When running on the same machine, discovery happens when the console passes its connection info the the agent it starts. What would it look like if an agent was on a remote machine? Would we assume in that scenario the agent instances would be started manually with `--listen --port=12345` and that the console runner would be manually given a list of agent endpoints (name/ip + port)? If started manually, the agent instances should be stopped manually. Should we have default port if not specified?

Or would we want some kind of autodiscovery?

#### Xamarin agent

This is a very real, very serious thought I've had as I used NUnit with a Xamarin project. We should eventually provide a standalone Xamarin agent application which receives assemblies over the wire from the NUnit engine running on your development machine and tests them, sending results back in realtime.

Let's make sure we don't block this scenario.

jnm2:
We could easily implement a UDP multicast autodiscovery protocol like SQL Server instance discovery uses.

(I have implemented the client side of the SQL protocol from scratch because `SqlDataSourceEnumerator` blocks for an arbitrary amount of time and I wanted a non-blocking `ObservableCollection` that notifies immediately as each server replies and keeps listening until you shut it down.)

jnm2:
[...]

PR #576, opened 2019-03-18
Run TCP server in agent with custom protocol rather than remoting

jnm2:
This is a draft PR so that I can get your feedback. `.\build -t test` is working. This will fix https://github.com/nunit/nunit-console/issues/266. 

The only thing left to do is to implement StopRun, which is called out in a comment.
And a lot of testing.

...and I clicked the wrong side of the button and didn't actually make a draft PR. Oh well, wipbot is still here. =)


CharliePoole:
Should it be "Replace remoting for ProcessRunner?" We still remote across AppDomains for .NET Framework.

CharliePoole:
What happens if the user test does a native write to console, for example, from a C or C++ program?

src/NUnitEngine/nunit.engine.api/TestFilter.cs
```diff
@@ -32,9 +29,6 @@ namespace NUnit.Engine
     /// In the console runner, filters serve only to carry this
     /// XML representation, as all filtering is done by the engine.
     /// </summary>
-#if !NETSTANDARD1_6
-    [Serializable]
```
jnm2: This must be put back so that AppDomain remoting still works on .NET Framework.

jnm2:
@CharliePoole 

> What happens if the user test does a native write to console, for example, from a C or C++ program?

Hmm, good thought. Or if someone writes to `Console.OpenStandardOutput()` in C#. That would completely break the connection. `SetStdHandle` might work, but I don't want to make this platform-dependent.

The reason I went with standard streams is that I found that initialization information had to be shared over them to initialize a TCP server in the agent process (telling it to bind to 127.0.0.1 and receiving back the port that Windows assigned). Down the road when the agent is started on another machine or before the console process, we'll want the TCP server to be in the agent process.

The next logical step would then be to only use stdio for initialization and make sure no user code is loaded into the process until a TCP connection is established and stdio is no longer used.

(A marginally easier option would be for me to set up in and out anonymous pipes and send the pipe handle to the agent. I think I should still go with TCP instead because we've been talking about wanting the user to be able to start the agent on another machine and connecting to it.)

jnm2:
Also, I should put the TCP server logic in the engine assembly so that e.g. a Xamarin project running in an emulator can run it.

jnm2:
@nunit/engine-team I've addressed the comments above in the two new commits. What do you think?

Any time the agent process is started, it starts a new TCP server bound to localhost and whatever free port Windows assigns. The agent prints this port to stdout when it starts up and the parent process reads it (just like a user would after manually starting `nunit-agent.exe`) and connects to it. The parent process writes `stop` to the stdin of the agent process to tell it to shut down the TCP server and exit (just like a user would).

`.\build -t test` is still working. The thing still left to do is to implement StopRun which is called out in a comment.

jnm2:
Looks like there's a failure on just Linux that I need to follow up then.

ChrisMaddock:
This is also very exciting! Hopefully will get back to having a proper look at these sooon.

jnm2:
Ooh, hi @dnfclas! @rprouse This is the first time @dnfclas hasn't recognized me since I first signed: `https://github.com/dotnet/reactive/pull/232#issuecomment-233788955`. Is this by design?

jnm2:
@nunit/engine-team I've completed the implementation and it's ready for review and testing whenever things settle down for you. The only thing I know of that I can work on is to add unit tests on AgentClient and AgentServerConnection.

Our `NUnit.Engine.Tests.Integration.RemoteAgentTests.Explore_does_not_throw_SocketException` test turns out to be extremely useful. It picked up two separate issues in CI, one of which is explained in that last commit.
The first fix that occurred to me is working, but after that I sank two hours into trying to understand _why_ the reentry is happening on Mono. As far as I'm concerned we shouldn't have to cope with reentry in that code; the original code should be fine. Gonna bet it's a Mono bug and maybe newer or older versions won't have this issue.

src/NUnitEngine/nunit-agent/Program.cs
```diff
@@ -130,58 +122,65 @@ public static void Main(string[] args)
...
-                if (Agent.Start())
-                    WaitForStop();
-                else
+                using (var server = AgentServer.Start(IPAddress.Loopback, engine.Services.GetService<ITestRunnerFactory>()))
                 {
-                    log.Error("Failed to start RemoteTestAgent");
-                    Environment.Exit(AgentExitCodes.FAILED_TO_START_REMOTE_AGENT);
+                    // Write to the console to let the parent process know which port Windows assigned
+                    Console.WriteLine("Listening on " + server.ListeningOn);
+                    log.Info("Listening on " + server.ListeningOn);
```
ChrisMaddock: I'm a little confused as to why writing to stdout is necessary, sorry. Can the engine not just bind to the port first, and pass the port number in on the agent command line args?

src/NUnitEngine/nunit-agent/Program.cs
```diff
@@ -130,58 +122,65 @@ public static void Main(string[] args)
...
+                        log.Info($"Ignoring unrecognized command '{line}' from standard input.");
+                        Console.WriteLine($"Unrecognized command '{line}'.");
+                    }
                 }
             }
             catch (Exception ex)
             {
-                log.Error("Exception in RemoteTestAgent", ex);
-                Environment.Exit(AgentExitCodes.UNEXPECTED_EXCEPTION);
+                log.Error("Unexpected exception", ex);
```
ChrisMaddock: Appreciate this is existing code - I think we might want to use `ExceptionHelper.Flatten()` or something here - make sure we do get all the logging info if things go wrong!

src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
```diff
@@ -1,5 +1,5 @@
 // ***********************************************************************
-// Copyright (c) 2011-2014 Charlie Poole, Rob Prouse
+// Copyright (c) 2011�2019 Charlie Poole, Rob Prouse
```
ChrisMaddock:
```suggestion
// Copyright (c) 2011-2019 Charlie Poole, Rob Prouse
```

Whichever type of dash this is, it borks in my VS diff tool! 😂 Swap for a normal hyphen?

src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
```diff
@@ -21,10 +21,10 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
-#if !NETSTANDARD1_6 && !NETSTANDARD2_0
+#if !NETSTANDARD // Dependency on RuntimeFrameworkService
```
ChrisMaddock: Thanks for adding the comment! I think we'll want to look at how we might be able to refactor to solve that maybe... (Outside of this PR, of course!)

src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
```diff
@@ -33,37 +33,35 @@ namespace NUnit.Engine.Runners
     /// <summary>
     /// ProcessRunner loads and runs a set of tests in a single agent process.
     /// </summary>
-    public class ProcessRunner : AbstractTestRunner
+    public partial class ProcessRunner : AbstractTestRunner
```
ChrisMaddock: partial?

src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
ChrisMaddock: Ahh, target framework version. Making that bit .NET Standard friendly is gonna be fun! 😄

src/NUnitEngine/nunit.engine/Agent/AgentProcess.CreateStartInfo.cs
```diff
@@ -0,0 +1,139 @@
+﻿// ***********************************************************************
```
ChrisMaddock: Nice extraction! I'm presuming all this class is just copied directly over and doesn't need reviewing!

src/NUnitEngine/nunit.engine/Agent/AgentClient.cs
```diff
@@ -0,0 +1,174 @@
...
+        public void Dispose()
+        {
+            _stream.Dispose();
+        }
+
+        public void Connect(TestPackage package)
+        {
+            if (package == null) throw new ArgumentNullException(nameof(package));
+
+            const byte version = 1;
```
ChrisMaddock: Nit: How about `protocolVersion`? What was your thoughts behind this by the way? At the point where we only package agent/engine together, this doesn't seem necessary - future proofing for potential separately packaged (remote?) agents?

src/NUnitEngine/nunit.engine/TestEngineResult.cs
```diff
@@ -152,5 +151,25 @@ public void Add(XmlNode xml)
...
+        {
+            var value = new TestEngineResult();
+
+            var count = reader.ReadByte();
+
+            for (var i = 0; i < count; i++)
+                value._xmlText.Add(reader.ReadString());
+
+            return value;
+        }
```
ChrisMaddock: Should this class still have the Serializable/NonSerialized attributes, now we're serializing 'by hand'?

src/NUnitEngine/nunit-agent/Program.cs
oznetmaster: I agree with @ChrisMaddock.  It would be better to not write to stdout.

jnm2: We'll get to the point hopefully very soon where the user is starting the TCP server first and passing the port number as command-line argument to the console runner. I'd like to keep them decoupled in this way.
Let's say we did that though. If we had a separate mode where the console runner bound to a port, we'd have to rely on the bound socket getting inherited into the child agent process. We'd have to duplicate the socket handle as inheritable and close the non-inheritable one. It felt much cleaner to me to have the console runner communicate with the agent just like a user would.

What is it about writing to stdout that could be problematic? I particularly liked this part and thought it would even be a selling point 😛, but of course I'm willing to do whatever.

src/NUnitEngine/nunit-agent/Program.cs
```diff
@@ -22,42 +22,32 @@
...
-        static Process AgencyProcess;
-        static RemoteTestAgent Agent;
         private static Logger log;
 
-        /// <summary>
-        /// The main entry point for the application.
-        /// </summary>
         [STAThread]
-        public static void Main(string[] args)
+        public static int Main(string[] args)
```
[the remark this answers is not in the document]
jnm2: Ah, good to know.

src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
jnm2: I've been using the n-dash (preferred for ranges) other places in the NUnit org, but I could swap it. The problem is that the file was saved with Western European codepage encoding instead of UTF8. Is it alright if I fix the problem by saving it as UTF8?

src/NUnitEngine/nunit-agent/Program.cs
CharliePoole: I'm confused by this as well. The existing remoting facility runs on top of tcp which US set up without other communication channel. The server has a single endpoint where clients register and are then given a custom connection on which they receive commands and return results. All that is done by remoting. It's just that we need to do it in our own code now.

src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
jnm2: AgentProcess.CreateStartInfo.cs was ProcessRunner.CreateStartInfo.cs for a few commits. Thanks for catching!

CharliePoole:
Is it really necessary to replace remoting, rather than encapsulating the comments channel and having multiple capabilities in the engine? Each agent only needs to implement one, of course.

src/NUnitEngine/nunit.engine/Agent/AgentProcess.CreateStartInfo.cs
jnm2: The class used to create the process rather than just the start info. I pulled that logic out which necessitated rearranging most of this, actually. I'll see if I can insert a commit showing the diff.

src/NUnitEngine/nunit-agent/Program.cs
oznetmaster: Many protocols do the same thing.  FTP, for instance.  They have a pre-defined port for initiating communications, and then negotiate a port to use for the commands and data.

We could do the same, and then allow a command line override for that port if really necessary.

src/NUnitEngine/nunit.engine/Agent/AgentClient.cs
jnm2: It's not super necessary, more to make sure this doesn't get forgotten till it's too late.

I'm hoping to do a proof-of-concept Xamarin app that references the engine and runs the agent TCP server in an Android emulator and have the console connect up to it. But nothing prevents the Xamarin app's referenced NUnit engine version from differing from the console runner's engine version.
tl;dr Even if the agent is packaged with the engine, people might make two versions of the engine talk to each other.

Maybe they should exchange engine versions instead of introducing a protocol version?

jnm2:
@CharliePoole For .NET Core where remoting does not exist, it is necessary to fully replace remoting in all communications with the agent. It's simpler then to have only one method of communicating with the agent rather than maintaining separate paths for .NET Framework and .NET Core.

Also, the motivation to get rid of cross-process remoting came before .NET Core IIRC. We still get the occasional socket exception due to remoting being a really bad application protocol that doesn't provide for clean shutdown.

As we discussed at the top of this PR, remoting still exists in the .NET Framework build, just not for the purpose of communicating with the agent.

src/NUnitEngine/nunit.engine/TestEngineResult.cs
jnm2: Yes, so that remoting can still move this between AppDomains for the .NET Framework build's AppDomain runner. (https://github.com/nunit/nunit-console/pull/576#discussion_r266556021)

src/NUnitEngine/nunit-agent/Program.cs
jnm2: That is still seriously complex compared to what it could be.

- Runner spins up a TCP server and passes the port it gets to the child agent process when it starts
- Agent connects back to the runner using TCP
- Agent spins up a TCP server and passes the port it gets to the parent process via the existing connection
- Runner and agent both disconnect from the original connection and begin talking on the second one

None of this seems necessary to me. In this scenario, the runner is not behaving like the thin client that I envision it to be. The runner only lasts as long as the test run it's executing. Once we start having agents that live longer, it will make less sense for all that to be part of the protocol.

It would be cool if we could keep it simplified so that we always have a client mindset when talking about the console runner and a server mindset when talking about the agent.
Speaking of what is confusing, this will best model people's expectations of such an agent, too. Other build and test agents seem to typically be structured this way.

oznetmaster:
I will note that I am actively watching this, since CF also does not support remoting, and I intend to modify my CF build to use this once it is working.

src/NUnitEngine/nunit-agent/Program.cs
jnm2: I'm picturing the responsibilities as linear, fitting both scenarios without complicating the inner part (in bold).
The first three points are analogous between the two scenarios, and the protocol responsibility is separated in such a way that the final point (our custom protocol) can be entirely independent of the scenario.

Scenario A: User wants tests to run tests in an existing agent 

- The user installs the agent as a service on a dedicated test server,
  or starts an Android emulator and opens an NUnit app hosting an agent,
  or starts an instance of VS with an agent extension to do UI tests in the VS process.

- In each of these cases, the user knows the IP address and port number due to either choosing the port number when starting/scheduling the agent, or seeing the port number and IP address when the Xamarin app starts, etc.

- The user passes the IP address and port number (or just port number if localhost) to the console runner.

- **The runner acts as a fairly thin client.**

Scenario B: User wants the runner to start and stop an agent

- The user starts the console runner without specifying an existing agent.

- The console runner acts on the user's behalf by managing the lifetime of the agent process.

- The console runner knows the IP address of the agent process because the runner process started it.

- **The runner acts as a fairly thin client.**

jnm2:
@oznetmaster Good to know, thanks! Your feedback on that is valuable.

src/NUnitEngine/nunit-agent/Program.cs
jnm2: I'm not sure where `ExceptionHelper.Flatten` lives but I'm happy to copy that over. Is that the framework repo?

src/NUnitEngine/nunit.engine/Agent/AgentProcess.CreateStartInfo.cs
jnm2: Figuring out how to retroactively preserve the diff in source control is a bit tricky, but I can do it.
Is it good enough for me to link https://github.com/nunit/nunit-console/pull/576/commits/520d6c5c280ae1cdb6d45f97416edb928177f999#diff-d00b5d3587d1cbdc42a626eb65d5513aL157? Enough changed that an inline diff won't be as good as side-by-side anyway.

I added my trusty `StringBuilder.EscapeProcessArgument` [extension method](https://github.com/nunit/nunit-console/pull/576/files/bf9b0812ba817b540b68cbfdc3599719b02f5eda#diff-01247bb8fc9f2620caf98937cfe7ffa5R37) and [tests](https://github.com/nunit/nunit-console/pull/576/files/bf9b0812ba817b540b68cbfdc3599719b02f5eda#diff-30360af2195b50bbf97ba3cca55947e9R30).

src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
jnm2: Yes :D I'm willing to help with that!

In the latest force push, I added `// Dependency` comments on the files in the Agent folder, too.

src/NUnitEngine/nunit-agent/Program.cs
CharliePoole: I take it that by agent, you are referring only to the existing nunit-agent executable used to run in a separate process. I was thinking more generally, in the sense that all execution is performed by an agent, but I'll stick with the restricted definition for now so we're on the same page.

In the relationship between the TestAgency and agents, it's the  agency which is the server, even though the agents are seen as "doing something" for the runner. This apparent reversal is fairly common in designs where the process to perform a task is actually spawned by the process that needs something done.

This isn't just a distinction of terms because it's the server's job, not the client's, to provide communication endpoints. The TestAgency "serves" multiple clients while the agent   only exists for short time to deal with a single runner.

We had always envisioned having other kinds of agents, on separate machines, functioning as actual servers, but we aren't doing that here - or if we are, then I'd say too much for one PR.

WRT encapsulation rather than replacement, I agree that the latter is simpler. But it leaves unsolved a fundamental problem in the engine, which is that it can only communicate in one way.

oznetmaster: It was my understanding that was the goal, to allow the agent to run on separate machines functioning as an actual server.

That is what I need for CF, and what I think @jnm2 meant when he discussed a Xamarian emulator and an Android device.

CharliePoole: @jnm2 Is that what you are trying to do? At one time we had a server project, which has been removed from the current repo. Back then, we used "server" to mean what @oznetmaster is talking about and "agent" for the current single-use, TestAgency-initiated processes.

I hadn't seen any indication that you were hoping to have one "agent" that works both ways, but since @oznetmaster 's comment seems to suggest that, I'll throw in my two cents, which is that you would be trying to do too much in one bit of code. Two such different "agents" would obviously share a lot of code - in fact, the code that I was calling "agent" or "core" in a separate spike I did. But they would manage their communications entirely differently. IMO, you'll have better result by limiting the PR to one thing.

src/NUnitEngine/nunit-agent/Program.cs
ChrisMaddock: Sorry - it's ExceptionHelper.BuildMessage/BuildStackTrace etc - I'd mis-remembered the names. (Flatten is the private methods in that class instead!)

It should actually probably be baked lower into the Logger classes I guess...probably out of scope for this PR. Let's leave this as it. 🙂

src/NUnitEngine/nunit.engine/Agent/AgentProcess.CreateStartInfo.cs
ChrisMaddock: Yeah, looks good! I just meant I wasn't going to worry about reviewing all the code, if it was just being moved around. 🙂

src/NUnitEngine/nunit-agent/Program.cs
```diff
@@ -130,58 +122,65 @@ public static void Main(string[] args)
...
-                    Environment.Exit(AgentExitCodes.FAILED_TO_START_REMOTE_AGENT);
+                    // Write to the console to let the parent process know which port Windows assigned
+                    Console.WriteLine("Listening on " + server.ListeningOn);
+                    log.Info("Listening on " + server.ListeningOn);
+
+                    Console.WriteLine("Type 'stop' to exit.");
+
+                    while (true)
+                    {
+                        var line = Console.ReadLine();
```
rprouse: You are using a TCP channel to send results back to the parent process, but you are using the captured STDIN and STDOUT to communicate too. For example the port number above and the stop command here. That seems to defeat your goal of allowing communications with remote  agents like mobile devices and emulators.

Am I misreading this? Is this just another way to stop? I see you have a stop command too.

src/NUnitEngine/nunit.engine/Agent/AgentClient.cs
rprouse: I think the engine version makes sense. We could allow different versions to talk to each other and just prevent it when we know we are making a breaking change.

src/NUnitEngine/nunit-agent/Program.cs
rprouse: I did think that the console would open the port and listen, then pass the port on the command line to the agent. I can see some of your points on wanting to do it the other way around, but it does introduce some difficulty. What if the agent is on a different machine or device?

src/NUnitEngine/nunit.engine/Agent/AgentClient.cs
```diff
@@ -0,0 +1,174 @@
...
+            _formatter.Serialize(_stream, package);
+            _stream.Flush();
+        }
+
+        public TestEngineResult Load()
+        {
+            WriteCommandType(AgentCommandType.Load);
+            _stream.Flush();
+
+            return ReadTestEngineResult();
```
rprouse: This shows why opening the port on the agent is easier, it fits well with the current architecture of the engine where we send commands to the agent and wait for the response. I'd done a much smaller spike on this before but with the engine opening the port and ended up jumping through hoops because it was somewhat backwards.

src/NUnitEngine/nunit.engine/Agent/AgentCommandType.cs
```diff
@@ -21,18 +21,16 @@
...
 
-namespace NUnit.Engine
+namespace NUnit.Engine.Agent
 {
-    /// <summary>
-    /// The ITestAgency interface is implemented by a TestAgency in 
-    /// order to allow TestAgents to register with it.
-    /// </summary>
-    public interface ITestAgency
+    internal enum AgentCommandType : byte
```
rprouse: This is odd that it got picked up as a file rename 😄

src/NUnitEngine/nunit.engine/Agent/AgentProcess.CreateStartInfo.cs
```diff
@@ -0,0 +1,139 @@
...
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+
+namespace NUnit.Engine.Agent
+{
+    internal partial class AgentProcess
+    {
+        private static ProcessStartInfo CreateAgentProcessStartInfo(TestPackage package, RuntimeFrameworkService runtimeFrameworkService)
```
rprouse: These methods are all static. Instead of breaking them out into a partial, should they be moved into a utility class? I like that they've been broken out though 😄

src/NUnitEngine/nunit-agent/Program.cs
CharliePoole: A bit of history, in case it helps...

We started out with two concepts... agents and servers...

Agents were to be those local processes initiated by the TestAgency, a server itself, which called back to the test agency at an endpoint contained in their argument list, received a job to run, ran it to completion, and exited when the tests were finally unloaded.

Servers, for which we had a separate project, were intended to be long-running and findable by the desktop runner. They were aimed at running on a separate machine although they could obviously be tested on the same machine. The desktop runner and engine would function as a client to such servers. They would clean up afterward and then accept new jobs.

I actually tried to do what @jnm2 is doing first. Having the short-term agents function as servers required extra communications. I saw some code written by @jcansdale for his original open source extension, where the roles were reversed, with the assembly running the tests functioning as a client to a little server embedded in the extension. I took that idea and never looked back. (Thanks @jcansdale ) 

For a long time we maintained code for what was basically a null server. It ran, you could contact it but it did nothing. We eventually dropped that project - I'm not sure if we did it for NUnit 3 or prior to that time - because nobody was working on int.

IMO, the long-running server and the short-running agent are two different things. The former is obviously more complicated. I think we need both. Most of the code will be the same, just combined in different ways and the separation will make the more complex server process just a little less complex.

BTW, this has nothing to do with whatever architecture current CI systems use for "agents". This goes back to something like 2004 and the terms were pure inventions at the time.

oznetmaster:
To complicate matters further, I have a totally new environment that I need to run NUnit in, that seemingly will also require the TCP connectivity.

Briefly, the entire nunit assembly will be loaded by, and running under, another process which is running as a server on mono on linux.

There is no way to communicate with the child process except through some form of TCP connectivity.  It can be a direct TCP connection, or through a REST API to a webpage, through WebSockets, or whatever.

There can be no agent involved on the test side, unless the agent can run as the child process, and then load the test assembly, which might be possible, if a custom agent was written.

This seems to imply that the test assembly must be the one listening for the TCP/WebSocket/HTML-REST connection.

There cannot be a STDIN/STDOUT redirection involved since linux services do not make stdin and stdout available to their child processes, as far as I can tell.

oznetmaster:
@jnm2 I would really like to work with your code as soon as you think it is appropriate.

Since I am sure I will have to fiddle with it because of the strangeness of my two environments, I can be flexible on what "appropriate" means.

src/NUnitEngine/nunit-agent/Program.cs
jnm2: The way I have it here, starting and stopping of the TCP server itself is set up using standard streams, and test runs (and stopping of test runs) is done over TCP. I was thinking of these as two distinct layers that don't mix.

ChrisMaddock:
Hey Joseph - what do you think's best to do with this PR? IIRC, you were looking at doing this in a series of incremental steps rather than one big bang now - does it make sense to close this one, or is it still useful to have an overall view of the changes? 🙂

ChrisMaddock:
Closing this for the same reason as #774. Thanks for all the work that went into it. 🙂

PR #579, opened 2019-03-18
Resolve MSBuild when only 16.0 or only a preview of VS is installed

jnm2:
I'm pulling this commit out of my two open PRs, similar to https://github.com/nunit/nunit/pull/3194.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -153,13 +153,37 @@ Task("UpdateAssemblyInfo")
 // BUILD ENGINE AND CONSOLE
 //////////////////////////////////////////////////////////////////////
 
-MSBuildSettings CreateMSBuildSettings(string target) => new MSBuildSettings()
-    .SetConfiguration(configuration)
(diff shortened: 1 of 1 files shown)


build.cake
```diff
@@ -153,13 +153,37 @@ Task("UpdateAssemblyInfo")
...
+
+        if (vsInstallation != null)
+        {
+            var msBuildPath = vsInstallation.CombineWithFilePath(@"MSBuild\Current\Bin\MSBuild.exe");
+
+            if (!FileExists(msBuildPath))
+                msBuildPath = vsInstallation.CombineWithFilePath(@"MSBuild\15.0\Bin\MSBuild.exe");
+
+            if (FileExists(msBuildPath))
+                settings.ToolPath = msBuildPath;
```
jnm2: Maybe Cake gets smarter in the future and e.g. resolves from NuGet.

PR #580, opened 2019-03-19
WIP - Split engine spike

CharliePoole:
This is a spike. That is, it's throw-away code written to learn something. It will never be merged - so __don't__ merge it! It's a PR only because that's the easiest way to see what I've done.

This PR splits the current engine assembly nto two parts, a lower and an upper part, with the upper part inheriting from the lower part. The anticipated advantage of this split is that the upper part (engine) and the lower part (core) can be built for differing targets. The upper part only needs to be built for platforms on which developers actually do their work, editing, compiling and testing. The lower part must be built for any targets on which we want the tests to run. In this spike, however, none of the targeting has been changed. The intent is merely to demonstrate that a split engine can work and pass all its tests.

BTW I'm a bit uncertain about what to call the lower part. I used nunit.engine.core because that's the name that Rob and I once came up with for a reduced-function engine. I'm tempted by `nunit.engine.agents` except that there is actually other stuff in their besides the agents.


rprouse:
Even though this isn't "Work in Progress", I added the WIP label for the WIP Bot as a backup to the draft state.

rprouse:
I need to read through the changes more closely, but I like the idea. It is good to see the conceptual parts for the engine split into actual projects. Thanks for doing the work as a proof of the idea. What platforms do you anticipate targeting for the upper part (Engine)? I could see it targeting Core and full framework, although I'm not sure which versions.

CharliePoole:
@rprouse Well good. I'm frankly surprised. Only did this 'cause @ChrisMaddock  "made me do it." We could discuss more on that privately if you care to.

GitHub actually does a nice job of simplifying the diff. Basically all I did was keep moving files around till it all compiled and then hacked it till it worked. Not how I would implement it for real.

Key problem is that the ExtensionService in the lower part needs to reference the upper part in order to find extension points. I solved that by adding a property with RootAssemblies to search. This is actually the germ of what's needed to allow runners to have extension points as well. In an actual impleentation, I'd want to define an API to speciry where extension points should be searched for.

I couldn't figure out how to make one of @jnm2 's Integration tests pass and just marked it Ignored.

jnm2:
> I couldn't figure out how to make one of @jnm2 's Integration tests pass and just marked it Ignored.

I see that's `Explore_does_not_throw_SocketException`. I'm working to make that test obsolete as part of removing remoting.

ChrisMaddock:
> Only did this 'cause @ChrisMaddock "made me do it." 

Objection! We came to the mutual agreement that it would be a shame for this work to not benefit the project in general. 😄 

I'd love to come back and have more of a look at this at some point, afraid work's a bit hectic at the moment. Hope to in the not-to-distant future.

ChrisMaddock:
Thanks for doing this Charlie! I'd be really keen to get this in, for two main reasons
- The smaller the agent, the easier to add/isolate/remove various platforms (e.g. .NET 2.0, .NET Core...)
- I think the current structure's confusing the new contributors, that the nunit-agent re-references the whole engine. That was of course never the intention.

The core library is bigger than I expected! After making the split, I'd be keen to see what we could perhaps refactor, to make the agent (and it's dependencies) as small as possible. The RuntimeFramework stuff jumps out immediately, as that'll be pretty platform specific, and off the top of my head, I can't think why the agent would require it. I'd also be interested to see if there's anything we could do to move Mono.Cecil out of the agents and into the engine - given the changing platforms that that supports.

On the name - I agree nunit.core is a little ambiguous. How about nunit.core.services, or nunit.engine.lib or something? I also think it could potentially be a little confusing that we have a specific nunit-agent assembly, but then the TestAgent class is referenced by both assemblies - maybe we should think about renaming something here? (I understand _why_ this is the case...just think the naming could be less confusing!)

Where do you want to go from here? 😄

CharliePoole:
@ChrisMaddock My thought is that some of that refactoring can take place before splitting. That's what I already did with some services that I removed, e.g. the project service.

We should take a look at how the runtime service is used. WRT mono.cecil, it's trickiy because it's used in selecting a driver. It's pretty hard to do that centrally, since it's up to the driver extensions to decide if they can handle a particular test assembly. However, we could experiment.

Another initial refactoring could be to introduce changes to the extension service so it's able to look for extension points dynamically in multiple assemblies.

Anyway, I'll think about this and write something. Right now I'm focused on making the GUI work with the 3.10 engine release.

CharliePoole:
@ChrisMaddock The `RuntimeFrameworkService` is in the engine assembly, not the agent/core. However `RuntimeFramework.CurrentFramework.DisplayName` is used in one log entry in the core.

There are probably a lot of classes like this that could be moved either before or after splitting.

CharliePoole:
The spike served it's purpose, was replaced by #689 and then #690. The latter is being merged. Closing this one and deleting the branch.

PR #582, opened 2019-03-21
[WIP] Add project result elements to XML

### LINKED ISSUE #53: Add project element to top-level sub-project before merging
This is a remaining issue from PR #52 which is intended to fix #30. As @CharliePoole stated in that PR,
## 

This looks much cleaner! The remaining problem I see has to do with where MakePackageResult is called. You moved it from the individual runners to MasterTestRunner. I like the way that removes duplication, but I'm not sure it works correctly now.

Here is what I think will happen, depending on what's on the command-line:

| Arguments | Outcome |
| --- | --- |
| One assembly | No project element, as expected |
| Multiple assemblies | No project element, as expected |
| One project | One project element, as expected |
| Multiple projects | No project elements rather than one per project |
| Project plus assemblies | No project elements rather than one wrapping just the project assemblies |

The problem is that the project element, if any, needs to be added to each top-level sub-project before they are all merged.

That said, this is a rather esoteric use case. I'd be OK with merging this and creating a new issue for the edge cases if @rprouse agrees.
## 

@CharliePoole is there more information you would like to add to this issue?

CharliePoole:
This will fix #53 but is currently partial only.

One thing leads to another! Currently, everything seems to be working correctly except for when you use `--process:Separate. When that happens with one or more projects in the command-line, the whole job of figuring out the proper hierarchy falls to the agent process. This is more work than the Process Runner and agent process were originally designed to handle so changes will be needed.

I see some choices for this issue...

1. Merge what's done and leave the `--process:Separate` case for fixing later. (3 PRs)
2. Put this on hold until the problem with `--process:Separate` can be resolved. (2 PRs)
3. Make resolving the --process:Separate part of this PR. (1 PR)

As anyone who knows me would guess, I prefer the 3 PR approach. I'd make them all steps in the same issue, however, since the problem really only comes up when trying to add XML elements for each project. Kind of depends on how comfortable folks are with merging code without knowing what the next step will be...

### DIFF
--- a/src/NUnitConsole/nunit3-console.tests/nunit3-console.tests.csproj
+++ b/src/NUnitConsole/nunit3-console.tests/nunit3-console.tests.csproj
@@ -13,7 +13,6 @@
   <ItemGroup>
     <PackageReference Include="NSubstitute" Version="2.0.3" />
     <PackageReference Include="NUnit" Version="3.11.0" />
-    <PackageReference Include="NUnit.Analyzers" Version="0.1.0-dev-00079" />
   </ItemGroup>
   <ItemGroup>
     <Compile Include="..\..\CommonAssemblyInfo.cs" LinkBase="Properties" />
--- a/src/NUnitEngine/nunit.engine/Internal/TestPackageExtensions.cs
+++ b/src/NUnitEngine/nunit.engine/Internal/TestPackageExtensions.cs
@@ -43,6 +43,13 @@ namespace NUnit.Engine.Internal
             return package.SubPackages.Count > 0;
         }
 
+        public static bool IsProjectPackage(this TestPackage package)
+        {
+            // We may not have the project service available, so we assume that any expanded package
+            // is a project package. Check name to verify that it's not the anonymous command-line package.
+            return package.FullName != null && package.HasSubPackages();
+        }
+
         public static IList<TestPackage> Select(this TestPackage package, TestPackageSelectorDelegate selector)
         {
             var selection = new List<TestPackage>();
--- a/src/NUnitConsole/nunit3-console.tests/ResultReporterTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/ResultReporterTests.cs
@@ -199,7 +199,7 @@ namespace NUnit.ConsoleRunner.Tests
 
         private static TestEngineResult AddMetadata(TestEngineResult input)
         {
-            return input.Aggregate("test-run start-time=\"2015-10-19 02:12:28Z\" end-time=\"2015-10-19 02:12:29Z\" duration=\"0.348616\"", string.Empty, string.Empty);
+            return input.Aggregate("test-run start-time=\"2015-10-19 02:12:28Z\" end-time=\"2015-10-19 02:12:29Z\" duration=\"0.348616\"", "2", string.Empty, string.Empty);
         }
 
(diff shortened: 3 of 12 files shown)


CharliePoole:
There is an alternative. The TestCentricGUI currently examines the XML returned by the engine and re-interprets it in light of the structure of nested test packages it is attempting to run. It injects XML elements so that the test hierarchy that is displayed in the GUI contains nodes for each of the projects.

The downside of the GUI doing this is twofold
 * It only works for the GUI
 * The nodes are fake. The engine knows nothing about them so any action that would normally invoke the engine passing one of these nodes has to be emulated in the GUI.

I'm wondering if the same code could be ported into the engine MasterTestRunner, eliminating both downsides and keeping the logic on one place, rather than distributed in each type of runner.

I'm going to create an alternate PR for this issue.

src/NUnitEngine/nunit.engine/Internal/TestPackageExtensions.cs
```diff
@@ -43,6 +43,13 @@ public static bool HasSubPackages(this TestPackage package)
             return package.SubPackages.Count > 0;
         }
 
+        public static bool IsProjectPackage(this TestPackage package)
+        {
+            // We may not have the project service available, so we assume that any expanded package
+            // is a project package. Check name to verify that it's not the anonymous command-line package.
+            return package.FullName != null && package.HasSubPackages();
```
rprouse: Could just be `return package?.HasSubPackages()`

mikkelbu: This is not the same. The code examines if `FullName` is null and will throw a NRE if `package` is null. Also `package?.HasSubPackages()` returns a nullable boolean and not a bool, so if the original code had been `package != null && package.HasSubPackages()` then we could have translated that into `package.?HasSubPackages() ?? false`.

src/NUnitConsole/nunit3-console.tests/nunit3-console.tests.csproj
```diff
@@ -13,7 +13,6 @@
   <ItemGroup>
     <PackageReference Include="NSubstitute" Version="2.0.3" />
     <PackageReference Include="NUnit" Version="3.11.0" />
-    <PackageReference Include="NUnit.Analyzers" Version="0.1.0-dev-00079" />
```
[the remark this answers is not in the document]
CharliePoole: Good illustration here. Dropping the analyzers got checked in by mistake. I generally drop them since I can't build in my admittedly strange environment, but avoid committing that change. It's another step to remember to take.

We dogfooded a lot at Microsoft, but it always meant using __production__ tools. For NUnit, we test the framework using the framework under development but using a released console or adapter. We test the adapter using a released framework, etc. That way we are testing only one thing at a time. The analyzer is the only exception to that I know of. OTOH, I'd really like to know why it causes a versioning conflict in my configuration that @mikkelbu can't duplicate.

src/NUnitEngine/nunit.engine/Internal/TestPackageExtensions.cs
CharliePoole: A package with a null name is essentially the command-line package, which contains all the other packages. So it's not a project.

CharliePoole:
Alternate fix was merged. Deleting this PR.

PR #584, opened 2019-03-21
Remove extra ..\ in the Addins file for NuGet v2 Extensions

### LINKED ISSUE #583: NUnit Console NuGet Package Doesn't Load Extensions
Found while testing the 3.10 release, see #574 

NuGet extensions do not work for the v2 format, but they do for the v3 format. It looks like an extra `..\` crept into the `addins` file.

rprouse:
Fixes #583 

@ChrisMaddock can you take a look at this? Did you add the extra `..\` to the v2 packages for a reason? Do we need to add a third line with each level of parents?

### DIFF
--- a/nuget/engine/nunit.nuget.addins
+++ b/nuget/engine/nunit.nuget.addins
@@ -1,2 +1,2 @@
-../../../NUnit.Extension.*/**/tools/     # nuget v2 layout
+../../NUnit.Extension.*/**/tools/        # nuget v2 layout
 ../../../../NUnit.Extension.*/**/tools/  # nuget v3 layout
--- a/nuget/engine/nunit.nuget.addins
+++ b/nuget/engine/nunit.nuget.addins
@@ -1,2 +1,3 @@
 ../../NUnit.Extension.*/**/tools/        # nuget v2 layout
+../../../NUnit.Extension.*/**/tools/     
 ../../../../NUnit.Extension.*/**/tools/  # nuget v3 layout


ChrisMaddock:
Extra redirection was required because NUnit.Engine package had gone from being packaged directly under `/lib/` to being packaged under `/lib/net20/`. That added the extra level which required paths being adjusted in the addins file.

The NUnit.ConsoleRunner package hasn't changed however, as there's no .NET Core console yet, hence the same redirection breaking that package. I suspect this PR would break use with the NUnit.Engine package, unfortunately.

Both packages currently share an addins file. Is the easiest solution to create two separate ones, for now? In the long run, the console package should maybe start to depend on the engine package instead? That would require logic to find an engine though...which is something we've been moving away from a little 🤔

ChrisMaddock:
Good find, by the way. This would have been a big issue!

ChrisMaddock:
The only unexplained bit here is why this differs per NuGet version. I suspect that means that the addins file currently in master isn't correct for either package - but I think the solution may need to be more involved than this.

Oh, for an integration test! 😬

CharliePoole:
The general idea is that each different package and package type (msi, nuget, etc.) needs its own addins file. For example, we use different files for nuget and chocolatey installations because each of them follows their own rules.

rprouse:
Just to be safe, I will add back in a third line. As I see it from the console, the packages are in the following directories;

**v2**

Version is part of the top level package directory

```
{SolutionRoot}\Packages\NUnit.Console.3.10.0\tools
```

**v3**

Version is a sub-directory under the top level package directory

```
C:\Users\{Users}\.nuget\nunit.console\3.10.0\tools
```

rprouse:
Changes pushed

ChrisMaddock:
Oh, of course. NuGet 3 still works, because the path intended for NuGet 2 is 'accidentally' the correct path for NuGet 3. 😅 

This works, but I think it would be good to separate out two separate addins files in future. Shall we merge this for the release, and file an issue to do that later?

------------------

Rob - not totally sure if I've explained what I see as the problem very well. Let me use your example to explain better.

Key is the location of the `nunit.nuget.addins` file within each package, as that sits besides the engine, and paths are relative to the engine location (or possibly the addins file location? It doesn't matter in this case anyway, as they're besides each other!)

**v2 NUnit.Console package**

```
{SolutionRoot}\Packages\NUnit.Console.3.10.0\tools\nunit.nuget.addins
```

**v3 NUnit.Console Package**
```
C:\Users\{Users}\.nuget\nunit.console\3.10.0\tools\nunit.nuget.addins
```


**v2 NUnit.Engine package**

```
{SolutionRoot}\Packages\NUnit.Engine.3.10.0\lib\net20\nunit.nuget.addins
```

**v3 NUnit.Engine Package**
```
C:\Users\{Users}\.nuget\nunit.engine\3.10.0\lib\net20\nunit.nuget.addins
```

After merging the .NET Standard engine back in to the main solution, the NUnit.Engine packages now have the addins file a directory level deeper than the NUnit.Console packages. Previously the addins file was at the same level in both packages, and therefore they were able to share an addins file, but that's no longer the case.

PR #597, opened 2019-03-31
Add RunnerSelectionTests for 'single project' package

ChrisMaddock:
A continuation of #514.

These tests reflect current behaviour. I believe they also expose some bugs - which was exactly what I was hoping to use these tests for! There's some more cases which aren't covered by these tests yet however, and I want to keep adding tests and get better coverage, before I start changing things here.


src/NUnitEngine/nunit.engine.api/TestPackage.cs
```diff
@@ -164,15 +164,5 @@ public T GetSetting<T>(string name, T defaultSetting)
...
-
-        #region Helper Methods
-
-        private static bool IsAssemblyFileType(string path)
-        {
-            string extension = Path.GetExtension(path).ToLower();
-            return extension == ".dll" || extension == ".exe";
-        }
-
-        #endregion
```
ChrisMaddock: Unrelated, but unused

src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
```diff
@@ -57,12 +57,6 @@ public ProcessRunner(IServiceLocator services, TestPackage package) : base(servi
             _agency = Services.GetService<TestAgency>();
         }
 
-        #region Properties
-
-        public RuntimeFramework RuntimeFramework { get; private set; }
-
-        #endregion
-
```
ChrisMaddock: Unrelated, but unused

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/RunnerSelectionTests.cs
```diff
@@ -53,6 +55,7 @@ public void RunnerSelectionTest(TestPackage package, RunnerResult expected)
         {
             var runner = _factory.MakeTestRunner(package);
             var result = GetRunnerResult(runner);
+            TestContext.Out.WriteLine(TestContext.CurrentContext.Test.Name);
```
ChrisMaddock:
```suggestion
```

ChrisMaddock:
Would someone on @nunit/engine-team mind having a quick scan through this and #599? They're only adding tests, so hopefully not too controversial. I'm keen to carry on with these, and start fixing bugs based on them! 😄

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20SingleProjectExpectedRunnerResults.cs
```diff
@@ -0,0 +1,236 @@
...
+        {
+            switch (domainUsage)
+            {
+                case DomainUsage.Default:
+                    return new RunnerResult
+                    {
+                        TestRunner = typeof(MultipleTestProcessRunner),
+                        SubRunners = new[]
+                        {
+                            new RunnerResult {TestRunner = typeof(ProcessRunner)}
```
jnm2: ReSharper overrides Visual Studio's default and removes the spaces from inside `{` and `}`. If you want to worry about it, we could add this to the .editorconfig to go back to the more common style:

```ini
resharper_space_within_single_line_array_initializer_braces = true
```

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20SingleProjectExpectedRunnerResults.cs
```diff
@@ -0,0 +1,236 @@
...
+                case DomainUsage.Single:
+                    return new RunnerResult
+                    {
+                        TestRunner = typeof(ProcessRunner)
+                    };
+                case DomainUsage.Multiple:
+                    return new RunnerResult
+                    {
+                        TestRunner = typeof(ProcessRunner)
+                    };
```
jnm2: Maybe stack the cases like this?

```cs
                case DomainUsage.Default:
                case DomainUsage.None:
                case DomainUsage.Single:
                case DomainUsage.Multiple:
                    return new RunnerResult
                    {
                        TestRunner = typeof(ProcessRunner)
                    };
```

Or if this was the intent, maybe this option?

```cs
        private static RunnerResult GetProcessModelSeparateResult(DomainUsage domainUsage)
        {
            if (!Enum.IsDefined(typeof(DomainUsage), domainUsage))
                ThrowOutOfRange(domainUsage);

            return new RunnerResult
            {
                TestRunner = typeof(ProcessRunner)
            };
        }
```

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20SingleProjectExpectedRunnerResults.cs
```diff
@@ -0,0 +1,236 @@
...
+                    return GetProcessModelSeparateResult(domainUsage);
+                case ProcessModel.Multiple:
+                    return GetProcessModelMultipleResult(domainUsage);
+                default:
+                    ThrowOutOfRange(processModel);
+                    break;
+            }
+
+            throw new ArgumentOutOfRangeException(
+                $"No expected Test result provided for this {nameof(ProcessModel)}/{nameof(DomainUsage)} combination.");
```
jnm2: It's not possible for this line to be hit, is it?

ChrisMaddock: No, but the method needs to throw or return something to compile. 😢 I'm not sure if inlining `ThrowOutOfRange` would have the same effect...I can try it.

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20SingleProjectExpectedRunnerResults.cs
ChrisMaddock: I think the first option is good. 🙂 I'd been keeping these tests as explicit as possible, but that would help on the maintenance front!

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20SingleProjectExpectedRunnerResults.cs
ChrisMaddock: I'm not too worried with these - they're inconsistent across the codebase. Would you prefer we went one way or the other? Which way? 🙂

ChrisMaddock:
Thanks Joseph - I'll come back when I'm at a desktop!

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20SingleProjectExpectedRunnerResults.cs
jnm2: Ah, then I would refactor ThrowOutOfRange to return the exception: `throw CreateOutOfRangeException(...);` or move this line to replace the `break;` line.

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20SingleProjectExpectedRunnerResults.cs
jnm2: I do have a preference for the editorconfig line, but I don't like to impose my preference unless y'all have the same preference. Just offering info on ReSharper's behavior in case it's useful.

ChrisMaddock: I _think_ I've added a change to the Resharper settings to solve that. 😄

ChrisMaddock:
Thanks Joseph! Review comments applied...will merge after CI, as I don't think there's anything controversial here. 😄

ChrisMaddock:
Oh...my changes invalidated the Github review. 😞 Could someone please review again? I force pushed, but only the last commit is new since Joseph's review, which is trivial.

@rprouse - how would you feel about removing the Github "1 review required" setting? I don't think our team policy should change, but enforcing it through GitHub seems to just keep catching us out in cases like this.

ChrisMaddock:
Thanks Mikkel!

PR #599, opened 2019-03-31
Add RunnerSelectionTests for 'multiple project' packages

ChrisMaddock:
N.B. To be reviewed after #597, branched off that one

-----
Next step in #514. As before, tests reflecting current behaviour, rather than correct behaviour!

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/Services/DefaultTestRunnerFactoryTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Services/DefaultTestRunnerFactoryTests.cs
@@ -55,48 +55,22 @@ namespace NUnit.Engine.Services.Tests
             Assert.That(_factory.Status, Is.EqualTo(ServiceStatus.Started));
         }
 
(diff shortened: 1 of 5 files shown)


src/NUnitEngine/nunit.engine.tests/Services/DefaultTestRunnerFactoryTests.cs
```diff
@@ -55,48 +55,22 @@ public void ServiceIsStarted()
             Assert.That(_factory.Status, Is.EqualTo(ServiceStatus.Started));
         }
 
+        //Note: These tests currently overlap with RunnerSelectionTests, and are being removed as they are
+        //re-implemented in that test class
```
jnm2: Least important thing imaginable- for consistency, you could add a space between `//` and the following text if you want to. :D

ChrisMaddock:
As https://github.com/nunit/nunit-console/pull/597 - this needs a GitHub review to merge - but the changes since the last review are trivial.

(Thanks Joseph for reviewing!)

PR #610, opened 2019-04-18
Re-implement Reload in the engine

### LINKED ISSUE #609: Run after Reload reports assemblies multipletimes
This is pretty strange! The following sequence of calls to the master test runner...

```c#
runner.Load();
runner.Reload();
var result = runner.Run(listener, TestFilter.Empty);
```

... returns a result in which the test assembly appears twice. If the reload is repeated, then the result will have an additional copy of the assembly for each repetition. 

This only appears after a Run. The result of the reload will seem quite normal.

Note that we don't use Reload in any of our runners, which is why this has gone undetected up to now.

CharliePoole:
Fixes #609

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
@@ -208,6 +208,36 @@ namespace NUnit.Engine.Runners.Tests
             CheckTestRunEvents();
         }
 
+        [Test]
+        public void RunAfterReload()
+        {
+            // Engine issue #609
+            _runner.Load();
+            _runner.Reload();
+            var result = _runner.Run(this, TestFilter.Empty);
+
+            Assert.That(result.Name, Is.EqualTo("test-run"));
+            CheckTestRunResult(result, _testRunData);
+
+            CheckThatIdsAreUnique(result);
+
+            CheckTestRunEvents();
+        }
+
+        [Test]
+        public void ExploreAfterReload()
+        {
+            // Engine issue #609
+            _runner.Load();
+            _runner.Reload();
+            var result = _runner.Explore(TestFilter.Empty);
+
+            Assert.That(result.Name, Is.EqualTo("test-run"));
+            CheckResult(result, _testRunData);
+
+            CheckThatIdsAreUnique(result);
+        }
+
 #if !NETCOREAPP1_1
         [Test]
         public void RunAsync()
--- a/src/NUnitEngine/nunit.engine/Runners/AbstractTestRunner.cs
+++ b/src/NUnitEngine/nunit.engine/Runners/AbstractTestRunner.cs
@@ -80,14 +80,10 @@ namespace NUnit.Engine.Runners
         protected abstract TestEngineResult LoadPackage();
 
         /// <summary>
-        /// Reload the currently loaded test package. Overridden
-        /// in derived classes to take any additional action.
+        /// Reload the currently loaded test package.
         /// </summary>
         /// <returns>A TestEngineResult.</returns>
-        protected virtual TestEngineResult ReloadPackage()
-        {
-            return LoadPackage();
-        }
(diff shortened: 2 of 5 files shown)


CharliePoole:
Not sure why the Azure build is failing... seems unrelated.

CharliePoole:
The implementation of Reload works in this PR provided a reload is possible. However, if settings are changed that cause it to require a different runner, the Reload will fail. For example, if you were to change the target runtime from .NET 2.0 to 4.5, while running in a separate process, the existing process would not be able to load the tests.

This makes the caller responsible for knowing whether it's possible to do a reload. That's how the GUI does it, by recognizing cases like the above that require an Unload / Load rather than a Reload.

I'm not sure this is a very good API however. Thoughts on how we should do this?

mikkelbu:
> Not sure why the Azure build is failing... seems unrelated.

@CharliePoole It is unrelated. I just triggered a build on master and that also failed. The failure is 
```
2019-04-18T19:35:11.3863893Z D:\a\1\s\src\NUnitEngine\nunit.engine.tests\nunit.engine.tests.csproj :
error NU3005: Package 'System.Xml.XPath 4.3.0' from source
'C:\Program Files\dotnet\sdk\NuGetFallbackFolder':
The package signature file entry is invalid.
The central directory header field 'compression method' has an invalid value (8). [D:\a\1\s\NUnitConsole.sln]
```
which is similar to the problem we had on Appveyor in #556, but there we only solved the problem for mock-assembly.csproj, as this was restored first in `appveyor.yml`. So my guess is that Azure has been updated and that we now run into the same problem here (just in nunit.engine.tests.csproj). I'll create a PR for both builds.

CharliePoole:
Here is our existing `ITestRunnerFactory.CanReuse` method, pretty much as inherited from V2. It only deals with changes to the process model andis never called.

https://github.com/nunit/nunit-console/blob/8ca1ac752b1d688ba85e1ad1525a097f307e0734/src/NUnitEngine/nunit.engine/Services/DefaultTestRunnerFactory.cs#L128

In NUnit V2, it is called by `TestLoader` in the `ReloadTests` method. If it returns false, then a new runner is created, changing the reload to a load.

So what do we want to do when the user asks us to reload after changing some settings that make it impossible to reuse the same runner? We could do as V2 does or just go ahead and let it error out.

CharliePoole:
I added a description of where I'd like to take this next in #609. I'll wait for feedback before proceeding.

CharliePoole:
Rebased to take advantage of @mikkelbu 's fix for the travis build.

ChrisMaddock:
> This makes the caller responsible for knowing whether it's possible to do a reload. That's how the GUI does it, by recognizing cases like the above that require an Unload / Load rather than a Reload.

I'm not _totally_ keen on merging an implementation that we know is broken from the behaviour we want to implement...even though the current functionality is also broken!

I assume your aim here is to implement a half-way house, until we can implement the "silent-full-reloads" you describe in #609? If that is the case, I think I'd be more comfortable with implementing a full reload every time for now, and then later enhancing that as a 'performance improvement'.

CharliePoole:
Yeah... I actually meant that sentence to indicate that I __didn't__ want to merge this as is, even though it would be an incremental improvement. :smile: 

I've done some further work, which goes a bit farther than you suggest. Since only the GUI currently uses this API I think it helps to get something in that works for most cases, so that new bugs can be discovered and worked on incrementally.

New commit coming "soon."

CharliePoole:
Setting this aside for a few days while we move from Croatia to Sicily.

ChrisMaddock:
Hi @CharliePoole - what's the status on this one? 🙂 If I'm reading correctly 'Reload' as a whole was never finished, but priorities have changed. Are we better closing this PR, and revisiting this at a later point?

CharliePoole:
@ChrisMaddock I'll put this on my list as the next thing to check (and answer) when I reach a break point on some other work I'm doing.

CharliePoole:
Replaced by PR #927

PR #612, opened 2019-04-21
Fix logging when including exception

ChrisMaddock:
There was a convention in this repo to log like this:

```
```

The assumption being that this commits both message and exception to the log.

That's not actually how our logger works - the first parameter is expected to be a format-string, so if it doesn't contain a `{0}`, then the exception information is never logged.

Given this is a public interface, I've just fixed our usages of it, rather than redefining the interface. I've also made use of ExceptionHelper, as per https://github.com/nunit/nunit-console/pull/576#discussion_r272793843. 🙂

### DIFF
--- a/src/NUnitEngine/nunit.engine/Internal/TcpChannelUtils.cs
+++ b/src/NUnitEngine/nunit.engine/Internal/TcpChannelUtils.cs
@@ -124,7 +124,7 @@ namespace NUnit.Engine.Internal
                 }
                 catch (Exception ex)
                 {
-                    Log.Error("Failed to create/register channel. {0}", ExceptionHelper.BuildStackTrace(ex));
+                    Log.Error("Failed to create/register channel. {0}", ExceptionHelper.BuildMessageAndStackTrace(ex));
                     Thread.Sleep(300);
                 }
 
--- a/src/NUnitEngine/nunit.engine/Internal/TcpChannelUtils.cs
+++ b/src/NUnitEngine/nunit.engine/Internal/TcpChannelUtils.cs
@@ -124,7 +124,7 @@ namespace NUnit.Engine.Internal
                 }
                 catch (Exception ex)
                 {
-                    Log.Error("Failed to create/register channel. {0}", ExceptionHelper.BuildMessageAndStackTrace(ex));
+                    Log.Error("Failed to create/register channel." + Environment.NewLine + ExceptionHelper.BuildMessageAndStackTrace(ex));
                     Thread.Sleep(300);
                 }
 
--- a/src/NUnitEngine/nunit.engine/Internal/TcpChannelUtils.cs
+++ b/src/NUnitEngine/nunit.engine/Internal/TcpChannelUtils.cs
@@ -29,6 +29,7 @@ using System.Runtime.Remoting.Channels;
 using System.Runtime.Remoting.Channels.Tcp;
 using System.Runtime.Serialization.Formatters;
 using System.Threading;
+using NUnit.Common;
 
 namespace NUnit.Engine.Internal
 {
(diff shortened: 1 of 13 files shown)


src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
```diff
@@ -80,7 +80,7 @@ public override TestEngineResult Explore(TestFilter filter)
             }
             catch (Exception e)
             {
-                log.Error("Failed to run remote tests {0}", e.Message);
```
ChrisMaddock: The changes in `ProcessRunner` are just to log additional information - these log calls were always formatted correctly.

CharliePoole:
Yes, our log calls are modeled after our assertion messages. Not misleading if your head is in a certain place, but possibly confusing when compared with other loggers. Would changing the name of the first arg to "format" make sense?

src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
```diff
@@ -80,7 +80,7 @@ public override TestEngineResult Explore(TestFilter filter)
...
+                log.Error("Failed to run remote tests {0}", ExceptionHelper.BuildStackTrace(e));
```
CharliePoole: Slightly surprised to learn that `BuildStackTrace` includes the message as well.

ChrisMaddock:
> Yes, our log calls are modeled after our assertion messages. Not misleading if your head is in a certain place, but possibly confusing when compared with other loggers. Would changing the name of the first arg to "format" make sense?

I think it'll be less confusing once there's not an incorrect convention in the code-base, but that sounds like a sensible idea anyway! I'll do that. 🙂

src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
ChrisMaddock: Me too, I'll rename

ChrisMaddock:
Changes made - I also made some additional updates to the Logger class inline with our conventions.

src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
```diff
@@ -279,7 +279,7 @@ protected override void Dispose(bool disposing)
                     }
                     catch (Exception e)
                     {
-                        var stopError = $"Failed to stop the remote agent. {ExceptionHelper.BuildMessage(e)}{Environment.NewLine}{ExceptionHelper.BuildStackTrace(e)}";
+                        var stopError = $"Failed to stop the remote agent. {ExceptionHelper.BuildMessage(e)}{Environment.NewLine}{ExceptionHelper.BuildMessageAndStackTrace(e)}";
```
jnm2: I'd probably have used concatenation over interpolation here.

src/NUnitEngine/nunit.engine/Internal/TcpChannelUtils.cs
```diff
@@ -123,7 +124,7 @@ public static TcpChannel GetTcpChannel(string name, int port, int limit, Current
                 }
                 catch (Exception ex)
                 {
-                    Log.Error("Failed to create/register channel", ex);
+                    Log.Error("Failed to create/register channel. {0}", ExceptionHelper.BuildMessageAndStackTrace(ex));
```
jnm2: I might have used concatenation instead of a format string, but this does stay consistent without changing everything.

ChrisMaddock:
Thanks both!

ChrisMaddock:
As #597 - this needs a GitHub review to merge - but the changes since the last review are trivial.

Issue #615, opened 2019-04-25
TestPackage(string) constructor fails under certain process/domain modes


ChrisMaddock:
Found under the tests from #514.

Under certain situations, the `TestPackage(string)` constructor, creates a different set of TestRunners to the `TestPackage(IList<string>)` ctor, for a single-assembly package. We decided in #598 that both of these constructors should leave to the same end result.

The combinations which don't work are listed below. None of these are high-priority, as using neither Process:Multiple or Domain:Multiple makes much sense when you only have a single assembly - however both ctor's should have the same behaviour here. 

```
ProcessModel:Single DomainUsage:Multiple
ProcessModel:Multiple DomainUsage:Default
ProcessModel:Multiple DomainUsage:None
ProcessModel:Multiple DomainUsage:Single
ProcessModel:Multiple DomainUsage:Multiple
```

This 'list' variant is the more popular option, which looks to be the one working correctly.

ChrisMaddock:
Behaviour also differs when running a single project file, in the .NET Standard 2.0 build only.

ChrisMaddock:
This ctor creates a different ITestRunner to the normal for projects too - I haven't tested, but I imagine a number of these options don't work. Getting this controlled under test in #616 before making changes...

CharliePoole:
The TestPackage(string) constructor no longer exists, nor do ProcessModel and DomainUsage, so this seems safe to close. :-)

PR #616, opened 2019-04-25
Complete Runner Selection Tests

### LINKED ISSUE #514: Add higher-level unit tests for structure of TestRunners

ChrisMaddock:
Fixes #514.

This adds the final combination of these tests (for now):
- Test packages created in the format of both TestPackage constructors (See #615)
- Test mixed assemblies and project packages
- Test behaviour of unknown extensions
- Remove tests made redundant by these ones (Previously testing in less detail)

The overall fixture currently contains 240 tests - but I think this is worthwhile, as there's so many possible  paths through this area of code. There's still further cases which might modify the TestPackage structure (e.g. x86 flag, framework flag) - but this covers everything we had before and more - so I'm planning to stop here for now! 

I'm hoping the number of unique result structures will shrink as we fix cases like the difference in package structure resulting from the two TestPackage ctors, and as the .NET Standard engine reaches feature parity with .NET Framework.

The next step, is to start using these tests to fix some bugs! 😄


ChrisMaddock:
Thanks Charlie!

Issue #606, opened 2019-04-15
Breaking change in filter functionality between framework 2.7 and 3.11


jbengaard:
We have a filter, --where="!(test != <something>)" which returned the appropriate tests on nunit framework 2,7.0.0. On 3.11.0.0, it always returns an empty set of tests.
This behavior was observed using NUnit Console Runner 3.10.0.

I attached a small example solution containing a few tests using 2.7,0.0 and a copy of the tests using 3.11.0.0. I also attached the output from running the tests with the same filter.

[nunit2example_1.zip](https://github.com/nunit/nunit-console/files/3080284/nunit2example_1.zip)
![Console output](https://user-images.githubusercontent.com/46798085/56135558-4e4bd500-5f91-11e9-8632-acbf752a52bb.PNG)

jbengaard:
I tried to use nunit framework  v3.0.5797 and console runner v3.0.5797,
with the same result.

Den man. 15. apr. 2019 kl. 16.57 skrev CharliePoole <
notifications@github.com>:

CharliePoole:
Sounds like the problem has been there for a while then. Can you avoid that specific syntax as a workaround?

CharliePoole:
This is probably a framework error, but someone should confirm before transferring it I think.

jbengaard:
Unfortunately, we have some tests running on framework 2.7 that need that specific type of filter. We have not found a workaround yet.

CharliePoole:
You should be aware that `!(test != X)` is not the same as `test == X` if X happens to be marked `Explicit`.

That is, a top level not-filter isn't allowed to select an Explicit test. For non-explicit tests, the normal equivalence is intended to apply.

cyanite:
> You should be aware that `!(test != X)` is not the same as `test == X` if X happens to be marked `Explicit`.

Well, `!(test != X)` is not the same as `test == X` in general, since in our case (I am a collegue of @jbengaard), I get 0 tests run with `--where="!(test!='Some.TopLevel.Namespace')"` and all tests run with the De Morgan equivalent `--where="test='Some.TopLevel.Namespace'"`.

CharliePoole:
Yes, I think we understand the problem you are seeing. I am only pointing out one small exception where Explicit is involved.

Someone will need to verify that the generated filter is correct before possibly moving the issue to the framework repo.

CharliePoole:
[...]

PR #618, opened 2019-05-02
Test double-negative filters

CharliePoole:
This PR simply adds some tests to verify that issue #606 is not an engine problem. Other tests are also added, but commented out, which are intended to verify that these filters actually work properly. 

They don't (work properly) but that's a framework issue. This can be merged to ensure that the console and engine keep working correctly, since there were not tests of this pattern before.


ChrisMaddock:
Looks good to me. 🙂

PR #619, opened 2019-05-04
WIP: Expand projects before building TestRunner tree

### LINKED ISSUE #617: Consider expanding projects before building ITestRunner structure
Currently, the MasterTestRunner builds the tree of ITestRunners, before expanding projects. (Done inside `InitializePackage()`.

https://github.com/nunit/nunit-console/blob/23fff2b0153f8cd238cea1ea8409b87d95e6def6/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs#L83-L85

This complicates both the `ITestRunnerFactory`'s, as code must consider whether both whether a package has subpackages, or is an unexpanded project. I think it might also be the underlying driver of a few bugs, such as #29.

Is there any reason not to expand to projects before building the ITestRunner tree? @CharliePoole - I think this is your area...any thoughts? This would be sorta inline with your recent PR https://github.com/nunit/nunit-console/pull/596

ChrisMaddock:
WIP, will fix #617. Branched off #616, so should be easier to look at once that's merged.

I've had a first attempt at working this out. Current status: it needs tidying up, unit tests need updating and I need to do a lot more manual testing.

After expanding the packages and tidying up, I found that MultipleProcessRunner and MultipleDomainRunner would only ever create their 'single' versions (e.g. ProcessRunner) as subrunners - I made a change there so a MultipleXRunner will create Multiple or Single version, depending on the subpackages. I think this will also fix some outstanding bugs.

An example path down the tree before:
```
AggregatingTestRunner (Command Line) -> ProcessRunner (a.nunit) -> MultipleTestDomainRunner (a.nunit) -> TestDomainRunner (a.dll) + TestDomainRunner (b.dll)
```
And after:
```
MultipleProcessRunner (Command Line) -> MultipleProcessRunner (a.nunit) -> ProcessRunner (a.dll) + ProcessRunner (b.dll)
```
These changes also mean that AggregatingTestRunner now has very limited use, as there's no need to "defer a decision until later". It's only remaining use is in the .NET Standard engine to aggregate multiple LocalTestRunners - I'm tempted to make a MultipleLocalRunner class, and make AggTestRunner abstract. I think that might make the inheritance structure a little clearer here.

cc @CharliePoole, if you're interested after our discussion on #617. Would be keen to hear about any issues you foresee.

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/Services/Fakes/FakeProjectService.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Services/Fakes/FakeProjectService.cs
@@ -56,5 +56,20 @@ namespace NUnit.Engine.Services.Tests.Fakes
         {
             return Path.GetExtension(path) == _supportedExtension;
         }
+
+        internal void ExpandAllProjects(TestPackage package)
+        {
+            if (package == null) throw new ArgumentNullException(nameof(package));
+
+            foreach (var subPackage in package.SubPackages)
+            {
+                ExpandAllProjects(subPackage);
+            }
+
+            if (package.Name != null)
+            {
+                ((IProjectService)this).ExpandProjectPackage(package);
+            }
+        }
     }
 }
--- a/src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/RunnerResult.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/RunnerResult.cs
@@ -35,19 +35,31 @@ namespace NUnit.Engine.Tests.Services.TestRunnerFactoryTests
 
         public override string ToString()
         {
+            return Environment.NewLine + ToString(0) + Environment.NewLine;
+        }
+
+        private string ToString(int level)
+        {
+            var tabs = new string('\t', level);
+
             var sb = new StringBuilder();
-            sb.AppendLine($"TestRunner: {TestRunner.Name}");
+            sb.AppendLine($"{tabs}TestRunner: {TestRunner.Name}");
 
             if (SubRunners.Count == 0)
                 return sb.ToString().Trim();
 
-            sb.AppendLine("SubRunners:");
+            level++;
+            tabs = new string('\t', level);
+
+            sb.AppendLine(tabs + "SubRunners:");
+
+            level++;
+            tabs = new string('\t', level);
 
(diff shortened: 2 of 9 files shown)


ChrisMaddock:
Hmm...actually, I'm not sure this process will work in the situation where a project defines it's own process settings. Might need more thought on how the aggregating runner fits in...

CharliePoole:
@ChrisMaddock This...

```C#
AggregatingTestRunner (Command Line) -> ProcessRunner (a.nunit) -> MultipleTestDomainRunner (a.nunit) -> TestDomainRunner (a.dll) + TestDomainRunner (b.dll)
```

Isn't strictly possible in actual execution. The runner (e.g. console or GUI) doesn't even know that `AggregatingTestRunner` exists. It only ever receives a `MasterTestRunner`. So the above would always be

```C#
MasterTestRunner (Command Line) -> AggregatingTestRunner (Command Line) -> ProcessRunner (a.nunit) -> MultipleTestDomainRunner (a.nunit) -> TestDomainRunner (a.dll) + TestDomainRunner (b.dll)
```

It might be a good move to rename all the engine runners (implementing ITestEngineRunner) so that they are called `XxxxEngineRunner`.

Also, the changed example you give above is different behavior. The "before" indicates that the user asked for the assemblies to run in a single process while the "after" uses two processes.

CharliePoole:
If you don't mind, I'll take a shot at some minimal changes that I think go in the right direction. Might take a while, however, since I can only catch an hour or less at a time to do it in the next week.

CharliePoole:
The reason we defer to the AggregatingTestRunner is that the project might contain settings of its own. Strictly speaking, that's only necessary when default values are used for domain and process, since the command-line settings override the project settings. However, I did it at all times for simplicity and because we may end up with other settings in the future, which would have to be checked along with domain and process.

ChrisMaddock:
Yeah, I'd left out the MasterTestRunner as a given sorry, behaviour there hasn't changed.

> Also, the changed example you give above is different behavior. The "before" indicates that the user asked for the assemblies to run in a single process while the "after" uses two processes.

Agree what you say is 'intended' behaviour - what I posted is current behaviour - with default process/domain modes. This was the bug I was describing. 🙂 

> If you don't mind, I'll take a shot at some minimal changes that I think go in the right direction. 

Please do!

ChrisMaddock:
Done in #622 instead. 😄

PR #622, opened 2019-05-07
Delay Creation of test engine runner and package expansion till actually needed

### LINKED ISSUE #22: Engine modifies TestPackage
@CharliePoole commented on [Thu Jan 07 2016](https://github.com/nunit/nunit/issues/1177)

In some situations, when project files are in use, the engine modifies any test package provided to it. This can cause unexpected side effects for the runner calling into the engine. See nunit/nunit-gui#69 for example.

We should review the code and, if possible, avoid modifying the original copy of a test package provided to the engine.

### LINKED ISSUE #617: Consider expanding projects before building ITestRunner structure

CharliePoole:
Fixes #22 and I believe that the first commit fixes #617. The engine runner is no longer created till needed.

@ChrisMaddock could you take a look and see if you agree with that last point? I would have prefered to explicitly initialize the engine runner in one spot, but without significant restructuring there is no such spot. I used a method rather than a property to do the initialization, because it seemed more expressive of what's happening to me.

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20ThreeItemExpectedRunnerResults.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20ThreeItemExpectedRunnerResults.cs
@@ -60,12 +60,12 @@ namespace NUnit.Engine.Tests.Services.TestRunnerFactoryTests.Results
                 case DomainUsage.Multiple:
                     return new RunnerResult
                     {
-                        TestRunner = typeof(AggregatingTestRunner),
+                        TestRunner = typeof(MultipleTestProcessRunner),
                         SubRunners = new[]
                         {
-                            new RunnerResult { TestRunner = typeof(MultipleTestProcessRunner) },
-                            new RunnerResult { TestRunner = typeof(MultipleTestProcessRunner) },
-                            new RunnerResult { TestRunner = typeof(MultipleTestProcessRunner) }
+                            new RunnerResult { TestRunner = typeof(ProcessRunner) },
+                            new RunnerResult { TestRunner = typeof(ProcessRunner) },
+                            new RunnerResult { TestRunner = typeof(ProcessRunner) }
                         }
                     };
                 default:
@@ -83,13 +83,7 @@ namespace NUnit.Engine.Tests.Services.TestRunnerFactoryTests.Results
                 case DomainUsage.Multiple:
                     return new RunnerResult
                     {
-                        TestRunner = typeof(AggregatingTestRunner),
-                        SubRunners = new[]
-                        {
-                            new RunnerResult { TestRunner = typeof(ProcessRunner) },
-                            new RunnerResult { TestRunner = typeof(ProcessRunner) },
-                            new RunnerResult { TestRunner = typeof(ProcessRunner) }
-                        }
+                        TestRunner = typeof(ProcessRunner)
                     };
                 default:
                     throw new ArgumentOutOfRangeException(nameof(domainUsage), domainUsage, ExceptionMessage);
@@ -103,7 +97,7 @@ namespace NUnit.Engine.Tests.Services.TestRunnerFactoryTests.Results
                 case DomainUsage.Default:
                     return new RunnerResult
                     {
-                        TestRunner = typeof(AggregatingTestRunner),
+                        TestRunner = typeof(MultipleTestDomainRunner),
                         SubRunners = new[]
                         {
                             new RunnerResult { TestRunner = typeof(TestDomainRunner) },
@@ -114,34 +108,22 @@ namespace NUnit.Engine.Tests.Services.TestRunnerFactoryTests.Results
                 case DomainUsage.None:
                     return new RunnerResult
                     {
-                        TestRunner = typeof(AggregatingTestRunner),
-                        SubRunners = new[]
-                        {
-                            new RunnerResult { TestRunner = typeof(LocalTestRunner) },
-                            new RunnerResult { TestRunner = typeof(LocalTestRunner) },
-                            new RunnerResult { TestRunner = typeof(LocalTestRunner) }
-                        }
+                        TestRunner = typeof(LocalTestRunner)
                     };
                 case DomainUsage.Single:
                     return new RunnerResult
                     {
-                        TestRunner = typeof(AggregatingTestRunner),
+                        TestRunner = typeof(TestDomainRunner)
+                    };
+                case DomainUsage.Multiple:
+                    return new RunnerResult
+                    {
+                        TestRunner = typeof(MultipleTestDomainRunner),
                         SubRunners = new[]
                         {
                             new RunnerResult { TestRunner = typeof(TestDomainRunner) },
                             new RunnerResult { TestRunner = typeof(TestDomainRunner) },
                             new RunnerResult { TestRunner = typeof(TestDomainRunner) }
-                        }
-                    };
-                case DomainUsage.Multiple:
-                    return new RunnerResult
-                    {
-                        TestRunner = typeof(AggregatingTestRunner),
-                        SubRunners = new[]
-                        {
-                            new RunnerResult { TestRunner = typeof(MultipleTestDomainRunner) },
-                            new RunnerResult { TestRunner = typeof(MultipleTestDomainRunner) },
-                            new RunnerResult { TestRunner = typeof(MultipleTestDomainRunner) }
                         }
                     };
                 default:
@@ -159,7 +141,7 @@ namespace NUnit.Engine.Tests.Services.TestRunnerFactoryTests.Results
                 case DomainUsage.Multiple:
                     return new RunnerResult
                     {
-                        TestRunner = typeof(AggregatingTestRunner),
+                        TestRunner = typeof(MultipleTestProcessRunner),
                         SubRunners = new[]
                         {
                             new RunnerResult { TestRunner = typeof(ProcessRunner) },
--- a/src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20TwoProjectExpectedRunnerResults.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20TwoProjectExpectedRunnerResults.cs
@@ -60,11 +60,11 @@ namespace NUnit.Engine.Tests.Services.TestRunnerFactoryTests.Results
                 case DomainUsage.Multiple:
                     return new RunnerResult
                     {
-                        TestRunner = typeof(AggregatingTestRunner),
+                        TestRunner = typeof(MultipleTestProcessRunner),
                         SubRunners = new[]
                         {
-                            new RunnerResult { TestRunner = typeof(MultipleTestProcessRunner) },
-                            new RunnerResult { TestRunner = typeof(MultipleTestProcessRunner) }
+                            new RunnerResult { TestRunner = typeof(ProcessRunner) },
+                            new RunnerResult { TestRunner = typeof(ProcessRunner) }
                         }
                     };
                 default:
@@ -82,12 +82,7 @@ namespace NUnit.Engine.Tests.Services.TestRunnerFactoryTests.Results
                 case DomainUsage.Multiple:
                     return new RunnerResult
                     {
-                        TestRunner = typeof(AggregatingTestRunner),
-                        SubRunners = new[]
-                        {
-                            new RunnerResult { TestRunner = typeof(ProcessRunner) },
-                            new RunnerResult { TestRunner = typeof(ProcessRunner) }
-                        }
+                        TestRunner = typeof(ProcessRunner)
                     };
                 default:
                     throw new ArgumentOutOfRangeException(nameof(domainUsage), domainUsage, ExceptionMessage);
@@ -101,7 +96,7 @@ namespace NUnit.Engine.Tests.Services.TestRunnerFactoryTests.Results
                 case DomainUsage.Default:
                     return new RunnerResult
                     {
-                        TestRunner = typeof(AggregatingTestRunner),
+                        TestRunner = typeof(MultipleTestDomainRunner),
                         SubRunners = new[]
                         {
                             new RunnerResult { TestRunner = typeof(TestDomainRunner) },
@@ -111,31 +106,21 @@ namespace NUnit.Engine.Tests.Services.TestRunnerFactoryTests.Results
                 case DomainUsage.None:
                     return new RunnerResult
                     {
-                        TestRunner = typeof(AggregatingTestRunner),
-                        SubRunners = new[]
-                        {
-                            new RunnerResult { TestRunner = typeof(LocalTestRunner) },
-                            new RunnerResult { TestRunner = typeof(LocalTestRunner) }
-                        }
+                        TestRunner = typeof(LocalTestRunner)
                     };
                 case DomainUsage.Single:
                     return new RunnerResult
                     {
-                        TestRunner = typeof(AggregatingTestRunner),
-                        SubRunners = new[]
-                        {
-                            new RunnerResult { TestRunner = typeof(TestDomainRunner) },
-                            new RunnerResult { TestRunner = typeof(TestDomainRunner) }
-                        }
+                        TestRunner = typeof(TestDomainRunner)
                     };
                 case DomainUsage.Multiple:
                     return new RunnerResult
                     {
-                        TestRunner = typeof(AggregatingTestRunner),
+                        TestRunner = typeof(MultipleTestDomainRunner),
                         SubRunners = new[]
                         {
-                            new RunnerResult { TestRunner = typeof(MultipleTestDomainRunner) },
-                            new RunnerResult { TestRunner = typeof(MultipleTestDomainRunner) }
+                            new RunnerResult { TestRunner = typeof(TestDomainRunner) },
+                            new RunnerResult { TestRunner = typeof(TestDomainRunner) }
                         }
                     };
                 default:
@@ -153,7 +138,7 @@ namespace NUnit.Engine.Tests.Services.TestRunnerFactoryTests.Results
                 case DomainUsage.Multiple:
                     return new RunnerResult
                     {
-                        TestRunner = typeof(AggregatingTestRunner),
+                        TestRunner = typeof(MultipleTestProcessRunner),
                         SubRunners = new[]
                         {
                             new RunnerResult { TestRunner = typeof(ProcessRunner) },
--- a/src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/RunnerResult.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/RunnerResult.cs
@@ -60,8 +60,6 @@ namespace NUnit.Engine.Tests.Services.TestRunnerFactoryTests
 
     public class RunnerResult
     {
-        public RunnerResult() { }
-
         public Type TestRunner { get; set; }
 
         public ICollection<RunnerResult> SubRunners { get; set; } = new List<RunnerResult>();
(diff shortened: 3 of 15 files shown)


ChrisMaddock:
Hi @CharliePoole - sorry this has taken me a few days to get to.

This looks good from the point of view of expanding the package early, however I'm concerned about the effects expanding the projects at this point will have downstream, on the ITestRunnerFactory's. These classes have lots of expectations about when a package has subpackages - and this change will change that, by expanding the projects before we created the TestRunner tree.

This is exactly the sort of change I'd wanted the new RunnerSelection tests to help us with - unfortunately they're pitched a level too low to help here - as they enter at the DefaultTestRunnerFactory, rather than the MasterTestRunner. What would you think to adjusting those tests first, to start at the MasterTestRunner, before trying this change - so we can be a little more certain about the downstream effects? We'd also need to fix the FakeProjectService in the RunnerSelection tests, which I believe isn't currently plugged in correctly. 😞

CharliePoole:
Can you point me to some examples of the expectations you mean? My understanding is that the runner factories require expansion to be complete.

ChrisMaddock:
@CharliePoole Essentially, the below method. This takes various different paths based on the existence of subpakages, or unexpanded projects. I'd be concerned that the changes in this PR might have some adverse effect on this logic.

https://github.com/nunit/nunit-console/blob/8ca1ac752b1d688ba85e1ad1525a097f307e0734/src/NUnitEngine/nunit.engine/Services/DefaultTestRunnerFactory.cs#L67-L125

CharliePoole:
@ChrisMaddock I thought I had cleaned that code up, but it may have been in a PR that was abandoned. IIRC, the code in my special build of the engine for the GUI was changed.

I think the code still works but it's very hard to follow as is and the checking of the project service is being done twice now... once here and once in mater test runner. Sometime in the next few days, I'll look through my branches and see if I have a cleaner version of it to add to this PR.

FYI I'm trying to get rid of my private builds for now and rely on the NUnit engine, at least through the 1.0 release of the GUI.

CharliePoole:
@ChrisMaddock I substituted some code from my private build (not all, because some does not yet apply to the NUnit version) and as you expected there is changed behavior. I fixed eight tests. Four of them have the correct behavior now IMO but four of them are questionable.

When we have two projects, the runner structure is now

```
MasterTestRunner
    AggregatingTestRunner
        MultipleTestProcessRunner
        MultipleTestProcessRunner
```

This would have made sense in earlier versions because the MultipleTestProcessRunner would be doing the aggregation for the project. Not sure it does any longer. What is more, two separate MultipleTestProcessRunners might mess up if we specified Agents, since the class was written on the assumption that there would be only one of them controlling the number of agents.

So let's talk about what we actually __want__ the structure to be in this case before I do another commit.

BTW the tests are helpful but hard to follow. I know I reviewed them but in actually trying to understand the test failures and make changes I felt I had to look in too many places in order to understand. I think I'm still a fan of repetition in tests, listing out each test case individually where it can be seen.

CharliePoole:
@ChrisMaddock Mulled this over a bit more tonight over pizza.

Since the entire purpose of AggregatingTestRunner's use for projects is to delay deciding on the second-level runners (because we don't yet know what settings the project may specify)

And since we are changing the order of evaluation here so we now __do__ know what settings were specified in the project

Then maybe we no longer need AggregatingTestRunner in this role.

What do you think?

Charlie

ChrisMaddock:
I think I went through the same process in #619, but then came across an case where it might be required. But I think that might just be because my implementation was wrong...I'm struggling to think of a case where we would require it.

Thinking aloud...what about the below situation:
- Two command line items: a.dll and b.nunit.
- b.nunit specifies `inprocess` in the NUnit project.
- No additional command line params - so we assume a.dll runs out-of-process, as normal.

Currently, I think we would (should) have:
```
AggregatingRunner
   -> ProcessRunner (a.dll)
   -> MultipleTestDomainRunner (b.nunit)
```

What would the top level runner be, if we didn't have AggregatingRunner anymore. Should it look like this?
```
MultipleTestDomainRunner
   -> ProcessRunner (a.dll)
   -> MultipleTestDomainRunner (b.nunit)
```

-------
*Extra: AggregatingRunner is currently necessary for the .NET Standard engine, although I think we should create a LocalAggregatingRunner subclass, so we can make the AggregatingRunner fully abstract..if we work out that we can do so for the main engine.

ChrisMaddock:
> BTW the tests are helpful but hard to follow. I know I reviewed them but in actually trying to understand the test failures and make changes I felt I had to look in too many places in order to understand. I think I'm still a fan of repetition in tests, listing out each test case individually where it can be seen.

Yeah, I thought they might introduce a maintenance cost, but I'm not sure of any better solution. Interested in any ideas you have to refactor?

In general - I think I need to think about this a little more than I can tonight. I'd be interested to see what things might look like with AggregatingRunner as solely an abstract class though...as I'm struggling to pinpoint exactly what it's purpose should be...

CharliePoole:
@ChrisMaddock  You're right on the structure. We still need an aggregating runner in order to aggregate. What I meant was we no longer need it to do the job of delayed sub runner selection let me try to be clearer... atr originally had these jobs...

1. Provide slots for multiple sub runners.

2. Delay creation of those sub runners.

3. Aggregate projects.

4. Aggregate the whole test run.

MTR now does 3 and 4.

Test runner factory could do 2.

That would just leave 1, which may not justify the class as more than a days structure.

Not saying to do this all together however. Step by step in separate PRs.

Regarding the .NET standard implementation... Rob created it in a way that did not allow the local test runner to handle multiple assemblies. I never understood why so I kept that part unchanged. If you know why it's that way, maybe we could bring it into line with the other build. I assume it's related to how the framework side works.

CharliePoole:
@ChrisMaddock More thinking on this... Here is a proposed set of rules for what kind of runner(s) the MasterTestRunner should use... tell me what you think...

1. Selection is based on the assemblies to be run, independent of whether they came from a project or were specified on the command-line. Engine runners only deal with assemblies.

2. If there is only one assembly, use the appropriate ProcessRunner, TestDomainRunner, etc. according to the settings.

2. If there are multiple assemblies, but they are all run in one separate process, then use a process Runner.

3. If there are multiple assemblies to run in process, but all in the same domain, use a domain runner.

4. In all other cases, we have either multiple processes, multiple domains or both. We need to use an aggregating test runner OR one derived from it.

5. Within each separate process, follow the same rules with the exception that we are now only dealing with the inprocess cases.

We should get these things working IMO in approximately the above order and merge each step as soon as we have something that makes an improvement without breaking anything that already works.

CharliePoole:
Latest changes implement most of the above. Needing more work: projects that include settings and .net standard differences.

ChrisMaddock:
Looks interesting! Afraid I'm struggling for time at the moment, but can hopefully come take a proper look next week. 🙂

CharliePoole:
@ChrisMaddock No problem. I'm only finding a little time to work myself.

I added a commit that restructures one of your test files to reflect my experience working with them. Essentially, I wanted to show all the alternatives in one file as much as possible.

As I already mentioned, my preference is to refactor production code to remove as much duplication as possible, but not so much in test code. One role of tests is documentation of how things work. If a programming book were refactored so that nothing was ever repeated, it would get shorter but not necessarily clearer. I think it's the same for tests.

Some people may not like my use of generics to create a DSL, but it avoids having to use typeof all the time, which I dislike.

I kept this all in a commit, which we can keep, drop or modify. Be good to decide that fairly soon, before other tests get built on top of it.

CharliePoole:
Latest commit fixes error when `--process:Multiple` and `--domain:Multiple` are both specified. All other combinations work when tested manually running the mock-assembly. Eventually we will want integration tests, which do automatically what I did manually.

This could have been fixed differently - by reporting an error when the user speciies those two options, which actually make no sense What do people think we should do here?

ChrisMaddock:
Hi Charlie, taking a look at this today.

My first concern is that the RunnerSelectionTests don't yet expand projects, so we're testing a different thing to what the production code is doing. (My bad, I wrote a bug...)

I want to fix that first, and then see what the test results look like. 🙂

ChrisMaddock:
Updating the tests to reflect new behaviour broke 18 tests - and it looks like the change actually fixed numerous cases. 😄 

I'm not totally happy with how the tests were done, but this PR is getting big, so I went for minimal change, and would like to refactor in a future PR. I'm now going to review what's here - but hopefully we can merge this pretty much as is, and continue in future PRs?

src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
```diff
@@ -388,6 +387,9 @@ private void ValidatePackageSettings()
                             "Cannot run {0} framework in process already running {1}.", frameworkSetting, currentFramework));
                 }
             }
+
+            if (runningInProcess && runAsX86 && IntPtr.Size == 8)
```
ChrisMaddock: I like this refactor. 🙂

src/NUnitEngine/nunit.engine/Services/DefaultTestRunnerFactory.cs
```diff
@@ -141,18 +113,50 @@ public override bool CanReuse(ITestEngineRunner runner, TestPackage package)
...
+        /// <summary>
+        /// Examine the immediate subpackages of a test package, if any
+        /// and count how many subpackages are found. Currently this is
+        /// only called (and only makes sense to call) from a top-level
+        /// package representing the command-line because that's the
+        /// only place packages may appear.
+        /// </summary>
+        /// <param name="package">A TestPackage</param>
+        /// <returns>Count of project subpackaegs</returns>
+        private int ProjectCount(TestPackage package)
```
ChrisMaddock: I think this is unused. 🙂

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/TestCases/NetStandardTestCases.cs
```diff
@@ -30,145 +30,177 @@ namespace NUnit.Engine.Tests.Services.TestRunnerFactoryTests.TestCases
...
-                    {
-                        new RunnerResult { TestRunner = typeof(LocalTestRunner) }
-                    }
-                };
-                yield return new TestCaseData(package, expected).SetName($"{{m}}({testName})");
+                yield return new TestRunnerFactoryData(
+                    "SingleProject (list ctor)",
+                    new TestPackage(new[] { "a.nunit" }),
+                    new RunnerResult<LocalTestRunner>()
+                );
```
ChrisMaddock: This test looks broken both before and after - I don't believe either the before or after TestRunner tree is correct here. Let's handle this in a separate issue.

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20ThreeItemExpectedRunnerResults.cs
```diff
@@ -114,34 +108,22 @@ private static RunnerResult GetProcessModelInProcessResult(DomainUsage domainUsa
                 case DomainUsage.None:
                     return new RunnerResult
                     {
-                        TestRunner = typeof(AggregatingTestRunner),
-                        SubRunners = new[]
-                        {
-                            new RunnerResult { TestRunner = typeof(LocalTestRunner) },
-                            new RunnerResult { TestRunner = typeof(LocalTestRunner) },
-                            new RunnerResult { TestRunner = typeof(LocalTestRunner) }
-                        }
```
ChrisMaddock: I think this change is also incorrect, there should be an Aggregating runner here, right?

ChrisMaddock: (This is only reinforcing my view that was should have a specific `LocalAggregatingTestRunner` - to distinguish the different needs here! 😅)

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20TwoProjectExpectedRunnerResults.cs
```diff
@@ -60,11 +60,11 @@ private static RunnerResult GetProcessModelMultipleResult(DomainUsage domainUsag
                 case DomainUsage.Multiple:
                     return new RunnerResult
                     {
-                        TestRunner = typeof(AggregatingTestRunner),
+                        TestRunner = typeof(MultipleTestProcessRunner),
                         SubRunners = new[]
                         {
-                            new RunnerResult { TestRunner = typeof(MultipleTestProcessRunner) },
-                            new RunnerResult { TestRunner = typeof(MultipleTestProcessRunner) }
```
ChrisMaddock: This is a change to one process per project, rather than one process per assembly - which isn't what we want. However, if I'm understanding correctly, this wouldn't have worked before anyway - as an AggregatingTestRunner or subclass should never be the leaf of the tree.

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20TwoProjectExpectedRunnerResults.cs
```diff
@@ -111,31 +106,21 @@ private static RunnerResult GetProcessModelInProcessResult(DomainUsage domainUsa
                 case DomainUsage.None:
                     return new RunnerResult
                     {
-                        TestRunner = typeof(AggregatingTestRunner),
-                        SubRunners = new[]
-                        {
-                            new RunnerResult { TestRunner = typeof(LocalTestRunner) },
-                            new RunnerResult { TestRunner = typeof(LocalTestRunner) }
-                        }
```
ChrisMaddock: As above - I think this is a break, and should still require an AggregatingTestRunner.

src/NUnitEngine/nunit.engine/Services/DefaultTestRunnerFactory.cs
ChrisMaddock: (I'll remove this now)

ChrisMaddock:
Actually - I think the `Domain.None` breaks are an easy fix. I'll just add in one final commit, which will restore them to previous behaviour, and delete the unused code.

ChrisMaddock:
> Actually - I think the Domain.None breaks are an easy fix. I'll just add in one final commit, which will restore them to previous behaviour, and delete the unused code.

Scrap that - fixing this requires breaking up the `Net20ThreeItemExpectedRunnerResults` class, which is more code than I want to add to this PR!

Latest commit just removes the unused code, will hand back over to you now, Charlie. 😄

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20ThreeItemExpectedRunnerResults.cs
CharliePoole: I would say no. LocalTestRunner derives from DirectTestRunner, which is capable of aggregating the result of multiple drivers. In this case, the result would be the same, whether using one runner with three drivers or three runners plus an aggregating runner. However, I think the latter is better because it's simpler and also conforms to the only acceptable solution when using DomainUsage.Single.

The key is that we only want an aggregating runner when we are aggregating runners. :-) And we only need separate runners when the settings require it (as for running in separate AppDomains).

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20TwoProjectExpectedRunnerResults.cs
CharliePoole: Not exactly. Since the project is not being expanded, I made the assumption for test purposes that it only includes one assembly. At some point, we should fake expansion of the project as in the MasterTestRunner tests and have projects with different numbers of assemblies.

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20TwoProjectExpectedRunnerResults.cs
CharliePoole: Same comment as above as well. :smile:

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/RunnerResult.cs
```diff
@@ -27,6 +27,37 @@
...
+        where TSUB1 : ITestEngineRunner
+        where TSUB2 : ITestEngineRunner
+        where TSUB3 : ITestEngineRunner
+    {
+        public RunnerResult()
+        {
+            SubRunners = new RunnerResult[] { new RunnerResult<TSUB1>(), new RunnerResult<TSUB2>(), new RunnerResult<TSUB3>() };
+        }
+    }
+
```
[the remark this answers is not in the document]
CharliePoole: Personally, I find the single line expression clearer than the one with multiplelines, but I can go either way. We can drop the commit with these changes if you like.

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/TestCases/NetStandardTestCases.cs
CharliePoole: I do think the after is correct, however, since LocalTestRunner can aggregate any number of assembies found in the project.

CharliePoole:
@ChrisMaddock See some added comments I made. I'll update my local repo and retest before merging.

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/RunnerResult.cs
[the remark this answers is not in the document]
ChrisMaddock: I presume it's not too easy to pull out, now it's in the middle of everything else? I'd be happy to merge as is and refactor later. 🙂

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20TwoProjectExpectedRunnerResults.cs
ChrisMaddock: > At some point, we should fake expansion of the project as in the MasterTestRunner tests and have projects with different numbers of assemblies.

That's what my most recent commit does...I think this does indicate a bug here. But only a change in behaviour to something which was already broken, so I think we can fix it as a separate issue.

ChrisMaddock:
Thanks @CharliePoole - I'd got mixed up that LocalTestRunner could only handle a single assembly. In that case, I'm happy with where things are at! 😄

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/RunnerResult.cs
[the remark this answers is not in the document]
CharliePoole: It depends on whether further changes hit that file. I'll take a look.

CharliePoole:
@ChrisMaddock I tried removing the changes to the one test file, but it got complicated. I'll go ahead and merge as you suggested.

PR #637, opened 2019-05-29
Refactor RunnerSelectionTests

ChrisMaddock:
As discussed in #622 

- Refactor tests to remove some boilerplate code and make more maintainable
- Pull out generic classes used for .NET Standard results, to maintain consistent 'tree' information for all tests.

@CharliePoole - what do you think to this, as a way of cutting down on the amount of code to type?

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20OneAssemblyOneProjectExpectedRunnerResults.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20OneAssemblyOneProjectExpectedRunnerResults.cs
@@ -63,8 +63,8 @@ namespace NUnit.Engine.Tests.Services.TestRunnerFactoryTests.Results
                         TestRunner = typeof(MultipleTestProcessRunner),
                         SubRunners = new[]
                         {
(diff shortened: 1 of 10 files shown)


CharliePoole:
@ChrisMaddock This looks good to me. The refactor is cleaner than the original and clearer than my generic version, which only makes sense if you wrote it.

I wasn't really primarily concerned with the amount of code to type - even though my revision shortened the code. My main gripe was that I had to read three different files in order to understand the tests. The structure in `NetStandardTestCases.cs` eliminates one of the three levels and I find it much easier to understand.

Most code is written once and read 100 times. Test code is written once and read 1000 or 10000 times. For product code, I want to remove duplication as much as possible but for test code I'll trade more duplication for better immediate visibility when read by people who don't know the code well.

ChrisMaddock:
Thanks Charlie.

> My main gripe was that I had to read three different files in order to understand the tests. The structure in NetStandardTestCases.cs eliminates one of the three levels and I find it much easier to understand.

I see what you mean. I broke them up due to the amount of content, and wanting to reuse a few things to reduce maintenance...the .NET Standard test cases are of course much more limited here. 

I'm hoping the number of variations will come down as we start to fix some things. (e.g. the differences between `TestPackage(string)` and `TestPackage(List<string>)`. Shall we revisit this idea once these are more stable?

PR #649, opened 2019-06-11
Move ProjectDependencies section from sln -> csproj's 

ChrisMaddock:
A workaround for https://github.com/dependabot/feedback/issues/520.

As I [understand from Stack Overflow](https://stackoverflow.com/q/5629981/1768779), the `ProjectDependencies` elements in the sln file can instead be specified in the csproj file. This is necessary in the NUnit Console's case, as the engine does not directly reference the agent, and as such, is not copied into the test dir by default. 

I also understand that `ProjectDependencies` is the older way of specifying such dependencies, and specifying them through the csproj instead hopefully follows the principle of least surprise!

### DIFF
--- a/src/NUnitConsole/nunit3-console.tests/nunit3-console.tests.csproj
+++ b/src/NUnitConsole/nunit3-console.tests/nunit3-console.tests.csproj
@@ -32,6 +32,8 @@
     <ProjectReference Include="..\..\NUnitEngine\nunit.engine\nunit.engine.csproj" />
     <ProjectReference Include="..\nunit3-console\nunit3-console.csproj" />
     <ProjectReference Include="..\..\NUnitEngine\nunit.engine.api\nunit.engine.api.csproj" />
+    <ProjectReference Include="..\..\NUnitEngine\nunit-agent\nunit-agent.csproj" />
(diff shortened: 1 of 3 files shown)


src/NUnitConsole/nunit3-console.tests/nunit3-console.tests.csproj
```diff
@@ -32,6 +32,8 @@
     <ProjectReference Include="..\..\NUnitEngine\nunit.engine\nunit.engine.csproj" />
     <ProjectReference Include="..\nunit3-console\nunit3-console.csproj" />
     <ProjectReference Include="..\..\NUnitEngine\nunit.engine.api\nunit.engine.api.csproj" />
+    <ProjectReference Include="..\..\NUnitEngine\nunit-agent\nunit-agent.csproj" />
+    <ProjectReference Include="..\..\NUnitEngine\nunit-agent\nunit-agent-x86.csproj" />
```
rprouse: Should you add the `<ReferenceOutputAssembly>false</ReferenceOutputAssembly>` from the SO post to these? They aren't directly referenced.

ChrisMaddock:
Closing this PR in line with calling time on the dependabot trial.

PR #660, opened 2019-07-02
Enable legacyCorruptedStateExceptionsPolicy

### LINKED ISSUE #336: Should legacyCorruptedStateExceptionsPolicy enabled=true in nunit3-console.exe.config?
I would argue that in the nunit3-console.exe.config,  the  legacyCorruptedStateExceptionsPolicy should be enabled by default.

The reason is that nunit console must cater to the widest range of audience possible, and it shouldn't present any surprises to them, or make their debugging lives difficult in the case of any scenario. In my case, a faulty third party native component throws an non-deterministic AccessViolationException after a few hours of running and thus causes the nunit runner to crash randomly ( similar to the case here (https://github.com/nunit/dotnet-test-nunit/issues/115) ).

If the Nunit just crashes like that because of AccessViolationException, I won't know whether the problem lies with the Nunit runner or my code ( or the code underlying mine). Worst of all, it will take me a long time to figure out what's the problem, as the problem doesn't always appear and even if it does, it only appears after quite a long time in running!

Furthermore, in nunit3-console 3.6.1, legacyUnhandledExceptionPolicy is already enabled by default, I don't see why legacyCorruptedStateExceptionsPolicy isn't.

jnm2:
Fixes #336.

Also, in order to fix the intermittent test run hangs investigated at https://github.com/nunit/nunit/issues/3295, at least one of the following things must happen:

- This PR gets merged (that was easy, let's go with this one for now 😅)

- Microsoft ports the .NET Core fix to .NET Framework 4.7 and 4.8 to fix the race condition that causes stack traces to throw AccessViolationException

- NUnit Framework drops support for partial trust (see comment in this PR and https://github.com/nunit/nunit/issues/3295)

- NUnit Framework implements a watchdog thread to notice the highly bizarre situation set up by the console's current app.config where the console process isn't killed but the NUnit framework test worker thread is killed (see comment in this PR and https://github.com/nunit/nunit/issues/3295)

### DIFF
--- a/src/NUnitConsole/nunit3-console/app.config
+++ b/src/NUnitConsole/nunit3-console/app.config
@@ -16,6 +16,14 @@
     <!-- Ensure that test exceptions don't crash NUnit -->
     <legacyUnhandledExceptionPolicy enabled="1" />
 
+    <!--
+     Since legacyUnhandledExceptionPolicy keeps the console from being killed even though an NUnit framework
+     test worker thread is killed, this is needed to prevent a hang. NUnit framework can only handle these
+     exceptions when this config element is present. (Or if future versions of NUnit framework drop support
+     for partial trust which would enable it to use [HandleProcessCorruptedStateExceptions].)
+    -->
+    <legacyCorruptedStateExceptionsPolicy enabled="true" />
+
     <!-- Run partial trust V2 assemblies in full trust under .NET 4.0 -->
     <loadFromRemoteSources enabled="true" />
 
--- a/src/NUnitEngine/nunit-agent/app.config
+++ b/src/NUnitEngine/nunit-agent/app.config
@@ -22,6 +22,14 @@
     <!-- Ensure that test exceptions don't crash NUnit -->
     <legacyUnhandledExceptionPolicy enabled="1" />
 
+    <!--
+     Since legacyUnhandledExceptionPolicy keeps the console from being killed even though an NUnit framework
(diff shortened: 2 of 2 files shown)


ChrisMaddock:
I agree with your conclusions here - I'd like to hear @rprouse opinion before we merge though. 🙂

jnm2:
Awesome! I fixed typos. Guess I was up too late.

KirillOsenkov:
Hi @jnm2 would you happen to remember a link to 

> Microsoft ports the .NET Core fix to .NET Framework 4.7 and 4.8 to fix the race condition that causes stack traces to throw AccessViolationException

We're hitting an AccessViolationException here:
https://github.com/microsoft/vstest/issues/3543#issuecomment-2302635138

and I feel like under some circumstances Assembly.LoadFrom() on .NET Framework can throw AccessViolationException but I can find almost no information about it anywhere, grasping at the straws here

jnm2:
@KirillOsenkov Yes, see https://github.com/nunit/nunit/issues/3295#issuecomment-626182901

KirillOsenkov:
Nice, thanks! Looks like it's a different one from ours, ours is in Assembly.LoadFrom(). Also the stack trace one was (fortunately!) fixed for .NET Framework (in fact I think I remember how we fixed it a few years back).

jnm2:
It does look different. Not sure what to make of this one. Good luck!

PR #669, opened 2019-08-10
nunit.console-runner-with-extensions.nuspec: Remove outdated release notes

oskarb:
It has been mentioning the teamcity event listener (as the only) change for
every version from 3.5.0 onwards. I am guessing that this is in fact not always a
change, nor the only change, so better to leave release notes empty if they are not updated.


oskarb:
@ChrisMaddock Alright.

ChrisMaddock:
Great, thanks for your help @oskarb!

Bypassing the broken CI, as it's been fixed in master, and I don't have the credentials to rerun from this machine. 🙂

Issue #667, opened 2019-08-08
Console Runner loads wrong .NET framework version when executing tests from multiple assemblies at once


PoByBolek:
When running tests from multiple assemblies with different target frameworks, the console runner seems to only load the highest framework version.

For example, I have two test projects: one with target framework version 3.5 and one with target framework version 4.0.
The single test case in the 3.5 project looks like this:

    using System;
    using NUnit.Framework;

    namespace Test35
    {
        [TestFixture]
        public class TestFor35
        {
            [Test]
            public void Example()
            {
                Assert.AreEqual(new Version(2, 0, 0, 0), typeof(int).Assembly.GetName().Version);
            }
        }
    }

The single test case in the 4.0 project looks like this:

    using System;
    using NUnit.Framework;

    namespace Test40
    {
        [TestFixture]
        public class TestFor40
        {
            [Test]
            public void Example()
            {
                Assert.AreEqual(new Version(4, 0, 0, 0), typeof(int).Assembly.GetName().Version);
            }
        }
    }

Now, when I run the test assemblies individually, everything is fine:

    D:\temp\Nunit.Test>nunit3-console.exe --noresult Test35\bin\Debug\Test35.dll
    NUnit Console Runner 3.10.0 (.NET 2.0)
    Copyright (c) 2019 Charlie Poole, Rob Prouse
    Freitag, 9. August 2019 00:37:25

    Runtime Environment
       OS Version: Microsoft Windows NT 10.0.18362.0
      CLR Version: 4.0.30319.42000

    Test Files
        Test35\bin\Debug\Test35.dll

    Run Settings
        DisposeRunners: True
        WorkDirectory: D:\temp\Nunit.Test
        ImageRuntimeVersion: 2.0.50727
        ImageRequiresX86: False
        ImageRequiresDefaultAppDomainAssemblyResolver: False
        NumberOfTestWorkers: 4

    Test Run Summary
      Overall result: Passed
      Test Count: 1, Passed: 1, Failed: 0, Warnings: 0, Inconclusive: 0, Skipped: 0
      Start time: 2019-08-08 22:37:25Z
        End time: 2019-08-08 22:37:27Z
        Duration: 1.746 seconds

    D:\temp\Nunit.Test>nunit3-console.exe --noresult Test40\bin\Debug\Test40.dll
    NUnit Console Runner 3.10.0 (.NET 2.0)
    Copyright (c) 2019 Charlie Poole, Rob Prouse
[...]

PR #670, opened 2019-08-10
Fix for running multiple test packages targetting different frameworks on the correct platform

ChrisMaddock:
Fixes https://github.com/nunit/nunit/issues/3338.

Previously, the chosen framework for a TestPackage was only set on the top level package. This meant that when subpackages got passed into the ProcessRunner, the information wasn't available, and they defaulted to run on the current framework. I think this is probably a long-standing bug, it just to a niche scenario like in #3338 for it to actually be noticed.

The fix calculates and sets the best framework on every package in the tree, so that sub runners can access the information. It's done on the basis of 'minimum change', as this area of code is a little messy already, and I'd _love_ to refactor it all one day. (And possibly one day soon, with the introduction of .NET Core targets.)

Note the ADO Linux build is currently failing here, as Mikkel reported on the framework: https://github.com/nunit/nunit/issues/3338. I'm hoping ADO may apply a fix here...

### DIFF
--- a/src/NUnitEngine/nunit.engine.api/IRuntimeFrameworkService.cs
+++ b/src/NUnitEngine/nunit.engine.api/IRuntimeFrameworkService.cs
@@ -45,6 +45,11 @@ namespace NUnit.Engine
         /// the settings in the package and the assemblies themselves.
         /// The package RuntimeFramework setting may be updated as a 
         /// result and the selected runtime is returned.
+        ///
+        /// Note that if a package has subpackages, the subpackages may run on a different
+        /// framework to the top-level package. In future, this method should
+        /// probably not return a simple string, and instead require runners
+        /// to inspect the test package structure, to find all desired frameworks.
         /// </summary>
         /// <param name="package">A TestPackage</param>
         /// <returns>The selected RuntimeFramework</returns>
--- a/src/NUnitEngine/nunit.engine.tests/Services/RuntimeFrameworkServiceTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Services/RuntimeFrameworkServiceTests.cs
@@ -96,6 +96,10 @@ namespace NUnit.Engine.Services.Tests
         [Test]
         public void RuntimeFrameworkIsSetForSubpackages()
         {
+            //Runtime Service verifies that requested frameworks are available, therefore this test can only currently be run on platforms with both CLR v2 and v4 available
+            Assume.That(new RuntimeFramework(RuntimeType.Net, new Version("2.0.50727")), Has.Property(nameof(RuntimeFramework.IsAvailable)).True);
+            Assume.That(new RuntimeFramework(RuntimeType.Net, new Version("4.0.30319")), Has.Property(nameof(RuntimeFramework.IsAvailable)).True);
+
             var topLevelPackage = new TestPackage(new [] {"a.dll", "b.dll"});
 
             var net20Package = topLevelPackage.SubPackages[0];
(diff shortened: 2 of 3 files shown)


src/NUnitEngine/nunit.engine.api/IRuntimeFrameworkService.cs
```diff
@@ -45,6 +45,11 @@ public interface IRuntimeFrameworkService
...
         /// The package RuntimeFramework setting may be updated as a 
         /// result and the selected runtime is returned.
+        ///
+        /// Note that if a package has subpackages, the subpackages may run on a different
+        /// framework to the top-level package. In future, this method should
+        /// probably not return a simple string, and instead require runners
+        /// to inspect the test package structure, to find all desired frameworks.
         /// </summary>
         /// <param name="package">A TestPackage</param>
         /// <returns>The selected RuntimeFramework</returns>
```
ChrisMaddock: As a future breaking change to the Engine API - I'd like this method to return void. This method is quite side-effect based, and with this change, returning a simple string here is a little mis-leading.

src/NUnitEngine/nunit.engine.tests/Services/RuntimeFrameworkServiceTests.cs
```diff
@@ -92,6 +92,30 @@ public void EngineOptionPreferredOverImageTarget(string framework, int majorVers
...
+
+            _runtimeService.SelectRuntimeFramework(topLevelPackage);
+
+            Assert.Multiple(() =>
+            {
+                Assert.That(net20Package.Settings[EnginePackageSettings.RuntimeFramework], Is.EqualTo("net-2.0"));
+                Assert.That(net40Package.Settings[EnginePackageSettings.RuntimeFramework], Is.EqualTo("net-4.0"));
+                Assert.That(topLevelPackage.Settings[EnginePackageSettings.RuntimeFramework], Is.EqualTo("net-4.0"));
+            });
+        }
```
ChrisMaddock: This stuff ain't great to test, hence the restriction. Along with the refactor, I'd love us to set up some better mocks for this service!

src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs
```diff
@@ -106,27 +106,37 @@ public string SelectRuntimeFramework(TestPackage package)
...
                 requestedFramework = new RuntimeFramework(RuntimeType.Any, RuntimeFramework.DefaultVersion);
+                log.Debug($"No specific framework requested for {package.Name}");
             }
 
-            log.Debug("Current framework is {0}", currentFramework);
-            if (requestedFramework == null)
-                log.Debug("No specific framework requested");
-            else
-                log.Debug("Requested framework is {0}", requestedFramework);
-
```
ChrisMaddock: Logging change to make the logic clearer, previously most cases where No specific framework was requested would actually fall through to the `else` statement.

src/NUnitEngine/nunit.engine.api/IRuntimeFrameworkService.cs
CharliePoole: Unless it were only called by runners that actually __need__ a single string, e.g. the ProcessRunner. I have a vague memory that that was the case at one time.

ChrisMaddock: Ah, maybe it was! This string is currently unused within the engine.

rprouse:
I am going to override the failing Linux build and merge since the fix for that has already been merged.

mikkelbu:
@ChrisMaddock I guess that this should have fixed #667? (and not nunit/nunit#3338 😄)

ChrisMaddock:
D’oh, thank you!

PR #681, opened 2019-09-15
Display path and version of extension assemblies

CharliePoole:
This PR is intended to resolve the problem discussed in TestCentric/testcentric-gui#395 for the engine and console.

The engine fixes are already in use in the private build that's part of the TestCentric GUI. The console changes are intended to display output as suggested by @Sharken03 in that issue.

### DIFF
--- a/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
+++ b/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
@@ -332,6 +332,13 @@ namespace NUnit.ConsoleRunner
                     if(node.TargetFramework != null)
                         _outWriter.Write(ColorStyle.Value, $"(.NET {node.TargetFramework?.FrameworkVersion})");
                     _outWriter.WriteLine(node.Enabled ? "" : " (Disabled)");
+
+                    _outWriter.Write("      Version: ");
+                    _outWriter.Write(ColorStyle.Value, node.AssemblyVersion.ToString());
+
+                    _outWriter.Write("      Path: ");
+                    _outWriter.Write(ColorStyle.Value, node.AssemblyPath);
+
                     foreach (var prop in node.PropertyNames)
                     {
                         _outWriter.Write("      " + prop + ":");
--- a/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
+++ b/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
@@ -334,10 +334,10 @@ namespace NUnit.ConsoleRunner
                     _outWriter.WriteLine(node.Enabled ? "" : " (Disabled)");
 
                     _outWriter.Write("      Version: ");
-                    _outWriter.Write(ColorStyle.Value, node.AssemblyVersion.ToString());
+                    _outWriter.WriteLine(ColorStyle.Value, node.AssemblyVersion.ToString());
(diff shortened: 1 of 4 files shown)


src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -332,6 +332,13 @@ private void DisplayExtensionList()
                     if(node.TargetFramework != null)
                         _outWriter.Write(ColorStyle.Value, $"(.NET {node.TargetFramework?.FrameworkVersion})");
                     _outWriter.WriteLine(node.Enabled ? "" : " (Disabled)");
+
+                    _outWriter.Write("      Version: ");
+                    _outWriter.Write(ColorStyle.Value, node.AssemblyVersion.ToString());
```
mikkelbu:
```suggestion
                    _outWriter.WriteLine(ColorStyle.Value, node.AssemblyVersion.ToString());
```

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -332,6 +332,13 @@ private void DisplayExtensionList()
...
+
+                    _outWriter.Write("      Path: ");
+                    _outWriter.Write(ColorStyle.Value, node.AssemblyPath);
```
mikkelbu:
```suggestion
                    _outWriter.WriteLine(ColorStyle.Value, node.AssemblyPath);
```

src/NUnitEngine/nunit.engine.api/Extensibility/IExtensionNode.cs
```diff
@@ -72,5 +72,15 @@ public interface IExtensionNode
...
         /// <returns>A collection of values</returns>
         IEnumerable<string> GetValues(string name);
+
+        /// <summary>
+        /// The path to the assembly implementing this extension.
+        /// </summary>
+        string AssemblyPath { get; }
+
+        /// <summary>
+        /// The Version of the assembly implementing this extension.
```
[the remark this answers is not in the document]
CharliePoole: I intended it to mean the C# Type `Version`, but I'm OK with just using the word "version".

CharliePoole:
Yes... the newlines were what I intended. I couldn't build or run locally. I'm still relying entirely on CI builds when I work on this project.

CharliePoole:
I don't understand the build failure.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -332,6 +332,13 @@ private void DisplayExtensionList()
                     if(node.TargetFramework != null)
                         _outWriter.Write(ColorStyle.Value, $"(.NET {node.TargetFramework?.FrameworkVersion})");
```
ChrisMaddock: Now we're adding extra stuff to this output, I think it would be good if the assembly target framework also displayed on a new line in the same style as Version/Path? Happy for this to be a separate PR though, if you'd rather just get this merged. 🙂

CharliePoole: @ChrisMaddock My time is a bit short so I'll leave tuning the output format to another PR.

Issue #676, opened 2019-08-23
Nunit3 Console throws System.Runtime.Remoting.RemotingException : Tcp transport error


vegesnacreativeworks:
am trying to run the Nunit3 test using Nunit 3.10 Console runner. i am getting below error.

Error :
System.Runtime.Remoting.RemotingException : Tcp transport error.
----> System.Runtime.Remoting.RemotingException : Connection closed
--RemotingException
Tcp transport error.
Server stack trace:
at System.Runtime.Remoting.Channels.Tcp.TcpMessageIO.ReceiveMessageStatus (System.IO.Stream networkStream, System.Byte[] buffer) [0x00017] in :0
at System.Runtime.Remoting.Channels.Tcp.TcpClientTransportSink.ProcessMessage (System.Runtime.Remoting.Messaging.IMessage msg, System.Runtime.Remoting.Channels.ITransportHeaders requestHeaders, System.IO.Stream requestStream, System.Runtime.Remoting.Channels.ITransportHeaders& responseHeaders, System.IO.Stream& responseStream) [0x0005e] in :0
at System.Runtime.Remoting.Channels.BinaryClientFormatterSink.SyncProcessMessage (System.Runtime.Remoting.Messaging.IMessage msg) [0x00066] in :0

Exception rethrown at [0]:
at System.Runtime.Remoting.Proxies.RealProxy.PrivateInvoke (System.Runtime.Remoting.Proxies.RealProxy rp, System.Runtime.Remoting.Messaging.IMessage msg, System.Exception& exc, System.Object[]& out_args) [0x0014d] in <7a62ef3b49f14d16b21221b9b5a4792c>:0
at (wrapper managed-to-native) System.Object.__icall_wrapper_mono_remoting_wrapper(intptr,intptr)
at (wrapper remoting-invoke) NUnit.Engine.Agents.RemoteTestAgent.Run(NUnit.Engine.ITestEventListener,NUnit.Engine.TestFilter)
at NUnit.Engine.Runners.ProcessRunner.RunTests (NUnit.Engine.ITestEventListener listener, NUnit.Engine.TestFilter filter) [0x00025] in :0
RemotingException
Connection closed
at System.Runtime.Remoting.Channels.Tcp.TcpMessageIO.StreamRead (System.IO.Stream networkStream, System.Byte[] buffer, System.Int32 count) [0x00011] in :0
[...]

PR #683, opened 2019-09-23
Safely encapsulating the atomic agent database operations 

jnm2:
Best reviewed a commit at a time.

While working on the new communication code for https://github.com/nunit/nunit-console/issues/266, I needed to make the agent database verify the registration atomically. This would be with an atomic AgentDatabase.TryRegister method. The need for such a method raises an immediate problem: the way to tell whether registration has happened is by reading and writing a completely unprotected public field, AgentRecord.Agent. There can't be safety and atomicity so long as that is done outside the same lock as the lock on the dictionary state.

The current scopes of the locks do not match the fundamental state transitions (nonexistent to started, started to registered, registered to terminated) and queries. Rather, only lower-level dictionary operations have been thread-safe. Since multiple dictionary operations must be composed in order to complete a state transition, race conditions are possible. Even worse, unprotected public fields of AgentRecord were mutated as part of these transitions. Since this wasn't done inside the same lock as the dictionary operations, there was no guarantee of atomicity whatsoever. 

This PR brings immediate benefit to the current codebase and paves the way for what comes next.

Also, I've been unable to run tests using Test Explorer on either of my machines, so I have to admit to having typed all this blind in a manner of speaking. There's a weird filtering error shown in the Output Window Tests view. This might also be related to me having only .NET Core 2.1 and 2.2 installed. 1.0, 1.1 and 2.0 are deprecated, so if that's the problem, we should get to a place where people can contribute without having to install any of these.

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/Services/AgentDatabaseTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Services/AgentDatabaseTests.cs
@@ -130,7 +130,7 @@ namespace NUnit.Engine.Services.Tests
         {
             var database = new AgentDatabase();
 
-            Assert.That(database.IsAgentRunning(DummyAgent.Id, out _), Is.False);
+            Assert.That(database.IsAgentProcessActive(DummyAgent.Id, out _), Is.False);
         }
 
         [Test]
@@ -139,7 +139,7 @@ namespace NUnit.Engine.Services.Tests
             var database = new AgentDatabase();
 
             database.Start(DummyAgent.Id, DummyProcess);
-            Assert.That(database.IsAgentRunning(DummyAgent.Id, out var process), Is.True);
+            Assert.That(database.IsAgentProcessActive(DummyAgent.Id, out var process), Is.True);
             Assert.That(process, Is.SameAs(DummyProcess));
         }
 
@@ -150,7 +150,7 @@ namespace NUnit.Engine.Services.Tests
 
             database.Start(DummyAgent.Id, DummyProcess);
             database.Register(DummyAgent);
-            Assert.That(database.IsAgentRunning(DummyAgent.Id, out var process), Is.True);
+            Assert.That(database.IsAgentProcessActive(DummyAgent.Id, out var process), Is.True);
             Assert.That(process, Is.SameAs(DummyProcess));
         }
 
@@ -162,7 +162,7 @@ namespace NUnit.Engine.Services.Tests
             database.Start(DummyAgent.Id, DummyProcess);
             database.Register(DummyAgent);
             database.MarkTerminated(DummyAgent.Id);
-            Assert.That(database.IsAgentRunning(DummyAgent.Id, out _), Is.False);
+            Assert.That(database.IsAgentProcessActive(DummyAgent.Id, out _), Is.False);
         }
 
         [Test]
@@ -176,19 +176,19 @@ namespace NUnit.Engine.Services.Tests
                 {
                     var id = Guid.NewGuid();
 
-                    Assert.That(database.IsAgentRunning(id, out _), Is.False);
+                    Assert.That(database.IsAgentProcessActive(id, out _), Is.False);
                     Assert.That(database.IsReady(id, out _), Is.False);
 
                     database.Start(id, DummyProcess);
-                    Assert.That(database.IsAgentRunning(id, out _), Is.True);
+                    Assert.That(database.IsAgentProcessActive(id, out _), Is.True);
                     Assert.That(database.IsReady(id, out _), Is.False);
 
                     database.Register(new DummyTestAgent(id));
-                    Assert.That(database.IsAgentRunning(id, out _), Is.True);
+                    Assert.That(database.IsAgentProcessActive(id, out _), Is.True);
                     Assert.That(database.IsReady(id, out _), Is.True);
 
                     database.MarkTerminated(id);
-                    Assert.That(database.IsAgentRunning(id, out _), Is.False);
+                    Assert.That(database.IsAgentProcessActive(id, out _), Is.False);
                     Assert.That(database.IsReady(id, out _), Is.False);
                 }
             }, threadCount: Environment.ProcessorCount);
--- a/src/NUnitEngine/nunit.engine.tests/Services/AgentDatabaseTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Services/AgentDatabaseTests.cs
@@ -30,7 +30,7 @@ using NUnit.Framework;
 
 namespace NUnit.Engine.Services.Tests
 {
-    public class AgentDataBaseTests
+    public class AgentDatabaseTests
     {
 #pragma warning disable 414
         private static int[] Counts = new int[] { 1, 3, 10 };
@@ -38,13 +38,13 @@ namespace NUnit.Engine.Services.Tests
 
         const string COUNTS = nameof(Counts);
 
-        AgentDataBase _data;
+        AgentDatabase _data;
         List<Guid> _generatedGuids;
 
         [SetUp]
         public void SetUp()
         {
-            _data = new AgentDataBase();
+            _data = new AgentDatabase();
             _generatedGuids = new List<Guid>();
         }
 
--- a/src/NUnitEngine/nunit.engine.tests/Services/AgentDatabaseTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Services/AgentDatabaseTests.cs
@@ -1,5 +1,5 @@
 ﻿// ***********************************************************************
-// Copyright (c) 2017 Charlie Poole, Rob Prouse
+// Copyright (c) 2017–2019 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
 // a copy of this software and associated documentation files (the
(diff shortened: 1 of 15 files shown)


ChrisMaddock:
Thanks for breaking this out Joseph, looks like a good wee chunk! I’m travelling a bunch for work this week, but will hopefully get a chance to look through this properly soon. 😊

ChrisMaddock:
Sorry I haven't been able to get back to this yet Joseph - I had a quick look, but decided I need to pull it down to understand properly - which I need to be at a proper desktop to do. Will hopefully be able to have a look this weekend. 😊

src/NUnitEngine/nunit.engine/Services/AgentStatus.cs
```diff
@@ -29,11 +29,8 @@ namespace NUnit.Engine.Services
     /// </summary>
     public enum AgentStatus
     {
-        Unknown,
```
CharliePoole: I routinely use something like `Unknown` as the first enum member because that's what you get if you forget to initialize a field. If you drop it, then the default will become `Starting`.

src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
```diff
@@ -240,7 +240,7 @@ protected override void Dispose(bool disposing)
                     log.Error(ExceptionHelper.BuildMessageAndStackTrace(ex));
                 }
 
-                if (_agent != null && _agency.IsAgentRunning(_agent.Id))
+                if (_agent != null && _agency.IsAgentRunning(_agent.Id, out var process))
```
CharliePoole: I can't tell what the type of process is by inspecting this line.

CharliePoole: Note: that's a hypothetical "I"! Obviously, __I__ do know what it is, but someone who does not might think, for example, that it's `bool process; // if true, then process this.` :smile:

src/NUnitEngine/nunit.engine/Services/AgentRecord.cs
```diff
@@ -29,14 +29,12 @@ namespace NUnit.Engine.Services
 {
     internal sealed class AgentRecord
     {
-        private AgentRecord(Guid id, Process process, ITestAgent agent)
+        private AgentRecord(Process process, ITestAgent agent)
```
CharliePoole: Yeah... originally, I think the idea I had was that the agency might start a process but some other, second-level process might actually register. That turned out never to happen. Good lesson in YAGNI.

CharliePoole:
Much cleaner!

src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
jnm2: What is the benefit to knowing what the type name is while reading this block of code? Hypothetically, it might not be the process class but rather some kind of info struct with an ExitCode property, or even a `bool` variable if you don't read the code below.

In the scenarios in which I picture myself beginning to care what type it is, I'm always looking at how the value is used and where the value came from originally, which means I'm not just reading this file anymore. But I bet I'm missing something.

CharliePoole: When reading someone else's code, I'm always trying to understand what's going on. Being primarily a static typing kind of guy, knowing the Type makes sense to me. So, when I see this, I want to page around and find the declaration of the method so as to know the Type. That takes time.

That's why our guidelines (IIRC) say to use var only when the Type is clear, i.e. clear to the reader, since it's obviously clear to the compiler.

jnm2: Okay, happy to help save time!

src/NUnitEngine/nunit.engine/Services/AgentStatus.cs
jnm2: :+1: I was thinking about which becomes default as well.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -273,14 +223,9 @@ private static string GetTestAgentExePath(bool requires32Bit)
...
 
-        private void OnAgentExit(object sender, EventArgs e)
+        private void OnAgentExit(Process process, Guid agentId)
         {
-            var process = sender as Process;
-            if (process == null)
-                return;
-
-            var agentRecord = _agentData.GetDataForProcess(process);
-            agentRecord.Status = AgentStatus.Terminated;
```
ChrisMaddock: Hmm...there's a little change of behaviour here isn't there, from a silent failure to a crash, if `OnAgentExit` is every called with a null Process. Are we happy that would never be the case? 🤔 

I'm wandering about things like the issue on mono [here](https://github.com/nunit/nunit-console/issues/676), where it appears the agent is crashing, but the exit code error message is never printed, which suggests the OnAgentExit method is either not being called, or is being called with a null process. I'm not sure we really have testing to cover that sort of area, do we?

Blegh...mono. :-p

src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
```diff
@@ -249,7 +250,15 @@ protected override void Dispose(bool disposing)
                     }
                     catch (SocketException se)
                     {
-                        var exitCode = _agency.GetAgentExitCode(_agent.Id);
+                        int? exitCode;
+                        try
+                        {
+                            exitCode = process.ExitCode;
+                        }
+                        catch (NotSupportedException)
```
ChrisMaddock: Nice future-proofing 👍 (I presume!)

src/NUnitEngine/nunit.engine.tests/Services/AgentDatabaseTests.cs
```diff
@@ -0,0 +1,228 @@
...
+                }));
+            }
+
+            foreach (var thread in threads)
+                thread.Start();
+
+            foreach (var thread in threads)
+                thread.Join();
+
+            if (exceptions.Count != 0) throw exceptions[0];
```
ChrisMaddock: I was going to suggest `AggregateException` here, but apparently that's .NET 4 😢

src/NUnitEngine/nunit.engine.tests/Services/AgentDatabaseTests.cs
```diff
@@ -0,0 +1,228 @@
...
+            var database = new AgentDatabase();
+
+            database.Start(DummyAgent.Id, DummyProcess);
+            database.Register(DummyAgent);
+            database.MarkTerminated(DummyAgent.Id);
+            Assert.That(database.IsAgentRunning(DummyAgent.Id, out _), Is.False);
+        }
+
+        [Test]
+        public static void ConcurrentOperationsDoNotCorruptState()
```
ChrisMaddock: Nice

src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
jnm2: The try...catch was present before I moved this code. 

https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.exitcode#exceptions:

> NotSupportedException
> You are trying to access the ExitCode property for a process that is running on a remote computer. This property is available only for processes that are running on the local computer.

I've also run into an exception when accessing the ExitCode property for an exited process that I didn't start, but I don't remember which exception it was. NotSupportedException sounds familiar though.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
jnm2: Sharp. Sorry I forgot to mention it! I did think about this. It is a framework bug for `sender` to not be the same Process instance that you use to subscribe to the event.

> which suggests the OnAgentExit method is either not being called, or is being called with a null process

I'd prefer either a NullReferenceException or some more specific message rather than silently ignoring. That way we find out for sure if there is a buggy framework that doesn't follow the pattern of `sender` matching the instance that raised the event?

src/NUnitEngine/nunit.engine/Services/AgentDatabase.cs
```diff
@@ -0,0 +1,111 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace NUnit.Engine.Services
+{
+    /// <summary>
+    /// Defines the agent tracking operations that must be done atomically.
+    /// </summary>
+    internal sealed partial class AgentDatabase
```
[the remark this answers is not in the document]
jnm2: Good thought.

The [most recent time](https://gist.github.com/jnm2/7cd52dce0b7f761c9bd496dabdeeccaa#file-asyncparallelqueue-atomicoperations-cs) I wrote a class like this, I made it a nested class called `AtomicOperations`. I'm still not sure if that was an okay name.

What about `AgentCollection`? I like `AgentStore` too.

Or `AgencyState`?

src/NUnitEngine/nunit.engine/Services/AgentDatabase.cs
```diff
@@ -0,0 +1,111 @@
...
+                    agent = record.Agent;
+                    return true;
+                }
+
+                agent = null;
+                return false;
+            }
+        }
+
+        public bool IsAgentRunning(Guid agentId, out Process process)
```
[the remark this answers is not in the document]
jnm2: Good point! We should also be checking `process` for null in ProcessRunner in case the status was Starting.

Folks often expect the `out` parameters to be set when the return value is `true`. What do you think of the change I made, `IsAgentProcessActive`?

src/NUnitEngine/nunit.engine/Services/AgentDatabase.cs
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        public void Register(ITestAgent agent)
+        {
+            lock (_agentsById)
+            {
+                if (!_agentsById.TryGetValue(agent.Id, out var record)
+                    || record.Status != AgentStatus.Starting)
+                {
+                    throw new ArgumentException($"Agent {agent.Id} must have a status of 'starting' in order to register. Status was {agent.Status}.", nameof(agent));
```
ChrisMaddock:
```suggestion
                    throw new ArgumentException($"Agent {agent.Id} must have a status of 'starting' in order to register.", nameof(agent));
```

Sorry Joseph, my suggestion failed to compile. 🙈 We can skip it, it was only a thought...

jnm2: I pushed an easy fix, `record` instead of `agent` 😀

src/NUnitEngine/nunit.engine/Services/AgentDatabase.cs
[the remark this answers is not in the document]
ChrisMaddock: Oops, missed that we hadn't resolved this one yesterday. AgentCollection or AgentStore sound good to me! AtomicOperations feels a little overly generic, given the class is specific to the TestAgents. 🙂

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -231,30 +182,29 @@ private Guid LaunchAgentProcess(TestPackage package)
             log.Debug("Launched Agent process {0} - see nunit-agent_{0}.log", p.Id);
             log.Debug("Command line: \"{0}\" {1}", p.StartInfo.FileName, p.StartInfo.Arguments);
 
-            _agentData.Add( new AgentRecord( agentId, p, null, AgentStatus.Starting ) );
-            return agentId;
+            _agents.Start(agentId, p);
```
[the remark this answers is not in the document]
ChrisMaddock: Just adding a reminder to this comment - based on the AgentsCollection/Store convo above! 😄

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
ChrisMaddock: Agree with the principle - I'd just prefer we tested that on all supported platforms before we put it out in a release...and I'm not sure whether our CI will or not!

I've just pushed a temp branch to simulate an error here, let's see if CI fails. If it does, this sounds good to go - if not, I'd be keen to try and work out a way to test this on mono first.
https://github.com/nunit/nunit-console/compare/tmpCItest?expand=1

Of course, the real solution to this is to have better CI testing. 😄

src/NUnitEngine/nunit.engine/Services/AgentDatabase.cs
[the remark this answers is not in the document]
ChrisMaddock: Looks good. 🙂

src/NUnitEngine/nunit.engine/Services/AgentDatabase.cs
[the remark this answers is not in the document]
jnm2: Cool, updated!

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
ChrisMaddock: Ok, this was a yak. Turns out exceptions thrown in this method are lost, rendering the entire method redundant...

Separate issue, I think. Also, it's 1.30am here...so tomorrows problem now! Let's get this merged. 🙂

jnm2: There are so many Mono versions though, each with its own idiosyncrasies. Let me know and I'd be happy to put this back.

jnm2: ^ commented without seeing your comment. Whatever you are comfortable with!

ChrisMaddock:
Thanks Joesph, this looks good to go! If you don't mind, I'd prefer to get #690 merged first then rebase this one over the top - I think that will be a little easier than trying to rebase #690! I can sort that once @CharliePoole's finished up over there. 🙂

ChrisMaddock:
@jnm2 Hey, sorry, I hadn't spotted this was in a fork rather than a branch. Would you mind rebasing? I tried doing it myself, but it made all the commits joint-authored, which I didn't particularly want to do when this is all your work! Sorry! 👎

jnm2:
@ChrisMaddock I appreciate your approach with this, thank you!

ChrisMaddock:
Thank you! Merging after CI. 😄

PR #687, opened 2019-10-05
Separate addins files for nuget packages

### LINKED ISSUE #586: Create Separate Addin File for the Engine NuGet Package
The engine is now one directory deeper in the nuget package. We've worked around this by adding an extra line to the Console addin file, but we should separate into two files, one for each package. 

See #584

ChrisMaddock:
Fixes #586.

For the last release, we made a quick fix with the addins files for the NuGet packages, to correct a path bug that had been brought in when combining the netfx and .NET Standard engines into the same solution. This removes that quick fix, and creates two separate files for each nuget package.

I've tested the install of each package by hand, and extensions appear to be loaded correctly.

### DIFF
--- a/nuget/engine/nunit.engine.nuget.addins
+++ b/nuget/engine/nunit.engine.nuget.addins
@@ -1,0 +1,2 @@
+../../../NUnit.Extension.*/**/tools/     # nuget v2 layout
+../../../../NUnit.Extension.*/**/tools/  # nuget v3 layout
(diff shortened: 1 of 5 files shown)


ChrisMaddock:
Hey @nunit/engine-team, please could someone give this a quick look over? I'm keen to get this into 3.11. 🙂

ChrisMaddock:
Thanks Joseph!

Issue #684, opened 2019-10-01
Split engine into upper and lower parts


CharliePoole:
I did PR #580 "Split engine spike" as a spike, i.e. code to prove a concept, intended to be discarded and re-developed more carefully under test. Based on the spike, I'm writing this issue as a proposal to go ahead and move to the actual implementation. Comments on the spike already indicated a general consensus that this would be a desirable evolution of the engine. Before starting to code for real I'd like to have some review of what I propose here.

PROPOSAL

Split the current engine assembly nto two parts, a lower and an upper part, with the upper part making use of the lower part. The upper and lower parts would be built for potentially different targets - more for the lower part than the upper.

I will continue to call the upper part the `engine`. It needs to be built for platforms on which developers actually edit, compile and test their programs. Initially, I'd retain .NET 2.0 and .NET Core 2.0 for this purpose.

I'm calling the lower part the "agent core" for the moment. It needs to be built for any targets on which we want the tests to run, i.e. all current targets. If, however, we are about to remove any existing targets, it would be convenient to do it before starting this work.

APPROACH

My preference is to do this as a series of PRs, including tests, so that the change takes place in smaller, easier to review steps. If preferred, I can use a "big bang" approach, however.

REVIEW

I'm looking for a review of the general direction before starting work. Obviously, the details will be looked at again in code review but I'd like to know that this is wanted before spending time on it.

ChrisMaddock:
I agree the general direction covers where we want to go, and I'm glad to hear you're willing to contribute this work to this project, rather than just have this change in a fork. 

I think we need to consider the name for the "engine lower part" from the point of view of new codebase contributors - "agent core" suggests to me something which is specifically tied to the engine. Something like "Engine services" or "common services" was suggested before, which I personally prefer.

On the approach - I'd be very much in favour of smaller PRs.

One last consideration you mentioned on email Charlie, this change has the potential to cause merge conflicts with @jnm2's current work on replacing the engine/agent communication protocol. Given Joseph's picked that up again just last week - I'd prefer we try and avoid merge conflicts with that work as far as possible - I think it's currently blocked awaiting review of #683, which I'm afraid I've had limited time to look at recently. I expect some of the early incremental PR's can still be completed with limited conflict?

jnm2:
[...]

PR #689, opened 2019-10-06
WIP - Issue 684a

CharliePoole:
**Do Not Merge**

I'm creating a PR so this attempt at #684 can be see by everyone, but I'm starting another PR, which will split the engine into only two parts rather than three. For the current PR, I'd like it if team members would take a look and confirm my view that this is a dead end.

It's probably best to follow along by looking at one commit at a time.

Why a dead end?

I think the idea, as described in #684, of having an upper (client) engine, a lower (agent) piece and a common core on which they both depend is a good one. It would probably work well if we were starting from scratch.

What I've seen here, particularly in the last commit of this PR, is that there is so much interdependency between services and runners that the three-part division, while probably possible, would be extremely messy and confusing. All three assemblies end up with both runners and services and it will most likely be unclear to later devs why a particular assembly is used for one class while a different one is used for another. It's probably hard enough to grok it with two assemblies!

If folks look at this and decide it's not as bad as I think and want me to continue with it I will. Meanwhile, I'm going to redo the whole thing in the same steps, but using only two assemblies as I did in the original spike.


CharliePoole:
Closing this without merging, since #690 has replaced it.

PR #694, opened 2019-10-19
Bump engine tests to run on .NET Core 2.1

ChrisMaddock:
Fixes #693. Explanation in the issue

Issue #475, opened 2018-10-05
Create .NET Core Console Runner


ChrisMaddock:
A simple .NET Core console runner, utilising the existing .NET Standard engine. This will initially be a restricted version of the main console, due to limitations of the .NET Standard engine. Over time, we hope both engine and console will develop, and we will have near feature parity (where appropriate!)

Blocked by: https://github.com/nunit/nunit-console/issues/388

CharliePoole:
If there's a specific use case or cases for why we want to do this, I think it helps to add it to the issue for those of us who don't know what the cases are. :smile:

ChrisMaddock:
The primary use case would be the ability to run NUnit tests on platforms where neither the .NET Framework or Mono are installed. 🙂 

The only way we can currently do that is via dotnet test. As we know, running via the VS adapter has a number of restrictions - having our own runner allows us to not be subject to these.

CharliePoole:
So, for example, on Linux? Allowing the user to avoid Mono? Is there a call for that?

I write this question in part for my own purposes. I've tried to always give preference to using Mono when there is an option. OTOH .NET Core 3.0 is starting to look pretty usable at long last.

ChrisMaddock:
Absolutely - at least where I work! 😄 

I disagree with "3.0 is starting to look pretty usable" - we've been running 2.x in production for a year now. Mono doesn't come into our picture - we're running on Linux machines which only have .NET Core installed. I don't want to have to install mono in to that environment, just to run the NUnit Console, and then in turn have that spin up a .NET Core process, when it could just run in .NET Core in the first place. 🙂

CharliePoole:
Bear in mind that I'm totally a desktop developer in .NET. Integration of SWF and WPF make it an option now.

rprouse:
We are investing in .NET Core heavily too. All new server side development is being done in .NET Core for portability and performance. Mono doesn't come into the picture anymore except for mobile development with Xamarin. Mono isn't installed on any of our Linux servers. 1.4 was usable for most ASP.NET apps, but 2.0 hit the sweet spot for us. I am looking forward to 3.0 to start migrating some of our legacy code.

CharliePoole:
Yes it's clear that's the way things will be going - and already have to some extent. I'm a bit mystified about why needing to install Mono is seen as a bad thing - I do it on every Linux system I have.

[...]

PR #696, opened 2019-10-19
Minimal compilation/test of .NET Core Console

ChrisMaddock:
Contributes to #475 

This PR is intended to be the minimum change required to get a .NET Core version of the console up and running and under test. (It currently looks a lot bigger than it is due to being based off #694) I might also break this up further and take out some of the unrelated fixes for review.

This is only the first step, there's still lots to do. This PR specifically excludes:
- Removing unsupported command line options
- Packaging

There's a last couple of bits I want to tidy up which will have to wait for another day, but this is nearly ready to go. 🙂

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -315,11 +315,7 @@ Task("TestNetStandard16Engine")
     {
         if (IsDotNetCoreInstalled)
         {
-            RunDotnetCoreTests(
-                NETCOREAPP11_BIN_DIR + ENGINE_TESTS,
-                NETCOREAPP11_BIN_DIR,
-                "netcoreapp1.1",
-                ref ErrorDetail);
+            RunTest(NETCORE21_CONSOLE, NETCOREAPP11_BIN_DIR, ENGINE_TESTS, "netcoreapp1.1", ref ErrorDetail);
         }
         else
         {
@@ -339,11 +335,7 @@ Task("TestNetStandard20Engine")
     {
         if (IsDotNetCoreInstalled)
         {
-            RunDotnetCoreTests(
-                NETCOREAPP21_BIN_DIR + ENGINE_TESTS,
-                NETCOREAPP21_BIN_DIR,
-                "netcoreapp2.1",
-                ref ErrorDetail);
+           RunTest(NETCORE21_CONSOLE, NETCOREAPP21_BIN_DIR, ENGINE_TESTS, "netcoreapp2.1", ref ErrorDetail);
         }
         else
         {
--- a/build.cake
+++ b/build.cake
@@ -47,6 +47,7 @@ var ENGINE_API_CSPROJ = PROJECT_DIR + "src/NUnitEngine/nunit.engine.api/nunit.en
 var ENGINE_TESTS_CSPROJ = PROJECT_DIR + "src/NUnitEngine/nunit.engine.tests/nunit.engine.tests.csproj";
 var CONSOLE_CSPROJ = PROJECT_DIR + "src/NUnitConsole/nunit3-console/nunit3-console.csproj";
 var CONSOLE_TESTS_CSPROJ = PROJECT_DIR + "src/NUnitConsole/nunit3-console.tests/nunit3-console.tests.csproj";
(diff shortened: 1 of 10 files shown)


src/NUnitConsole/nunit3-console/nunit3-console.csproj
```diff
@@ -15,6 +14,9 @@
     <Compile Include="..\..\NUnitEngine\nunit.engine\Internal\ExceptionHelper.cs" LinkBase="Utilities" />
     <Content Include="..\..\..\nunit.ico" Link="nunit.ico" />
   </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)'=='netcoreapp2.1'">
+    <PackageReference Include="Mono.Options" Version="5.3.0.1" />
+  </ItemGroup>
```
ChrisMaddock: I chose to add this in as a dependency instead of a copy/paste when I first started on this, months ago. I've gone off the idea a little now - will investigate copying it over, as per the netfx version.

src/NUnitConsole/nunit3-console/Options.cs
```diff
@@ -2,9 +2,14 @@
 // Options.cs
```
ChrisMaddock: I did a wholesale update of Mono Options, to the latest version from [here](https://github.com/mono/mono/blob/master/mcs/class/Mono.Options/Mono.Options/Options.cs), for a version that compiles under both .NET 2.0 and .NET Core 2.1. This is potentially a little risky, in case mono have made any breaking changes. An alternative would be to keep the existing .NET 2.0 version of the file, and use different options.cs files per build - which should avoid any breaking changes, but could potentially lead to different command line usage options between the two builds of the console. 🤔 

What do people think?

ChrisMaddock:
This is now ready for review, would appreciate any feedback. 🙂

build.cake
```diff
@@ -257,18 +274,44 @@ Task("TestNet20Engine")
...
+            RunDotnetCoreTests(
+                NETCORE21_CONSOLE,
+                NETCOREAPP21_BIN_DIR, 
+                CONSOLE_TESTS,
+                "netcoreapp2.1",
+                ref ErrorDetail);
+        }
+        else
+        {
+            Warning("Skipping .NET Standard tests because .NET Core is not installed");
```
CharliePoole: S/b .NET Core?

src/NUnitConsole/nunit3-console/Options.cs
mikkelbu: I think it is fine to update it (hopefully we/they have tests), but it should probably be stated in the release notes, so people are aware of this.

ChrisMaddock:
Thanks both! I'll come back to these soon. 😄 

> I'm a bit surprised that so little had to change to make this work.

This was just to get it to compile and pass the tests...plenty more still to do! Realistically the console itself is fairly simple though - and the .NET Core APIs are getting pretty complete now, it seems. It was smaller than I expected too. though. 🙂

build.cake
```diff
@@ -192,29 +196,42 @@ Task("Build")
...
              MSBuild(ENGINE_API_CSPROJ, CreateMSBuildSettings("Publish")
                 .WithProperty("TargetFramework", framework)
                 .WithProperty("NoBuild", "true") // https://github.com/dotnet/cli/issues/5331#issuecomment-338392972
                 .WithProperty("PublishDir", BIN_DIR + framework));
 
-        Information("Publishing netcoreapp tests so that dependencies are present...");
+        foreach(var framework in NETCORE_FRAMEWORKS)
+             MSBuild(ENGINE_TESTS_CSPROJ, CreateMSBuildSettings("Publish")
+                .WithProperty("TargetFramework", framework)
+                .WithProperty("NoBuild", "true") // https://github.com/dotnet/cli/issues/5331#issuecomment-338392972
```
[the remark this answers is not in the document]
ChrisMaddock: Oops, thanks! I'll take out that line too, while I'm here. 😄

ChrisMaddock:
Thanks both - changes made. They were all minor, so I'll merge after CI has passed. 😄

PR #704, opened 2019-10-31
Agent in nupkg should not be referenced and causes warnings in consuming projects

jnm2:
Putting the agent in the `lib` folder of the nupkg is semantically wrong because that is only for assemblies that should be added to the consuming project's `/reference:` list when it compiles its source code. It's also been causing this warning to hang out in the error list and build log for any project that references the nunit.engine package:

> MSB3270 There was a mismatch between the processor architecture of the project being built "MSIL" and the processor architecture of the reference "C:\Users\Joseph\.nuget\packages\nunit.engine\3.10.0\lib\net20\nunit-agent-x86.exe", "x86". This mismatch may cause runtime failures. Please consider changing the targeted processor architecture of your project through the Configuration Manager so as to align the processor architectures between your project and references, or take a dependency on references with a processor architecture that matches the targeted processor architecture of your project.

After this is published, we'll be able to pull in the new version in the adapter project to address the warning in the error list.

To test, I built the nunit.engine package locally, added a nuget.config to the adapter repo adding the folder containing the nupkg as a package source, deleted the bin folder, and built the solution. The warning disappeared and all the same files showed up in build output as before.

Reference for the changes that were made: https://docs.microsoft.com/en-us/nuget/reference/nuspec#including-content-files


nuget/engine/nunit.engine.nuspec
```diff
@@ -45,20 +48,27 @@
...
-    <file src="../../nuget/engine/nunit.nuget.addins" target="lib/netstandard1.6"/>
+    <file src="../../nuget/engine/nunit.nuget.addins" target="content/any/netstandard1.6"/>
+    <file src="../../nuget/engine/nunit.nuget.addins" target="contentFiles/any/netstandard1.6"/>
     <file src="bin/netstandard2.0/nunit.engine.dll" target="lib/netstandard2.0" />
     <file src="bin/netstandard2.0/nunit.engine.core.dll" target="lib/netstandard2.0" />
     <file src="bin/netstandard2.0/nunit.engine.api.dll" target="lib/netstandard2.0" />
     <file src="bin/netstandard2.0/Mono.Cecil.dll" target="lib/netstandard2.0" />
-    <file src="../../nuget/engine/nunit.nuget.addins" target="lib/netstandard2.0"/>
+    <file src="../../nuget/engine/nunit.nuget.addins" target="content/any/netstandard2.0"/>
+    <file src="../../nuget/engine/nunit.nuget.addins" target="contentFiles/any/netstandard2.0"/>
```
jnm2: The addins file isn't moving because it's following the agent but rather because only assemblies to reference belong in the lib folder.

jnm2:
@ChrisMaddock That's good to check. This is all I know about it without having tried packages.config instead of PackageReference, from https://docs.microsoft.com/en-us/nuget/reference/nuspec#including-content-files:

> Content files are included in a package using the <files> element, specifying the content folder in the target attribute. However, such files are ignored when the package is installed in a project using PackageReference, which instead uses the <contentFiles> element.
>
> For maximum compatibility with consuming projects, a package ideally specifies the content files in both elements.

> I want to check the behaviour in nuget 3 and 4 when installing the package outside of a project, e.g. for CI.

Would the `nunit.engine` package ever be installed outside of a project? I was thinking that scenario would only use `nunit.runners` and `nunit.consolerunner`.

ChrisMaddock:
> Would the nunit.engine package ever be installed outside of a project? I was thinking that scenario would only use nunit.runners and nunit.consolerunner.

Ah, sorry - I'd missed this was just for the nunit.engine package and not every package.

Afraid I didn't get round to this last weekend, but I'd still like to test a little more if that's ok. Will get back to it. 🙂

ChrisMaddock:
Ok, I eventually got round to testing this with packages.config.

As far as I can tell, this change will break all users who are still on packages.config. Below's a screenshot of my bin dir after building a project referencing the NUnit.Engine package built by this PR.

image

I can't keep up with the NuGet ecosystem too well at the moment, but as far as I'm aware packages.config is still a supported method, and there are still [various limitations of PackageReference](https://docs.microsoft.com/en-gb/nuget/consume-packages/migrate-packages-config-to-package-reference#limitations) which prevent people upgrading. (We hit some of the package-based-limitations ourselves at my old job.)

I'm not too keen to merge a change which has the potential to be breaking for so many users, just to fix a warning message in a build log. Although it is frustrating that that's the case - once again, trying to support both the old and the new is biting us.

@nunit/engine-team Anyone else any thoughts here?

jnm2:
@ChrisMaddock I definitely share your concerns. Here's a fix I tested that preserves the copying to output when used with packages.config.

ChrisMaddock:
Hey Joseph - FYI I'll try this again on my project as soon as I'm back on that machine. 🙂

PR #715, opened 2019-12-12
Recognize all supported versions of the .NET Framework 4.x

### LINKED ISSUE #713: Engine will not recognize .NET Framework versions beyond 4.5
I found this when working on the GUI, which lists installed frameworks in it's menu. There's a bit of a complication in that some later framework versions use three-components, like 4.6.1. This breaks code in `RuntimeFramework`, which assumes it's dealing with a CLR version.

See more info in testcentric/testcentric-gui#447, which I'm working on. If desired, we can figure out how to backport the fix to NUnit when I finish it.

CharliePoole:
Fixes #713

### DIFF
--- a/src/NUnitEngine/nunit.engine.core/Internal/RuntimeFrameworks/DotNetFrameworkLocator.cs
+++ b/src/NUnitEngine/nunit.engine.core/Internal/RuntimeFrameworks/DotNetFrameworkLocator.cs
@@ -59,7 +59,7 @@ namespace NUnit.Engine.Internal.RuntimeFrameworks
                         }
                         else if (CheckInstallDword(versionKey))
                         {
-                            // Versons 1.1, 2.0, 3.0 and 3.5 are possible here
+                            // Versions 1.1, 2.0, 3.0 and 3.5 are possible here
                             yield return new RuntimeFramework(RuntimeType.Net, new Version(name.Substring(1, 3)));
                         }
                     }
--- a/src/NUnitEngine/nunit.engine.core/Internal/RuntimeFrameworks/DotNetFrameworkLocator.cs
+++ b/src/NUnitEngine/nunit.engine.core/Internal/RuntimeFrameworks/DotNetFrameworkLocator.cs
@@ -91,7 +91,6 @@ namespace NUnit.Engine.Internal.RuntimeFrameworks
 
         private static readonly MinimumRelease[] ReleaseTable = new MinimumRelease[]
         {
-            // TODO: Make 3-component versions work correctly
             new MinimumRelease(378389, new Version(4, 5)),
             new MinimumRelease(378675, new Version(4, 5, 1)),
             new MinimumRelease(379893, new Version(4, 5, 2)),
--- a/src/NUnitEngine/nunit.engine.core/Internal/RuntimeFrameworks/DotNetFrameworkLocator.cs
+++ b/src/NUnitEngine/nunit.engine.core/Internal/RuntimeFrameworks/DotNetFrameworkLocator.cs
@@ -59,8 +59,8 @@ namespace NUnit.Engine.Internal.RuntimeFrameworks
                         }
                         else if (CheckInstallDword(versionKey))
                         {
-                            // Versions 1.1 through 3.5
-                            yield return new RuntimeFramework(RuntimeType.Net, new Version(name.Substring(1)));
+                            // Versons 1.1, 2.0, 3.0 and 3.5 are possible here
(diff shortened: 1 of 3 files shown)


CharliePoole:
I wondered about that... for a long time we only distinguished net-4.0 and net-4.5, with the latter actually meaning "whatever version of 4.x you happen to have installed." I remember back when emulating 4.0 behavior under 4.5 was important to some people, causing us to make that distinction, but nobody has ever actually asked for this. OTOH, if we start checking TargetFrameworkAttribute, as @ChrisMaddoc has discussed, we'll find things like ".NETFramework,Version=v4.7.2" and end up saying we can't run the code without this fix.

src/NUnitEngine/nunit.engine.core/Internal/RuntimeFrameworks/DotNetFrameworkLocator.cs
```diff
@@ -59,8 +59,8 @@ public static IEnumerable<RuntimeFramework> FindDotNetFrameworks()
                         }
                         else if (CheckInstallDword(versionKey))
                         {
-                            // Versions 1.1 through 3.5
-                            yield return new RuntimeFramework(RuntimeType.Net, new Version(name.Substring(1)));
+                            // Versons 1.1, 2.0, 3.0 and 3.5 are possible here
```
ChrisMaddock:
```suggestion
                            // Versions 1.1, 2.0, 3.0 and 3.5 are possible here
```

🙂

src/NUnitEngine/nunit.engine.core/RuntimeFramework.cs
```diff
@@ -85,84 +85,87 @@ public RuntimeFramework(RuntimeType runtime, Version version)
...
-        private void InitFromClrVersion(Version version)
+        private Version GetFrameworkVersionForClr(Version clrVersion)
         {
-            this.FrameworkVersion = new Version(version.Major, version.Minor);
-            this.ClrVersion = version;
-            if (Runtime == RuntimeType.Mono && version.Major == 1)
-                this.FrameworkVersion = new Version(1, 0);
+            return Runtime == RuntimeType.Mono && clrVersion.Major == 1
+                ? new Version(1, 0)
+                : new Version(clrVersion.Major, clrVersion.Minor);
```
ChrisMaddock: Am I reading this right, that this is to detect Mono 1.0? This is the sort of complexity I'd love to strip out, in a future iteration! 😅

src/NUnitEngine/nunit.engine.core/Internal/RuntimeFrameworks/DotNetFrameworkLocator.cs
```diff
@@ -77,6 +77,33 @@ private static IEnumerable<RuntimeFramework> FindExtremelyOldDotNetFrameworkVers
...
+            public MinimumRelease(int release, Version version)
+            {
+                Release = release;
+                Version = version;
+            }
+        }
+
+        private static readonly MinimumRelease[] ReleaseTable = new MinimumRelease[]
+        {
+            // TODO: Make 3-component versions work correctly
```
ChrisMaddock: Can you elaborate on this comment, please? What isn't working correctly with them?

src/NUnitEngine/nunit.engine.core/RuntimeFramework.cs
CharliePoole: in principle, the CLR version doesn't necessarily have obvious relation to the framework version. You could have SomeFramework 1.0 with a CLR of 5.1234.5678. Mono was merely the first place where this showed up - Mono 1.0 is essentially equivalent to MS .NET 1.1.

Easiest way to simplify IMO is to figure out how to do without the CLR version entirely. In the meantime, I'm thinking of a change in my own engine where the individual Runtimes figure out the CLR version, so e.g. Mono.Runtime would have a method that allowed deducing the CLR version from the framework version. Otherwise, this code will get a lot more complicated for .NET Core and other runtimes we add.

src/NUnitEngine/nunit.engine.core/Internal/RuntimeFrameworks/DotNetFrameworkLocator.cs
CharliePoole: Obsolete comment... removing.

CharliePoole:
```suggestion
```

PR #716, opened 2019-12-20
Add a Dockerfile to the project to allow building on Linux

### LINKED ISSUE #453: build-mono-docker.ps1 fails to run out the box
I don't generally touch mono, but just tried out this script to debug https://github.com/nunit/nunit-console/pull/452. I followed the instructions in BUILDING.MD, which just say to install docker, and run `.\build-mono-docker.ps1 -t test`.

The script failed with the below message:

```
PS C:\Users\Chris\Documents\git\nunit-console> .\build-mono-docker.ps1 -t test
C:\ProgramData\chocolatey\lib\docker\tools\docker\docker.exe: error during connect: Post
 http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.33/containers/create: open //./pipe/docker_engine: 
The system cannot find the file specified. In the default daemon configuration on Windows, 
the docker client must be run elevated to connect. This error may also indicate that the 
docker daemon is not running.
See 'C:\ProgramData\chocolatey\lib\docker\tools\docker\docker.exe run --help'.
```

I've never touched Docker before, so have no idea what to do with that information! 😄 Anyone more familiar than I am have any guesses as to a fix we could put in?

rprouse:
Fixes #453 

This removes the dependency on my external docker image by adding the `Dockerfile` to this repository. The first build is slow because it must download and build the image, but after that it is just as quick as on Windows.

The docker image that is built is close to the build images that we use in Azure DevOps except that it uses a new version of Mono.


ChrisMaddock:
Turns out I don't have premium-enough Windows to install Docker any more, but let's get this merged. 🙂 Rob - I've pushed another commit to git-ignore the deb file, hope you don't mind. Will merge once CI has passed. 🙂

PR #724, opened 2020-01-04
[WIP] Workaround to load assemblies not located in same dir as engine under .NET Core

### LINKED ISSUE #710: .NET Core engine only works when located in same directory as test assembly

My code is:

            {
            }

I am using:
.NET Core 3.0
NUnit.Engine 3.10.0
NUnit.Engine.Api 3.10.0

ChrisMaddock:
Fixes #710. This is a hack for 3.11, rather than the proper solution.

This uses `Assembly.LoadFrom` in place of `Assembly.Load` to allow the engine to load assemblies which are not located in the same directory as itself. (I expect this is why we haven't seen this issue before, as the engine is always located in the same dir as the test assembly when used with the adapter.) It uses a bit of a hack to assume the framework also sits in the same dir.

The proper solution, I believe, would be to use `AssemblyLoadContext`. This is only available in .NET Core, and not .NET Standard - meaning we'd need to either convert the engine to .NET Core, or create a .NET Core agent. The latter is in our pipeline anyway - but there's a few bigger items in the way first (Replacing .NET remoting being the key one.)

Note this is only possible for the .NET Standard 2.0 build - .NET Standard 1.6 is stuck with the restriction that the test assembly and engine must be copied to the same directory. Given .NET Standard 1.6 is no longer supported by the framework (and it's already been mentioned before) - I'm hoping we can support from the next engine release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ matrix:
         - sudo apt-get install apt-transport-https
         - sudo apt-get update
         - sudo apt-get install dotnet-sdk-2.2
+        - sudo apt-get install dotnet-sharedframework-microsoft.netcore.app-1.0.16
 
 # macOS package restore currently taking >30 mins on Travis
 #    - os: osx
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -58,7 +58,8 @@ jobs:
       sudo dpkg -i packages-microsoft-prod.deb
       sudo apt-get install apt-transport-https
       sudo apt-get update
-      sudo apt-get install dotnet-sdk-2.2
+      sudo apt-get install dotnet-sdk-1.2
+      sudo apt-get install dotnet-sharedframework-microsoft.netcore.app-1.0.16
       ./build.sh --target=Test --configuration=Release
     displayName: Build and test
 
--- a/build.cake
+++ b/build.cake
@@ -317,9 +317,9 @@ Task("TestNetStandard16Engine")
         if (IsDotNetCoreInstalled)
         {
             RunDotnetCoreTests(
-                NETCORE21_CONSOLE,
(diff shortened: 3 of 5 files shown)


ChrisMaddock:
The last commit was required to fix the .NET Standard 1.6 tests. A little odd, as the .NET Standard build hasn't been changed at all by this PR.

They were being run under the .NET Core 2.1 console - I'm not too sure why that wouldn't work, but given we're not planning to support running .NET Core 1.1 on the console, I didn't put much effort into working out why.

ChrisMaddock:
Well hasn't this turned into a huge rabbit hole...

- We can't run out .NET Standard 1.6 tests on .NET Core 2.1. 
- We can't run them with NUnitLite like we used to, as .NET Core 1.2 is no longer installed on either of our Linux CI systems, and it's so old the packages can no longer be installed from the package manager.

I feel like I'm putting a whole lot of time into getting something working that nobody uses. I'm going to come back to this another day and remove the .NET Standard 1.6 build, and then put this PR on top of that. 😩

ChrisMaddock:
Ah, I think I get it. When tests are run under the .NET Core runner, they're actually loading the assemblies next to the .NET Core Console that they're running on, rather than the assembly that we think we're passing in. We're only getting away with it because the assemblies are named/signed the same.

So with this change our CI is actually running the .NET Standard 2.0 test assemblies when we think it's running .NET Standard 1.6. The failures I think are caused by when it tries to load the references from the test assembly - and finds the differences in the two builds in the nunit.engine.core assembly. 

That makes this a "bad change" - it's basically just obscuring the problem. I wonder instead if we can detect .NET Core at runtime (easy enough), and call the correct AssemblyLoadContext methods via reflection. One to try another day...

Issue #730, opened 2020-01-29
NUnit project file settings are ignored


cyanite:
The way TestPackages deal with settings means that settings always propagate down to sub packages, but never up. At launch, all console line settings are set on the top level package, and are propagated down to the next package, which could be an .nunit project.

This subpackage later (at expand time) loads its settings, such as process model and domain model, from the project file. These settings are never propagated back up to the top level TestPackage, however, so they never take proper effect.

This was already a problem in 3.10, but it still managed to work because NUnit would (incorrectly, I think) create an AggregateTestRunner for the single .nunit project, which would then in turn instantiate the correct runner (say a multi process version).

Since the aggregate test runner has been removed, this is no longer the case and NUnit will always use a single process runner even when the .nunit project specifies otherwise. In my case, this causes all my tests to fail.

So, this is a regression since 3.10, but in my opinion the settings handling was already broken earlier.

The simplest fix is this: Whenever GetSetting is called and we have exactly one sub package, forward the call to the sub package instead of handling it ourselves. I've created a pull request for this fix.

If you'd prefer a different fix that's fine, but it will leave us in the odd situation of taking action based on only console settings in several places.

cyanite:
It's unfortunate that I didn't debug this before the 3.11 release.

cyanite:
Other things are broken as well by this bug, I think, such as selecting the correct default configuration. This is also read on the top level package, but will always have the default value (or from command line options) there.

cyanite:
An alternative approach is to only do this at the top level, although this requires some further code change since a TestPackage currently doesn't know if it's the top level package.

CharliePoole:
UPDATE: OK... on the first reading, I missed "when we have exactly one subpackage." So I see you are already concerned about the problem of conflicting settings. However, I'll leave my more general comments because, as you will see, "exactly one subpackage" is not a sufficient qualification.

The engine passes settings downward if they are added with AddSetting. If you use Settings.Add, they only apply to the specific package.

So why (and when) do we want to pass settings downward? Answer: when we are overriding settings from the command-line. The command-line is the only place where we create settings and pass them downward, because the meaning of the options like --process is that they override individual settings, which would otherwise be used, on specific assemblies.

So where do the "natural" settings for each assembly come from? That is, the settings we are overriding?

[...]

PR #731, opened 2020-01-29
Propagate settings up from single sub package

cyanite:
For instance, this makes settings from nunit project files available at top level. This fixes a regression since 3.10.

I guard both SubPackages and Settings against null, since there is a constructor that doesn't always initialize them.

### DIFF
--- a/src/NUnitEngine/nunit.engine.api/TestPackage.cs
+++ b/src/NUnitEngine/nunit.engine.api/TestPackage.cs
@@ -158,6 +158,8 @@ namespace NUnit.Engine
         /// <returns></returns>
         public T GetSetting<T>(string name, T defaultSetting)
         {
+            if (SubPackages != null && SubPackages.Count == 1 && SubPackages[0].Settings != null)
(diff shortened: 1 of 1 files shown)


cyanite:
This fixes #730

cyanite:
> The "upward propogation" approach will break the engine in cases that you may not be encountering in your own work.

Are you sure? I only propagate (rather, call GetSetting on the sub package) when there is just one sub package, in which case we already down-propagated all settings to it (either at set time or at package expand time). So it will only break if a single sub package intentionally has different settings than its parent. I don't think that can happen at top level, but it can perhaps happen for assemblies? Not sure.

By the way, there are other things that are broken by this bug (already in 3.10) as well. I'll comment in the issue.

ChrisMaddock:
Hey @cyanite - thanks for digging into this. If we can, I'd much prefer to fix this in the overall architecture, rather than 'special-casing' this particular instance.

I'll comment further on the issue, to keep discussion in one place. 🙂

ChrisMaddock:
Closing in favour of #735 - thanks @cyanite!

PR #735, opened 2020-02-10
Use AggregatingTestRunner for non-trivial cases

### LINKED ISSUE #730: NUnit project file settings are ignored

cyanite:
This means cases involving at least one project file.

Fixes #730

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20OneAssemblyOneProjectExpectedRunnerResults.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/Results/Net20OneAssemblyOneProjectExpectedRunnerResults.cs
@@ -127,15 +127,9 @@ namespace NUnit.Engine.Tests.Services.TestRunnerFactoryTests.Results
                 case DomainUsage.None:
                 case DomainUsage.Single:
                 case DomainUsage.Multiple:
-                    return new RunnerResult
-                    {
-                        TestRunner = typeof(MultipleTestProcessRunner),
-                        SubRunners = new[]
-                        {
-                            RunnerResult.ProcessRunner,
-                            RunnerResult.ProcessRunner
-                        }
-                    };
+                    return new RunnerResult(typeof(AggregatingTestRunner),
+                        RunnerResult.MultiRunnerWithTwoSubRunners,
+                        RunnerResult.ProcessRunner);
                 default:
                     throw new ArgumentOutOfRangeException(nameof(domainUsage), domainUsage, ExceptionMessage);
             }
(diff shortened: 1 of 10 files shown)


cyanite:
See discussion in #730

CharliePoole:
@cyanite Your arguments for putting in this fix also work as arguments for it's being a critical bug - which we have always defined as one requiring an immediate pre-release.

@ChrisMaddock classified it as merely high priority, which would mean fix right away but don't release. That said, I could be reading too much into the classification. For me, it's good to have that classification right up front, because I would review a critical bug differently from any other. (FWIW I feel personally that this one is critical, because it broke a key feature.

That said, if it's not handled as critical, I would lean toward taking another week or come up with a fix that would survive new features that might come along. For example... nested projects, pre-expanded saved packages, etc. To me, that means actually checking to see if there are any projects involved, at any depth in the tree, using the ProjectService to recognize projects. That's not terribly hard to do, in fact, so i would lean to tossing the logic that uses subpackage count at all... btw, I think I may have been the one to come up with that logic, but things have moved on since then in other parts of the code, just not at the particular point where we are now looking.

@ChrisMaddock FWIW, in the long term, I think we probably want a bottom-up selection of runners, rather than top-down. That is, look at the lowest level subpackages first. Then, if a package has two subpackages with different settings, select the appropriate combining runner for those two packages. That would be a complete rewrite however, probably for NUnit 4 or later.

Bottom line, let's decide if this is critical. If a fix is going to be released in a matter of days, then the one @cyanite has done is perfect. But if it's just going to sit in master for a few months, I'd like to see a better one. I recognize that anyone can use the dev build if they need this fix, but I don't think that happens for a lot of organizations.

Finally, if we do this as a critical fix, can we leave the issue (or create a new one) so the fix gets fixed at a later time?

ChrisMaddock:
Thanks for digging into this! Will try and take a proper look at the weekend. 🙂

ChrisMaddock:
Hmm...this is irritating. Turns out we actually picked up on this change causing a potential problem when it was implemented - we just missed what the actual effect would be.

https://github.com/nunit/nunit-console/pull/622/files#r287597965


Thanks for looking into this, @cyanite. Merging this as a temporary fix.

At some point, we need to rewrite this area fully, it's proving unmaintainable. I'd like to do that as part of v4 - and remove domain isolation at the same time.

PR #741, opened 2020-02-18
Create NET40 agent and eliminate use of COMPLUS_Version variable

### LINKED ISSUE #740: Create separate agents for .NET 2.0-3.5 and .NET 4.x
We had some offline discussion related to the direction taken in the TestCentric GUI with issue TestCentric/testcentric-gui#518. At @ChrisMaddock request, I'll implement the same fix for NUnit.

Steps planned:

 - [x] Move current NET20 agents to a separate subdirectory under "agents"
 - [x] Create NET40 agent in a sibling directory
 - [x] Modify TestAgency to use correct agent and eliminate use of environment variable to distinguish
 - [ ] Modify packaging to use the same structure

Initially, we have to come up with a directory structure that works both in development, when tests run out of the bin directory and in packages where the package structure is what counts.

CharliePoole:
~Please review but don't merge - or make comments on #740~
Ready fo review and merge. Fixes #740

This commit is only partway there. I need some discussion on the issue regarding the file structure the team wants to use both in the bin directory and the packages and then I can finish. See that issue for a more extended explanation of what I'm asking.


jnm2:
I like this.

src/NUnitEngine/nunit-agent/nunit-agent-x86.csproj
```diff
@@ -2,16 +2,20 @@
...
     <ApplicationIcon>..\..\..\nunit.ico</ApplicationIcon>
     <PlatformTarget>x86</PlatformTarget>
     <IntermediateOutputPath>obj\$(Configuration)\x86\</IntermediateOutputPath>
+    <OutputPath>..\..\..\bin\$(Configuration)\agents\</OutputPath>
     <GenerateSupportedRuntime>false</GenerateSupportedRuntime>
   </PropertyGroup>
   <ItemGroup Condition="'$(TargetFramework)'=='net20'">
     <Reference Include="System.Runtime.Remoting" />
   </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)'=='net40'">
```
mikkelbu: I would prefer to have an or in the condition rather that two separate ItemGroups unless we expect that the references are going to be different on the two targets.

src/NUnitEngine/nunit-agent/nunit-agent.csproj
```diff
@@ -2,14 +2,18 @@
...
+    <TargetFrameworks>net20;net40</TargetFrameworks>
     <ApplicationManifest>app.manifest</ApplicationManifest>
     <ApplicationIcon>..\..\..\nunit.ico</ApplicationIcon>
     <GenerateSupportedRuntime>false</GenerateSupportedRuntime>
+    <OutputPath>..\..\..\bin\$(Configuration)\agents\</OutputPath>
   </PropertyGroup>
   <ItemGroup Condition="'$(TargetFramework)'=='net20'">
     <Reference Include="System.Runtime.Remoting" />
   </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)'=='net40'">
```
mikkelbu: Same as for src/NUnitEngine/nunit-agent/nunit-agent-x86.csproj

build.cake
```diff
@@ -373,7 +373,15 @@ Task("CreateImage")
...
-        CopyDirectory(BIN_DIR, CURRENT_IMG_DIR + "bin/");
+        // Temporary fix to keep packages stable until we settle on a new structure
+        //CopyDirectory(BIN_DIR, CURRENT_IMG_DIR + "bin/");
+        CopyDirectory(BIN_DIR + "net20", CURRENT_IMG_DIR + "bin/net20/");
+        CopyDirectory(BIN_DIR + "net35", CURRENT_IMG_DIR + "bin/net35/");
+        CopyDirectory(BIN_DIR + "netstandard1.6", CURRENT_IMG_DIR + "bin/netstandard1.6/");
+        CopyDirectory(BIN_DIR + "netstandard2.0", CURRENT_IMG_DIR + "bin/netstandard2.0/");
+        CopyDirectory(BIN_DIR + "netcoreapp1.1", CURRENT_IMG_DIR + "bin/netcoreapp1.1/");
+        CopyDirectory(BIN_DIR + "netcoreapp2.1", CURRENT_IMG_DIR + "bin/netcoreapp2.1/");
+        CopyFiles(BIN_DIR + "agents/net20/nunit-agent*", CURRENT_IMG_DIR + "bin/net20/");
```
mikkelbu: Probably obvious, but we are missing copy of net40 agent

src/NUnitEngine/nunit-agent/nunit-agent-x86.csproj
CharliePoole: It's interesting that with the new SDK-style projects, readability of project files becomes a thing, just as it was once a thing for Nant scripts in the distant past.

Of course, the question is "What do we think is most readable here?" I thought separating each one would be better, although it's not a very strong preference for me.

@ChrisMaddock @jnm2 What do you think? What's the best practice for combining or not combining these in your views.

One suggestion. Let's avoid __negative__ tests. The problem with them is that they impact runtimes that are added __later on__.

src/NUnitEngine/nunit-agent/nunit-agent.csproj
CharliePoole: Same again.

build.cake
CharliePoole: Ah! The tests all pass because we don't test or check content of packages! (referencing earlier discussion with you on TestCentric project) :smiling_imp: I'll fix.

src/NUnitEngine/nunit-agent/nunit-agent-x86.csproj
CharliePoole: I'll leave this as is until there's some more feedback and change as needed in a later commit.

build.cake
CharliePoole: Actually, this too depends on further comments on the structure proposed in #740

src/NUnitEngine/nunit-agent/nunit-agent-x86.csproj
ChrisMaddock: I would expect an 'or' by convention, but don't feel strongly. I guess it would highlight the similarities of the two platform builds to a future contributor.

> readability of project files becomes a thing

One of the biggest steps forward in my time in .NET, in my opinion! 😄 The less cryptic magic buried in enormous XML files, the better. If anything, I wish the VS experience was a little more oriented towards editing the xml directly, rather than still be GUI-first.

CharliePoole: OK... easy fix.

CharliePoole: Ah! Not quite so easy because VS wants to create the runtime subdirectories. I guess that's why I made a parallel agents subdirectory (it's been a few weeks, I'm afraid).

jnm2: VS should not behave differently at all when reading this versus the original:

```cs
  <ItemGroup Condition="'$(TargetFramework)'=='net20' or '$(TargetFramework)'=='net40'">
    <Reference Include="System.Runtime.Remoting" />
  </ItemGroup>
```

CharliePoole: Correct. But people have stylistic preferences, so I'll change it. Of course, the original didn't have net40 at all.

build.cake
CharliePoole: Resolved in last commit.

CharliePoole:
The second commit deals with earlier comments, updates the packaging and adds a Task to do simple checks on the content of the packages. Third commit deals with a problem due to the zip and msi package naming convention being different from that of the nuget and chocolatey files. The zip and msi don't include any pre-release suffix. I'm not sure why that is so, but I fixed the checks to expect it since any naming change would be out of scope for this PR.

ChrisMaddock:
Hi @CharliePoole - just FYI this is still on my radar to take a look at, afraid things are a little busy at the moment. :-)

rprouse:
I'll leave the final approval to @ChrisMaddock, but the code and the idea look good to me.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -134,13 +134,11 @@ private Process LaunchAgentProcess(TestPackage package, Guid agentId)
                     string.Format("The {0} framework is not available", targetRuntime),
                     "framework");
 
-            string agentExePath = GetTestAgentExePath(useX86Agent);
+            string agentExePath = GetTestAgentExePath(targetRuntime, useX86Agent);
```
ChrisMaddock, line 137: I think this code needs moving down a few lines - it looks like target runtime isn't finalised until the code on line 150. Now we need to know exactly which framework version we want to use before selecting the agent, we need to do that a little later, I think.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -211,16 +203,27 @@ private ITestAgent CreateRemoteAgent(TestPackage package, int waitTime)
...
             string engineDir = NUnitConfiguration.EngineDirectory;
             if (engineDir == null) return null;
 
+            string agentsDir = Path.Combine(engineDir, "agents");
+            log.Debug($"Checking for agents at {agentsDir}");
+
+            if (!Directory.Exists(agentsDir))
+            {
+                agentsDir = Path.Combine(Path.GetDirectoryName(engineDir), "agents");
+                log.Debug($"Directory not found! Using {agentsDir}");
```
ChrisMaddock, line 217: Minor: How about removing the second calculation of agentsDir? Just thinking this may break in future, if someone alters one line, but not the other. 🙂 (I also think this will actually log the wrong path - `engineDir` is already a directory, so calling `Path.GetDirectoryName` on it will get the directory a level up)


e.g.
```suggestion
                log.Debug($"Directory not found! Using {agentsDir}");
```

nuget/runners/nunit.console-runner.nuspec
```diff
@@ -26,10 +26,24 @@
...
+    <file src="..\..\nuget\runners\nunit.console.nuget.addins" target="tools/agents/net20"/>
+    <file src="bin/agents/net40/nunit-agent.exe" target="tools/agents/net40" />
+    <file src="bin/agents/net40/nunit-agent.exe.config" target="tools/agents/net40" />
+    <file src="bin/agents/net40/nunit-agent-x86.exe" target="tools/agents/net40" />
+    <file src="bin/agents/net40/nunit-agent-x86.exe.config" target="tools/agents/net40" />
+    <file src="bin/agents/net40/nunit.engine.api.dll" target="tools/agents/net40" />
+    <file src="bin/agents/net40/nunit.engine.api.xml" target="tools/agents/net40" />
+    <file src="bin/agents/net40/nunit.engine.core.dll" target="tools/agents/net40" />
+    <file src="bin/agents/net40/Mono.Cecil.dll" target="tools/agents/net40" />
+    <file src="..\..\nuget\runners\nunit.console.nuget.addins" target="tools/agents/net40"/>
```
ChrisMaddock, line 46: I know this was just copy-paste, but while we're here, please can we switch them to cross-platform friendly path separators? 😁

Good job remembering to include the addins files though...not sure I would have. 

```suggestion
    <file src="../../nuget/runners/nunit.console.nuget.addins" target="tools/agents/net20"/>
    <file src="bin/agents/net40/nunit-agent.exe" target="tools/agents/net40" />
    <file src="bin/agents/net40/nunit-agent.exe.config" target="tools/agents/net40" />
    <file src="bin/agents/net40/nunit-agent-x86.exe" target="tools/agents/net40" />
    <file src="bin/agents/net40/nunit-agent-x86.exe.config" target="tools/agents/net40" />
    <file src="bin/agents/net40/nunit.engine.api.dll" target="tools/agents/net40" />
    <file src="bin/agents/net40/nunit.engine.api.xml" target="tools/agents/net40" />
    <file src="bin/agents/net40/nunit.engine.core.dll" target="tools/agents/net40" />
    <file src="bin/agents/net40/Mono.Cecil.dll" target="tools/agents/net40" />
    <file src="../../nuget/runners/nunit.console.nuget.addins" target="tools/agents/net40"/>
```

build.cake
```diff
@@ -41,6 +42,7 @@ var IMAGE_DIR = PROJECT_DIR + "images/";
 var MSI_DIR = PROJECT_DIR + "msi/";
 var CURRENT_IMG_DIR = IMAGE_DIR + $"NUnit-{productVersion}/";
 var CURRENT_IMG_NET20_BIN_DIR = CURRENT_IMG_DIR + "bin/net20/";
+var CURRENT_IMG_AGENTS_BIN_DIR = CURRENT_IMG_DIR + "bin/agents";
```
ChrisMaddock, line 45:
```suggestion
```

Looks like this was unused in the end. 🙂

nuget/runners/nunit.console-runner.nuspec
ChrisMaddock: Ack - on second thoughts, the addin's file is going to break isn't it, cos the relative paths inside it will be starting from a directory a level deeper. 

In the long run, we need to work on that solution to having agents not load their own extensions. In the short term - how about creating a separate `nunit.agents.nuget.addins` file?

msi/nunit/engine-files.wxi
```diff
@@ -10,7 +10,10 @@
             <Component Id="NUNIT.ENGINE.CORE" Location="local" Guid="F4F14D75-6096-467E-82E1-BD513645C0EF">
```
ChrisMaddock, line 10: Changes to the msi look good - but we need to sort out an addins file for the agents I think, or we'll break some existing extensions that are currently packaged with the msi. Same comments on the NuGet package also apply here, I think we'll need a separate addins file.

msi/nunit/engine-files.wxi
```diff
@@ -19,33 +22,69 @@
...
+                      ProcessorArchitecture="x86"
+                      Source="$(var.InstallImage)bin/agents/net20/nunit-agent-x86.exe" />
+                <File Id="nunit_agent_net40_x86.exe.config"
+                      Source="$(var.InstallImage)bin/agents/net20/nunit-agent-x86.exe.config" />
+            </Component>
+            <!--<Component Id="NUNIT_AGENT_NET40_DEPENDENCIES" Location="local" Guid="CCAE29F5-4767-4A63-BB55-28C99B28E4DC">
+                <CopyFile Id="nunit.agent.net40.engine.api.dll" FileId="nunit.engine.api.dll" />
+                <CopyFile Id="nunit.agent.net40.engine.core.dll" FileId="nunit.engine.core.dll" />
+                <CopyFile Id="nunit.agent.net40.Mono.Cecil.dll" FileId="Mono.Cecil.dll" />
+            </Component>-->
```
ChrisMaddock, line 87: This can be removed now, right? 🙂

package-checks.cake
```diff
@@ -0,0 +1,237 @@
...
+            HasFiles("LICENSE.txt", "license.rtf", "NOTICES.txt", "CHANGES.txt"),
+            HasDirectory("bin/net20").WithFiles("nunit3-console.exe", "nunit3-console.exe.config").AndFiles(ENGINE_FILES),
+            HasDirectory("bin/net35").WithFiles("nunit3-console.exe", "nunit3-console.exe.config").AndFiles(ENGINE_FILES),
+            HasDirectory("bin/netstandard1.6").WithFiles(ENGINE_FILES),
+            HasDirectory("bin/netstandard2.0").WithFiles(ENGINE_FILES),
+            HasDirectory("bin/netcoreapp1.1").WithFiles(ENGINE_FILES),
+            HasDirectory("bin/netcoreapp1.1").WithFiles(ENGINE_FILES),
+            HasDirectory("bin/agents/net20").WithFiles(AGENT_FILES),
+            HasDirectory("bin/agents/net40").WithFiles(AGENT_FILES)) &
+        CheckMsiPackage("NUnit.Console", HasDirectory("NUnit.org/nunit-console")); // Placeholder for later implementation
```
ChrisMaddock, line 62: This is a great step in the right direction, thanks for adding these! Can I suggest adding on the addins files? Given they're about to get a little more complicated, with different files for different levels of different packages, this would be a great way to validate that they're being gathered up correctly. 😄

src/NUnitEngine/nunit-agent/Program.cs
```diff
@@ -95,22 +95,6 @@ public static void Main(string[] args)
...
-                log.Debug("Agent process has the COMPLUS_Version environment variable value \"{0}\" overridden with \"{1}\", restoring the original value.", cpvOriginal, Environment.GetEnvironmentVariable("COMPLUS_Version"));
-                Environment.SetEnvironmentVariable("TestAgency_COMPLUS_Version_Original", null, EnvironmentVariableTarget.Process); // Erase marker
-                Environment.SetEnvironmentVariable("COMPLUS_Version", (cpvOriginal == "NULL" ? null : cpvOriginal), EnvironmentVariableTarget.Process); // Restore original (which might be n/a)
-              }
-            }
-            catch(Exception ex)
-            {
-              log.Warning("Failed to restore the COMPLUS_Version variable. " + ex.Message); // Proceed with running tests anyway
-            }
-
```
ChrisMaddock, line 113: Hooray for removing cryptic old code! 🎉

src/NUnitEngine/nunit.engine.tests/Integration/RemoteAgentTests.cs
```diff
@@ -7,7 +7,7 @@ namespace NUnit.Engine.Tests.Integration
 {
     public sealed class RemoteAgentTests : IntegrationTests
     {
-        [Test]
+        [Test, Ignore("Temporary, while changing directory structure")]
```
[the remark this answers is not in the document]
CharliePoole: Fixed. It actually took an additional fix to make the test run.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
CharliePoole: I didn't notice that. I suspect that line 150 may no longer be needed but I'll have to check. I think we want all the decisions about target runtime to have been made before we get to this point if possible!

nuget/runners/nunit.console-runner.nuspec
CharliePoole: Right on both counts!

msi/nunit/engine-files.wxi
CharliePoole: Agreed!

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
CharliePoole: What you say about getting the parent dir is exactly what's intended! We run in two environments. When running tests in the output directory, we need a different relative path from what we do in the packages. I'll verify that this is correct first and then add a comment since it is a bit confusing.

nuget/runners/nunit.console-runner.nuspec
CharliePoole: Oh! Were you referring to the discussion we had about having the extensions located by the engine, which would leave info in the `TestPackage` for the agent to find?

build.cake
CharliePoole: Yes, it was a remnant of my earlier proposal to change the structure under BIN_DIR so that it would match the intended package structure.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
CharliePoole: Hmm... I don't call that method at all in the TestCentric engine. In NUnit it's only called here. Assuming the comment on the method definition is correct, it will only change `targetRuntime` if no version has been specified. We don't have any tests to verify that's what it does, however, so I'll add those first.

The only way the situation can arise is if the user specifies something like `--framework:net` on the command-line. IMO, this should get resolved long before we are at the point of launching a process, but I need to test whether that's true.

IOW this is a __really__ good catch!

CharliePoole: @ChrisMaddock OK, I did some experimenting. It turns out that RuntimeFrameworkService never passes a runtime type without version (like `net`) to the `TestAgency`. It's all resolved in that service beforehand so `TestAgency` can never change the target at line 150 and the call does nothing.

I checked it out in two ways: Running tests with an argument of `--framework:Net` and stepping through the debug build. That's what I expected, because I knew I had removed this call in the TestCentric engine

Based on that, I'll remove the line. If you're worried, I can add some Guard statements that affirm what we believe: that targetFramework is set, includes a version and is available at the start of the method.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
CharliePoole: Confirmed and added a comment. The two calculations give two diferent results, as intended.

nuget/runners/nunit.console-runner.nuspec
CharliePoole: OK, I added a file into each of the agent directories containing only `../../` in order to forward to the addins file in the root directory. I'm going to build and test that.

package-checks.cake
```diff
@@ -0,0 +1,237 @@
...
+
+    if (!FileExists(path))
+    {
+        WriteError("Package was not found!");
+        return false;
+    }
+
+    bool allOK = true;
+
+    // TODO: Figure out how to check content of msi
```
jnm2, line 102: We could shell out to msiexec:

```cs
var exitCode = StartProcess(
    "msiexec",
    new ProcessArgumentBuilder()
        .Append("/a")
        .AppendQuoted(path)
        .AppendQuoted("TARGETDIR=" + someTemporaryDirectory)
        .Append("/quiet")
        .Render());
```

jnm2: https://docs.microsoft.com/en-us/windows/win32/msi/command-line-options

CharliePoole: @jnm2 Yup, we could install it. I was trying to figure out how to do it without installing, however, as I do for the nuget and chocolatey packages. I did it that way because you need to run as administrator to install a chocolatey package.

I also hesitate to get too deep into the msi, with which I'm no longer very familiar.

jnm2: I didn't realize this would install the msi. Are you sure it does?

nuget/runners/nunit.console-runner.nuspec
CharliePoole: This works as expected. By using `../../` in each agent, the actual content of the the main addins file doesn't have to be replicated.

package-checks.cake
CharliePoole: Well, /a is the administrative install option, which basically creates a share from which others install. It may be worth trying so as to get a directory where a check could be performed.

CharliePoole: Yeah, that works! It gives me a temp directory with a copy of the msi __and__ an NUnit.org directory containing the entire installation. So, I guess I'll add it to the checks.

CharliePoole: @jnm2 OK... it works programmatically... __finally__!

I decided to just roll my own with strings rather than using `ProcessArgumentBuilder`... 'cause... that's me! :smiley: 

In terms of efficiency, probably not the best idea but actually a good one for learning. I am amazed to discover that `msiexec` does not tolerate forward slashes in paths. And the error message it gave me about an invalid character replaced the '/' with a '.' because (apparently) it things a slash is non-representable!

Yet another reason to hate the whole msi thing!

CharliePoole:
This is ready for review again. The latest commit implements the suggestions in the first reviews, except where my comments indicated otherwise. I used @jnm2's suggestion of doing a temporary administrative install of the msi and verifying it's contents in the `CheckPackages` task.

ChrisMaddock:
Hi Charlie - sorry, this is still on my radar, afraid I've been focusing on other unblocking issues today.

I want to give it a proper look through as some of your conclusions above don't match my memory. It's more than likely I'm just remembering how things work wrong - but I want to be sure of that before merging! 🙂 

Sorry this one's been out here a while.

CharliePoole:
@ChrisMaddock If those are conclusions in some comments I resolved, perhaps you could unresolve them when you get a chance so I know which ones.

ChrisMaddock:
Hi Charlie - the bit I'm not sure about is the redundant line. That doesn't match my memory of how things worked here, but it has been a while. I'd like to have a proper look at that when I get a chance, if that's ok. 🙂

CharliePoole:
@ChrisMaddock That's cool. I'll wait for this to clear.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
ChrisMaddock: Thanks, that makes more sense. Thanks for adding the comment

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
ChrisMaddock: Thanks for holding on for this one. I checked, and you're right, I was remembering the code incorrectly. But I'm much happier for double checking!

ChrisMaddock:
Hi @CharliePoole - thanks for waiting on this one. Everything looks good to me!

Just wanted to check - did you already verify agent-extension-loading for the three package types manually? Just aware that we don't have any automated testing around that, and would rather find it now than at release!

If that's all been sorted already, then feel free to merge. 😄

CharliePoole:
@ChrisMaddock Let me double-check first. Have you looked at the package tests I run in the TestCentric build script? There are currently nine different test runs for each of the three package types.

CharliePoole:
Yup. Re-verified all 3 manually.

Issue #720, opened 2020-01-01
Deprecate deprecated packages on nuget.org


ChrisMaddock:
https://docs.microsoft.com/nuget/nuget-org/deprecate-packages

Nuget.org's package deprecation functionality is now available. 😄 We should make use of it for our two deprecated packages.

ChrisMaddock:
The nunit-console-with-extensions Choco package needs to be added to the list for removal from the repo, it catches me out every time. 🙁

CharliePoole:
Two packages are deprecated and Chris added a third package, which he recommended we deprecate. Currently deprecated packages use the convention-based approach of indicating they are deprecated in the title and having a dependency on the replacement package where there is one.

Currently, we are re-publishing the deprecated packages, using a new version number, every time we publish. I don't think this makes sense, because the whole idea is that we don't want to keep maintaining these packages. @nunit/engine-team What do you think? Am I missing some reason why this is useful?

## Packages

Details follow on each of the three packages. They are all packages for nuget.org unless otherwise stated.

### nunit.engine.netstandard

This is an old alternative engine, which we no longer use. The NUnit.Engine package replaces it. We currently release this with an new version number every time we publish packages.

__Actions:__ Stop Publishing; Convert to NuGet Deprecation Feature (?)

### NUnit.Runners

This is an old naming for the current NUnit.Console package, which includes both NUnit.Console.Runner and our five standard extensions.  We currently release this with an new version number every time we publish packages.

__Actions:__ Stop Publishing; Convert to NuGet Deprecation Feature (?); Update Description

### nunit-console-with-extensions (Chocolatey package)

This is an old package with no exact equivalent. In the Chocolatey environment, bundling extensions doesn't make a lot of sense because it's so trivial to add extensions from the command line. This package was added in error and we don't have any precise equivalent - users should install nunit-console-runner plus whichever extensions they want.

__Actions:__ Stop Publishing; Deprecate on Chocolatey.org; Update Description

## Explanation of Actions

### Stop Publishing

We would remove the publication of the deprecated packages from the build script. The `nuspec` files would be kept and moved to a special folder in case a modified version needs to be pushed in the future. We would add instructions about how to push them manually in that case.

### Convert to NuGet Deprecation Feature

Although this is basically the main point of the issue, I have marked this action with a question mark in each case.

@nunit/engine-team What advantage do you see in this conversion, as applied to packages we have already deprecated in the convention-based way.

### Deprecate on Chocolatey.org

[...]

PR #746, opened 2020-04-06
Issue 733 : Add <icon> in nuspec files

### LINKED ISSUE #733: iconUrl is deprecated in NuGet packages
The `<iconUrl>` element used in our nuspec files has been deprecated. We should switch to using `<icon>` instead. See the docs [here](https://docs.microsoft.com/en-us/nuget/reference/nuspec#iconurl).

Contributions very welcome here, if anyone's willing! 😁 I'd recommend copying our existing images into the `/nuget/` folder in the directory, for the simplest solution, and then upgrading the nuspec files accordingly.

mano-si:
Fixes #733 

- Pulled the previously used icon file from nunit/resources repository and checked in a copy.
- Changes to only nuspec files to replace <iconUrl> with <icon>.
- Tested by :
   1. Package build : Succeeded. 
   2. Installed nuget package from local source to a vs project. : Icon visible. Installed successfully.


mano-si:
@CharliePoole @ChrisMaddock Please review this PR.  :)

nuget/deprecated/nunit.engine.netstandard.nuspec
```diff
@@ -9,7 +9,7 @@
     <license type="file">LICENSE.txt</license>
     <projectUrl>https://nunit.org</projectUrl>
     <repository type="git" url="https://github.com/nunit/nunit-console"/>
-    <iconUrl>https://cdn.rawgit.com/nunit/resources/master/images/icon/nunit_256.png</iconUrl>
+    <icon>images\nunit_256.png</icon>
```
CharliePoole, line 12: Not much point here, since I assume we don't issue new versions of the deprecated package.

mano-si: Yes, I guess we can skip here. I had modified it just to be consistent.

mano-si:
@mikkelbu Thanks for letting me know about the vm issue. I was already wandering what had gone wrong :D . So, I will wait for the Azure DevOps build to pass as that issue gets resolved. 

Regarding the backward compatibility, I guess we can keep `iconUrl`. But I am not too sure here.

nuget/deprecated/nunit.engine.netstandard.nuspec
mikkelbu: Personally, I would go for consistency.

CharliePoole: But if you don't publish it isn't that an even worse inconsistency?

mano-si: Am ok with either way. But I think @CharliePoole  has a point. Let me revert the change for the deprecated package.

CharliePoole: @mano-si To be clear, I was __asking__ whether the project releases new versions of the deprecated package. I don't know the answer but just assumed they don't. It's a manual process, BTW.

mikkelbu: The package was also published in the latest release - version 3.11.1- https://www.nuget.org/packages/nunit.engine.netstandard/.

CharliePoole: I see you're maintaining the packages even though they are deprecated. That's not what I assumed, and it's not how I thought deprecation was supposed to work. If you update those packages every release, then exactly what does it mean that they are deprecated? Rather than deprecated, it seems like they are just aliases for the newer package names. That's certainly an approach you can take, but I wouldn't call it deprecation.

In any case, since my assumption was wrong, I'll back out of this discussion. Now that GitHub has an official notion of deprecating packages, you might want to look at that at some point in the future.

ChrisMaddock: Deprecated packages are still being released for now as a low-cost solution to keeping people on the latest version. 🙂 #720 tracks removing them properly.

Thanks for updating these as well, @mano-si!

mano-si:
> @mano-si This looks great - thanks for pulling it together!
> 
> I agree with @mikkelbu that it would be good to keep `iconUrl` for now, as well as `icon`. That also matches the official advice from NuGet. ([link](https://docs.microsoft.com/en-us/nuget/reference/nuspec#icon))
> 
> Would you mind putting the iconUrl lines back in, along with the new icon format? We should be able to merge then. 😄

Added iconUrl back. Checked with a Package build and it was successful. We are good here then. 😄 

Thanks @ChrisMaddock  @mikkelbu @CharliePoole  for helping me do my first os contribution. 😃

ChrisMaddock:
Brilliant - thanks @mano-si! Welcome to open source!

Let us know if you're interested in working on some more issues, I have a couple of ideas. 😄

PR #747, opened 2020-04-06
Change macOS image version

mikkelbu:
From macOS-10.13 to macOS-10.15 - see e.g. https://devblogs.microsoft.com/devops/removing-older-images-in-azure-pipelines-hosted-pools/

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -102,7 +102,9 @@ jobs:
       packageType: runtime
       version: 2.1.x
 
-  - bash: ./build.sh --target=Test --configuration=Release
+  - bash: |
+     dotnet tool install --global Cake.Tool
+     dotnet cake --target=Test --configuration=Release
     displayName: Build and test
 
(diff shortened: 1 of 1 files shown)


ChrisMaddock:
Thanks for tackling this Mikkel! 🙌

azure-pipelines.yml
```diff
@@ -90,10 +90,23 @@ jobs:
...
+      
+  - task: UseDotNet@2
+    displayName: 'Install .NET Core runtime 2.1'
+    inputs:
+      packageType: runtime
+      version: 2.1.x
+
+  - bash: |
+     dotnet tool install --global Cake.Tool
+     dotnet cake --target=Test --configuration=Release
```
ChrisMaddock, line 109: I haven't seen this before. Cute! 🙂

mikkelbu: I could have gotten some inspiration from https://github.com/nunit/nunit/blob/master/azure-pipelines.yml#L84-L87 😄 (I had to do this otherwise cake could not use the installed SDK),

PR #749, opened 2020-04-24
Issue-391:To provide useful error message when agent crashes with StackOverflow exception

mano-si:
Added a case to handle exit code of `-1073741571` and provide meaningful message to user.

### DIFF
--- a/src/NUnitEngine/nunit.engine/Services/TestAgency.cs
+++ b/src/NUnitEngine/nunit.engine/Services/TestAgency.cs
@@ -252,50 +252,23 @@ namespace NUnit.Engine.Services
                 case AgentExitCodes.UNABLE_TO_LOCATE_AGENCY:
                     errorMsg = "Remote test agent unable to locate agency process.";
                     break;
-                //case AgentExitCodes.STACK_OVERFLOW_EXCEPTION:
-                //    errorMsg = "Remote test agent was terminated due to a StackOverFlow exception in one of the tests.";
-                //    break;
+                case AgentExitCodes.STACK_OVERFLOW_EXCEPTION:
+                    if (Environment.OSVersion.Platform == PlatformID.Win32NT)
+                    {
+                        errorMsg = "Remote test agent was terminated due to a stack overflow.";
+                    }
+                    else
+                    {
+                        errorMsg = $"Remote test agent exited with non-zero exit code {process.ExitCode}";
+                    }
+                    break;
                 default:
-                    //errorMsg = $"Remote test agent exited with non-zero exit code {process.ExitCode}";
-                    errorMsg = GetDefaultError(process.ExitCode);
+                    errorMsg = $"Remote test agent exited with non-zero exit code {process.ExitCode}";
                     break;
             }
 
             throw new NUnitEngineException(errorMsg);
         }
-
-        private string GetDefaultError(int errorCode)
-        {
-            string defaultErrorMessage = $"Remote test agent exited with non-zero exit code {errorCode}";
-
-            switch (Environment.OSVersion.Platform)
-            {
-                case PlatformID.Win32NT:
-                    return GetWindowsError(errorCode);
-
-                case PlatformID.Unix:
-                    return defaultErrorMessage;
-
-                case PlatformID.MacOSX:
-                    return defaultErrorMessage;
-
-                default:
-                    return defaultErrorMessage;
-            }
-        }
-
-        private string GetWindowsError(int errorCode)
-        {
-            switch (errorCode)
-            {
-                case AgentExitCodes.STACK_OVERFLOW_EXCEPTION:
-                    return "Remote test agent was terminated due to a stack overflow.";
-
-                default:
-                    return $"Remote test agent exited with non-zero exit code {errorCode}";
-            }
-        }
-
 
         public IServiceLocator ServiceContext { get; set; }
 
--- a/src/NUnitEngine/nunit.engine/Services/TestAgency.cs
+++ b/src/NUnitEngine/nunit.engine/Services/TestAgency.cs
@@ -252,6 +252,9 @@ namespace NUnit.Engine.Services
                 case AgentExitCodes.UNABLE_TO_LOCATE_AGENCY:
                     errorMsg = "Remote test agent unable to locate agency process.";
                     break;
+                case AgentExitCodes.STACK_OVERFLOW_EXCEPTION:
+                    errorMsg = "Remote test agent was terminated due to a StackOverFlow exception in one of the tests.";
+                    break;
                 default:
                     errorMsg = $"Remote test agent exited with non-zero exit code {process.ExitCode}";
                     break;
--- a/src/NUnitEngine/nunit.engine/Services/TestAgency.cs
+++ b/src/NUnitEngine/nunit.engine/Services/TestAgency.cs
@@ -252,16 +252,50 @@ namespace NUnit.Engine.Services
                 case AgentExitCodes.UNABLE_TO_LOCATE_AGENCY:
                     errorMsg = "Remote test agent unable to locate agency process.";
                     break;
-                case AgentExitCodes.STACK_OVERFLOW_EXCEPTION:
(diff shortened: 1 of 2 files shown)


mano-si:
> Do we have documentation that this exit code only occurs on stack overflow?

I could not find any resource to be sure.

CharliePoole:
If we are not sure what the return code means, then the message shouldn't imply that we are sure. We could reword it to say it's an unknown may be a stack overflow in one of your tests. BTW, these exit codes may be different depending on the platform we run... that will be important in the future, when we start running .NET Core tests.

An idea I just thought of, which is ugly but might work...

You said that a time delay allowed the proper message to display. What if we put a time delay in the default case? That would hopefully happen less frequently and we might learn something.

mano-si:
> If we are not sure what the return code means, then the message shouldn't imply that we are sure. We could reword it to say it's an unknown may be a stack overflow in one of your tests. BTW, these exit codes may be different depending on the platform we run... that will be important in the future, when we start running .NET Core tests.
> 
> An idea I just thought of, which is ugly but might work...
> 
> You said that a time delay allowed the proper message to display. What if we put a time delay in the default case? That would hopefully happen less frequently and we might learn something.

Agreed. We could print a generic message with a hint that its probably a stackoverflow. That makes sense.

Regarding, putting a time delay in default case. That's interesting. When I did that and tested, it didn't work as I would have thought. The `OnAgentExit()` method is unable to throw exception in that case and also we don't get the stack trace of `NUnit.Engine.NUnitEngineUnloadException`. I suspect this happens because the first time when I put a Thread.Sleep() in `OnAgentExit()` it was at the start of the method which meant `_agents.MarkTerminated(agentId);` was never executed and the main thread exited before. I could be totally wrong here. 

Also, `NUnit.Engine.NUnitEngineUnloadException` full stacktrace doesn't necessarily provide valuable information. It did provide me the exit code which anyway we did in the default case handling.

It was something like this

```
NUnit.Engine.NUnitEngineUnloadException : System.Net.Sockets.SocketException : No connection could be made because the target machine actively refused it 127.0.0.1:57385
Agent connection was forcibly closed. Exit code was -1073741571.
--SocketException
No connection could be made because the target machine actively refused it 127.0.0.1:57385

Server stack trace:
   at System.Net.Sockets.Socket.DoConnect(EndPoint endPointSnapshot, SocketAddress socketAddress)
   at System.Net.Sockets.Socket.Connect(EndPoint remoteEP)
   at System.Runtime.Remoting.Channels.RemoteConnection.CreateNewSocket(EndPoint ipEndPoint)
   at System.Runtime.Remoting.Channels.SocketCache.GetSocket(String machinePortAndSid, Boolean openNew)
   at System.Runtime.Remoting.Channels.Tcp.TcpClientTransportSink.SendRequestWithRetry(IMessage msg, ITransportHeaders requestHeaders, Stream requestStream)
   at System.Runtime.Remoting.Channels.Tcp.TcpClientTransportSink.ProcessMessage(IMessage msg, ITransportHeaders requestHeaders, Stream requestStream, ITransportHeaders& responseHeaders, Stream& responseStream)

Exception rethrown at [0]:
   at NUnit.Engine.Agents.RemoteTestAgentProxy.Stop() in C:\GitHub\nunit-console\src\NUnitEngine\nunit.engine\Agents\RemoteTestAgentProxy.cs:line 60
   at NUnit.Engine.Runners.ProcessRunner.Dispose(Boolean disposing) in C:\GitHub\nunit-console\src\NUnitEngine\nunit.engine\Runners\ProcessRunner.cs:line 251
```

I am honestly not quite sure of a possible great solution here. 

@CharliePoole  @ChrisMaddock  thoughts ??

ChrisMaddock:
Hi folks,

-1073741571 translates to 0xC00000FD which is documented [here](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/596a1078-e883-4972-9bbc-49e60bebca55?redirectedfrom=MSDN) to be reserved for stack overflows.

Charlie makes a good point about this potentially being platform specific. Can I suggest adding a check here that we're running on Windows ([https://docs.microsoft.com/en-us/dotnet/api/system.platformid?view=netframework-2.0](example)), and only output the error message if so? I also agree the message should perhaps be "less certain" - I'll suggest some alternative wording in a minute.

On the race condition, is it worth having a play around and see what works? Ideally we want to show the error message from `OnAgentExit` whenever we exit with something other than `AgentExitCode.Ok` - but of course we don't want to add any delay to the normal case where everything works as it should! I'm not too familiar with the order of events here, that may not be possible.

Let's not worry about suppressing the SocketException in this issue - we can leave that for another time, if we even think it's worthwhile. 🙂

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -252,6 +252,9 @@ private void OnAgentExit(Process process, Guid agentId)
                 case AgentExitCodes.UNABLE_TO_LOCATE_AGENCY:
                     errorMsg = "Remote test agent unable to locate agency process.";
                     break;
+                case AgentExitCodes.STACK_OVERFLOW_EXCEPTION:
+                    errorMsg = "Remote test agent was terminated due to a StackOverFlow exception in one of the tests.";
```
ChrisMaddock, line 256:
```suggestion
                    errorMsg = "Remote test agent was terminated due to a stack overflow.";
```

How about something as simple as this? That way we're not saying any more than we definitely know.

mano-si:
> Hi folks,
> 
> -1073741571 translates to 0xC00000FD which is documented [here](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/596a1078-e883-4972-9bbc-49e60bebca55?redirectedfrom=MSDN) to be reserved for stack overflows.
> 
> Charlie makes a good point about this potentially being platform specific. Can I suggest adding a check here that we're running on Windows ([example](https://docs.microsoft.com/en-us/dotnet/api/system.platformid?view=netframework-2.0)), and only output the error message if so? I also agree the message should perhaps be "less certain" - I'll suggest some alternative wording in a minute.
> 
> On the race condition, is it worth having a play around and see what works? Ideally we want to show the error message from `OnAgentExit` whenever we exit with something other than `AgentExitCode.Ok` - but of course we don't want to add any delay to the normal case where everything works as it should! I'm not too familiar with the order of events here, that may not be possible.
> 
> Let's not worry about suppressing the SocketException in this issue - we can leave that for another time, if we even think it's worthwhile. 🙂

Excellent! Thanks for finding the documentation. 😃  I also like the check for windows. We can do that.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
mano-si: Yes it would be better. I will change it.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -252,14 +252,51 @@ private void OnAgentExit(Process process, Guid agentId)
                 case AgentExitCodes.UNABLE_TO_LOCATE_AGENCY:
                     errorMsg = "Remote test agent unable to locate agency process.";
                     break;
+                //case AgentExitCodes.STACK_OVERFLOW_EXCEPTION:
+                //    errorMsg = "Remote test agent was terminated due to a StackOverFlow exception in one of the tests.";
+                //    break;
                 default:
-                    errorMsg = $"Remote test agent exited with non-zero exit code {process.ExitCode}";
+                    //errorMsg = $"Remote test agent exited with non-zero exit code {process.ExitCode}";
+                    errorMsg = GetDefaultError(process.ExitCode);
```
mano-si, line 260: This is just a draft and an afterthought. My intention is to create something, given the environment and exit code would give me back the most likely error message.

ChrisMaddock:
I think separate error codes files per environment is probably overkill at the moment - we only have one code on one environment that needs that right now. 🙂 I think an if statements fine for now, and we can always build it out more if we need to in future.

Issue #748, opened 2020-04-22
Make Project config information available to runners


CharliePoole:
The TestCentric GUI needed to be able to list all the configurations available for a project. In general, any GUI runners would probably need this in order to list the available configs so the user could select one. It also needed to be able to recognize projects in the first place in order to know when such a menu should be displayed. While this could be done in other ways, the simplest thing was to call the `ProjectService` taking advantage of whatever project loader extensions were installed in the engine.

Because `IProjectService` was not originally provided in the API assembly, I added it there originally. Now that TestCentric is using the packaged NUnit API assembly, it is no longer available. The features that made use of it have now been disabled. Fortunately, they were never released, so it causes no problems to the users right now.

While there are other approaches, such as saving the information in the `TestPackage`, exposing `IProjectService` in the API assembly seems like the simplest thing all around. If this seems like a reasonable thing to do I can provide a PR. I've marked it high priority from my perspective of needing the interface. :smile: 

Please comment @nunit/engine-team !

ChrisMaddock:
Sounds sensible to me. Could you propose an API, Charlie?

CharliePoole:
@ChrisMaddock All I did in my (old) testcentric code was move the current api definition out of the internal namespace in `nunit.engine` and into `nunit.engine.api`. But you're right - this may actually be an opportunity to redefine the API if we need to. I'll take a look at that.

CharliePoole:
Here's the API I was using. It matches NUnit's, except for the `LoadFrom` method, which I added.

The `IProject` interface returned from `LoadFrom` is already in the API assembly because it is used by extenders. In fact, it's a bit odd that I never made `IProjectService` public. I think that was because I wanted runners to not have to worry about projects. In fact, they don't unless they want some internal info like what configurations are supported.

```
    /// <summary>
    /// The IProjectService interface is implemented by ProjectService.
    /// It knows how to load projects in a specific format and can expand
    /// TestPackages based on projects.
    /// </summary>
    public interface IProjectService
    {
        /// <summary>
        /// Returns true if the file indicated is one that this
        /// loader knows how to load.
        /// </summary>
        /// <param name="path">The path of the project file</param>
        /// <returns>True if the loader knows how to load this file, otherwise false</returns>
        bool CanLoadFrom(string path);

        /// <summary>
        /// Loads a project of a known format.
        /// </summary>
        /// <param name="path">The path of the project file</param>
[...]

PR #753, opened 2020-04-27
[WIP] Add new interface with project config information

CharliePoole:
Option A for fixing #748

This is the approach that @ChrisMaddock and I discussed on the issue itself. It's no longer my preferred approach. See discussion on #784.


CharliePoole:
Closing this PR in favor of continuing with #754.

PR #754, opened 2020-04-27
Add config info to expanded project packages

CharliePoole:
This is option B to issue #748 

This is now my preferred approach, which doesn't involve changing the interface. See comparison on #748

### DIFF
--- a/src/NUnitEngine/nunit.engine.core/EnginePackageSettings.cs
+++ b/src/NUnitEngine/nunit.engine.core/EnginePackageSettings.cs
@@ -38,6 +38,11 @@ namespace NUnit
         public const string ActiveConfig = "ActiveConfig";
 
         /// <summary>
+        /// A list of the avaialble config names for this project.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


ChrisMaddock:
I'm quite happy with this solution.

In a different world - I wonder if the package should actually include all configs, expanded and ready to go - and then the config the user passes in just acts as a filter. 🤔 That's way out of scope for this issue though, just a shower-thought!

CharliePoole:
@ChrisMaddock Thanks for the review. I'll merge.

WRT your comment, I don't think it calls for a different world, just a different PR. :smile: 

Main problem with proliferating agents in the currentt environment is that they all are making private copies of the engine assemblies they use. That's not necessary but a little bit of work is needed to avoid it. I didn't do that work because it's not needed now with just two agents.

CharliePoole:
Since the console runner doesn't use this feature, I tested in the GUI by applying the identical changes in the TestCentric engine. Works fine.

PR #759, opened 2020-05-03
Use same current directory for agent as is being used by the runner

### LINKED ISSUE #758: Carry CurrentDirectory over to agent Processes
This issue is identical to TestCentric/testcentric-gui#577 which will may end up creating an additional incompatibiility with the nunit engine.

Beginning with NUnit 3, we made it a policy that we wouldn't change the user's working directory (aka CurrentdDirectory) and that we wouldn't rely on the working directory for much of anything - currently only as a default location for saving logs and result files.

We followed through with that promise by removing all NUnit code that changes the working directory and providing the `TestDirectory` property to users who need to know where the current test assembly is located.

However, from a naive user point of view, there is __one place__ where the CD is changed. When we launch an agent process, the CD of that process is the location of the agent assembly. This is true no matter what directory the user was in when the test was launched.

This is easy to fix. `TestAgency` merely has to set `Process.StartInfo.WorkingDirectory` to the value of `Environment.CurrentDirectory`. Then, tests run with `--process:Multiple` will have the same working directory as those run in process.

I don't consider this a breaking change, since it's highly unlikely that anyone is counting on the cd being set to directory where the agent is installed. However if the @nunit/engine-team feels it is breaking, we should track it as a potential V4 change.

CharliePoole:
Fixes #758 

The first commit refactors the private method `TestCentric.LaunchAgentProcess` to a class, `AgentProcess`, in order to enable testing. Tests of existing function are added.

The second commit sets one integration test to Ignore. See subsequent comment about this.

The third commit actually makes the desired change.

### DIFF
--- a/src/NUnitEngine/nunit.engine.core/Internal/ServerBase.cs
+++ b/src/NUnitEngine/nunit.engine.core/Internal/ServerBase.cs
@@ -54,7 +54,7 @@ namespace NUnit.Engine.Internal
         }
         
         // Currently virtual to allow NSubstitute to mock it. In fact, it will
-        // probably be virtual or abstract at some point, if we implemment
+        // probably be virtual or abstract at some point, if we implement
         // alternate transports for different runtime targets
         public virtual string ServerUrl
         {
--- a/src/NUnitEngine/nunit.engine.core/Internal/ServerBase.cs
+++ b/src/NUnitEngine/nunit.engine.core/Internal/ServerBase.cs
@@ -52,8 +52,11 @@ namespace NUnit.Engine.Internal
             this.uri = uri;
             this.port = port;
         }
-
-        public string ServerUrl
+        
+        // Currently virtual to allow NSubstitute to mock it. In fact, it will
+        // probably be virtual or abstract at some point, if we implemment
+        // alternate transports for different runtime targets
+        public virtual string ServerUrl
         {
             get { return string.Format("tcp://127.0.0.1:{0}/{1}", port, uri); }
         }
--- a/src/NUnitEngine/nunit.engine/Services/AgentProcess.cs
+++ b/src/NUnitEngine/nunit.engine/Services/AgentProcess.cs
@@ -48,6 +48,7 @@ namespace NUnit.Engine.Services
 
             StartInfo.UseShellExecute = false;
             StartInfo.CreateNoWindow = true;
+            StartInfo.WorkingDirectory = Environment.CurrentDirectory;
             EnableRaisingEvents = true;
 
             if (TargetRuntime.Runtime == RuntimeType.Mono)
--- a/src/NUnitEngine/nunit.engine/Services/AgentProcess.cs
+++ b/src/NUnitEngine/nunit.engine/Services/AgentProcess.cs
@@ -1,0 +1,108 @@
+// ***********************************************************************
+// Copyright (c) Charlie Poole and TestCentric Engine contributors.
+// Licensed under the MIT License. See LICENSE.txt in root directory.
+// ***********************************************************************
+
+#if NETFRAMEWORK
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+using NUnit.Engine;
+using NUnit.Engine.Internal;
+
+namespace NUnit.Engine.Services
+{
+    public class AgentProcess : Process
+    {
+        private static readonly Logger log = InternalTrace.GetLogger(typeof(AgentProcess));
+
+        public AgentProcess(TestAgency agency, TestPackage package, Guid agentId)
+        {
+            // Get target runtime
+            string runtimeSetting = package.GetSetting(EnginePackageSettings.RuntimeFramework, "");
+            TargetRuntime = RuntimeFramework.Parse(runtimeSetting);
+
+            // Access other package settings
+            bool runAsX86 = package.GetSetting(EnginePackageSettings.RunAsX86, false);
+            bool debugTests = package.GetSetting(EnginePackageSettings.DebugTests, false);
+            bool debugAgent = package.GetSetting(EnginePackageSettings.DebugAgent, false);
+            string traceLevel = package.GetSetting(EnginePackageSettings.InternalTraceLevel, "Off");
+            bool loadUserProfile = package.GetSetting(EnginePackageSettings.LoadUserProfile, false);
+            string workDirectory = package.GetSetting(EnginePackageSettings.WorkDirectory, string.Empty);
+
+            AgentArgs = new StringBuilder($"{agentId} {agency.ServerUrl} --pid={Process.GetCurrentProcess().Id}");
+
+            // Set options that need to be in effect before the package
+            // is loaded by using the command line.
+            if (traceLevel != "Off")
+                AgentArgs.Append(" --trace=").EscapeProcessArgument(traceLevel);
+            if (debugAgent)
+                AgentArgs.Append(" --debug-agent");
+            if (workDirectory != string.Empty)
+                AgentArgs.Append(" --work=").EscapeProcessArgument(workDirectory);
+
+            AgentExePath = GetTestAgentExePath(TargetRuntime, runAsX86);
+
+            log.Debug("Using nunit-agent at " + AgentExePath);
+
+            StartInfo.UseShellExecute = false;
+            StartInfo.CreateNoWindow = true;
+            EnableRaisingEvents = true;
+
+            if (TargetRuntime.Runtime == RuntimeType.Mono)
+            {
+                StartInfo.FileName = RuntimeFramework.MonoExePath;
+                string monoOptions = "--runtime=v" + TargetRuntime.ClrVersion.ToString(3);
+                if (debugTests || debugAgent) monoOptions += " --debug";
+                StartInfo.Arguments = string.Format("{0} \"{1}\" {2}", monoOptions, AgentExePath, AgentArgs);
+            }
+            else if (TargetRuntime.Runtime == RuntimeType.Net)
+            {
+                StartInfo.FileName = AgentExePath;
+                StartInfo.Arguments = AgentArgs.ToString();
+                StartInfo.LoadUserProfile = loadUserProfile;
+            }
+            else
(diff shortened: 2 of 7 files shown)


CharliePoole:
@jnm2 

As noted in the description, I marked the single integration test to be ignored. If it were to be run, it would need to set up the directory to contain an `agents` subdirectory with at least one agent target subdirectory. I didn't want to do it in this PR although I'd be glad to merge in any change you want to make.

I know this was an approach integration testing you were experimenting with. You set up temp directories to look like the environment in which the packaged software actually runs and then run some tests. Depending on the level of the tests, these are integration tests or possibly system tests.

I have some thoughts of my own on this level of testing and have implemented some of it in the TestCentric GUI. I call them "package tests" and they are really at the system level because they run after packaging. They are separate from the unit tests and are written as part of the Cake script.

Why I like this approach:

1. They are run on the actual packages, so I don't have to simulate the directory structure of the packages. Simulating it creates a double update situation, that is, when the packaging changes  the integration setup has to change as well.

2. By testing packages, the code that does the packaging is also being tested.

3. By running these tests separately, the developer can choose when to run them.

4. By doing it in the Cake script, I can readily change which tests are run without rebuilding. This is partly psychological, I guess. I think I feel more ready to just make temporary changes in a "script" than I am in changing the actual code.

6. I've set up the script so there are three levels of testing. Level 1 runs any time we package. Level 2 runs immediately before and after merging a dev build (i.e. before == on the PR). Level 3 is currently empty but intended to run before actual publication to nuget.org, chocolatey.org or github.

This is a work in progress and I wouldn't mind some help on it. One thing that would be very handy would be a way for the developer to specifically select a single package test to run. Because a full build runs the GUI 27 times, it makes for a bit of a wait otherwise. I currently just edit the script, commenting out the ones I don't want, but that's not perfect.

I know you have put quite a bit of effort into what currently exists in the engine project, so I'd like to have your ideas before creating an issue on this.

CharliePoole:
A comment on what __isn't__ here:

I ported `AgentProcess` from TestCentric, modifying it so it would fit within the current engine code. There were a few things I might have liked to change but avoided in order to keep the PR specifically on topic...

1. Using FrameworkName rather than our own RuntimeFrameowk identifiers to decide on the target runtime. I think a pretty big change in how the target runtime is represented will be needed at some point.

2. Distinguishing between __requested__ ( by the user) runtime and __selected__ (by the engine) runtime. A small change but one that can easily break things so probably best separated if we want to do it. Could be a PR without an issue.

3. Merging `EnginePackageSettings` and `InternalEnginePackageSettings`. The distinction was originally based on what the console runner needed to know about, but it turns out other runners may need to know the "internal" settings even if they don't set them. It's also a pain to remember which settings are where. This is trivial and could be done here or separately if desired. Could also be a separate PR without an issue.

src/NUnitEngine/nunit.engine/Services/AgentProcess.cs
```diff
@@ -0,0 +1,109 @@
...
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+using NUnit.Engine;
+using NUnit.Engine.Internal;
+
+namespace NUnit.Engine.Services
+{
+    public class AgentProcess : Process
```
jnm2, line 16: I strongly prefer composition over inheritance when layering domain-specific functionality on top of existing types.

src/NUnitEngine/nunit.engine.core/Internal/ServerBase.cs
```diff
@@ -52,8 +52,11 @@ protected ServerBase(string uri, int port)
             this.uri = uri;
             this.port = port;
         }
-
-        public string ServerUrl
+        
+        // Currently virtual to allow NSubstitute to mock it. In fact, it will
+        // probably be virtual or abstract at some point, if we implemment
```
jnm2, line 57:
```suggestion
        // probably be virtual or abstract at some point, if we implement
```

jnm2:
> This is a work in progress and I wouldn't mind some help on it. One thing that would be very handy would be a way for the developer to specifically select a single package test to run. Because a full build runs the GUI 27 times, it makes for a bit of a wait otherwise. I currently just edit the script, commenting out the ones I don't want, but that's not perfect.
>
> I know you have put quite a bit of effort into what currently exists in the engine project, so I'd like to have your ideas before creating an issue on this.

I think the solution to enable the developer to have control over selecting a single package test is to move away from using a script and use a standard NUnit test for each single package test, the way 
I set up the acceptance tests in the adapter project. That seems like the most friendly starting point in terms of developer control over what runs when. I'm sure there's room for improvement.

src/NUnitEngine/nunit.engine/Services/AgentProcess.cs
CharliePoole: Me too! But there's a narrow band where either one may work well. In this case, I wanted the `AgentProcess` to actually __be__ a process. You'll notice that I also use composition. That is, the `AgentProcess` figures out where it's agent is located and does a few other things required to make the process work, while TestAgency sets up some other stuff, like monitoring when the process exits. I think the balance here is pretty reasonable, if I do say so myself. :smile:

CharliePoole:
@jnm2 As you know, I haven't been in the adapter code for two or three years, but I'll take a look. Is it a separate project with acceptance tests? One thing that's also important to me is to be able to run the all the unit tests by themselves, without the acceptance tests. That's easiest if they are in a separate assembly. It also works best that way if you're an old-time XP guy: unit tests always have to pass 100% while acceptance tests start out failing until the feature is implemented.

jnm2:
@CharliePoole 

Yes, it's in a separate project: https://github.com/nunit/nunit3-vs-adapter/tree/master/src/NUnit.TestAdapter.Tests.Acceptance

> PS: @jnm2 Since you approved this, I'm assuming you'll take care of the test I ignored when you're ready to get back to it. 😺

I am going to work on getting rid of remoting next.

src/NUnitEngine/nunit.engine/Services/AgentProcess.cs
ChrisMaddock: Niggle - could this class be internal? I know we don't have a good history of doing that for the engine, but I'd like to start. 🙂

ChrisMaddock:
I'd also be a fan of a separate-acceptance-tests project. I even made a start on it a couple of weeks ago, but never finished the branch. There's a whole bunch of ideas I'd like to add to one.

Charlie - we talked about tests-in-cake on our call, and I expressed my reservations then. I've thought more about it since, and I'm afraid I'm still not convinced of the benefits vs an NUnit project. I'm still concerned about how this would scale. I'd also be worried about Cake alienating our casual contributors, vs NUnit tests, which I think we can expect the enormous majority to be familiar with. (I know I personally squirm if I have to contribute an xUnit test to a project!) More than anything however, as part of the NUnit project, I really think we should be dog-fooding and exhibiting our own stuff, as much as possible! 🙂 

Charlie, on your points about, 1 and 2 I'd love to see - if you'd be interested in contributing them. I think we have an issue for point 1 knocking around somewhere. Point 3, from memory, I quite like the division, but I'd need to review the list properly to make an informed decision. Good shout keeping those for separate PRs, however!

---

On the PR, I haven't had a chance to look through the code properly - but by all means get it merged now Joseph's reviewed. 😄 Leave it with you to merge when ready, Charlie.

ChrisMaddock:
> I am going to work on getting rid of remoting next which will make this moot :D

That's great news @jnm2 - it was on my list to catch up with you about that. I'm hoping to get an in-process version of the .NET Core Console in the next release, but running different platforms off both consoles I'd love to see as the next goal. 😄

CharliePoole:
@jnm2 Could you clarify why getting rid of remoting will make integration testing moot?

@ChrisMaddock I expect we'll identify the direction on communication as one of the biggest incompatibilities between testcentric and nunit. As it's an internal matter, however, I see it as blocking only the third level of re-converging we talked about... i.e. getting rid of testcentric engine entirely.

jnm2:
@CharliePoole Not integration testing, this test:

> ```cs
> [Test, Ignore("No longer works with multiple agents")]
> public void Explore_does_not_throw_SocketException()
> ```

I'm expecting it won't survive when the dust is settled. We'll need a whole set of new integration tests written from scratch either way.

ChrisMaddock:
Agreed! 

> @jnm2 Could you clarify why getting rid of remoting will make integration testing moot?

I think Joseph meant that this particular test being ignored will shortly become redundant - not "integration testing" in general! 😄

CharliePoole:
@ChrisMaddock  I think an NUnit project can be equally good, provided it's run on the packages afer they are produced. Run in any other way, it opens the door of doing white-box testing rather than black-box. 

Of course, I didn't know I'd like doing it in a cake script until I did it. :smile: That's why I like trying things out rather than deciding what's best before you try. Using Cake (which is still just C# of course) I am forced to treat the packages as they are, looking only at inputs and outputs. I'll keep pushing on this in testcentric till I figure out what the limits of the technique are. So far, it has helped me do two releases very rapidly.

As far as contributors go, I wasn't seeing this as something contributors would write but tests that their contributions would have to pass... sort of an acceptance test for the contributions beyond the simple functionality. Of course, my assumption is that everyone runs the cake scripts, which may not be your assumption.

CharliePoole:
Sounds like we at least agree they should be separate from the unit tests.

jnm2:
@CharliePoole 

> I think an NUnit project can be equally good, provided it's run on the packages afer they are produced. Run in any other way, it opens the door of doing white-box testing rather than black-box.

The acceptance tests project in the adapter repo doesn't reference any of the other projects, but rather relies on the last output from `.\build.ps1 -t pack` (I think that's the command). This could be automated if we wanted.

CharliePoole:
@jnm2 I looked at it and got reminded how different the GUI is from the adapter. That is, you can take a GUI package and run the GUI... it doesn't have to run under anything else. 

That would apply to the console as well, but if you want to acceptance test an engine package, then that's more like the adapter - you need something to call it.

Goes to show... it's in the details.

PR #762, opened 2020-05-04
Simplify agent communication in preparation for new wire protocol

jnm2:
This cuts down on the amount of work needed to build the new protocol and it enhances the separation between ProcessRunner and all agency/agent concerns.

Review a commit at a time to follow the incremental changes.

- Removes the ITestAgent.Start and ITestAgent.Id properties since they will not be remote procedure calls in the new protocol

- Reinvents RemoteTestAgentProxy as IAgentLease so that TestAgency is in full control both of starting and stopping of agents and is the only thing with access to ITestAgent instances in the console process.

Looks like it nets six lines fewer 😊

### DIFF
--- a/src/NUnitEngine/nunit.engine.core/Agents/RemoteTestAgent.cs
+++ b/src/NUnitEngine/nunit.engine.core/Agents/RemoteTestAgent.cs
@@ -42,6 +42,7 @@ namespace NUnit.Engine.Agents
     {
         private static readonly Logger log = InternalTrace.GetLogger(typeof(RemoteTestAgent));
 
+        private readonly Guid _agentId;
         private readonly string _agencyUrl;
 
         private ITestEngineRunner _runner;
@@ -56,8 +57,9 @@ namespace NUnit.Engine.Agents
         /// Construct a RemoteTestAgent
         /// </summary>
         public RemoteTestAgent(Guid agentId, string agencyUrl, IServiceLocator services)
-            : base(agentId, services)
+            : base(services)
         {
+            _agentId = agentId;
             _agencyUrl = agencyUrl;
         }
 
@@ -93,7 +95,7 @@ namespace NUnit.Engine.Agents
 
             try
             {
-                _agency.Register(this);
+                _agency.Register(_agentId, this);
                 log.Debug("Registered with TestAgency");
             }
             catch (Exception ex)
--- a/src/NUnitEngine/nunit.engine.core/Agents/RemoteTestAgent.cs
+++ b/src/NUnitEngine/nunit.engine.core/Agents/RemoteTestAgent.cs
@@ -73,7 +73,7 @@ namespace NUnit.Engine.Agents
             return this;
         }
 
-        public override bool Start()
+        public bool Start()
         {
             log.Info("Agent starting");
 
--- a/src/NUnitEngine/nunit.engine.core/ITestAgent.cs
+++ b/src/NUnitEngine/nunit.engine.core/ITestAgent.cs
@@ -36,12 +36,6 @@ namespace NUnit.Engine
         Guid Id { get; }
 
         /// <summary>
-        /// Starts the agent, performing any required initialization
-        /// </summary>
-        /// <returns>True if successful, otherwise false</returns>
-        bool Start();
-
-        /// <summary>
         /// Stops the agent, releasing any resources
         /// </summary>
         void Stop();
--- a/src/NUnitEngine/nunit.engine.core/ITestAgent.cs
+++ b/src/NUnitEngine/nunit.engine.core/ITestAgent.cs
@@ -21,8 +21,6 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
-using System;
-
 namespace NUnit.Engine
 {
     /// <summary>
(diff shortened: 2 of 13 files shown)


src/NUnitEngine/nunit.engine/Services/TestAgency.AgentLease.cs
```diff
@@ -23,41 +23,34 @@
...
 
-namespace NUnit.Engine.Agents
+namespace NUnit.Engine.Services
 {
-    /// <summary>
-    /// RemoteTestAgentProxy wraps a RemoteTestAgent so that certain
-    /// of its properties may be accessed without remoting.
-    /// </summary>
-    internal class RemoteTestAgentProxy : ITestAgent
+    public partial class TestAgency
```
jnm2, line 29: Nice, it even followed it.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -201,6 +194,47 @@ public void StartService()
...
+                        var stopError = $"Agent connection was forcibly closed. Exit code was {exitCode?.ToString() ?? "unknown"}. {Environment.NewLine}{ExceptionHelper.BuildMessageAndStackTrace(ex)}";
+                        log.Error(stopError);
+                        throw new Exception(stopError, ex);
+                    }
+                }
+                catch (Exception ex)
+                {
+                    var stopError = "Failed to stop the remote agent." + Environment.NewLine + ExceptionHelper.BuildMessageAndStackTrace(ex);
+                    log.Error(stopError);
+                    throw new Exception(stopError, ex);
```
jnm2, line 234: Throwing bare Exception is a smell, but so is catching bare Exception and we're already doing that here and in the calling method. Figured there was no reason to introduce a new exception type.

src/NUnitEngine/nunit.engine.core/ITestAgent.cs
```diff
@@ -21,26 +21,13 @@
...
-        /// <summary>
-        /// Gets a Guid that uniquely identifies this agent.
-        /// </summary>
-        Guid Id { get; }
-
-        /// <summary>
-        /// Starts the agent, performing any required initialization
-        /// </summary>
-        /// <returns>True if successful, otherwise false</returns>
-        bool Start();
```
jnm2, line 42: Do I understand correctly that all public engine types are changeable in every release because consumers of the engine should only be referencing the engine API assembly?

ChrisMaddock: Yes, that's my opinion. I want to enforce that in v4. 🙂

ChrisMaddock:
Leave it with you as to whether you want to merge now, Joseph, or if you want other people to look over it first. 🙂

src/NUnitEngine/nunit.engine.core/ITestAgent.cs
CharliePoole: In V3, we could at least tell people this.

src/NUnitEngine/nunit.engine.core/Agents/RemoteTestAgent.cs
```diff
@@ -56,8 +57,9 @@ public class RemoteTestAgent : TestAgent, ITestEngineRunner
         /// Construct a RemoteTestAgent
         /// </summary>
         public RemoteTestAgent(Guid agentId, string agencyUrl, IServiceLocator services)
-            : base(agentId, services)
+            : base(services)
```
CharliePoole, line 60: If I understand you, you are saying that only Remote agents need an agent ID, as apposed to all agents. Of course, it makes no difference since we have not yet created any other agent types. Are you willing to give up the option of other agents?

src/NUnitEngine/nunit.engine.core/ITestAgent.cs
ChrisMaddock: A fair point. I went to take a look at what our current docs say. I think the first paragraph of this page covers that the API is the supported way to integrate with the engine? 

https://github.com/nunit/docs/wiki/Test-Engine-API

I'd love to spin out the engine and console docs at some point, and build a little more structure around them. One day...

ChrisMaddock: A fair point. I went to take a look at what our current docs say. I think the first paragraph of this page covers that the API is the supported way to integrate with the engine? 

https://github.com/nunit/docs/wiki/Test-Engine-API

I'd love to spin out the engine and console docs at some point, and build a little more structure around them. One day...

src/NUnitEngine/nunit.engine.core/Agents/RemoteTestAgent.cs
jnm2: That's not quite it. All agents still need an ID and still use it to register with the agency. I moved the ID out of the ITestAgent interface so that ITestAgent members can represent remote procedure calls. I'm not planning for the protocol to have a "Start" message to the agent or a "Get ID" message. But even though its storage location moved, the ID is still being used in all the same ways as before this PR, I think.

jnm2: (And the reason to make ITestAgent look just like the wire protocol is to enable incremental changes to get there, rather than doing the whole thing in one PR and implementing our current ITestAgent interface on top of that.)

src/NUnitEngine/nunit.engine.core/ITestAgent.cs
jnm2: @ChrisMaddock your last comment showed up twice, so I deleted one of the copies.

CharliePoole: I think that doc page says it well. I was only reacting to the "modern" view sometimes expressed that any public method being changed constitutes a breaking change.

PR #765, opened 2020-05-04
Split RuntimeFramework package setting into two: Requested and Target

CharliePoole:
As discussed in PR #759, we are using the setting "RuntimeFramework" to mean two things: the framework the user requested and the framework selected as best by the RuntimeFrameworkService. This works in some places, but in others we have to check whether the framework indicated is valid and is actually available. These checks are only needed for a pending user request, so having two names is more useful.

### DIFF
--- a/src/NUnitConsole/nunit3-console/EnginePackageSettings.cs
+++ b/src/NUnitConsole/nunit3-console/EnginePackageSettings.cs
@@ -101,7 +101,13 @@ namespace NUnit
         /// are strings like "net-4.5", "mono-4.0", etc. Default is to
         /// use the target framework for which an assembly was built.
         /// </summary>
-        public const string RuntimeFramework = "RuntimeFramework";
+        public const string RequestedRuntimeFramework = "RequestedRuntimeFramework";
+
+        /// <summary>
+        /// Indicates the Target runtime selected for use by the engine,
+        /// based on the requested runtime and assembly metadata.
+        /// </summary>
+        public const string TargetRuntimeFramework = "TargetRuntimeFramework";
 
         /// <summary>
         /// Bool flag indicating that the test should be run in a 32-bit process 
--- a/src/NUnitConsole/nunit3-console/EnginePackageSettings.cs
+++ b/src/NUnitConsole/nunit3-console/EnginePackageSettings.cs
@@ -21,6 +21,8 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
+using System;
+
 namespace NUnit
 {
     /// <summary>
@@ -95,6 +97,14 @@ namespace NUnit
         /// for more than one assembly, "Separate" for a single assembly.
         /// </summary>
         public const string ProcessModel = "ProcessModel";
+
+        /// <summary>
+        /// Indicates the desired runtime to use for the tests. Values 
+        /// are strings like "net-4.5", "mono-4.0", etc. Default is to
+        /// use the target framework for which an assembly was built.
+        /// </summary>
+        [Obsolete("Use 'RuntimeFramework.RequestedRuntimeFramework' instead.")]
+        public const string RuntimeFramework = "RequestedRuntimeFramework";
 
         /// <summary>
         /// Indicates the desired runtime to use for the tests. Values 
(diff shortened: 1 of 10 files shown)


ChrisMaddock:
Hi Charlie - will this not break any runners using the existing "RuntimeFramework" package setting?

I'm all for the change in general, but I think we need to keep supporting the string "RuntimeFramework" as well as the new version in the engine, for now. Removing the ambiguous string RuntimeFramework could be a v4 thing - what do you think?

ChrisMaddock:
I can't track down the place where we discussed it, but I'm guessing we didn't quite understand each other here. 

My view is that this is a good change for the internals of the engine - but we need the external interface to remain the same for now - and I guess that package settings are part of that interface, really. We should move them to the API project in future, so that there's not ambiguity around that. 🙂

CharliePoole:
Good point. I see two options...

1. Put Requested run time back to its original form. We would use the new
target one where appropriate.

2. Allow the original form as an alternate, which we immediately replace
with Requested.

ChrisMaddock:
I like version 2. That moves us towards a simpler solution within the engine, and maintains existing support for runners.

CharliePoole:
I'll do that.

CharliePoole:
I did this in the simplest way possible, but I'm OK if we want more.

The last commit adds back the old `RuntimeFramework` setting, marked as obsolete. It's value is "RequestedRunitmeFramework". This will work for anyone who uses our file or a copy of it __which they update__.

However, anyone with an out of date copy of the file or using just a string will fail. i can fix either of those cases but I need to know if people think it's worth doing. My own view is that anyone either using a copy of the file or string constants should know they are out on a limb and will soon figure out what the problem is.

CharliePoole:
@ChrisMaddock Can you let me know whether you think an even more defensive approach is needed here?

ChrisMaddock:
Hi @CharliePoole - sorry this has taken me a while to get back to.

I think the Engine should continue to support the original form, as you suggested above, until v4. In my eyes, as we don't expose a referencable list of known package settings in the Engine API, we encourage consumers to have their own files with strings in instead (like we do ourselves in the console), which means we need to continue to support 'old strings', in my opinion. Would you mind putting that in, so the engine converts the package setting at the highest level to the new version?

I'd like to improve this for v4, but it'll be a little unclear how to do that in a clean way, given some package settings 'belong' to the framework, and others to the engine. I wonder if they engine should expose a list of "known" package settings, which the driver could potentially convert for different frameworks? For the NUnit3 driver, those would essentially be one-to-one, but we may wish to convert these differently for other frameworks. This is a bigger topic though. 🙂

CharliePoole:
Will do!

ChrisMaddock:
@CharliePoole - did you want to get this one into 3.12? I'm hoping to do the beta release on Friday. 🙂

CharliePoole:
Unfortunately, I have a family thing this afternoon, and of course there's the time zone difference. :-(

But I'll take a look to see how close it is.

CharliePoole:
OK, I'm back working on this one. Just rebased on current master.

CharliePoole:
Last commit deals with anyone who is using an old private copy of the source file or a simple string constant. If an obsolete key is present, the replacement key is added with the same value. I don't remove the old one because the user could have code of their own that looks for it. I wish there were a useful mechanism to issue a warning in this case, but we don't really have one.

src/NUnitConsole/nunit3-console/EnginePackageSettings.cs
```diff
@@ -101,7 +103,21 @@ public static class EnginePackageSettings
         /// are strings like "net-4.5", "mono-4.0", etc. Default is to
         /// use the target framework for which an assembly was built.
         /// </summary>
-        public const string RuntimeFramework = "RuntimeFramework";
+        [Obsolete("Use 'RuntimeFramework.RequestedRuntimeFramework' instead.")]
```
ChrisMaddock, line 106:
```suggestion
        [Obsolete("Use 'RequestedRuntimeFramework' instead.")]
```

This isn't qualified, is it?

src/NUnitEngine/nunit.engine.core/EnginePackageSettings.cs
```diff
@@ -106,7 +108,21 @@ public static class EnginePackageSettings
         /// are strings like "net-4.5", "mono-4.0", etc. Default is to
         /// use the target framework for which an assembly was built.
         /// </summary>
-        public const string RuntimeFramework = "RuntimeFramework";
+        [Obsolete("Use 'RuntimeFramework.RequestedRuntimeFramework' instead.")]
```
ChrisMaddock, line 111:
```suggestion
        [Obsolete("Use 'RequestedRuntimeFramework' instead.")]
```

As above

src/NUnitConsole/nunit3-console/EnginePackageSettings.cs
CharliePoole: No idea how I got that in there !

PR #773, opened 2020-05-17
Only allow one runner at a time per connection to the agent by the protocol design

### LINKED ISSUE #763: Agent protocol design: allow concurrent runners for the same agent?
/cc @nunit/engine-team
/cc @CharliePoole because I want to make sure we're not blocking a UI scenario
/cc @oznetmaster because you also had a use case that I'd like to be aware of

The current design of ITestAgent has `ITestEngineRunner CreateRunner(TestPackage package)`. This means that you can create as many runner instances as you like for the same agent instance and use them all at the same time.

We do not make use of this functionality. ProcessRunner calls ITestAgent.CreateRunner immediately upon getting an agent. It proceeds to make all its calls through that single runner instance before calling ITestAgent.Stop().

Implementing this functionality in our new wire protocol seems like a lot of work for something we currently don't use. It would require either 1) a new TCP connection per runner or 2) multiplexing all the commands and test results between multiple runners in the same connection. With multiplexing, we'd have to manage a header on every runner message in our protocol specifying which runner session it's talking about. We'd need a system to be sure one runner didn't hog the whole TCP connection.

Allowing the same agent to run more than one runner does sound like it has potential to me, but mainly from the perspective of inverting the whole thing so that the agent is the listener/server and the console is the client. One runner per client connection is much more natural; clients can open up multiple connections if they want multiple runners. However, we weren't interested in inverting the server/client relationship between the console and the agent. See https://github.com/nunit/nunit-console/pull/576 for that implementation and discussion.

If we decide not to support concurrent runners in the same timeframe as moving off of remoting, I will do a simplification PR while we're still using PR before moving forward. It will make reviews significantly easier to follow if we make the remoting interfaces match the design of the replacement protocol before actually replacing it.

It would look something very roughly like the following. Before:

```cs
public interface ITestAgent
{
    void Stop();
    ITestEngineRunner CreateRunner(TestPackage package);
}
```

After:

```cs
public interface ITestAgent
{
    void ShutDown(); // Modified, was: void Stop();
    
    // Copied from from ITestEngineRunner:
    
    TestEngineResult Load(TestPackage package); // Modified, was: TestEngineResult Load();
    void Unload();
    TestEngineResult Reload();
    int CountTestCases(TestFilter filter);
    TestEngineResult Run(ITestEventListener listener, TestFilter filter);
    TestEngineResult Explore(TestFilter filter);
    // *Eventually* would likely disappear or change form:
    AsyncTestEngineResult RunAsync(ITestEventListener listener, TestFilter filter);
    void StopRun(bool force);
}
```

jnm2:
Closes #763
Contributes to #266 

Review a commit at a time to follow the incremental changes.


src/NUnitEngine/nunit.engine.core/Agents/RemoteTestAgent.cs
```diff
@@ -107,7 +121,7 @@ public bool Start()
             return true;
         }
 
-        public override void Stop()
+        public void ShutDown()
```
CharliePoole, line 124: Agents used `Stop` to indicate a distinction with `ShutDown`, which has a special meaning for Services. Services youse `StartUp` and `ShutDown`, whereas communications elements have used `Start` and `Stop`. Where a comm element is also a service, `Start` and `Stop` are typically called as part of `StartUp` and `ShutDown`. Having these as separate methods makes some unit-testing easier.

src/NUnitEngine/nunit.engine.core/ITestAgent.cs
```diff
@@ -24,18 +24,53 @@
...
+    /// Defines the current communication protocol between the engine and the agent. This is an implementation detail
+    /// which is in the process of changing as the dependency on .NET Framework's remoting is removed.
     /// </summary>
     public interface ITestAgent
     {
         /// <summary>
         /// Stops the agent, releasing any resources
         /// </summary>
-        void Stop();
+        void ShutDown();
```
CharliePoole, line 35: Others may not be confused, but I guess I find it confusing to see an interface that implements `ShutDown` even though it's not an `IService` and which implements it without also implementing `StartUp`.

src/NUnitEngine/nunit.engine.core/ITestAgent.cs
```diff
@@ -24,18 +24,53 @@
...
 
         /// <summary>
-        ///  Creates a test runner
+        /// Loads a package which will be used for all subsequent calls to the other methods (except
+        /// <see cref="ShutDown"/>). This method must be called at least once before calling methods that operate on a
+        /// test package. It may be called any number of times after those methods.
         /// </summary>
-        ITestEngineRunner CreateRunner(TestPackage package);
+        TestEngineResult Load(TestPackage package);
```
CharliePoole, line 42: Reading this, I feel the need to read all the code so as to see if all the methods of `ITestEngineResult` are present. They are, so why not just inherit from that interface?

src/NUnitEngine/nunit.engine.core/Polyfill.cs
```diff
@@ -0,0 +1,11 @@
+﻿#if NET20
```
CharliePoole, line 1: I'm not much liking the name of this class. :smile: Why not "Func.cs"?

src/NUnitEngine/nunit.engine.core/Agents/RemoteTestAgent.cs
ChrisMaddock: I will say, I've never been aware of that design intent! 😅

I have no strong opinion on the final naming. Joseph - re: Charlie's comment about unit testing, do you think that will cause an issue, based on your final goal here?

src/NUnitEngine/nunit.engine.core/ITestAgent.cs
```diff
@@ -34,8 +35,51 @@ public interface ITestAgent
...
+
+        /// <summary>
+        /// Unloads any loaded package. If none is loaded, the call is ignored.
+        /// </summary>
+        void Unload();
```
ChrisMaddock, line 47: I expect your working on a minimal-change-basis here, but before we move on from this project entirely, I think we should consider if ignoring an empty Unload call rather than throwing is the behaviour we want here. It feels inconsistent with Reload's behaviour below...

src/NUnitEngine/nunit.engine.core/Agents/RemoteTestAgent.cs
CharliePoole: @ChrisMaddock Well, to put it differently, the intent is seen in the definition of `IService`, which has `StartUp` and `Shutdown` methods. There is no rule against using the same method names outside of the interface, but it's potentially confusing... or maybe only confusing to me. :smile:

src/NUnitEngine/nunit.engine.core/ITestAgent.cs
CharliePoole: @ChrisMaddock It could have been different, but it's public behavior. I don't think we should change what has been a safe call into one that throws.

ChrisMaddock:
Hi Joseph,

As #774, I think this has been superceded by #937 too. Thanks again though!

PR #774, opened 2020-05-18
Lower communication to byte arrays while keeping remoting

jnm2:
The next smallest step towards getting rid of remoting seemed to be to add a new method to replace the existing methods on ITestAgent: `byte[] SendMessage(byte[])`.

➡ Please let me know what you like and don't like about it. Ideas always welcome too.

🚧 Based on https://github.com/nunit/nunit-console/pull/773. Look at the commits after the empty `🚧 base 🚧` commit.

### Goal

Evolving the communication model in the future should be straightforward for someone who doesn't know or remember how all this works.

Right now I'm not entirely happy with where it ended up.
❌ There's duplication between the implementations of request and response model structs.
❌ It's not yet possible to write tests of AgentLease and RemoteTestAgent which drive all the new code.
❌ Handling of errors is too awkward (and not comprehensive enough once remoting is gone).
❌ BinaryWriterExtensions and ProtocolReader are cross-cutting. If they aren't filled ahead of time with unused methods, I'm afraid they might create friction when those methods are needed. Maybe what is actually wrong is that the new concept count is too high.

### Missing from this draft

- Run and StopRun haven't been converted. They will definitely be the most complex.
- Exhaustive tests on every new class or method.


ChrisMaddock:
Had a quick skim through this - the direction looks good, but I think I need to get in pulled down in an IDE to properly follow through. Will try and do that this weekend. 🙂 Exciting to see the progress though!

ChrisMaddock:
Hi Joseph,

I'm going to close this one now, as it's been superseded by #937. Thanks for all the work you did on this, regardless. 😄

PR #776, opened 2020-05-19
Fix paths in addins files for global nuget package location

### LINKED ISSUE #775: Extension loading broken on Linux when installed from nuget package
NUnit.ConsoleRunner comes with a `nunit.console.nuget.addins` file that tries to find extensions that are installed in directories parallel to the on nunit3-console is in. However, this fails on Linux with the global nuget package location (`$HOME/.nuget/packages`), because nuget converts the path names to lowercase whereas NUnit looks for the exact camelcase directoryname.

These are some of the file I have that show the problem:

```
$HOME/.nuget/packages/nunit.consolerunner/3.11.1/tools/nunit3-console.exe
$HOME/.nuget/packages/nunit.consolerunner/3.11.1/tools/nunit.console.nuget.addins
$HOME/.nuget/packages/nunit.extension.nunitv2resultwriter/3.6.0/tools/nunit-v2-result-writer.dll
```

ermshiperete:
Paths on Linux are case-sensitive and nuget converts the paths
to lowercase when installing them in the global package location.
However, when installing the packages in a local directory the
paths don't get changed.

This change adds the lowercase paths in addition to the camel case
ones so that we find extensions that are installed parallel to
the NUnit.ConsoleRunner nuget package regardless of whether it is
installed in a local directory or in the global package location.

This fixes #775.

---
This change is [Reviewable](https://reviewable.io/reviews/nunit/nunit-console/776)

### DIFF
--- a/nuget/engine/nunit.engine.nuget.addins
+++ b/nuget/engine/nunit.engine.nuget.addins
@@ -1,2 +1,4 @@
 ../../../NUnit.Extension.*/**/tools/     # nuget v2 layout
 ../../../../NUnit.Extension.*/**/tools/  # nuget v3 layout
+../../../nunit.extension.*/**/tools/     # nuget v2 layout
+../../../../nunit.extension.*/**/tools/  # nuget v3 layout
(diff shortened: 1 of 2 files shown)


CharliePoole:
@ChrisMaddock This may be something to look at for V4.0. Perhaps the `ExtensionService` should have some special knowledge of how nuget stores packages and supplement the paths in the `.addins` file.

Since Chocolatey standards already call for lower-case, dash-sparated package ids, I don't think we need to do anything with those packages.

PR #777, opened 2020-05-31
Remove unused code to find engine from registry

ChrisMaddock:
We removed the functionality to look up an engine in the registry a few years back - this method was left over.


ChrisMaddock:
Definitely. Been battling the reflection in TestEngineActivator today, and I think we need to consider the cost vs benefit there for v4, now we've moved away from the original intention of a central engine being loaded by different runners.

I'll have more of a think about that...

Issue #725, opened 2020-01-04
Remove .NET Standard 1.6 build


ChrisMaddock:
.NET Standard 1.6 is no longer supported by the framework or Micrsoft - and is has various limitations in the engine. I propose removing it in the next release of the engine.

mano-si:
@ChrisMaddock Starting on this. You can assign it to me. 😊

CharliePoole:
@ChrisMaddock This means that people using an older version of the framework, with .NET standard 1.6 support, will no longer be able to run there tests using that engine. Of course, they will run under the 2.0 build, but isn't that breaking to some degree? And what happens if one of those folks has a test machine with only .NET Core 1.1 installed, for some reason?

Personally, I don't feel strongly about this either way, but I've been very confused about the degree of "breaking" required before we say we won't do something. In the past, I was quite willing to "break" minor things but there has been a lot of team pushback against that idea for the past few years. I'm using this as a sort of test case to try to understand what level of backward compatibility people really want to support.

ChrisMaddock:
Great - thanks @mano-si!

---

@CharliePoole I'd agree this is a breaking change. This isn't a definition you'll like I'm afraid, but my personal view is that each breaking change is a trade off, where we weigh up the benefits vs the potential cost to users. Here's some of my thoughts on why we should go for this one.

1. The benefit to the project is high. There's a currently high cost in adapting code to support this legacy platform.
2. The potential impact to users is low. Two reasons:
 a. .NET Core 1.x never went into full LTS. My experience of people's use is that 1.x was very much the experimental warm-up to 2.x.
b. .NET Core appears to have a different upgrade culture around it, with no need to rely on a user's own installation. If you control the infrastructure, you can upgrade as required. If you're running on a user's machine, you can deploy a self-contained application, and still upgrade as required. That's very different from running on .NET Framework of course - and reflected in Microsoft's support and release cadence.
3. The NUnit framework has set a precedent here, but dropping .NET Standard 1.x support already.
4. The .NET Standard 1.x build has been around for a relatively short time, and then engine has a much lower direct user base than console. Vastly the main consumer of the .NET Standard engine (in user numbers) is the adapter, which doesn't the build.
5. It's a clear-cut break for a end-user - no subtle change in behaviour. A user with a .NET Standard 1.6 project will no longer be able to upgrade the NuGet package. 

All that said, I opened this issue to invite discussion, so thank you for that!

[...]

PR #778, opened 2020-06-01
Add .NET Core 3.1 build of engine to access APIs for loading .NET Core assemblies correctly

### LINKED ISSUE #710: .NET Core engine only works when located in same directory as test assembly

ChrisMaddock:
To sort #710 properly, we need to make use of the AssemblyLoadContext APIs, which are only available in .NET Core 2.1 and above.

In our 'full vision', this functionality would live in a .NET Core specific agent, and not require a full build of the engine. However, I don't want to delay the .NET Core Console further while we work on .NET Core agents - and I consider #710 a major enough bug that I don't want to release the .NET Core Console without a solution.

Accordingly, this PR adds a .NET Core 3.1 build of the engine. We use v3.1 specifically to make use of the `AssemblyDependencyResolver` class, which was first available in .NET Core 3. (Of which, 3.1 was the first LTS release). There are workarounds for earlier frameworks - however my priority now is getting the latest version of .NET Core working. If we end up with separate agents for older versions of .NET Core, it would be possible for someone to look at adding the workarounds into those - if there's sufficient demand.

I expect this build to be a temporary build, until we have .NET Core agents, at which point we can remove it and go back to just packaging .NET Standard 2.0. I can't think of any way that could be a breaking change, on the basis that any .NET Core 3.1 runners can consume the .NET standard package instead?

---

For ease of review, this PR just includes adding the .NET Core 3.1 build. The actual fix will be in a second PR.


src/NUnitEngine/nunit.engine.tests/Drivers/NUnit3FrameworkDriverTests.cs
```diff
@@ -51,21 +51,6 @@ public void CreateDriver()
...
-        //    Assert.That(_controller.AssemblyPath, Is.EqualTo(MOCK_ASSEMBLY));
-        //    Assert.That(_controller.Settings, Is.SameAs(_settings));
-        //}
-
-        public void ConstructController_MissingFile_ThrowsArgumentInvalid()
-        {
-            var assemblyName = typeof(NUnit.Framework.TestAttribute).Assembly.GetName();
-            Assert.That(new NUnit3FrameworkDriver(AppDomain.CurrentDomain, assemblyName), Throws.ArgumentException);
-        }
-
```
ChrisMaddock, line 68: This is unrelated, but this test was missing its TestAttribute so not running. The test is no longer valid.

src/NUnitEngine/nunit.engine.tests/Services/ExtensionServiceTests.cs
```diff
@@ -256,8 +258,8 @@ public static IEnumerable<TestCaseData> InvalidTargetFrameworkCombos()
 
         public static IEnumerable<TestCaseData> InvalidRunnerCombos()
         {
-#if NETCOREAPP2_1
-            Assembly netstandard = typeof(ExtensionService).Assembly;
+#if NETCOREAPP
+            Assembly netstandard = Assembly.LoadFile(Path.Combine(GetSiblingDirectory("netstandard2.0"), "nunit.engine.dll"));
```
ChrisMaddock, line 262: These tests were using the executing engine as an example .NET Standard assembly - which of course caused problems when that assembly is build for .NET Core. I changed them to fetch the .NET Standard build specifically, in the same way the other tests in this class do.

Ideally, I think we need a set of test assemblies for this. (Or perhaps to use mock-assembly?)

src/NUnitConsole/nunit3-console/nunit3-console.csproj
```diff
@@ -3,7 +3,7 @@
     <OutputType>Exe</OutputType>
     <RootNamespace>NUnit.ConsoleRunner</RootNamespace>
     <AssemblyName>nunit3-console</AssemblyName>
-    <TargetFrameworks>net20;netcoreapp2.1</TargetFrameworks>
```
ChrisMaddock, line 6: The console is as-yet unreleased, but will now target .NET Core 3.1 as a minimum. With the long-term aim of making it self-contained anyway, I'm not too worried about this.

ChrisMaddock:
Just added the blocked label, as I'll need to rebase this over #779 once that's merged, and make the same change for the new .NET Core 3.1 build. That's only a minor change however, the rest is still ready for review. 🙂

ChrisMaddock:
Reverting to draft until I can figure out what's causing the test failures seen in #781

ChrisMaddock:
Ok - I'm happy this is the right approach now and everything's working - I'm just going to tidy up the commits, and it'll be ready for review. 🙂

ChrisMaddock:
Ok - this is now ready for review. I've tried to break things up so it's easy to look at commit-by-commit.

As a reminder, this PR only includes adding the .NET Core 3.1 build itself - the fix will come in a later PR.

build.cake
```diff
@@ -202,28 +201,28 @@ Task("Build")
...
                 .WithProperty("PublishDir", BIN_DIR + framework));
 
-        foreach(var framework in NETSTANDARD_FRAMEWORKS)
+        foreach(var framework in new [] { "netstandard1.6", "netstandard2.0" })
              MSBuild(ENGINE_API_CSPROJ, CreateMSBuildSettings("Publish")
                 .WithProperty("TargetFramework", framework)
                 .WithProperty("PublishDir", BIN_DIR + framework));
 
-        foreach(var framework in NETCORE_FRAMEWORKS)
+        foreach(var framework in new [] { "netcoreapp1.1", "netcoreapp2.1", "netcoreapp3.1" })
```
jnm2, line 214: Could we give this list of frameworks a descriptive name since it repeats?

build.cake
```diff
@@ -667,7 +692,7 @@ FilePath GetResultXmlPath(string testAssembly, string framework)
 {
     var assemblyName = System.IO.Path.GetFileNameWithoutExtension(testAssembly);
 
-    // Workaround for https://github.com/nunit/nunit-console/issues/501
+    // Required for Test suites running under NUnitLite
```
jnm2, line 695:
```suggestion
    // Required for test suites running under NUnitLite
```

ChrisMaddock:
Thanks Joseph. 😄 

> Did I miss the place where the .NET Core APIs begin to be used to load assemblies, or is that for a later PR?

Yeah, that bit is in https://github.com/nunit/nunit-console/pull/781. 🙂 It's depends on three other open PRs, so I'll rebase and tidy up once all those are merged. 

Looks like there's a little more battle required yet, to get .NET Core 1.1 running on Azure. Might have to come back to that later, will sort your feedback at the same time. Thanks!

build.cake
ChrisMaddock: So the lists of targets for each build are now different. The engine publishes for netstandard1.6, netstandard2.0 and netcoreapp3.1. The API dll for the two netstandard targets, and the tests for three net core targets. 

That was my original reason for getting rid of the variables - and I'm not too sure how we could name the three different lists helpfully - being explicit and inlining the lists was the best I could come up with. Any suggestions?

ChrisMaddock:
So the old Azure image won't build .NET Core 3.1, and the new one won't build .NET Core 1.1. 😤

Guess this is a task for another day...

jnm2:
Just a suggestion and I might be missing some context, but how sure are we that .NET Core 1.1 is worth building for? It's past end of life.

build.cake
jnm2: Oh, I misread. This seems fine!

ChrisMaddock:
Yeah, I'm inclined to agree, see https://github.com/nunit/nunit-console/issues/725. It's too much of a timesink for the benefit, in my opinion. I just feel like that should be done separately and intentionally, rather than taken out because something else broke it - but it does pull in to question how much of my time it's worth investing in trying to get it back up and running again...

ChrisMaddock:
It's a little frustrating, as it would be a breaking change for runner-maintainers...however I expect the number of actual end users is critically small.

ChrisMaddock:
Finally got CI working again. 

Thanks Joseph for reviewing this. 🙂  As it's a significant change, I'll just leave the PR up a few more days before merging, to give anyone else on the team the chance to raise any concerns.

PR #779, opened 2020-06-01
[CI] Revert to running .NET Standard Engine Tests via NUnitLite

ChrisMaddock:
Unfortunately #710 seems to have also affected our own tests. Since https://github.com/nunit/nunit-console/pull/696, I don't think we've actually been running the .NET Standard 1.6 tests, as the console runner has instead been picking up the assembly of the same name in it's own directory - which is the .NET Standard 2.0 test suite. I only noticed as I was getting some weird results working on #710 - and then spotted the two test runs were reporting the same number of tests.

This reverts to running the .NET Standard tests under NUnitLite, as we previously did. (With some added CI changes to install a .NET Core 1.1 runtime, which have been lost along the way.)

It raises an interesting question about the resilience of our own tests. Perhaps we should instead start using the latest NuGet release of the NUnit Console to run the tests...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ matrix:
         - sudo dpkg -i packages-microsoft-prod.deb
         - sudo apt-get install apt-transport-https
         - sudo apt-get update
+        - sudo apt-get install dotnet-sharedframework-microsoft.netcore.app-1.1.2
         - sudo apt-get install dotnet-sdk-2.2
 
 # macOS package restore currently taking >30 mins on Travis
--- a/NUnitConsole.sln
+++ b/NUnitConsole.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27004.2002
+# Visual Studio Version 16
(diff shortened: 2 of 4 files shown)


ChrisMaddock:
Ok - there's still something wrong here. 

It's now reporting the correct number of tests - but if I run the .NET Standard 1.6 tests manually I get six failures, which isn't happening when it's run via Cake. (The failures look to be correct.) I'll need to dig into this some more...

ChrisMaddock:
Ok, this is quite unpleasent.

`dotnet tests.dll` appears to default to run with the latest installed version of dotnet. To target a specific framework, we would need to use the `--fx-version` flag, which means the cake script needs to not target a specific version of the .NET Core runtime - which means contributors now need to have a specific runtime version installed to run the Cake script...

That said, if I run `dotnet tests.dll` myself, I see the test failures as I would expect. I don't yet understand how my command is different to Cake's, and I'm out of time for now. Will have to come back to this another time.

ChrisMaddock:
Ok - I believe this is now the correct behaviour for CI.

The failures I mentioned previously are incorrect, which I take to mean there is a further issue with the .NET Core console. I'll need to come back to that one another day - but this particular PR is now ready for review.

One small step at a time. 🙂

PR #781, opened 2020-06-01
Use AssemblyLoadContext to correctly load .NET Core assemblies 

### LINKED ISSUE #710: .NET Core engine only works when located in same directory as test assembly

ChrisMaddock:
I believe this fixes #710. It uses .NET Core's AssemblyLoadContext API's to ensure .NET Core dependencies are loaded in the intended way. (It also means that we no longer need to publish as test assembly, to be able to run tests on it.)

~~This is based on top of #778, so only the last commit is actually new here: https://github.com/nunit/nunit-console/commit/45769e2630d4ab2aae196f250a6d94e47e1b9081~~

~~Note this will currently fail CI due to #779. (I believe this is actually running the tests correctly now, and exposing test errors which need to be fixed!)~~

~~@beloquintana @CharliePoole @schrufygroovy - once we've got CI fixed up, would you be able to test this package with your respective runners, and check it solves the problems you were seeing? I'll shout when it's ready. 🙂~~

### DIFF
--- a/src/NUnitEngine/nunit.engine.core/Drivers/NUnit3DriverFactory.cs
+++ b/src/NUnitEngine/nunit.engine.core/Drivers/NUnit3DriverFactory.cs
@@ -61,14 +61,16 @@ namespace NUnit.Engine.Drivers
         /// Gets a driver for a given test assembly and a framework
         /// which the assembly is already known to reference.
         /// </summary>
-        /// <param name="domain">The domain in which the assembly will be loaded</param>
         /// <param name="reference">An AssemblyName referring to the test framework.</param>
         /// <returns></returns>
         public IFrameworkDriver GetDriver(AssemblyName reference)
         {
             Guard.ArgumentValid(IsSupportedTestFramework(reference), "Invalid framework", "reference");
-
+#if NETSTANDARD
             return new NUnitNetStandardDriver();
+#elif NETCOREAPP3_1
+            return new NUnitNetCore31Driver();
+#endif
         }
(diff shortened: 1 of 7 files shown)


ChrisMaddock:
Ok - I've made more progress with this, but will have to come back to it another day. There are two outstanding causes of Test Failures:

1. XmlTransformResultWriterTests - throws a null reference exception, I expect when the test tries to initialise another engine inside the engine. Unsurprisingly, this isn't actually necessary to test the XML transform functionality - I'll look at refactoring that.

2. MasterTestRunnerTests - duplicate test ID's with a project and two mock assemblys. This one's a little more concerning - although I'm really not sure why it's only that specific case which is failing. Need to look more closely into that one.

ChrisMaddock:
Diagnosis: both the above problems are unrelated issues - I believe this is now the correct solution. I'll come back and rebase this functionality once the three other dependent PRs have been merged. 😄

ChrisMaddock:
This has now been rebased, and is ready for review. 

@beloquintana @CharliePoole @schrufygroovy You should be able to download a NuGet package containing this build from the link below - I'd appreciate if you could test it, and confirm it solves your issues. I also plan to release this as a beta release on NuGet within the next few days.
https://ci.appveyor.com/project/CharliePoole/nunit-console/builds/33647940/artifacts

PR #783, opened 2020-06-07
Refactor XMLTransformResultWriterTests to avoid initialising entire engine

ChrisMaddock:
Another step towards #710 - separated out of the main PRs.

XMLTransformResultWriterTests have previously been initialising an entire extra engine to test the XML transform functionality. Initialising an engine inside a test run has caused us problems before, and I don't think will work inside the new .NET Core assembly loading functionality. (Although that's not a restriction I'm particularly concerned about.)

Initialising an entire engine isn't necessary for what these tests are looking to do. This PR restructures the tests to be more light-weight. It's separated out into a separate PR, as it ended up requiring a few more changes to make the MasterTestRunner more testable.


src/NUnitEngine/nunit.engine.api/IExtensionService.cs
```diff
@@ -58,6 +58,11 @@ public interface IExtensionService
         /// <param name="typeName"></param>
         /// <param name="enabled"></param>
         void EnableExtension(string typeName, bool enabled);
+
+        /// <summary>
+        /// Get an enumeration of all registered extensions of a specific type.
+        /// </summary>
+        IEnumerable<T> GetExtensions<T>();
```
ChrisMaddock, line 65: I'm adding a new method to the public IExtensionService API here, primarily or the purpose of making MasterTestRunner more testable - however this seems a sensible method to expose in general. If anyone can think of a way this could bite us in future, I'd appreciate the feedback!

ChrisMaddock:
Thanks Charlie. 👍

PR #784, opened 2020-06-07
Fix DirectTestRunner to not give all drivers the same ID.

ChrisMaddock:
This (I believe) is another separate bug which was exposed with the .NET Core work.

Each TestPackage (including sub-package) has a unique ID, and that unique ID should be passed on to each driver. There's currently a bug in the DirectTestRunner under which gives driver the ID from the Top-level test package, meaning all drivers end up with the same ID.

This was picked up by the MasterTestRunnerTests with the conversion to .NET Core. It was not visible previously as:
1) In the .NET 2.0 build, tests go through the same MultipleTestDomainRunner, which does not have the same bug.
2) In the .NET Standard builds, multiple test assemblies are currently loaded in the same AssemblyLoadContext, meaning that test ID's across multiple assemblies continued going up sequentially, and were not reset for each assembly.

In the second case, the MasterTestRunnerTests should be stronger, to assert that the ID prefix is different for each test assembly, rather than simply that all ID's are unique. That's the change I've made for the 'failing test' in this commit.

### DIFF
--- a/src/NUnitEngine/nunit.engine.core/Runners/DirectTestRunner.cs
+++ b/src/NUnitEngine/nunit.engine.core/Runners/DirectTestRunner.cs
@@ -148,7 +148,7 @@ namespace NUnit.Engine.Runners
 #else
                 IFrameworkDriver driver = driverService.GetDriver(testFile, skipNonTestAssemblies);
 #endif
-                driver.ID = TestPackage.ID;
+                driver.ID = subPackage.ID;
                 result.Add(LoadDriver(driver, testFile, subPackage));
                 _drivers.Add(driver);
             }
--- a/src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Runners/MasterTestRunnerTests.cs
@@ -265,18 +265,31 @@ namespace NUnit.Engine.Runners.Tests
 
         private void CheckThatIdsAreUnique(XmlNode test)
         {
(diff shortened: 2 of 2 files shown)


ChrisMaddock:
@CharliePoole - would you mind reviewing this one? I ask as I want to be sure I'm understanding how these id-prefix's are intended to be set correctly - and I expect you'll remember the intial design intention!

Much appreciated. 🙂

CharliePoole:
The idea of the prefixes is that they are known in the process and load context where the TestPackages are created. Packages are not created anywhere else, although they may be modified.

Consequently, a simple increasing number can be used to assign each package a unique id, which never changes during the lifetime of the package. Since packages map one-to-one to assemblies, that guarantees a unique prefix for each assembly for the life of the package.

If any of the underlying assumptions should change, then of course it will no  longer work correctly.

src/NUnitEngine/nunit.engine.core/Runners/DirectTestRunner.cs
```diff
@@ -148,7 +148,7 @@ protected override TestEngineResult LoadPackage()
 #else
                 IFrameworkDriver driver = driverService.GetDriver(testFile, skipNonTestAssemblies);
 #endif
-                driver.ID = TestPackage.ID;
+                driver.ID = subPackage.ID;
```
CharliePoole, line 151: Right! The two will end up being the same IFF there is only one assembly in the package, which is why the error slipped through. IIRC I had to change it in the GUI engine as well.

ChrisMaddock:
Great - thanks for the extra explanation, Charlie. 😄

PR #790, opened 2020-06-21
Fix agent debug logging

ChrisMaddock:
The argument to set the debug log level was not being read by the agent.

Now standardised as an equals sign, same as all other arguments.

### DIFF
--- a/src/NUnitEngine/nunit-agent/Program.cs
+++ b/src/NUnitEngine/nunit-agent/Program.cs
@@ -67,7 +67,7 @@ namespace NUnit.Agent
                 {
                     debugArgPassed = true;
                 }
-                else if (arg.StartsWith("--trace:"))
+                else if (arg.StartsWith("--trace="))
                 {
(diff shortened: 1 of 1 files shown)


src/NUnitEngine/nunit-agent/Program.cs
```diff
@@ -67,7 +67,7 @@ public static void Main(string[] args)
                 {
                     debugArgPassed = true;
                 }
-                else if (arg.StartsWith("--trace:"))
+                else if (arg.StartsWith("--trace="))
```
CharliePoole, line 70: I wish this were not so fragile. Should we add a test that guarantees all args created for the agent process use this form?

ChrisMaddock: This one broke due to the method to write the arguments being standardised. I feel like now there's a standard method rather than each one being added individually, it will hopefully be more robust?

We might want to build on that when we start making agents-in-extensions, of course.

PR #797, opened 2020-07-18
Add NUnit.Console-Runner.NetCore packages

### LINKED ISSUE #750: .NET Core Console Packaging
One of the last remaining tasks before releasing the .NET Core Console is deciding how to package it. At this point, I think we should be looking to do this in a backwards-compatible way, however in my mind this is one of the major motivations to move forward to an NUnit Console v4, so we can redo the packaging to better cope with this.

Below is some ideas about how to handle each of the packages we currently do - would be great to here ideas on this, as I'm keen we get it right first time. 🙂 Would also love some help in writing the code for this one too - I think we just need some discussion time first, to make sure we're doing the right thing.

## NuGet
NuGet is tricky to do in a backwards-compatible way, as we currently just put the console under `/tools/`. That doesn't work too well when we're introducing a second console. The best I can think of is putting the netcore console under a subdirectory (`tools/netcore`), and leave the .NET 2.0 version in root, for backwards compatibility.

Anyone got any nicer ideas, here?

## Zip package
For this, I suggest we introduce a separate `NUnit.Console.NetCore-3.11.1.zip` file, which contains a published console with all dependencies. I suggest the current `NUnit.Console-3.11.1.zip` remains unnamed for backwards-compat, I know we had issues changing that package name before, with CI scripts, and update-checking scrapers.

This one would also be revised for a v4.0 I think.

## msi installer / Chocolatey
I suggest we leave both of these for now, possibly pending the below

## Looking Forwards
Looking forwards, I'd be super-interested in looking to deploying the Console as a [self-contained executable](https://docs.microsoft.com/en-us/dotnet/core/deploying/), with runtime-dependent-executables as the agents.

This gives us a simple way to remove the .NET Framework console and have a single build, with no cost to end users. It also removes any issues about user's being required to install dependencies on build/production machines specifically for the purposes of running tests. (Although I think it might be best for the agents to remain runtime-dependent - to ensure users test assemblies run on the same platform as their production code would, as far as possible.)

What we'd lose, is inprocess running on netfx - so I think this is a v4 sort of solution. This really needs more thought - which is for a separate issue. Just throwing the idea out there!

---

Thoughts from anyone on the above very welcome. 😄 also cc @mano-si, this was one of the issues we talked about a couple of weeks ago. 🙂

ChrisMaddock:
Fixes #750 and #475 

This is the final part of the .NET Core Console Runner MVP.  🎉 Once this is merged, I aim to get a beta release out shortly, so people can trial the new console.

There is a new NuGet package added under the name nunit.console-runner.netcore. (word separators to try and keep convention with existing packages)

For the zip, I realised the .NET Standard engine was already included in the main zip, and didn't want to break backwards compatibility by separating that out. So I kept everything in a single zip file. The zip was 11MB before this change and is now 13MB - so it's no big change, but we probably want to look at breaking that up for v4.

Both packages have been tested manually.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -440,6 +440,14 @@ Task("PackageNuGet")
             NoPackageAnalysis = true
         });
 
+        NuGetPack("nuget/runners/nunit.console-runner.netcore.nuspec", new NuGetPackSettings()
+        {
+            Version = productVersion,
+            BasePath = CURRENT_IMG_DIR,
+            OutputDirectory = PACKAGE_DIR,
+            NoPackageAnalysis = true
+        });
+
         NuGetPack("nuget/deprecated/nunit.runners.nuspec", new NuGetPackSettings()
         {
             Version = productVersion,
--- a/nuget/runners/nunit.console-runner.netcore.nuspec
+++ b/nuget/runners/nunit.console-runner.netcore.nuspec
@@ -1,0 +1,53 @@
+<?xml version="1.0"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+  <metadata>
+    <id>NUnit.ConsoleRunner.NetCore</id>
+    <title>NUnit Console Runner (.NET Core)</title>
+    <version>$version$</version>
+    <authors>Charlie Poole, Rob Prouse</authors>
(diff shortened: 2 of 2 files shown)


ChrisMaddock:
Thanks Mikkel! 🙂

Issue #662, opened 2019-07-03
Mono: Stacktrace missing files and line numbers


ArnaudValensi:
Hi, 

For my CI and my TDD workflow, I need to display the stacktrace with the line number on the console, but I cannot figure out how.

Here what I have with nunit3-console:
```
$ mono --debug --optimize=-inline nunit3-console.exe ../../../Ecs.Tests/bin/Debug/EcsTests.dll
NUnit Console Runner 3.10.0 (.NET 2.0)
Copyright (c) 2019 Charlie Poole, Rob Prouse
mercredi 3 juillet 2019 04:06:23

Runtime Environment
   OS Version: MacOSX 18.6.0.0
  CLR Version: 4.0.30319.42000

Test Files
    ../../../Ecs.Tests/bin/Debug/EcsTests.dll


Errors, Failures and Warnings

1) Failed : EcsTests.BagTest.Capacity
  Expected: 25
  But was:  24
  at EcsTests.BagTest.Cap () [0x00000] in <7cec980c48ea4a1880f73954ca19d517>:0
  at EcsTests.BagTest.Capacity () [0x00000] in <7cec980c48ea4a1880f73954ca19d517>:0

...
```

But it works in Visual Studo for Mac 2019 (which is Xamarin rebranded and use mono under the hood for .net framework projects):

<img width="788" alt="image" src="https://user-images.githubusercontent.com/604486/60558373-49552100-9d49-11e9-9bf5-76e61f7a9134.png">

I have the same problem with xunit, https://github.com/xunit/xunit/issues/1978

I am on a mac.

My config:
```
$ mono --version
Mono JIT compiler version 5.20.1.19 (2018-10/886c4901747 Tue Apr  9 12:37:29 EDT 2019)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:
	SIGSEGV:       altstack
	Notification:  kqueue
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug
	Interpreter:   yes
	LLVM:          yes(600)
	Suspend:       hybrid
	GC:            sgen (concurrent by default)
```

My csproj file:
```
$ cat Ecs.Tests.csproj
<?xml version="1.0" encoding="utf-8"?>
<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
[...]

PR #799, opened 2020-07-22
Fix issue with missing filenames and linenumbers in stacktraces

tdctaz:
Fix issue #662 with missing filenames and line numbers in stacktraces when running under mono.

Did some benchmarks for just adding `--debug` to every agent but it added a 10-15% runtime overhead to the tests (it varies a lot depending on the tests). But as mentioned by @ChrisMaddock that might not be acceptable for everyone.

Did try and see if there was anyway to detect if the current program was launched under mono using `--debug`, but could not find any clean way of detecting it.

So this adds a new option `--debugmono` (not sure if the name is good enough) when specified forces mono to run with `--debug`.

### DIFF
--- a/src/NUnitConsole/nunit3-console.tests/NetCoreConsoleOptionsTest.cs
+++ b/src/NUnitConsole/nunit3-console.tests/NetCoreConsoleOptionsTest.cs
@@ -93,6 +93,7 @@ namespace NUnit.ConsoleRunner.Tests
             "-loaduserprofile",
             "-agents=1",
             "-debug",
+            "-debugmono",
             "-pause",
             "-set-principal-policy=NoPrincipal",
 #if DEBUG
--- a/src/NUnitConsole/nunit3-console.tests/NetCoreConsoleOptionsTest.cs
+++ b/src/NUnitConsole/nunit3-console.tests/NetCoreConsoleOptionsTest.cs
@@ -93,7 +93,6 @@ namespace NUnit.ConsoleRunner.Tests
             "-loaduserprofile",
             "-agents=1",
             "-debug",
-            "-debugmono",
             "-pause",
             "-set-principal-policy=NoPrincipal",
 #if DEBUG
--- a/src/NUnitConsole/nunit3-console/ConsoleOptions.cs
+++ b/src/NUnitConsole/nunit3-console/ConsoleOptions.cs
@@ -173,6 +173,8 @@ namespace NUnit.Common
 
         public bool DebugAgent { get; private set; }
 
+        public bool DebugMono { get; private set; }
+
         public bool ListExtensions { get; private set; }
 
         public bool PauseBeforeRun { get; private set; }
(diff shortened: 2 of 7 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/nunit/nunit-console?pullRequest=799) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/nunit/nunit-console?pullRequest=799) before we can accept your contribution.<br/><br/>:x: tdctaz [sign now](https://cla.dotnetfoundation.org/nunit/nunit-console?pullRequest=799)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/nunit/nunit-console?pullRequest=799) it.</sub>

CharliePoole:
@tdctaz This looks good to me except for the naming, as you suggest. "DebugMono" suggests to me that we want to debug mono.

There are two names here: (1) the name used for the setting in `EnginePackageSettings.cs` and (2) the option the user types on the command-line.

The setting name can be as long as we like, since only we need to type it in our code -- and Intellisense helps with that. So something like `UseMonoDebugOption` or `MonoUseDebugOption` would be perfectly fine. You'll notice that many of the setting names are quite long.

The command-line option should be __reasonably__ short, although we have deliberately used longer option names where we don't expect many users to make use of the option. We can also have both a long and an abbreviated form for the option. Both forms show up in the console help.

One convention we could consider is to have options aimed at mono alone all begin with "--mono". Although it's a subjective opinion, `--monodebug` seems better to me than `--debugmono`. 

If we want a long and short pair, we could do `--monoUseDebugOption, --monoDebug`. A benefit of having two forms is that the user sees the two forms right next to one another  in the console help.

There are lots of possibilities, so please give it a think and hang on to see what the rest of the @nunit/engine-team thinks!

ChrisMaddock:
Great, thanks for digging in to this @tdctaz, and taking a look at the performance impact too!

If we're only talking 10-15%, I'm tempted to say we should just have this always-on, and not worry about the flag? I personally think that's an acceptable level of slowdown for adding important debug-logic. It also makes tests on mono feature-comparable to tests on netfx too.

Not too sure, what do you and @CharliePoole think?

CharliePoole:
@ChrisMaddock We could reverse it and have an option to turn it off.

Or maybe we should have it always on and wait for somebody to ask for a way to turn it off.

tdctaz:
Thanks, @ChrisMaddock and @ChrisMaddock.

I went back to just having `--debug` always on. If we want an option to disable it let me know and i'll add that, also if it something that gets requested later down the line.

tdctaz:
I got a what I suspect is a rare instability on the PR. I remember seeing a very similar error to this before as a random instability for the Unity unittests (6 times in the last year, all on Ubuntu 16.04).

Did do a force push, with no changes, to quickly verify that this properly is an instability. Not sure how you are handling situations like this.

```
System.Runtime.Remoting.RemotingException : Tcp transport error.
  ----> System.Runtime.Remoting.RemotingException : Connection closed
```

Here is the failed build in question:
https://travis-ci.org/github/nunit/nunit-console/builds/711130476

CharliePoole:
We usually just restart the build, which is what I did before seeing you had already done a force push.

ChrisMaddock:
I think we now only see the very occasional instability specifically on mono - we've all but eradicated it on .NET FX now, I think. We're also on the verge of a complete rewrite of our communications code which should also solve this problem. 🙂 

Thanks for contributing this one! I'm hoping to get a beta of the next release out soon. 👍

Issue #800, opened 2020-07-25
TypeLoadException: Method 'get_AssemblyPath' in type 'NUnit.Engine.Extensibility.ExtensionNode' does not have an implementation.


jnm2:
I'm getting a persistent failure to run tests in ReSharper's test runner (ReSharper 2020.1.4 on VS 16.6.5) with this error in the log:

> ERROR System.TypeLoadException: Method 'get_AssemblyPath' in type 'NUnit.Engine.Extensibility.ExtensionNode' from assembly 'nunit.engine, Version=3.10.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb' does not have an implementation.
System.TypeLoadException: Method 'get_AssemblyPath' in type 'NUnit.Engine.Extensibility.ExtensionNode' from assembly 'nunit.engine, Version=3.10.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb' does not have an implementation.
   at NUnit.Engine.Services.ExtensionService..ctor()
   at NUnit.Engine.TestEngine.Initialize()
   at NUnit.Engine.TestEngine.GetRunner(TestPackage package)
   at JetBrains.ReSharper.UnitTestRunner.nUnit30.NUnitRunner.Run(String assembly, IDictionary`2 testParameters) in /Product.Root/Psi.Features/UnitTesting/nUnit/Runner30/Src/NUnitRunner.cs:line 41
   at JetBrains.ReSharper.UnitTestRunner.nUnit30.BuiltInNUnitRunner.WithExtensiveErrorHandling(IRemoteTaskServer server, Action action) in /Product.Root/Psi.Features/UnitTesting/nUnit/Runner30/Src/BuiltInNUnitRunner.cs:line 73

This is what the test project looks like:

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net48</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.6.1" />
    <PackageReference Include="NUnit" Version="3.12.0" />
    <PackageReference Include="NUnit3TestAdapter" Version="3.17.0" />
  </ItemGroup>

</Project>
```

This is all it takes to repro the problem:

```cs
using NUnit.Framework;

public class Repro
{
    [Test]
    public void Test() { }
}
```

[...]

PR #802, opened 2020-07-29
Increment nunit.engine.api assembly version inline with other engine dlls

### LINKED ISSUE #801: Begin incrementing EngineApiVersion with every release, as per Engine version
To prevent issues like #800 and #424.

Historically, the engine api assembly version has remained static at 3.0, whilst the engine version increments with each release. The intention of this was to support a [centrally installed engine](https://github.com/nunit/nunit-console/issues/800#issuecomment-664633972), which is an concept that didn't end up gaining much traction, as the engine developed over the years.

I suggest from 3.12, we start incrementing the API assemblies version in line with the main engine version.

ChrisMaddock:
Fixes #801. See discussion on #800 for reasonings behind this change.


ChrisMaddock:
Hmm, Travis failed again. This hasn't happened in months, but has now happened 2/3 runs since merging #799. I wonder if we should back it out...

ChrisMaddock:
Thanks Joseph!

ulrichb:
When will this be released? Is there a workaround in the meantime? 

I tried different combination of NUnit and NUnit3TestAdapter package dependencies and all of them fail with R# :(

ChrisMaddock:
Hi @ulrichb,

The latest NUnit3TestAdapter and latest R# should work together. If you're finding that not to be the case, please open a new issue - I suspect you might be encountering a different problem.

On this change, not totally sure if it will be released yet, as it's caused another issue - see #807. Not totally sure what the solution is to all this yet...

ChrisMaddock:
Hi @ulrichb,

The latest NUnit3TestAdapter and latest R# should work together. If you're finding that not to be the case, please open a new issue - I suspect you might be encountering a different problem.

On this change, not totally sure if it will be released yet, as it's caused another issue - see #807. Not totally sure what the solution is to all this yet...

Issue #770, opened 2020-05-10
v4 Console/Engine Platform support 


ChrisMaddock:
NUnit has a long history of supporting legacy platforms. For v4, I wanted to think about how we can continue to do that, whilst reducing the burden on the development team. This post is intended to start that discussion.

Some high level thoughts:
- I'd like to see all the code in the main Engine/Console project targeting modern platforms.
- I'd like us to be able to continue to run on legacy platforms in an arms-length manner.
- I'd like to do whatever possible to reduce platform lock-in, and allow us to be more flexible in our commitment to supporting platforms.

Here's how I was thinking we could address that, component-by-component:

## The Console

The NUnit Console has historically been restricted to targeting .NET 2.0, as we've committed to being able to run tests on platforms with only .NET 2.0 installed.

.NET Core brought about [self-contained executables](https://docs.microsoft.com/en-us/dotnet/core/deploying/), which allow us to run on any platform, with no dependency on any runtimes being installed. To me, a truly portable test runner exactly where we want to be. This would then mean that in future, we can upgrade the console's platform as we see fit, as opposed to being restricted to targeting the lowest-supported platform.

## The Engine Agents

It is necessary to have an engine agent for every platform we wish to run tests on. I would like us to look at making it possible to add agents via the engine's extensibility functionality, so that users can set up the NUnit Engine can support a range of platforms, and the Engine Team aren't required to support all of these.

I'd like us to look at moving the CLR v2 agent out to an extension, and removing it from the main codebase, as we did with e.g. NUnit 2 functionality at the start of NUnit 3. .NET 2.0 is nearly 20 years old, and Microsoft's support for it ended just under 10 years ago. Doing this would mean there was a viable method for users to continue running tests on CLR v2, but it was not part of the main supported codebase.

.NET Framework and .NET Core agents should of course part of the main distribution - I hope the latter will be added during the 3.x series. 

## The Engine

The main engine assembly appears to be the main place where we will still be restricted on the platforms we support, as the platform the engine targets of course affects runners that rely on it.

I would like to see two ideas considered:

1. Removing in-process test running entirely. This would provide complete separation between the target platform of the engine, and the target platform of tests we support running. 

2. Aiming for a single-build, which can be utilised by runners on all operating systems. Right now .NET Standard 2.0 may be the most suitable option, but that picture may well change over the time it takes to develop v4 and as .NET 5 is released - and I think it would be sensible to review the exact platform further down the line.

[...]

Issue #424, opened 2018-05-24
System.TypeLoadException nunit3-console 3.8


esyang97:
I have my test file using NUnit 3.10.1

Here's the output I'm receiving after running nunit3-console.exe
```
NUnit Console Runner 3.6.0
Copyright (C) 2017 Charlie Poole

Runtime Environment
   OS Version: Microsoft Windows NT 10.0.16299.0
  CLR Version: 4.0.30319.42000

Test Files
    .\dll_test.dll


Unhandled Exception: NUnit.Engine.NUnitEngineException: Failed to start remote test agent.
   at NUnit.Engine.Services.TestAgency.OnAgentExit(Object sender, EventArgs e)
   at System.Diagnostics.Process.OnExited()
   at System.Diagnostics.Process.RaiseOnExited()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading._ThreadPoolWaitOrTimerCallback.PerformWaitOrTimerCallback(Object state, Boolean timedOut)
System.TypeLoadException: Could not load type 'NUnit.Engine.NUnitEngineUnloadException' from assembly 'nunit.engine.api, Version=3.0.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb'.
   at NUnit.Engine.Runners.ProcessRunner.Dispose(Boolean disposing)
   at NUnit.Engine.Runners.AbstractTestRunner.Dispose()
   at NUnit.Engine.Runners.MasterTestRunner.Dispose(Boolean disposing)
   at NUnit.Engine.Runners.MasterTestRunner.Dispose()
   at NUnit.ConsoleRunner.ConsoleRunner.RunTests(TestPackage package, TestFilter filter)
   at NUnit.ConsoleRunner.Program.Main(String[] args)
PS C:\Users\eyang\Documents\Work\Demonstration\dll_test\dll_test\bin\x64\Debug>

```

I found this thread (https://github.com/nunit/nunit/issues/1348) but it did not help much.

Cheers!

ChrisMaddock:
Closing this, as there's not enough information to understand/reproduce the problem. 

Happy to reopen if anyone has more to add. 🙂

Aleksanderis:
[...]

PR #813, opened 2020-09-01
Fix loading of extensions referencing Engine API v3.0.0

### LINKED ISSUE #807: Could not load file or assembly 'nunit.engine.api, Version=3.0.0.0,
#475 

I trying out 3.12 beta of the console-runner. Only change is to 
`<PackageReference Include="NUnit.Console" Version="3.11.1" />`
to 
`<PackageReference Include="NUnit.Console" Version="3.12.0-beta1" />`, 
and referencing 'nuni3-console.exe' from the beta-folder. 
All other packages is latest version, and targetting .NET Framework 4.8

Full output: 
```
Running only unit-tests
NUnit Console Runner 3.12.0-beta1 (.NET 2.0)
Copyright (c) 2020 Charlie Poole, Rob Prouse
mandag 3. august 2020 15:20:33

Runtime Environment
   OS Version: Microsoft Windows NT 6.2.9200.0
   Runtime: .NET Framework CLR v4.0.30319.42000

Test Files
    .\ProjectTests3.nunit

Test Filters
    Where: cat != MANUAL and cat != DATABASE and cat != NIGHTLY

Could not load file or assembly 'nunit.engine.api, Version=3.0.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb' or one of its dependencies. The system cannot find the file specified.

```

ChrisMaddock:
Fixes #807. This is a pretty unsatisfying fix to the problem. It feels like we're going down a rabbit hole, but I'm struggling to come up with a cleaner solution.

I'd appreciate some feedback on the approach, or any ideas of other solutions.

## Root cause
Since Engine version 3.0, the nunit.engine.api.dll assembly has not had the version number incremented. We've intended to only make what we thought were backwards-compatible changes to this assembly, however we've recently found that these changes have actually been causing a number of issues: ref: #6, #424, #800. 

There seems to be two viable solutions to this problem: 1) Stop making any changes to nunit.engine.api.dll or 2) start versioning the assembly when changes are made. I personally consider option 1 to be too restrictive to future development.

For that reason, we started versioning nunit.engine.api.dll in v3.12-beta1.

## Extension loading issue
Issue #807 highlighted that changing the version number of nunit.engine.api.dll had broken extension loading, for the NUnit-Org packaged extensions. The reason for this is that these extensions reference nunit.engine.api v3.0, but are not packaged with that assembly. Previously, this hadn't caused a problem as api v3.0 was always supplied with the engine, however with the above change this is no longer the case.

## Workaround-fix
This PR works around the above problem by forcing nunit.engine.api.dll v3.0 to always resolve to the same version as the engine, which allows the existing extensions to still be loaded. 

I _think_ this is acceptable as it's restricted to only apply to nunit.engine.api v3.0, which will be the only version affected by this issue, once we start incrementing the assembly version. It should also only affect cases where the engine is loading the API assembly itself, which as far as I can think, would only be in an extension loading scenario.

## Interim "proper" fix
Extensions should start packaging their referenced version of nunit.engine.api.dll alongside the extension assembly.

## v4 fix
For v4, I think we should split the extension API off from the main API assembly used for controlling the engine. The former is stable and sees a very limited number of changes, whereas the latter we wish to change more regularly. This is touched on in #770. Extensions should also continue to package the appropriate API version.

-----

Would appreciate all thoughts on this - I've not found it a nice problem to reason about...

### DIFF
--- a/src/NUnitEngine/nunit.engine.api/TestEngineActivator.cs
+++ b/src/NUnitEngine/nunit.engine.api/TestEngineActivator.cs
@@ -89,6 +89,11 @@ namespace NUnit.Engine
         /// <returns>An <see cref="ITestEngine"/></returns>
         public static ITestEngine CreateInstance(Version minVersion, bool unused = false)
         {
+            //Workaround for to support loading extensions referencing the original 3.0.0 version of this API, the version of which was
+            //not incremented from Engine versions 3.0-3.11. 
+            //See https://github.com/nunit/nunit-console/pull/813 for further details.
+            AppDomain.CurrentDomain.AssemblyResolve += EmbeddedAssemblyResolver;
+
             try
             {
                 Assembly engine = FindNewestEngine(minVersion);
@@ -151,6 +156,15 @@ namespace NUnit.Engine
             }
             catch (Exception){}
             return newestAssemblyFound;
+        }
+
+        private static Assembly EmbeddedAssemblyResolver(object sender, ResolveEventArgs args)
+        {
+            if (args.Name.StartsWith("nunit.engine.api, Version=3.0.0.0", StringComparison.Ordinal))
+            {
+                return Assembly.GetAssembly(typeof(TestEngineActivator));
+            }
+            return null;
         }
 #endif
     }


CharliePoole:
@ChrisMaddock I share your reservations about using this as a fix. For one thing, I'm not sure what will happen once a single extension is upgraded to use the new API assembly. Are we locking in extensions to only work with a specific engine version?

ChrisMaddock:
> Are we locking in extensions to only work with a specific engine version?

This is exactly what I'm trying to avoid doing here. But I'm also not an expert on the complexities of .NET assembly loading logic, so still concerned this may cause unintended side effects.

---

I'm considering delaying this change until we can break up the API in v4.0, and instead putting a freeze on changes to the nunit.engine.api assembly for the remainder of 3.x. That sounds like it will really impact development and move us further away from our desired architecture, but I'm struggling to think of an alternative solution to safety give us stability here.

I think the key issue is that modifying a production assembly while keeping the same version number works against the principles of .NET's build system, and continuing to do that will keep leading to unexpected and tricky to diagnose problems. There's all sorts of arguments as to whether the adapter should be copying to `bin`, and whether ReSharper should be loading assemblies from `bin` - but I think ultimately versioning the API assembly in the normal way would have avoided all these issues.I do think that's something we need to work out how we can get away from in 4.0. 😞

ChrisMaddock:
I'm going to close this PR for now, see #826 for the reasoning.

PR #816, opened 2020-09-05
Eliminate use of Mono.Cecil

### LINKED ISSUE #718: Eliminate use of Mono.Cecil
I'm writing this as a suggestion or idea for the NUnit Engine. As it happens, I'm following up in the TestCentric engine and will be glad to share the code if I'm successful.

Back in 2008 or 2009, I wrote a class to read PE headers into memory and examine the various assembly flags and other header fields. We used it for a while.

At the time, we were working on engine extensions and (originally) were using `Mono.Addins` to implement them. That meant we would be loading `Mono.Cecil` anyway, because `Mono.Addins` used it. So, we eliminated the `AssemblyReader` class in favor of `Mono.Cecil`. That was done here: https://github.com/nunit/nunit-console/commit/370d6081a38aee21eea332b4ea28026158bbee3c#diff-6f6d92c3a7b16401c609d5413097d4d4

However, we no longer use `Mono.Addins` but still use `Mono.Cecil`. This is problematic for two main reasons:

 * Newer versions of Mono.Cecil are not available for all the platforms where we need it. As a result, we mix two versions in our builds.
 * For our usage, Mono.Cecil is overkill. We probably use 5% of what it's capable of doing.

What we need to do is as follows:

1. Examine an assembly even if we are not currently running under the runtime it targets.
2. Make sure it's a .NET assembly.
3. See if 32-bit execution is required.
4. Get the runtime version (CLR version) used for the assembly
5. Look at assembly references to find what test framework it uses.
6. Examine assembly attributes, in particular the `TargetFrameworkAttribute`.

The `AssemblyReader` class already did the first four of these items at the time we dropped it. I'm looking at it to see about adding items 5 and 6 for the TestCentric engine.

CharliePoole:
Fixes #718 

See issue for details


ChrisMaddock:
Thanks Charlie - I’m travelling at the moment, but will try and take a look at these soon.

rprouse (approved):
I would be very happy to see the end of `Mono.Cecil`, so this is very welcome. The PR assumes that `testcentric.engine.metadata` is correct for all the cases we use it for and all the frameworks. If so, then this looks good.

Eventually, should we bring this code into the NUnit project so that it can be maintained by a wider communty if there are problems with less used platforms or architectures?

CharliePoole:
@rprouse I'd say that the PR does not assume anything except our __ability__ to deal with any frameworks we need to deal with. Clearly, it doesn't support frameworks that do not yet exist but the engine and console tests do exercise it fairly well.

I agree that it may be advisable to eventually bring this under NUnit and noted that in the issue itself. However, for me, the decision is tightly bound with the eventual outcome of the TestCentric / Engine Compatibility project, a cooperative project between Chris and myself. The issue being resolved here is, in fact, part of that project.

If  the new assembly were forked by NUnit __prematurely__, that would have several negative consequences:

 1. The Engine project would have more work to do.

 2. The Engine project would have traded an __actual__ maintainer of this assembly for the __theoretical__ possibility of finding more community help.

 3. There would be two versions of the assembly, with the result of actually increasing the divergence, which this PR is trying to reduce.

In this context, "prematurely" means "before the conclusion of the compatibility project" or at least before we have a clear idea of where that project is going to end up, i.e. whether the TC engine can be eliminated or not.

msi/nunit/engine-files.wxi
```diff
@@ -32,11 +32,11 @@
                     <CopyFile Id="nunit.agent.net40.engine.api.xml" DestinationDirectory="NET40_AGENT_DIR" />
                 </File>
             </Component>
-            <Component Id="MONO.CECIL" Location="local" Guid="D4A15752-6685-3494-9346-F623175CEFFA">
-                <File Id="Mono.Cecil.dll"
-                      Source="$(var.InstallImage)bin/net20/Mono.Cecil.dll">
-                    <CopyFile Id="nunit.agent.net20.Mono.Cecil.dll" DestinationDirectory="NET20_AGENT_DIR" />
-                    <CopyFile Id="nunit.agent.net40.Mono.Cecil.dll" DestinationDirectory="NET40_AGENT_DIR" />
+            <Component Id="NUNIT.ENGINE.METADATA" Location="local" Guid="D4A15752-6685-3494-9346-F623175CEFFA">
```
ChrisMaddock, line 35: Shouldn't we be changing the GUID here, as this is a new and different assembly? I think that would be best practice.

ChrisMaddock (commented):
Thanks Charlie - LGTM. Just one point - please go ahead and merge when it's resolved. 🙂

msi/nunit/engine-files.wxi
CharliePoole: I'm not sure I know the right answer. The GUID represents the component, not the assembly. I figured that it's essentially a replacement component and so should have the same GUID. I believe that would be necessary if we did partial updates, e.g. replacing just the one component.

Since we always do a full install, I believe either will work, but I'm not 100% sure which approach is technically more correct. I sent Rob Mensching a note to see what he thinks.

CharliePoole: Got an answer... Per Rob, I should change the component GUID if the name of the assembly it contains is changing. I'll go ahead and do that.

FWIW, he also suggests we stop maintaining GUIDs on components and let WIX take care of it for us. That seems like a separate issue, however.

ChrisMaddock: I didn’t know that was an option - I’ve always done my own GUIDs. Cool!

CharliePoole: Done

PR #824, opened 2020-11-13
Remove redundant Microsoft.DotNet.InternalAbstractions dependency

### LINKED ISSUE #818: Seemingly redundant dependency on Microsoft.DotNet.InternalAbstractions for platforms other than .Net Standard 1.6
This line https://github.com/nunit/nunit-console/blob/0d037ca9d569282c14b798ed6a69b8532fc88a4e/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs#L451 makes the only use of anything from Microsoft.DotNet.InternalAbstractions in the entire codebase. This line only gets to the compiler when the package is compiled to target .Net Standard 1.6 because all other platforms have `Environment.Version` which is well known and is used almost everywhere.

However the .nuspec and .csproj files are crafted such that the dependency on Microsoft.DotNet.InternalAbstractions is not restricted to .Net Standard 1.6 Those files also list this dependency for .Net Standard 2.0 and .Net Core 3.1 

The problem is Microsoft.DotNet.InternalAbstractions is published in Nuget with no license specified which is is long known issue https://github.com/dotnet/sdk/issues/9889 This makes Microsoft.DotNet.InternalAbstractions formally "unlicensed" and raises concerns when conducting a codebase licensing audit. This codebase depends on Microsoft.DotNet.InternalAbstractions and so is also flagged as having a dependency on an "unlicensed" piece of software. Same goes for NUnit3TestAdapter which make use of NUnit.Engine - https://github.com/nunit/nunit3-vs-adapter/issues/785

It would be nice if the dependency was restricted to .Net Standard 1.6 only because this is the only case it is actually needed. This would greatly reduce impact of the licensing issue.

ChrisMaddock:
Fixes #818

### DIFF
--- a/nuget/engine/nunit.engine.nuspec
+++ b/nuget/engine/nunit.engine.nuspec
@@ -30,7 +30,6 @@
       </group>
       <group targetFramework="netstandard2.0">
         <dependency id="NETStandard.Library" version="2.0.0" />
-        <dependency id="Microsoft.DotNet.InternalAbstractions" version="1.0.0" exclude="compile" />
         <dependency id="System.ComponentModel.TypeConverter" version="4.3.0" exclude="compile" />
         <dependency id="System.Reflection" version="4.3.0" exclude="compile" />
         <dependency id="System.Diagnostics.Process" version="4.3.0" exclude="compile" />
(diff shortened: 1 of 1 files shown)


Dmitry-Me:
@ChrisMaddock Please note that this only tells NuGet that the dependency is not needed, but .csproj files still have the dependency for both netstandard2.0 and netcoreapp3.1 targets. So it is possible that after some changes made later the package will build fine but a runtime failure occurs when the code actually runs because a compile-time dependency is still there.

ChrisMaddock:
Whoops! Thanks @Dmitry-Me - you're spot on.

PR #848, opened 2020-12-27
Fix extension loading for extensions which target multiple runtimes

### LINKED ISSUE #844: .NET Core console runner fails to load extensions when netfx and netstandard versions conflict
It does not work when a file like "NUnit.Extension.TeamCityEventListener\1.0.9\tools\.addins" contains:
```
net20/
netstandard2.0/
```

It works when it contains:
```
netstandard2.0/
net20/
```

[InternalTrace.26616.log](https://github.com/nunit/nunit-console/files/5709755/InternalTrace.26616.log)

ChrisMaddock:
Fixes #844. 

There was an error in the extension selection logic, where an extension of a higher version could replace an extension of a lower version, even if the higher version could not be loaded in the current runtime. This changes brings the check of where an extension is valid in the current runtime forward to prevent that.

The actual change is in the first commit, the second commit is subsequent reformatting and removal of the old check. Unfortunately this whole piece of logic currnently relies on 'real files' and is untested - something I think we'll need to refactor in future.

### DIFF
--- a/src/NUnitEngine/nunit.engine.core/Services/ExtensionService.cs
+++ b/src/NUnitEngine/nunit.engine.core/Services/ExtensionService.cs
@@ -410,7 +410,13 @@ namespace NUnit.Engine.Services
         /// </summary>
         internal void FindExtensionsInAssembly(ExtensionAssembly assembly)
         {
-            log.Info("Scanning {0} assembly for Extensions", assembly.FilePath);
+            log.Info($"Scanning {assembly.FilePath} for Extensions");
+
+            if (!CanLoadTargetFramework(Assembly.GetEntryAssembly(), assembly))
+            {
+                log.Info($"{assembly.FilePath} cannot be loaded on this runtime");
+                return;
+            }
 
             IRuntimeFramework assemblyTargetFramework = null;
 #if NETFRAMEWORK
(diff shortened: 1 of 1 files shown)


ChrisMaddock:
Ah - turns out this was better tested than I thought...although that doesn't stop the check being in the wrong place! This needs a bit more refactoring that I really have time for today. Will come back to it at some point.

src/NUnitEngine/nunit.engine.core/Services/ExtensionService.cs
```diff
@@ -352,43 +352,46 @@ private void ProcessAddinsFile(DirectoryInfo baseDir, string fileName, bool from
...
+
+            Visit(filePath);
+
+            try
             {
-                Visit(filePath);
+                var candidate = new ExtensionAssembly(filePath, fromWildCard);
+
+                if (!CanLoadTargetFramework(Assembly.GetEntryAssembly(), candidate))
+                    return;
```
ChrisMaddock, line 366: The addition of lines 364-6 is the only substantial change in this PR.

ChrisMaddock:
Apologies folks, I've done more re-formtting in this PR than is helpful. Have highlighted the only important change with a comment, this is ready for review now, so long as CI passes this time.

PR #851, opened 2021-01-02
Stop building engine assembly for .NET Standard 1.6

### LINKED ISSUE #849: Eliminate .NET Standard 1.6 build of nunit.engine assembly
This came up in some discussons within the team on other topics...

Now that the engine has been split in to `nunit.engine` and `nunit.engine.core`, the `nunit.engine` assembly is only loaded 
in the main process and therefore only needs to support the runtime used by that process. The .NET Standard 1.6 build is no longer needed and may be removed. 

The 1.6 build of `nunit.engine.core` is not being removed at this time. It remains to be seen whether we will continue to maintain the .NET Standard 2.0 build of the engine or if we will switch to a .NET Core build.

CharliePoole:
Fixes #849

PR #854 should be reviewed before this one, because it removes .NET Standard 1.6 completely. If that's approved, we can drop this one, which only removes the 1.6 build for the primary engine.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -142,6 +142,17 @@ Task("Clean")
         CleanDirectory(PACKAGE_DIR);
     });
 
+// Not currently used in CI but useful for cleaning your local obj
+// directories, particularly after changing the target of a project.
+Task("CleanAll")
+    .Description("Cleans obj directories in additon to standard clean")
+    .IsDependentOn("Clean")
+    .Does(() =>
+    {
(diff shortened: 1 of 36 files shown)


ChrisMaddock:
See comments on #854. I remain of the opinion that removing .NET Standard 1.6 completely is the way we want to go, but will take a proper look through these soon. 🙂

ChrisMaddock:
@CharliePoole - this is superseded by #871, right?

CharliePoole:
Right!

PR #856, opened 2021-01-03
Issue warning if .NET 5.0 not installed for local build

### LINKED ISSUE #853: Allow local build to succeed even if all runtimes are not installed
One of my machines has an older version of VS2019 and fails the build when it comes time to run .NET 5.0 tests (it has a pre-release). It seems to me that we should that all runtimes we use be available on CI systems, and set up the agents so that it works, as we now do. OTOH, on a local developer machine, it may be more convenient to just skip any build or test that's not supported.

@nunit/engine-team What do you think?

CharliePoole:
Fixes #853


mikkelbu (approved):
LGTM

PR #854, opened 2021-01-02
Remove netstandard1.6 and netcoreapp1.1 builds

CharliePoole:
Fixes #725

PR #863, opened 2021-01-14
chore: Use released version of NUnit 3.13.0

mikkelbu:


ChrisMaddock:
Thanks Mikkel - this was on my to-do list before doing the release. 🙂 It's conflicting with #854 now, but I'm hoping after #864 is merged we'll be able to merge this in on top. Thanks for sorting it! 😄

PR #866, opened 2021-01-17
Point changes.txt to docs site

### LINKED ISSUE #803: Remove CHANGES.TXT to simplify release process
The release process is currently quite manual and time consuming. There's lots of improvements we could make here, but this one I think is low-hanging fruit.

Currently we have to create two versions of the changelog, one in text format and one in markdown. I'd like to suggest removing  CHANGES.TXT from the repo, and just have on docs site. This would involve updating the various packages that currently include CHANGES.txt to link to the change log instead.

Thoughts welcome. 🙂

SeanKilleen:
Resolves #803.


CHANGES.txt
```diff
@@ -1,1516 +1,3 @@
...
-Bug Fixes
-
- * 400502 	NUnitEqualityComparer.StreamsE­qual fails for same stream
- * 400508 	TestCaseSource attirbute is not working when Type is given
- * 400510 	TestCaseData variable length ctor drops values
- * 417557 	Add SetUICultureAttribute from NUnit 2.5.2
- * 417559 	Add Ignore to TestFixture, TestCase and TestCaseData
- * 417560 	Merge Assert.Throws and Assert.Catch changes from NUnit 2.5.2
- * 417564 	TimeoutAttribute on Assembly
+Instead, please see the [NUnit docs](https://docs.nunit.org), and specifically the [Console and Engine Release Notes](https://docs.nunit.org/articles/nunit/release-notes/console-and-engine.html) page there.
```
ChrisMaddock, line 3:
```suggestion
Please see the [NUnit docs](https://docs.nunit.org), and specifically the [Console and Engine Release Notes](https://docs.nunit.org/articles/nunit/release-notes/console-and-engine.html) page there.
```

ChrisMaddock:
Hi Sean, thanks for doing this!

Sorry, I wanted to add this as a suggestion, but multi-line suggestions don't seem to be working for me. (Or they behave differently to how GitHub documents them doing...)

What would you think to taking out the "this file is no longer used..." bit? My thinking: people will get to this file via two routes. 1. Via old links to it, which is I think the usecase you were intending, but also 2. when this file is distributed with our zip/msi and other packages.

Your wording makes complete sense for the first usecase, but I think it suits the second case less, when we'll still be distributing CHANGES.TXT with future packages. How about rewording to just something simple like "Full release notes can be found at XXX"?

SeanKilleen:
@ChrisMaddock should be ready to go; feel free to make any changes the wording as you see fit.

ChrisMaddock (approved):
Looks great - thanks for that!

PR #869, opened 2021-01-17
Link to release notes on docs and Review uses of CHANGES.TXT

ChrisMaddock:
I was about to ask @SeanKilleen if he would mind updating the NuGet packages release notes in line with his changes in #866, but then found out most of them were blank anyway!

This PR changes some references from CHANGES.TXT to a direct link, to avoid multiple levels of redirection. CHANGES.TXT is retained for the msi and zip distributions, but removed from choco/nuget - with the logic that users would use the respective gallery's and interfaces to access this information instead.


CharliePoole (approved):
Good idea! LGTM

ChrisMaddock:
Thanks Charlie. 🙂

PR #870, opened 2021-01-18
WIP: Add Documentation Analyzers and Remediate

### LINKED ISSUE #782: Add DotNetAnalyzers.DocumentationAnalyzers
The doccomments in this repo haven't always kept in sync with the code. As recommended by @jnm2 in https://github.com/nunit/nunit/issues/3475 - this will help keep things tidy. In the framework project they chose to suppress DOC100 - I think it makes sense for us to do the same.

> I would recommend that we add the design-time tool https://www.nuget.org/packages/DotNetAnalyzers.DocumentationAnalyzers to our Directory.Build.props and follow all its prompts. It's been awesome in every project where I've seen it. I know it catches this particular case.

SeanKilleen:
Resolves #782.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -13,7 +13,9 @@ indent_size = 4
 
 # Default severity for analyzer diagnostics with category 'DocumentationAnalyzers.Refactorings'
 dotnet_analyzer_diagnostic.category-DocumentationAnalyzers.Refactorings.severity = suggestion
+dotnet_diagnostic.DOC100.severity = none # DOC100: Place text in paragraphs
 
 [*.{proj,csproj,vcxproj,xproj,json,config,nuspec,xml,yml}]
 indent_style = space
 indent_size = 2
+
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,6 +14,7 @@ indent_size = 4
 # Default severity for analyzer diagnostics with category 'DocumentationAnalyzers.Refactorings'
 dotnet_analyzer_diagnostic.category-DocumentationAnalyzers.Refactorings.severity = suggestion
 dotnet_diagnostic.DOC100.severity = none # DOC100: Place text in paragraphs
+dotnet_diagnostic.DOC900.severity = none # DOC900: Render as Markdown (not working correctly at time of initial use)
 
 [*.{proj,csproj,vcxproj,xproj,json,config,nuspec,xml,yml}]
 indent_style = space
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,6 +11,9 @@ root = true
 indent_style = space
 indent_size = 4
 
+# Default severity for analyzer diagnostics with category 'DocumentationAnalyzers.Refactorings'
+dotnet_analyzer_diagnostic.category-DocumentationAnalyzers.Refactorings.severity = suggestion
+
 [*.{proj,csproj,vcxproj,xproj,json,config,nuspec,xml,yml}]
 indent_style = space
 indent_size = 2
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -9,6 +9,10 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="DotNetAnalyzers.DocumentationAnalyzers" Version="1.0.0-beta.59"/>
+  </ItemGroup>
+  
+  <ItemGroup>
     <Compile Include="..\..\CommonAssemblyInfo.cs" LinkBase="Properties" />
   </ItemGroup>
 
--- a/src/NUnitConsole/nunit3-console/nunit3-console.csproj
+++ b/src/NUnitConsole/nunit3-console/nunit3-console.csproj
@@ -15,6 +15,12 @@
     <Content Include="..\..\..\nunit.ico" Link="nunit.ico" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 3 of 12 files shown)


SeanKilleen:
A little ways in on this, I'm not sure if it's the right call to go down this path because:

* The analyzer package seems to be a little older (not having been updated in 2 years). 
* The refactoring for DOC900 didn't work when suggested
* I don't seem to be able to get a solution-wide analysis to appear in Visual Studio, and so would have to open every file.

Is it still a good idea to proceed with this? Unsure, but it might make more sense to abandon. 

CC @ChrisMaddock.

ChrisMaddock:
Hmm, ok - sounds like it might be more effort than it's worth. This was firmly in the 'nice-to-have' category, and I'd assumed it would be a simple install-and-process.

Let's give it a miss, if it's not working out so well. The vast majority of our doccomments are for internal use only, so the need isn't as big as with the framework. Thanks for trying it out anyway! 🙂

sharwell:
> I don't seem to be able to get a solution-wide analysis to appear in Visual Studio, and so would have to open every file.

You can enable Full Solution Analysis in Tools&rarr;Options, or you can right click the solution in Solution Explorer and manually run the full solution analysis with Analyze and Code Cleanup &rarr; Run Code Analysis on Solution.

SeanKilleen:
@sharwell thanks for showing up here! I'll definitely consider re-opening this especially if we've got a hand. Don't want to write it off prematurely. 👍

src/Directory.Build.props
```diff
@@ -8,6 +8,10 @@
     <DisableImplicitNuGetFallbackFolder>true</DisableImplicitNuGetFallbackFolder>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="DotNetAnalyzers.DocumentationAnalyzers" Version="1.0.0-beta.59"/>
```
sharwell, line 12:
```suggestion
    <PackageReference Include="DotNetAnalyzers.DocumentationAnalyzers" Version="1.0.0-beta.59" PrivateAssets="all" />
```

src/NUnitConsole/nunit3-console/nunit3-console.csproj
```diff
@@ -14,6 +14,12 @@
     <Compile Include="..\ConsoleVersion.cs" LinkBase="Properties" />
     <Content Include="..\..\..\nunit.ico" Link="nunit.ico" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="DotNetAnalyzers.DocumentationAnalyzers" Version="1.0.0-beta.59">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+  </ItemGroup>
```
sharwell, line 22: ❔ Why is this not just inherited from **Directory.Build.props**?

src/NUnitEngine/nunit.engine.core/Internal/AssemblyHelper.cs
```diff
@@ -68,7 +68,7 @@ private static bool IsFileUri(string uri)
         /// <summary>
         /// Gets the assembly path from code base.
         /// </summary>
-        /// <remarks>Public for testing purposes</remarks>
+        /// <remarks><para>Public for testing purposes</para></remarks>
```
sharwell, line 71: 💡 Unrelated to this change, but I highly recommend using the `TestAccessor` pattern for this. It eliminates the need for testing concerns to impact the accessibility of members.

.editorconfig
```diff
@@ -11,6 +11,12 @@ root = true
 indent_style = space
 indent_size = 4
 
+# Default severity for analyzer diagnostics with category 'DocumentationAnalyzers.Refactorings'
+dotnet_analyzer_diagnostic.category-DocumentationAnalyzers.Refactorings.severity = suggestion
+dotnet_diagnostic.DOC100.severity = none # DOC100: Place text in paragraphs
+dotnet_diagnostic.DOC900.severity = none # DOC900: Render as Markdown (not working correctly at time of initial use)
```
sharwell, line 17: 💭 Well it sure won't work like this. Since the default severity is `refactoring`, maybe leave this one at its default?

src/NUnitEngine/nunit.engine.api/ITestRunner.cs
```diff
@@ -42,9 +42,9 @@ public interface ITestRunner : IDisposable
         /// </summary>
         /// <returns>An XmlNode representing the loaded package.</returns>
         /// <remarks>
-        /// This method is normally optional, since Explore and Run call
+        /// <para>This method is normally optional, since Explore and Run call
         /// it automatically when necessary. The method is kept in order
-        /// to make it easier to convert older programs that use it.
+        /// to make it easier to convert older programs that use it.</para>
```
sharwell, line 47: I believe this one is the result of DOC100, which you disabled?

You likely don't need to use `<para>` elements unless the `<remarks>` element has _multiple_ paragraphs.

ChrisMaddock:
Hi @SeanKilleen  - I'm going to close this one for now, until someone gets a chance to revist. Thanks for giving it a shot however, and by all means would be happy to look again if you feel like coming back to it in future. 😄

PR #871, opened 2021-01-19
Remove netstandard1.6 and netcoreapp1.1 builds

### LINKED ISSUE #725: Remove .NET Standard 1.6 build

CharliePoole:
Fixes #725 

This is essentially the same commit as was previously approved... but different in that it's the result of reverting @ChrisMaddock 's reversion of the change. So it probably deserves a second look... just in case.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,13 +11,6 @@ jobs:
   
   steps:
 
-  - task: UseDotNet@2
-    displayName: 'Install .NET Core 1.1.2'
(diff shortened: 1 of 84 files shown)


ChrisMaddock (approved):
Thanks Charlie! This should make development a lot easier. 🙂

PR #875, opened 2021-01-20
Correct packaging of x86 .NET 4.0 agent in msi

### LINKED ISSUE #872: NUnitEngineException by updating from 3.11.1 to 3.12
I get the attached exception after updating NUnit-Console from 3.11.1 to 3.12.
With NUnit-Console 3.11.1, there was no exception thrown.

I have NUnit 3.13 installed via Visual Studio 2017 NuGet. 

[exception.txt](https://github.com/nunit/nunit-console/files/5835965/exception.txt)

ChrisMaddock:
Fixes #872 - fixes a copy-paste error in the msi installer.

@nunit/engine-team - this built off the 3.12.0 release branch. As this is just a packaging issue, unless anyone can think of a good reason not to, I plat to just rebuild the msi and update the GitHub artifact for this issue, rather than doing a full 3.12.1 release. Would appreciate a second opinion on that however...before something goes wrong!

### DIFF
--- a/msi/nunit/engine-files.wxi
+++ b/msi/nunit/engine-files.wxi
@@ -79,9 +79,9 @@
             <Component Id="NUNIT_AGENT_NET40_X86" Location="local" Guid="8EB0C529-56D8-4746-804E-6679744699F6">
                 <File Id="nunit_agent_net40_x86.exe"
                       ProcessorArchitecture="x86"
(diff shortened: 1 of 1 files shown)


CharliePoole (commented):
Looks good. And just replacing the msi in the release makes sense along with posting a notice.

PR #855, opened 2021-01-03
Improved testability of DirectoryFinder

x789:
Instead of using `System.IO`, the `DirectoryFinder` uses a small abstraction layer to access the file-system. This allows its logic to be tested thoroughly in unit tests that are not dependent on the real file system.

More about the _how_ and _why_ can be found in the comments of issue #789.

PR #876, opened 2021-01-21
Simplifying file headers

### LINKED ISSUE #793: Simplify headers on source files
I'm proposing that we shorten and simplify the file headers as follows:

1. Eliminate the license text in order to make it shorter.
2. Eliminate copyright dates so we no longer need to update them.
3. Add mention of contributors just as a matter of fairness.

The actual text to be used should be agreed upon. As a start, I propose the following:

```
// *****************************************************************************
// Copyright (c) Charlie Poole, Rob Prouse and Contributors.
// Licensed under the MIT License. See LICENSE.txt in root directory.
// *****************************************************************************
```
Based on experience doing this for TestCentric, I suggest that we review
the name of the license file now. If we want to change it later, as I did, we'll
have to edit all the headers! (Mine is now LICENSE)

CharliePoole:
Fixes #793

### Approach

* Using the one-line header suggested by @ChrisMaddock ...
```
// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
```
This will be easy to change before we finally merge.

* Added a target to the build, only intended to be run locally. It gives separate lists of files with no header and files with a non-standard header. It's located in a separate cake file dedicated to local-only targets.

* I considered adding a task to add headers to files without them but it didn't seem to have a big payoff once the initial conversion is complete.

* I also considered adding a task to replace non-standard headers, but didn't for a different reason: it requires too much thought to be automated. IMO we should not automatically replace the name of a copyright holder. A human has to decide.

* I exempted AssemblyInfo.cs from checking. It can be added back if desired.

* All targeted files now have the standard header, with three exceptions. See further comment later.

@nunit/engine-team What do you think?


CharliePoole:
### Exceptions

Three exceptions are currently found...

1. Options.cs - this is a file we have always kept with its original header and which we try not to modify.
2. ProcessUtils.cs - Copyright Joseph Musser
3. ProcessUtilsTests.cs - Copyright Joseph Musser

I assume that items 2 and 3 are inadvertent, but I won't replace them without first hearing from @jnm2.

The three exceptions are a great illustration of why I don't believe we should ever automatically replace headers.

CharliePoole:
BTW although totally unrelated to #793 I moved other targets that are local only to the new cake file.

CharliePoole:
I have to say, I don't __love__ the new header as it's a bit too terse for me. My personal preference would be a four-line version, keeping the lines of asterisks. My second favorite would be just the two lines with the actual info. That said, I acknowledge that 99.9% of readers will probably figure out what the one-line version is saying. :confused: 

One line:
```
// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
```

Two lines:
```
// Copyright (c) Charlie Poole, Rob Prouse and Contributors.
// Licensed under the MIT License. See LICENSE.txt in root directory.
```

Four lines:
```
// *****************************************************************************
// Copyright (c) Charlie Poole, Rob Prouse and Contributors.
// Licensed under the MIT License. See LICENSE.txt in root directory.
// *****************************************************************************
```

local-tasks.cake
```diff
@@ -0,0 +1,121 @@
...
+    return header;
+}
+
+private void ListFilesWithNoHeader()
+{
+}
+
+private void ListFilesWithNonStandardHeader()
+{
+}
```
ChrisMaddock, line 115:
```suggestion
```
Suspect these are accidental?

local-tasks.cake
```diff
@@ -0,0 +1,121 @@
...
+
+static readonly int CD_LENGTH = Environment.CurrentDirectory.Length + 1;
+
+// Standard Header. Change this for each project as needed.
+static readonly string[] STD_HDR = new [] {
+    "// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt"
+};
+
+Task("CheckHeaders")
+    .WithCriteria(BuildSystem.IsLocalBuild)
```
ChrisMaddock, line 42: I wonder about adding this to CI as a check, or maybe also to the TestAll target? My thinking is it would be good if new contributors hit this automatically before they did a pull request - so it's gets sorted from the start rather than us having to ask people to amend. We'd need to make the task throw an error, of course.

local-tasks.cake
```diff
@@ -0,0 +1,121 @@
...
+        Information($"\nFiles Examined: {examined}");
+        Information($"Missing Headers: {NoHeader.Count}");
+        Information($"Non-Standard Headers: {NonStandard.Count}");
+    });
+
+private List<FilePath> GetFilesToCheck()
+{
+    var files = new List<FilePath>();
+    foreach(var file in GetFiles("src/**/*.cs"))
+        if (file.GetFilename().ToString().ToLower() != "assemblyinfo.cs")
```
ChrisMaddock, line 88: Would it simplify things just to add the header to these files? There's no particular reason for them to be exempted, is there?

ChrisMaddock (commented):
>I considered adding a task to add headers to files without them but it didn't seem to have a big payoff once the initial conversion is complete.
> I also considered adding a task to replace non-standard headers, but didn't for a different reason: it requires too much thought to be automated. IMO we should not automatically replace the name of a copyright holder. A human has to decide.

Agreed and agreed.

> I have to say, I don't love the new header as it's a bit too terse for me. My personal preference would be a four-line version, keeping the lines of asterisks. My second favorite would be just the two lines with the actual info. That said, I acknowledge that 99.9% of readers will probably figure out what the one-line version is saying. 😕

I don't have strong feelings if we'd prefer a longer version. My preference would be the two-lines over the four-lines - I'm not convinced the astrisks add vale! 😄 

Thanks for sorting this, Charlie.

local-tasks.cake
CharliePoole: I can do that but I feel like we need a stronger capability to eliminate files we do __not__ want checked, e.g. Options.cs. I'll add a pattern of files to __not__ check in addition to the ones we check and eliminate the criterion.

local-tasks.cake
CharliePoole: Mostly, this is in so I can test the ability to exempt files, which I think we need. I'll remove before merge if you like.

CharliePoole:
The latest commit fixes issues from the review and adds a cleaner way of identifying exempt files. Currently exempt are Options.cs and AssemblyInfo.cs.

It's still local-only because running it in CI would potentially break the build, so I think we have to get all our ducks in a row before doing that... either in this PR or later on in a different one.

@ChrisMaddock and @nunit/engine-team ... do we definitely want to run this in CI? If so, would it be before building, after building, after testing, etc? Do we want a failure to break the build? 

Once we decide, I'll make the changes. Obviously, we can change again later on.

ChrisMaddock:
I think now we've written the code, we might as well run it! As above, I think this would be best if new contributors hit this before creating a PR, so how about putting it under the "Test" target? (It's a test of the source code, right?)

CharliePoole:
I can do that but we're still blocked by @jnm2's two files.

jnm2:
> 2. ProcessUtils.cs - Copyright Joseph Musser
> 3. ProcessUtilsTests.cs - Copyright Joseph Musser
> 
> I assume that items 2 and 3 are inadvertent, but I won't replace them without first hearing from @jnm2.

This was on purpose but I don't know what I'm doing. I've been copying both files into various unrelated projects from https://gist.github.com/jnm2/c5c840bf317605a40f5f56f944db4892. I was thinking this meant that the copyright was not owned by you and Rob. Is there a better way to do it than what I did?

jnm2:
(If I had submitted with the copyright notice claiming ownership by you and Rob, would that have significance in any of the other places I've submitted and used the same code, minus the namespace change?)

CharliePoole:
Nothing wrong with what you've done, but if it's on purpose, then we need to decide whether to exempt it from the header check or stop using it. IMO that's a @ChrisMaddock decision. 

Ideally, someone would have asked about this at the PR stage. In future, you would have to add such a file to the list of exemptions yourself, so it would get reviewed automatically.

I have lots of code myself that I've contributed to multiple projects under their own terms, both with and without my name.

CharliePoole:
It occurs to me that all the headers we are looking at, including the old one, are missing one thing. There's nothing to say that the file is part of any NUnit project. So if it gets separated, where did the file come from?

CharliePoole:
Ready for review again...

The latest commits did the following...

* The check task runs in the CI build. If any file is missing a header or has a nonstandard header, the build fails.
* I tried adding the check to the Test target but files generated by msbuild caused it to fail so I moved it to before Build.
* I exempted Joseph's two files, Options.cs and AssemblyInfo.cs. exempted files are listed but don't cause a failure.

I'm still not 100% happy with the one-line header, but I can live with it.

ChrisMaddock:
Exemption is fine by me. 

> There's nothing to say that the file is part of any NUnit project. So if it gets separated, where did the file come from? 

I think this is a fair point - it would be nice to include the name NUnit in the files. Want to suggest an alternative Charlie? I'd be fine if this pushed it to multiple lines - maybe something similar to your second example above?

CharliePoole:
How about just adding a modifier to Contributors? That tells you
implicitly what it's part of.

```
// Copyright (c) Charlie Poole, Rob Prouse and NUnit Engine Contributors.
// Licensed to you under the MIT License. See LICENSE in root directory.

// Copyright (c) Charlie Poole, Rob Prouse and NUnit Console Contributors.
// Licensed to you under the MIT License. See LICENSE in root directory.
```

It's only a tiny bit harder to change them all using a slightly different wording
for engine and console. Otherwise, we can just say NUnit, but that covers
the framework, adapter, etc as well.

BTW I have found it easiest to work in VS Code when changing these.

ChrisMaddock:
I think "NUnit Contributors" works ok? 🙂 Having two different copyright headers for different parts of the project sounds like it will lead to confusion!

ChrisMaddock:
Charlie - this will clash with #855. Would you mind if we got that merged in first, and just did a final sweep of this after? I'm hoping the contributor will be able to make the last couple of changes soon - my fault for leaving it sitting while we got the release out! 🙂

CharliePoole:
@ChrisMaddock I came in here to remind you that this is still pending but then saw  that you changed the label in the last day to waiting for contributor. Are we no longer waiting for #855?

ChrisMaddock:
855 is merged now. 😊 Sorry, I meant to ping you when I changed the labels - I clearly got distracted half way through!

how about we merge this and then rebase the 4.0 branch on top of this? Just as this will touch literally every file. 😅

CharliePoole:
Sure. I'll need to do another pass since I now have conflicts. But I'll do that and merge once the CI passes.

CharliePoole:
I fixed all conflicts and added headers to new files. I also had to add an exemption for any source files in an `obj` directory since a lot of them are now being generated for us. Since it's only the headers, I plan to merge this as soon as CI passes.

PR #883, opened 2021-01-29
Add .netcoreapp3.1 files to package checks

### LINKED ISSUE #852: No tests of .NET Core 3.1 packages
When the .NET Core 3.1 package was added, no checks were added to package-checks.cake. The package checks aren't real tests but they at least confirm that the right files are in each package, avoiding the need for manual verification.

CharliePoole:
Fixes #852


ChrisMaddock:
Thanks Charlie!

Issue #894, opened 2021-02-03
Test runner does not close fired up proceseses. 


TimonPost:
When I fire up a test runner, it starts two processes: 

![image](https://user-images.githubusercontent.com/19969910/106756341-7ce7e380-662f-11eb-8549-93c234faee8a.png)

After the runner has performed the test round it should close the processes. I try to accomplish this by setting one of the following configurations:

```
static void Main(string[] args)
        {
            // Get an interface to the engine
            using (ITestEngine engine = TestEngineActivator.CreateInstance())
            {
                var settings = engine.Services.GetService<ISettings>();

                settings.SaveSetting("DisposeRunners", true);
                settings.SaveSetting("StopOnError", true);
                settings.SaveSetting("DebugAgent", false);
                settings.SaveSetting("DebugTests", false);

                // Create a simple test package - one assembly, no special settings
                TestPackage package = new TestPackage(@"E:\programming\FaultifyNew\Faultify\Benchmark\Faultify.Benchmark.NUnit\bin\Debug\netcoreapp3.1\test-duplication-0\Faultify.Benchmark.NUnit.dll");
                package.AddSetting("DisposeRunners", true);
                package.AddSetting("StopOnError", true);
                package.AddSetting("DebugAgent", false);
                package.AddSetting("DebugTests", false);

                // Get a runner for the test package
                using (ITestRunner runner = engine.GetRunner(package))
                {
                    // Run all the tests in the assembly
                    XmlNode testResult = runner.Run(null, TestFilter.Empty);

                    runner.StopRun(true);
                    runner.Unload();
                }
            };

            Console.ReadLine();
        }
```

None of them seem to work and have no effect. How can I make sure the processes fired-up by the runner are closed? They are locking my files which prevents me to use them.'

**Info**
    "NUnit.Engine" Version="3.12.0" 
    netcoreapp3.1

**Files**

[NUnitRunner.zip](https://github.com/nunit/nunit-console/files/5918885/NUnitRunner.zip)

CharliePoole:
The `ISettings` service is provided by the engine for use by runners, wanting to save their settings. The engine does not use any settings you add to the service. Those setting are not used by the engine, whose behavior is entirely under the control of the TestPackage and its settings. However, that didn't cause your problem, since you duplicated the same settings in the package.

In fact, I would normally perform my initial tests of a runner without any setting added to the TestPackage. You should be able to run tests in a simple test assembly using default settings before proceeding further.

[...]

PR #895, opened 2021-02-05
Make sure `AssemblyDefinitions` are disposed after use.

TimonPost:
NUnit does not correctly release the file lock for some of the assemblies. This PR makes sure the `Dispose` functionality is called thereby releasing all the underlying file resources.

### Mono Cecil (Fixed in this PR)
 `Mono.Cecil.AsseblyDefinition` is not disposed of, this locks the assembly until the caller application closes.

- https://github.com/nunit/nunit-console/blob/master/src/NUnitEngine/nunit.engine.core/Services/DriverService.cs#L76
- https://github.com/nunit/nunit-console/blob/master/src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs#L233
- https://github.com/nunit/nunit-console/blob/master/src/NUnitEngine/nunit.engine.core/Drivers/NUnitNetStandardDriver.cs#L84
- https://github.com/nunit/nunit-console/blob/master/src/NUnitEngine/nunit.engine.core/Extensibility/ExtensionAssembly.cs#L37

### CustomAssemblyLoadContext (Not fixed in this PR)

Currently, an assembly is loaded by this [code](https://github.com/nunit/nunit-console/blob/master/src/NUnitEngine/nunit.engine.core/Drivers/NUnitNetCore31Driver.cs#L84). 

```csharp
var assemblyLoadContext = new CustomAssemblyLoadContext(assemblyPath);
_testAssembly = assemblyLoadContext.LoadFromAssemblyPath(assemblyPath);
```

Next, the following private fields are initialized:

```csharp
Assembly _testAssembly;
Assembly _frameworkAssembly;
object _frameworkController;
Type _frameworkControllerType;
```

Those have to be released in order for `CustomAssemblyLoadContext.Unload()` to be able to release its resources. 

_microsoft [article](https://docs.microsoft.com/en-us/dotnet/standard/assembly/unloadability):_
image

### Related Issues
Fixes: #894 

### Work Left for other PR
- Make `CustomAssemblyLoadContext.Unload()` working see [this article](https://docs.microsoft.com/en-us/dotnet/standard/assembly/unloadability). Lets discuss this issue in the related issue.

### DIFF
--- a/src/NUnitEngine/nunit.engine.core/Drivers/NUnitNetCore31Driver.cs
+++ b/src/NUnitEngine/nunit.engine.core/Drivers/NUnitNetCore31Driver.cs
@@ -78,7 +78,7 @@ namespace NUnit.Engine.Drivers
 
             assemblyPath = Path.GetFullPath(assemblyPath);  //AssemblyLoadContext requires an absolute path
             var assemblyLoadContext = new CustomAssemblyLoadContext(assemblyPath);
-
+            
             try
             {
                 _testAssembly = assemblyLoadContext.LoadFromAssemblyPath(assemblyPath);
--- a/src/NUnitEngine/nunit.engine.core/Drivers/NUnitNetCore31Driver.cs
+++ b/src/NUnitEngine/nunit.engine.core/Drivers/NUnitNetCore31Driver.cs
@@ -78,7 +78,6 @@ namespace NUnit.Engine.Drivers
 
             assemblyPath = Path.GetFullPath(assemblyPath);  //AssemblyLoadContext requires an absolute path
             var assemblyLoadContext = new CustomAssemblyLoadContext(assemblyPath);
-            
             try
             {
                 _testAssembly = assemblyLoadContext.LoadFromAssemblyPath(assemblyPath);
--- a/src/NUnitEngine/nunit.engine.core/Drivers/NUnitNetStandardDriver.cs
+++ b/src/NUnitEngine/nunit.engine.core/Drivers/NUnitNetStandardDriver.cs
@@ -81,29 +81,30 @@ namespace NUnit.Engine.Drivers
         {
             var idPrefix = string.IsNullOrEmpty(ID) ? "" : ID + "-";
 
-            var assemblyRef = AssemblyDefinition.ReadAssembly(testAssembly);
-            _testAssembly = Assembly.Load(new AssemblyName(assemblyRef.FullName));
-            if(_testAssembly == null)
-                throw new NUnitEngineException(string.Format(FAILED_TO_LOAD_TEST_ASSEMBLY, assemblyRef.FullName));
+            using (var assemblyRef = AssemblyDefinition.ReadAssembly(testAssembly)) {
+                _testAssembly = Assembly.Load(new AssemblyName(assemblyRef.FullName));
+                if(_testAssembly == null)
+                    throw new NUnitEngineException(string.Format(FAILED_TO_LOAD_TEST_ASSEMBLY, assemblyRef.FullName));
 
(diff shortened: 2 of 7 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/nunit/nunit-console?pullRequest=895) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/nunit/nunit-console?pullRequest=895) before we can accept your contribution.<br/><br/>:x: TimonPost [sign now](https://cla.dotnetfoundation.org/nunit/nunit-console?pullRequest=895)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/nunit/nunit-console?pullRequest=895) it.</sub>

CharliePoole:
@TimonPost As I mentioned on the issue, the entire reason we use `AssemblyDefinition`, which I extracted and modified from `Mono.Cecil`, is to __avoid__ loading the assembly. So while your fix may work for the immediate problem, it bothers me to do it without understanding why it's really necessary. One thing that would help would be to have a test that fails before and passes after your fix. If my `AssemblyDefinition` is holding the assembly loaded, I would prefer to fix that directly in the metadata assembly if possible.

Note that I'm only talking about the AssemblyDefinition part of this. I didn't work on the .NET Core driver, so I'll leave it to @ChrisMaddock to comment on that.

TimonPost:
The `AssemblyDeffinitions` in the following files are only used locally in method scope, therefore I think it's perfect to dispose of them directly after use:

NUnitNetStandardDriver.cs
DriverService.cs 
RuntimeFrameworkService.cs


The `AssemblyDeffinition`  in the `ExtensionAssembly.cs` is a bit different since it is stored as a local variable, and exposes public member to request access to various parts of this `AssemblyDefinition`. I am not sure if it is the correct way to implement `Disposable` for this class. Any ideas?

CharliePoole:
I've been wanting to get rid of ExtensionAssembly, but that's probably a bigger job than is called for by this issue. I agree with the approach you have taken by making it disposable, but we need to make sure that the code using it actually calls Dispose. I'll do a more in-depth code review tomorrow ... well today, but in the morning. :smile:

TimonPost:
Okay, I think it might be a nice idea to separate the `AssemblyDefinition` issue and `CustomAssemblyLoadContext` into two PR's since resolving the `CustomAssemblyLoadContext` issue sounds a bit more difficult.

CharliePoole:
I like the use of StopService - that's what it's for. :smile: 

Since you plan to modify this PR, post a comment on it when it's ready for review and we'll see it.

TimonPost:
I pushed the changes which introduce `StopService`, also every using statement that is implemented. Unfortunately, those using statements cause code to be intended which makes it look like more changes than it actually is. I think it is ready for review. Removed the WIP status.

CharliePoole:
Going back to bed now. :wink: I'll go through the code tomorrow.

CharliePoole (approved):
This looks good to me. You'll need one more committer to approve it before we merge. Thanks for your work!

ChrisMaddock:
This looks good - thanks for digging into this Timon! Will discuss the remaining problem on the issue.

@CharliePoole - this sort of issue is an interesting argument towards including the Mono.Cecil files directly in the engine, which didn't really cross my mind when we were considering it before. Maybe something to consider for the future. 🙂

CharliePoole:
@ChrisMaddock Not sure why you say that about including directly. The same issue exists no matter how the assembly is used.

PR #896, opened 2021-02-06
Implement Package Tests for NUnit Console Runner

CharliePoole:
Fixes #852 and Fixes #892

NOTE: I inadvertently kept working on the issue-852 branch, which should have been a separate PR. However #852 was a very small change, so I'm leaving it here. 

At this point, I'm creating the PR so we can discuss whether further work is needed and, if so, whether to keep going on this PR or plan to do it in a separate one.

INCLUDED:

1. Package content checking for the .NET Core console runner, which was previously excluded.

2. Restructured our scripts to make use of parameterized setup and tasks. This is a big change but was required in order to port my package testing code from the TestCentric GUI.

3. Added testing for five console packages: NuGet for .NET FX, NuGet for .NET Core, Chocolatey, Zip and MSi. There are currently four test runs for each .NET FX package and three for the .NET Core package. The tests are not extensive but they verify that the executable can run without crashing and that the top-level result counters match what is expected. The .NET FX tests include running a `.nunit` project file.

4. Added a .NET 4.0 build of mock-assembly so it could be used in one of the .NET FX tests.

5. Added an NUnit V2 build of mock-assembly for use in the tests, but ran into an issue and commented out that test for now.

NOT INCLUDED:

1. Tests of .nunit project under .net core - or of any other extensions - since no extension builds exist for that runtime.

2. No test of NUnit V2 test execution under .NET FX. The test actually works for all builds but the zip. The zip build is able to load all extensions __except__ the V2 framework driver because that extension is needed by the agent, not the engine. It looks like the zip creation needs to be substantially revised so I'm putting that off for another issue.

3. The package test code has the ability to set different levels of testing but this is not yet used. All tests are run on every build.

5. There is no summary report of test results. With 4 packages and 3 or four tests each, I was sometimes confused, so I think this may be a useful extension.

6. No automatic package uploading or release. That's a separate issue.

NOTE: there's currently commented-out code that's used in the GUI version of the scripts. They relate to points (2), (3) and (5). I prefer to leave the code there for when I take up those issues.

PR #902, opened 2021-02-26
Implement package tests

### LINKED ISSUE #892: Implement Package tests for engine and console runner
This will initially be a set of tests implemented in the cake script itself. Since the script is c#, we can port to a test assembly later if we want to test more deeply. However, I've been testing this way in three or four projects and haven't seen any big limitations so far.

CharliePoole:
Fixes #892 (originally #852)

This PR replaces #896, with a simpler implementation more in keeping with the existing, primarily procedural, structure of the build script. It contains only the code needed to do what it does, with no "hooks" in place for further expansion.

INCLUDED

1. Added testing for five console packages: NuGet for .NET FX, NuGet for .NET Core, Chocolatey, Zip and MSi. There are currently four test runs for each .NET FX package and three for the .NET Core package. The tests are not extensive but they verify that the executable can run without crashing and that the top-level result counters match what is expected.

2.  The tests of the MSI and ZIP packages include running a `.nunit` project file. The required extension is already present in those packages.

3. Added a .NET 4.0 build of mock-assembly so it could be used in one of the .NET FX tests.

4. A summary report of test results for each package tested is displayed.

NOT INCLUDED:

1. Tests of .nunit project under any other packages, which would require a separate installing the extension. This is reserved for a separate issue if desired.

2. No other extensions are currently tested.

3. No automatic package uploading or release. That's a separate issue.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -373,7 +373,7 @@ Task("CreateImage")
         CopyDirectory(BIN_DIR, CURRENT_IMG_DIR + "bin/");
     });
 
-Task("PackageNuGet")
+Task("BuildNuGetPackages")
     .Description("Creates NuGet packages of the engine/console")
     .IsDependentOn("CreateImage")
     .Does(() =>
@@ -437,7 +437,13 @@ Task("PackageNuGet")
         });
     });
 
-Task("PackageChocolatey")
+Task("TestNugetPackages")
+    .Does(() =>
+    {
+
+    });
+
+Task("BuildChocolateyPackages")
     .Description("Creates chocolatey packages of the console runner")
     .Does(() =>
     {
@@ -498,6 +504,12 @@ Task("PackageChocolatey")
             });
     });
 
+Task("TestChocolateyPackage")
+    .Does(() =>
+    {
+
+    });
+
 //////////////////////////////////////////////////////////////////////
 // PACKAGE COMBINED DISTRIBUTIONS
 //////////////////////////////////////////////////////////////////////
@@ -533,7 +545,7 @@ Task("CreateCombinedImage")
     }
 });
 
-Task("PackageMsi")
+Task("BuildMsiPackage")
 .IsDependentOn("CreateCombinedImage")
 .Does(() =>
 {
@@ -549,7 +561,13 @@ Task("PackageMsi")
         );
 });
 
-Task("PackageZip")
+Task("TestMsiPackage")
+    .Does(() =>
+    {
+
+    });
+
+Task("BuildZipPackage")
 .IsDependentOn("CreateCombinedImage")
 .Does(() =>
 {
@@ -568,6 +586,12 @@ Task("PackageZip")
     var zipPath = string.Format("{0}NUnit.Console-{1}.zip", PACKAGE_DIR, productVersion);
     Zip(ZIP_IMG, zipPath);
 });
+
+Task("TestZipPackage")
+    .Does(() =>
+    {
+
+    });
 
 Task("InstallSigningTool")
     .Does(() =>
@@ -624,8 +648,8 @@ Task("SignPackages")
         }
     });
 
-Task("CheckPackages")
-    .Description("Check content of NuGet packages")
+Task("CheckPackageContent")
+    .Description("Check content of all the packages we build")
     .Does(() =>
     {
         CheckAllPackages();
@@ -825,14 +849,26 @@ Task("Test")
     .IsDependentOn("TestEngine")
     .IsDependentOn("TestConsole");
 
+Task("BuildPackages")
+    .Description("Builds all packages for distribution")
+    .IsDependentOn("BuildNuGetPackages")
+    .IsDependentOn("BuildChocolateyPackages")
+    .IsDependentOn("BuildMsiPackage")
+    .IsDependentOn("BuildZipPackage");
+
+Task("TestPackages")
+    .Description("Tests the packages")
+    .IsDependentOn("TestNugetPackages")
+    .IsDependentOn("TestChocolateyPackage")
+    .IsDependentOn("TestMsiPackage")
+    .IsDependentOn("TestZipPackage");
+
 Task("Package")
-    .Description("Packages the engine and console runner")
+    .Description("Builds and tests all packages")
     .IsDependentOn("CheckForError")
-    .IsDependentOn("PackageNuGet")
-    .IsDependentOn("PackageChocolatey")
-    .IsDependentOn("PackageMsi")
-    .IsDependentOn("PackageZip")
-    .IsDependentOn("CheckPackages");
+    .IsDependentOn("BuildPackages")
+    .IsDependentOn("CheckPackageContent")
+    .IsDependentOn("TestPackages");
 
 Task("Appveyor")
     .Description("Builds, tests and packages on AppVeyor")
--- a/build.cake
+++ b/build.cake
@@ -1,5 +1,6 @@
 #load ci.cake
 #load package-checks.cake
+#load package-tests.cake
 
 // Install Tools
 #tool NuGet.CommandLine&version=5.3.1
@@ -440,7 +441,9 @@ Task("BuildNuGetPackages")
 Task("TestNugetPackages")
     .Does(() =>
     {
+        new NuGetNetFXPackageTester(Context, productVersion).RunTests();
 
+        new NuGetNetCorePackageTester(Context, productVersion).RunTests();
     });
 
 Task("BuildChocolateyPackages")
@@ -507,7 +510,7 @@ Task("BuildChocolateyPackages")
 Task("TestChocolateyPackage")
     .Does(() =>
     {
-
+        new ChocolateyPackageTester(Context, productVersion).RunTests();
     });
 
 //////////////////////////////////////////////////////////////////////
@@ -564,7 +567,7 @@ Task("BuildMsiPackage")
 Task("TestMsiPackage")
     .Does(() =>
     {
-
+        new MsiPackageTester(Context, version).RunTests();
     });
 
 Task("BuildZipPackage")
@@ -590,7 +593,7 @@ Task("BuildZipPackage")
 Task("TestZipPackage")
     .Does(() =>
     {
-
+        new ZipPackageTester(Context, productVersion).RunTests();
     });
 
 Task("InstallSigningTool")
--- a/build.cake
+++ b/build.cake
@@ -107,7 +107,7 @@ Setup(context =>
             else
                 suffix += "-" + System.Text.RegularExpressions.Regex.Replace(branch, "[^0-9A-Za-z-]+", "-");
 
-            // Nuget limits "special version part" to 20 chars. Add one for the hyphen.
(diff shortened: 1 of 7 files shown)


package-tests.cake
```diff
@@ -0,0 +1,312 @@
...
+        var reporter = new ResultReporter(PackageName);
+
+        foreach (var packageTest in PackageTests)
+        {
+            var resultFile = _outputDir + "TestResult.xml";
+            // Delete result file ahead of time so we don't mistakenly
+            // read a left-over file from another test run. Leave the
+            // file after the run in case we need it to debug a failure.
+            if (_context.FileExists(resultFile))
+                _context.DeleteFile(resultFile);
```
ChrisMaddock, line 74: Would this also delete the results files from the main nunit-engine unit test run? Or it outputDir something else here?

package-tests.cake
```diff
@@ -0,0 +1,312 @@
...
+            new ExpectedResult("Failed")
+            {
+                Total = 37,
+                Passed = 23,
+                Failed = 5,
+                Warnings = 0,
+                Inconclusive = 1,
+                Skipped = 7
+            }));
+
```
ChrisMaddock, line 137: Suggestion - how about extracting these to static objects called "StandardResult" and "TwoAssemblyResult" or something, so it's only in one place? Just thinking it could become a lot to update whenever a new test is added to mock assembly, if these package tests grow more.

package-tests.cake
```diff
@@ -0,0 +1,312 @@
...
+            // Administrative install doesn't copy these files to
+            // their final destination, so we do it.
+            _context.CopyFiles(
+                PackageBinDir + "*.dll",
+                PackageBinDir + "agents/net20/");
+            _context.CopyFiles(
+                PackageBinDir + "*.dll",
+                PackageBinDir + "agents/net40/");
+        }
+    }
```
ChrisMaddock, line 286: I like this method! As a future enhancement, it might be nice to do a `nuget install` and `choco install` of those packages in a similar way, and check the final structure.

package-tests.cake
```diff
@@ -0,0 +1,312 @@
...
+        // Msiexec does not tolerate forward slashes!
+        string package = PackageUnderTest.ToString().Replace("/", "\\");
+        string testDir = PackageInstallDirectory.Replace("/", "\\");
+        int rc = _context.StartProcess("msiexec", $"/a {package} TARGETDIR={testDir} /q");
+        if (rc != 0)
+            Console.WriteLine($"  ERROR: Installer returned {rc.ToString()}");
+        else
+        {
+            // Administrative install doesn't copy these files to
+            // their final destination, so we do it.
```
ChrisMaddock, line 278: Would you mind explaining what this bit is doing? I'm not sure I quite understand what the file copying is making up for. Does msiexec now unpack properly, in this situation?

package-tests.cake
```diff
@@ -0,0 +1,312 @@
...
+    protected override string PackageName => $"NUnit.Console-{_packageVersion}.msi";
+    protected override string PackageInstallDirectory => _packageDir + "test/msi/";
+    protected override string PackageBinDir => PackageInstallDirectory + "NUnit.org/nunit-console/";
+
+    protected override void CreatePackageInstallDirectory()
+    {
+        // Msiexec does not tolerate forward slashes!
+        string package = PackageUnderTest.ToString().Replace("/", "\\");
+        string testDir = PackageInstallDirectory.Replace("/", "\\");
+        int rc = _context.StartProcess("msiexec", $"/a {package} TARGETDIR={testDir} /q");
```
ChrisMaddock, line 272: I'm not sure - but I wonder if this will throw and terminate the script if msiexec cannot be found? Should we instead try/catch, show an error/fail the tests but allow the other tests to continue running?

I checked the Cake docs to try and find out, but it's not documented unfortunately.

test-results.cake
```diff
@@ -0,0 +1,182 @@
...
+
+    private string GetAttribute(XmlNode node, string name)
+    {
+        return node.Attributes[name]?.Value;
+    }
+
+    private int IntAttribute(XmlNode node, string name)
+    {
+        string s = GetAttribute(node, name);
+        return s == null ? 0 : int.Parse(s);
```
ChrisMaddock, line 61: Should we throw an exception instead, if `s == null`? Just thinking that zero could be a valid value - and these attributes should always be present, I think?

test-results.cake
```diff
@@ -0,0 +1,182 @@
...
+
+        foreach (var error in Errors)
+            Console.WriteLine(error);
+
+        Console.WriteLine(Errors.Count == 0
+            ? "   SUCCESS: Test Result matches expected result!"
+            : "\n   ERROR: Test Result not as expected!");
+    }
+
+    private void ReportMissingFiles()
```
ChrisMaddock, line 114: I'm a little confused by what this method's doing sorry. It looks to me like it's recording errors for all the failed test suites, but if I'm right about that (I suspect I'm not!), then I don't understand why it's named `ReportMissingFiles`. Do you think it might be worth commenting this method?

test-results.cake
```diff
@@ -0,0 +1,182 @@
...
+            {
+                string message = suite.SelectSingleNode("reason/message")?.InnerText;
+                Errors.Add($"     {message}");
+            }
+        }
+    }
+
+    private void CheckCounter(string label, int expected, int actual)
+    {
+        if (expected > 0 && expected != actual)
```
ChrisMaddock, line 136: I think this ties into my comment above on zero being a valid value here, I'm not sure expected needs to be > 0, does it?

ChrisMaddock (approved):
Hi Charlie,

Looks good! Thanks for breaking things up, this was much easier to review. I've left a a few comments throughout, all minor things - see what you think to them whether it's worth doing anything about. 🙂

package-tests.cake
CharliePoole: My 'outputDir' is basically bin\Release. I think that's also where our own test results are saved, but we already overwrite them so I didn't figure it was a problem. If we wanted to, we could come back and save __every__ result file to a different place but I'd rather do that as a small, separate PR.

package-tests.cake
CharliePoole: We already have more than two sets. Each derived tester can add tests. The msi and zip testers add more test cases which the nuget and chocolatey runs cannot handle.

I agree with you  that it would be nicer to have all these in one place or at least in one file but I couldn't think of a clean way
to do that. Basically, it seems to require two matching hierarchies. That calls for the Bridge pattern, which seems like a bit of overkill for what is basically a script.

Again, I'd prefer to let this idea percolate rather than trying to solve it all at once. If you like, we can create "idea" issues for some of this stuff.

test-results.cake
CharliePoole: You're right in principle. I had a vague memory that we sometimes omitted the attribute in the past, but it may no longer apply. If you want a true check, I would use -1 because I don't want the constructor of the ActualResult to throw. A -1 would cause an error to be generated when the TestReport compared values anyway.

test-results.cake
CharliePoole: What it looks at is not all test suites, but all those that may occur at or above the assembly level. Project and Assembly are self-explanatory, while Unknown is what we get if the file is either not found or is of an unknown type. (BTW I've tested all these cases manually or I wouldn't have known that myself.) Within those suites, we only look at those that failed __themselves__ (not due to a child test failing) and display the error message.

Without the errors, you can't figure out why all your actuals are zero. With it you get to see messages like "File not found", "Not a supported extension", "Invalid XML", etc. without scrolling back to look at the actual test run or in some cases running the debugger. It saved me a lot of time working on this.

I'll write all that in a comment.

test-results.cake
CharliePoole: Slightly different here. However, it should be >= 0 because ExpectedValue uses -1 to mean "this doesn't need to be checked." If we wanted to run one of our __actual__ tests as a package test, I'd just require it to pass and wouldn't check all the counts because it would require too much maintenance.

CharliePoole (commented):
See my inline comments. I'll make the changes where indicated and push another commit.

package-tests.cake
ChrisMaddock: Ah ok. Didn't realise we overwrote them already - fair enough, separate issue!

package-tests.cake
ChrisMaddock: Cool, sounds good!

package-tests.cake
ChrisMaddock: Not sure about Windows - but presumably it wouldn't be on Linux/OSX? Seems sensible to me. 

When I tried to run the previous PR, I thought missing msiexec was the reason it failed, but I think it was missing wix instead. Haven't been able to pull this one yet to recheck, I'm afraid!

test-results.cake
ChrisMaddock: Great, thanks! 😄 Makes complete sense.

package-tests.cake
ChrisMaddock: Actually, I guess choco is Windows-only too, right? Maybe we should just add IsRunningOnWindows to these two tasks, so people could still theoretically run the nuget tests on Linux, if required.

test-results.cake
CharliePoole: I tried -1 here and got an error. I think that's why I used zero in the first place. This time I dug deeper and found that it's an error in the engine. It only affects `test-run` and Project type `test-suilte` elements, which is why it has gone unnoticed. I'll file a separate issue.

I'll leave in the workaround for now.

CharliePoole: See issue #904

package-tests.cake
CharliePoole: I had assumed we were running the packaging on Linux, but I just checked and see we aren't. I'll run some tests locally to see if it builds and if not I'll prevent the msi and chocolatey stuff from executing.

I had assumed we were testing this stuff under Azure but it appears that we use the Package target only on the windows build. Since I'm not an Azure guy, I think any fix to that (if you want one) should be separate.

build.cake
```diff
@@ -825,14 +853,26 @@ Task("Test")
...
+Task("BuildPackages")
+    .Description("Builds all packages for distribution")
+    .IsDependentOn("BuildNuGetPackages")
+    .IsDependentOn("BuildChocolateyPackages")
+    .IsDependentOn("BuildMsiPackage")
+    .IsDependentOn("BuildZipPackage");
+
+Task("TestPackages")
+    .Description("Tests the packages")
+    .IsDependentOn("TestNugetPackages")
```
ChrisMaddock, line 865: Niggle I've just spotted - can we case the `BuildNuGetPackages` and `TestNugetPackages` targets consistently? I'm not too fussy which way round - `NuGet` would be consistent with the branding and Cake Methods etc.

ChrisMaddock:
Thanks Charlie. By all means merge this in once the last couple of bits are in - unless you wanted to wait for Mikkel to take a look over as well.

build.cake
CharliePoole: Fixed!

CharliePoole:
I'll commit a few trivial changes from the review and then merge.

PR #907, opened 2021-03-01
Remove SettingsService

### LINKED ISSUE #873: Remove UserSettings From Engine
In testcentric/testcentric-gui#580 we agreed that both projects would drop user settings from their engines for compatibility. Handling settings is application-specific. This change is breaking.

CharliePoole:
Fixes #873


src/NUnitEngine/EngineApiVersion.cs
```diff
@@ -4,5 +4,5 @@
 
 [assembly: AssemblyProduct("NUnit Engine API")]
 [assembly: AssemblyVersion("3.0.0.0")]
-[assembly: AssemblyInformationalVersion("3.13.0")]
+[assembly: AssemblyInformationalVersion("4.0.0")]
 [assembly: AssemblyFileVersion("3.13.0")]
```
ChrisMaddock, line 8: Think you've missed a version here. 🙂

src/NUnitEngine/nunit.engine.tests/Services/ServiceManagerTests.cs
```diff
@@ -9,63 +9,63 @@ namespace NUnit.Engine.Services.Tests
...
+            IService service = _serviceManager.GetService(typeof(FakeService2));
+            Assert.That(service, Is.SameAs(_service2));
         }
 
-        [Test]
-        public void AccessServiceByInterface()
-        {
-            IService service = _serviceManager.GetService(typeof(ISettings));
-            Assert.That(service, Is.SameAs(_settingsService));
-        }
```
ChrisMaddock, line 69: Did you mean to remove this test, rather than refactor in the same way as above?

src/NUnitEngine/nunit.engine/Services/SettingsService.cs
```diff
@@ -1,50 +0,0 @@
...
-{
-    /// <summary>
-    /// Summary description for UserSettingsService.
-    /// </summary>
-    public class SettingsService : SettingsStore, IService
-    {
-        private const string SETTINGS_FILE = "Nunit30Settings.xml";
-
-        public SettingsService(bool writeable)
-            : base(Path.Combine(NUnitConfiguration.ApplicationDirectory, SETTINGS_FILE), writeable) { }
```
ChrisMaddock, line 17: I think `NUnitConfigurations.AppDirectory` can also be removed now - it was only used to store settings, I think? 🙂

src/NUnitEngine/EngineApiVersion.cs
ChrisMaddock: I think there's also a version stored in the Cake file, which should be updated in sync with these?

src/NUnitEngine/nunit.engine.tests/Services/ServiceManagerTests.cs
CharliePoole: I realized it duplicated a check made in the InitializeServices test.

src/NUnitEngine/EngineApiVersion.cs
CharliePoole: Your right that I missed one. I'll fix.

However, the cake file was updated when I created dev-4.0 so doesn't need to be changed here.

PR #911, opened 2021-03-03
Drop NUnitConfiguration.cs

CharliePoole:
Just a small refactoring. The NUnitConfiguration class was left with only one property accessed from only one place.

OK to just merge and delete branch if it looks OK.


ChrisMaddock:
Think this might be targeting the wrong branch Charlie, could you redo? Don’t think that can be changed from within a PR.

CharliePoole:
Will do. At least the package test caught it!

PR #912, opened 2021-03-03
Drop NUnitConfiguration class

CharliePoole:
@ChrisMaddock Retrying the PR to merge into the 4.0 branch.

I initially tried simply deleting the PR and branch and re-pushing the same branch id. But when I created a PR it just restored the deleted one. So an entirely new branch was needed.

I'm afraid this two-branch approach may lead to a lot of problems unless there's a way to use a hook to default the target differently for a certain branch naming pattern.


mikkelbu (approved):
LGTM. As far as I can tell not much is lost by not caching the result.

mikkelbu:
> @ChrisMaddock Retrying the PR to merge into the 4.0 branch.
> 
> I initially tried simply deleting the PR and branch and re-pushing the same branch id. But when I created a PR it just restored the deleted one. So an entirely new branch was needed.
> 
> I'm afraid this two-branch approach may lead to a lot of problems unless there's a way to use a hook to default the target differently for a certain branch naming pattern.

I think you can just change the target of the PR as described here - https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/changing-the-base-branch-of-a-pull-request

CharliePoole:
@mikkelbu I looked at that and thought it was just for the title. I'll try it if the need arises again.

ChrisMaddock:
Thanks Mikkel - didn't know about that feature. 🙂

Issue #909, opened 2021-03-02
Can we use "pluggable agents" to replace native in-process execution?


CharliePoole:
@ChrisMaddock @mikkelbu @rprouse @jnm2 

I'm calling this a question, but I mean it as a question we are asking ourselves.

Here's a rough outline of what I want to try with some spike code. Some of this is new ground so it would be nice to have all our brains on it.

FIRST, here's how things generally work. Except as noted, they are the same in both the current NUnit engine and the testcentric fork. Also, for clarity, I've simplified things a lot here.

1. When the user asks the engine for a runner, passing a package, it always gets a `MasterTestRunner`.
2. That runner is initially "bare" without any lower-level runners to back it up. But as soon as the tests need to be loaded, `MasterTestRunner` uses various services to fill in the appropriate internal runner (`ITestEngineRunner`).
3. If the `ITestRunner` is one of the in-process runners, it finds a driver (`IFrameworkDriver`) that knows how to run the tests. On platforms that support `AppDomains`, it creates one in which to run the test.
4. OTOH. if the runner is a `ProcessRunner`, things work differently. `ProcessRunner` uses the `TestAgency` service to get an agent. Although we usually just call them "agents", think of them as "process proxy agents" for this discussion, because the serve as an in-process proxy for the runner executing in a separate process.
5. `TestAgency` launches that process and waits for it to establish communication. Once it does it returns the proxy to the `ProcessRunner`.
6. The `ProcessRunner` issues commands to the remote runner and the remote runner goes through steps 1 to 3 itself.

The current GUI engine deviates from the above by
* Not supporting in-process execution - so step 3 never takes place.
* Supporting "pluggable agents", i.e. agents as extensions. However, they are still always proxies for a remote process, which they launch.

Because the GUI engine has already taken those steps toward generalized agents, I'm using that engine for my spike.

SECOND, here's how I'd like to see agents work...

1. Continue to use a `MasterTestRunner` as we now do.
2. Rather than holding internal  runners directly, `MasterTestRunner` will acquire an agent. A new service will be needed, which knows what agents are available and can decide which ones to use based on settings in the package.
3. An agent will return an internal runner on request. The engine will not actually know where that runner is located - whether in-process, in a separate process or on a separate machine.
4. Once `MasterTestRunner has a set of agents with runners, it will continue to call the runners directly, as it now does.
5. I'll wait to see what the code tells us, but I think we'll see some structural changes as a consequence of all this:
[...]

PR #913, opened 2021-03-04
Remove DomainUsage options

### LINKED ISSUE #860: Eliminate DomainUsage setting
In a separate discussion, @ChrisMaddock expressed the desire to eliminate user control of how the runner and engine use AppDomains. I've been through the codebase to look at the impact and I can see it would enormously simplify a lot of code, particularly the test code. This issue is for the purpose of opening discussion on making the change. I'm flagging it as an idea for discussion by the @nunit/engine-team.

Some notes...

1. We can't eliminate AppDomains themselves for some platforms, they are needed. However, we can make their use an internal implementation detail outside of the control of users.
2. The command-line `--domain` option would be dropped.
3. The `DomainUsage` enumeration would be dropped.
3. Both the saved setting and the package setting for DomainUsage would disappear.
4. All the tests that utilize DomainUsage would be eliminated.
5. I believe that the nunit project loader would simply ignore the domain setting in `.nunit` files, but this would have to be tested.
6. The nunit project loader should probably be modified at some point.
7. This is a breaking change for the console, engine and nunit project loader extension.

FWIW, I favor doing this as part of a 4.0 release that resolves all the breaking issues we currently have, without necessarily adding a lot of new functionality. The new stuff can come in 4.1, 4.2, etc.

@ChrisMaddock Hope you don't mind that I added a new "breaking" label. :smile:

CharliePoole:
Fixes #860 Prerequisite to #909


ChrisMaddock:
Hi Charlie - I'll try and take a look through this over the weekend, unless anyone else beats me too it! 👍

ChrisMaddock:
All looks good to me. 🙂

CharliePoole:
Great. I'll get on to the next step.

Issue #789, opened 2020-06-21
DirectoryFinder.GetDirectories throws for a path with the drive specified


ChrisMaddock:
`DirectoryFinder.GetDirectories()` is a method used when locating engine extensions. (See [locating addins](https://github.com/nunit/docs/wiki/Engine-Extensibility#locating-addins) in the docs, for background) It searches file paths for wildcard characters, and returns all possible directories that could match the wildcard.

It currently throws when passed a path containing a drive, e.g. `C:\Users\Chris\mock-event-listener.dll`. It should be able to handle such paths in the normal way.

Tagging this as a nice, approachable issue for a first time contributor. If I was tackling this, my first step would be to create a unit test which reproduces the issue, and look at why the exception is thrown, and how this could be prevented.

Below is the stack trace currently thrown:

```
System.ArgumentException
  HResult=0x80070057
  Message=Second path fragment must not be a drive or UNC name.
Parameter name: path2
  Source=mscorlib
  StackTrace:
   at System.IO.Path.InternalCombine(String path1, String path2)
   at System.IO.FileSystemEnumerableIterator`1.GetFullSearchString(String fullPath, String searchPattern)
   at System.IO.FileSystemEnumerableIterator`1..ctor(String path, String originalUserPath, String searchPattern, SearchOption searchOption, SearchResultHandler`1 resultHandler, Boolean checkHost)
   at System.IO.DirectoryInfo.InternalGetDirectories(String searchPattern, SearchOption searchOption)
   at NUnit.Engine.Internal.DirectoryFinder.ExpandOneStep(IList`1 dirList, String pattern) in C:\Users\Chris\Documents\git\nunit-console\src\NUnitEngine\nunit.engine.core\Internal\DirectoryFinder.cs:line 147
   at NUnit.Engine.Internal.DirectoryFinder.GetDirectories(DirectoryInfo baseDir, String pattern) in C:\Users\Chris\Documents\git\nunit-console\src\NUnitEngine\nunit.engine.core\Internal\DirectoryFinder.cs:line 76
   at NUnit.Engine.Internal.DirectoryFinder.GetFiles(DirectoryInfo baseDir, String pattern) in C:\Users\Chris\Documents\git\nunit-console\src\NUnitEngine\nunit.engine.core\Internal\DirectoryFinder.cs:line 102
[...]

PR #914, opened 2021-03-07
ExtensionService/DirectoryFinder can handle absolute paths

x789:
This fixes issue #789. Absolute paths are now properly handled and the logic is covered by unit-tests,

If the `ExtensionService` runs on Windows and a Unix-style absolute path is found (and vice-versa), an `NUnitEngineException` is thrown that the user informs that the path cannot be resolved.
I think this is better than to silently ignore the corresponding entry. (But I can change that if desired...)


src/NUnitEngine/nunit.engine.core/Internal/PathUtils.cs
```diff
@@ -166,6 +166,56 @@ public static string Combine(string path1, params string[] morePaths)
             return result;
         }
 
+        /// <summary>
+        /// Returns a value that indicates whether the specified file path is absolute or not on Windows operating systems.
+        /// </summary>
+        /// <param name="path">Path to check</param>
+        /// <returns><see langword="true"/> if <paramref name="path"/> is an absolute or UNC path; otherwhise, false.</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="path"/></exception>
```
ChrisMaddock, line 174:
```suggestion
        /// <exception cref="ArgumentNullException"><paramref name="path"/> is <see langword="null"/>.</exception>
```

Just to match the format of your other doc-comments. 🙂

src/NUnitEngine/nunit.engine.core/Services/ExtensionService.cs
```diff
@@ -325,35 +338,57 @@ private int ProcessAddinsFiles(IDirectory startDir, bool fromWildCard)
...
+                }
+                else
+                {
+                    return new Backports.Tuple<IDirectory, string>(_fileSystem.GetDirectory(System.IO.Path.GetDirectoryName(path)), System.IO.Path.GetFileName(path));
                 }
             }
+            else if (!IsPathRelative(path))
+            {
+                throw new NUnitEngineException($"The path '{path}' is not relative.");
+            }
```
ChrisMaddock, line 382: I'm a little confused by this clause sorry. This should never be hit, right - a path that's neither fully qualified nor relative? 

An exception thrown here could be displayed to the end user, so if this being thrown would be an indication of an error in the code, how about we go with something more descriptive? Maybe something like "Error when discovering extensions for path: {path}. Unexpected path format."

src/NUnitEngine/nunit.engine.core/Internal/AddinsFileReader.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            var result = new List<string>();
+            using (var reader = new StreamReader(stream))
+            {
+                for(var line = reader.ReadLine(); line != null; line = reader.ReadLine())
+                {
+                    line = line.Split(new char[] { '#' })[0].Trim();
+                    if (line != string.Empty)
+                    {
+                        result.Add(line.Replace('\\', '/'));
```
ChrisMaddock, line 48: Observation: this is a change in logic I think - previously the substitution would only happen on windows systems. I can't think of an issue that would cause - unless someone has a valid unix path with a backslash in...which would be very weird!

Thinking about it, is the substitution even necessary? I would have thought the .NET path classes would have taken care of that.

src/NUnitEngine/nunit.engine.tests/Internal/AddinsFileReaderTests.cs
```diff
@@ -0,0 +1,70 @@
...
+            "# note that an absolute path is allowed, but is probably not a good idea in most cases",
+            "c:\\windows\\absolute\\directory",
+            "/unix/absolute/directory",
+            "\\transform\\backslash\\to\\slash"
+        };
+
+        [Test]
+        public void Inheritance()
+        {
+            Assert.True(typeof(IAddinsFileReader).IsAssignableFrom(typeof(AddinsFileReader)));
```
ChrisMaddock, line 36:
```suggestion
            Assert.That(typeof(IAddinsFileReader), Is.AssignableFrom(typeof(AddinsFileReader)));
```

How about this format, for a more specific error message? 🙂

src/NUnitEngine/nunit.engine.tests/Internal/AddinsFileReaderTests2.cs
```diff
@@ -0,0 +1,69 @@
...
+using NUnit.Framework;
+using System.IO;
+
+namespace NUnit.Engine.Internal.Tests
+{
+    /// <summary>
+    /// Tests the implementation of <see cref="AddinsFileReader"/>.
+    /// </summary>
+    /// <remarks>All tests in this fixture modify the file-system. Therefore they need to be marked explicitly to run.</remarks>
+    [TestFixture, Explicit, Category("WritesToDisk"), NonParallelizable]
```
ChrisMaddock, line 14: If I recall correctly, we have a few other tests that write real files. I think we can enable this - my concern with it being explicit is that it would never get run! 😅

src/NUnitEngine/nunit.engine.tests/Internal/DirectoryFinderTests.cs
```diff
@@ -609,6 +613,12 @@ public void GetFiles_PatternIsEmpty()
             Assert.That(() => finder.GetFiles(Substitute.For<IDirectory>(), string.Empty), Throws.ArgumentException.With.Message.Contains(" pattern "));
         }
 
+        [Test]
+        public void Inheritance()
+        {
+            Assert.True(typeof(IDirectoryFinder).IsAssignableFrom(typeof(DirectoryFinder)));
```
ChrisMaddock, line 619: As above on Is.AssignableFrom() 🙂

src/NUnitEngine/nunit.engine.tests/Internal/DirectoryFinderTests.cs
```diff
@@ -455,9 +455,13 @@ public void GetDirectories_PatternIsNull()
...
         {
-            var finder = new DirectoryFinder(Substitute.For<IFileSystem>());
+            var directory = Substitute.For<IDirectory>();
+            var sut = new DirectoryFinder(Substitute.For<IFileSystem>());
+
+            var directories = sut.GetDirectories(directory, string.Empty);
 
-            Assert.That(() => finder.GetDirectories(Substitute.For<IDirectory>(), string.Empty), Throws.ArgumentException.With.Message.Contains(" pattern "));
+            Assert.That(directories, Has.Count.EqualTo(1));
+            Assert.That(directories.First(), Is.EqualTo(directory));
```
ChrisMaddock, line 464: Spotted this one above - out of interest, why the change in behaviour here? Not saying I disagree with it - I just don't understand why? 😄

ChrisMaddock (commented):
Hey @x789 - sorry it's taken a while for me to get to this. This all looks great - thanks for putting the time into such a robust solution! I had a few questions going through where I wasn't sure about the reason behind a couple of changes and a couple of small suggestions, but overall it looks good to go - hopefully we can get it merged soon. 🙂

src/NUnitEngine/nunit.engine.tests/Internal/AddinsFileReaderTests.cs
x789: I tried it but somehow when using `Assert.That` with `Is.AssignableFrom` the test fails. For now, I extended the `Assert.True` with a message to improve readability in case of an error.

But I will try to figure out why
`Assert.True(typeof(IAddinsFileReader).IsAssignableFrom(typeof(AddinsFileReader)));` passes and `Assert.That(typeof(IAddinsFileReader), Is.AssignableFrom(typeof(AddinsFileReader)));` fails.

src/NUnitEngine/nunit.engine.tests/Internal/DirectoryFinderTests.cs
x789: As above... 😉

src/NUnitEngine/nunit.engine.core/Internal/AddinsFileReader.cs
x789: My first thought was that the substitution is harmless on non-Windows systems and that's why I changed the behavior.
But when I read your comment I remembered that macOS uses backslashes to escape special character in paths (e.g. ' '). I will change the behavior back to the original. 🤦‍♂️

src/NUnitEngine/nunit.engine.tests/Internal/DirectoryFinderTests.cs
x789: An empty pattern is used when a directory is specified in the addins-file. I think to express "do not scan for subdirectories",  an empty pattern is a better way than using `null`.

The using code in `ExtensionService` is:
```c#
if (path.EndsWith("/"))
{
  return new Backports.Tuple<IDirectory, string>(_fileSystem.GetDirectory(path), string.Empty);
}
```

src/NUnitEngine/nunit.engine.core/Services/ExtensionService.cs
x789: That exception is thrown when an absolute Unix path is found on a Windows system and vice versa. The exception is intended to help the user isolate the cause of the error.
I had the following scenarios in my mind when I added that additional check:

- A user copies an addins-file with an absolute path from a Windows to a Unix machine. The expected addins would not load because the absolute path is treated as a relative path and will probably not be resolved.
- A user copies an addins-file with an absolute path from a Unix to a Windows machine. The expected addins would not load because the absolute path is treated as a relative path on the current working drive.

I absolutely agree that the exception-message is not useful in this case. It would be better to explain what to do to solve the error.
On the other hand, maybe I should remove the check because it is not really helpful... 🤔

src/NUnitEngine/nunit.engine.tests/Internal/AddinsFileReaderTests2.cs
x789: Removed the 'explicit' attribute and the tests run fine on the build-servers. 😀

CharliePoole:
See  the (sparse) docs for `AssignableFromConstraint`: https://docs.nunit.org/articles/nunit/writing-tests/constraints/AssignableFromConstraint.html

The first argument after `That` isn't a Type, but an instance of the type.

However, I don't really see the point of testing the inheritance structure of a class you are writing yourself.

src/NUnitEngine/nunit.engine.tests/Internal/AddinsFileReaderTests.cs
x789: Thank you @CharliePoole for explaining `Is.AssignableFrom` to me. But as Charlie said, since the point of an inheritance test is questionable, I removed it.

src/NUnitEngine/nunit.engine.tests/Internal/DirectoryFinderTests.cs
x789: Removed the test.

src/NUnitEngine/nunit.engine.core/Internal/AddinsFileReader.cs
x789: I reverted the original behavior.

@ChrisMaddock: You are right there is no difference in behavior on Windows-systems between the [DirectorySeparatorChar](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.directoryseparatorchar?view=net-5.0) and the [AltDirectorySeparatorChar](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.altdirectoryseparatorchar?view=net-5.0). The only advantage I see by replacing the backslash is that you would only have to consider one character during development.

CharliePoole:
@ChrisMaddock When I designed the `.addins` format, I unfortunately didn't document it anywhere. Maybe I need to correct that somewhere! Any idea where it should go in the docs?

Once small item: paths were intended to use a forward slash, as they do in many cross-platform formats. It would be an error to convert back-slash to forward-slash, because there are platforms that allow back-slash in file names. Converting forward-slash to back-slash internally would be OK to do internally if running on a platform that doesn't understand forward slash as a separator, but all the platforms we run on do understand it - which is why many cross-platform apps use it.

x789:
@CharliePoole As far as I know there are two documents that describe `.addins` files. "[Engine Addins Spec](https://docs.nunit.org/articles/nunit/technical-notes/nunit-internals/specs/Engine-Addins-Spec.html#locating-addins)" has a brief description and "[Installing Engine Extensions](https://docs.nunit.org/articles/nunit-engine/extensions/Installing-Extensions.html)" describes it in more detail.

The original code checked if a backslash is the directory-separator and then replaced all backslashes to forward-slashes. This PR introduces an `AddinsFileReader` where the system's directory-separator is replaced by a forward-slash internally. That means backslashes on Linux/macOS/Unix systems will not be changed and therefore the original behavior is reproduced. But on a (hypothetical) system that uses a different character (e.g. a pipe `|`), the new implementation would change that character to a forward-slash without touching any forward- or back-slashes.

@CharliePoole @ChrisMaddock, even though it was intended to only use forward-slashes to separate directories in an `.addins` file, I propose that backslashes are still supported on Windows systems. I think there are some people out there who simply copy&paste a path (or a part of it) from Windows Explorer to an `.addins` file. Additionally, prohibiting a backslash in an `.addins` file would be an incompatible change.
But if you think it is better to throw an exception when a line contains a backslash, let me now and I will change the behavior accordingly.

src/NUnitEngine/nunit.engine.core/Services/ExtensionService.cs
ChrisMaddock: That makes sense. 🙂 

I don't think I have a particularly strong opinion on whether we skip or throw here. I guess one thought is that any such path would get caught as a Missing Directory slightly later on anyway (I think?), which potentially means we're not gaining a lot from this extra check?

Happy to go with whatever you think is best here - you have a better understanding of the details of this part of the codebase than I do! My only thinking is that if we do keep the exception, I think it would be good for it to have a more specific error message.

ChrisMaddock:
> It would be an error to convert back-slash to forward-slash, because there are platforms that allow back-slash in file names. 

@CharliePoole This is the source of my confusion - it looks like that's what the code does. Admittedly only for platforms when a back-slash _isn't_ valid in a file name, but I don't understand why the change is necessary at all.

https://github.com/nunit/nunit-console/pull/914/files#diff-592aaa6543fa280eb2a563a266b72e05cb61293a251565584fc75a112e501313L40-L42

As far as I can see it's only done for the `pattern.IndexOf('/');` line in DirectoryFinder - but I would have thought it would be better for that to instead be `pattern.IndexOf(Path.DirectorySeparatorChar)` and the NUnit code to instead leave paths exactly as the user specified them, and not try and modify anything.

-----

@x789 I agree with your conclusion about allowing Windows users to use backslashes, which I think is the current behaviour in this PR, right?

I'd be happy for us to merge this once we've sorted that exception in the ExtensionService one way or the other. I think you're best placed of all of us to judge the best solution there! 

If you do want to dig further into the slash-replacing that would be great - but appreciate that's pre-existing code to this PR, and not directly related. Leave that one up to you as to how much more your interested in digging in this area! 😄

CharliePoole:
@ChrisMaddock Well, yes, the "if" statement is exactly my point. Also, we are only doing it with an argument passed by value, so the change is thrown away after it's used. The reason you can't use `IndexOf(Path.DirectorySepatorChar)` is that (1) the user may have used the alternate separator and (2) all the separators in a path do not have to be the same!

src/NUnitEngine/nunit.engine.core/Services/ExtensionService.cs
x789: Unfortunately there is no message or exception thrown when an `.addins`-file contains a reference to a non-existing item.

But to always handle that "unresolvable reference" scenario in the same way, I removed the check and the exception. This way we achieve a better readability for developers and users don't wonder why sometimes an exception is thrown for an invalid entry and sometimes not.

x789:
@ChrisMaddock You are right. The code in this PR preserves the original behavior and allows users to use backslashes in paths.

@CharliePoole I'm sorry, but I don't think I understood your point correctly. The modified pattern (where the backslashes are replaced) is required by the subsequent logic inside `DirectoryFinder.GetDirectories`. Therefore, it is irrelevant whether `pattern` was passed into that method by reference or by value.

Replacing slashes makes the code that uses paths easier, because it can rely on forward-slashes.
But in my opinion, I think the cleanest way to handle the paths is to not modify them at all and to use [DirectorySeparatorChar](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.directoryseparatorchar?view=net-5.0) AND [AltDirectorySeparatorChar](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.altdirectoryseparatorchar?view=net-5.0) in any logic that splits or checks for directories. 
A quick peek showed that (at least) `AddinsFileReader`, `DirectoryFinder` and `ExtensionService` need to be adapted. I think it is feasible but I would separate it from this PR.

CharliePoole:
@x789 My comment was directed at something you wrote... about what __users__ might use in an addins file. What I meant was that we should allow users to use whatever they want (but see below) and only make replacements internally as needed.

But now I am having second thoughts. If we are a cross-platform application, then perhaps we should require users to use cross-platform paths. This would be a breaking change, so I'm wondering if we should think about that for the 4.0 release. I have always used forward slashes in `.addins` files for exactly that reason.

@ChrisMaddock what do you think about that for 4.0? I know of other cross-platform apps that do the same thing. We might want to add some validation of the files as well. It would simplify our logic as well.

ChrisMaddock:
Thanks for resolving the last couple of bits @x789! Let's merge this before we go any further, than I'll continue the conversation... 👍

ChrisMaddock:
> What I meant was that we should allow users to use whatever they want (but see below) and only make replacements internally as needed.

So - I think we're all in agreement about allowing users to use whatever they want! The bit that I think is in question is whether we need to make any replacements internally, or if the code should handle the paths in whatever form the user provided them throughout. I think the latter is possible. 🙂 

> If we are a cross-platform application, then perhaps we should require users to use cross-platform paths. This would be a breaking change, so I'm wondering if we should think about that for the 4.0 release. I have always used forward slashes in .addins files for exactly that reason.

I'm not such a fan of this idea. I think we're better to interpret the different paths of all our platforms rather than force our users to only use paths which are valid on all the platforms other users might be running on. The "cross-platform" onus should be on the software rather than the users, in my opinion.

> But in my opinion, I think the cleanest way to handle the paths is to not modify them at all and to use DirectorySeparatorChar AND AltDirectorySeparatorChar in any logic that splits or checks for directories.

@x789 I agree! If you're interested in taking a look, I think that would be a good improvement. 🙂

CharliePoole:
@ChrisMaddock That approach of allowing users to use what they like makes sense. Maybe what's needed is some guidance for people creating extensions and distributing addins files. IMO they ought to be completely compatible, which generally means using relative paths with a forward slash, as we have always done. Of course, you could just say "what's needed is some guidance." :smile_cat:

x789:
I vote to keep the support for absolute paths in v4.0.
Even though platform independence is interesting for the NUnit contributors (since it avoids multiple implementations), it is an implementation detail. I think a user of NUnit simply expects that NUnit can access the file-system as she is used to on the particular operating system.

@ChrisMaddock: I think I will find the time to remove the path-modification. But #915 is still on my list and IMO has a higher priority.

CharliePoole:
I didn't think we were considering removing it!!! But it's definitely an error for someone to use it in an extension they plan to publish and distribute, just as it's bad practice to have hard-coded absolute paths in almost any application.

CharliePoole:
@x789 I've been away from NUnit for a while but I'm helping out now and I was reviewing some work you did a few years back. You managed to get pretty familiar with the engine internals and made a significant contribution. Are you still working in .NET? With NUnit? Would you be interested in helping out further? You can contact me directly by email, which is visible in my profile here.

PR #917, opened 2021-03-13
Replace Gitter badge with Slack invite link

jnm2:
Copy of https://github.com/nunit/nunit/pull/3794.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![AppVeyor Build status](https://ci.appveyor.com/api/projects/status/81uhucr7tlq2kwup/branch/master?svg=true)](https://ci.appveyor.com/project/CharliePoole/nunit-console/branch/master) [![Travis Build Status](https://travis-ci.org/nunit/nunit-console.svg?branch=master)](https://travis-ci.org/nunit/nunit-console) [![Azure Pipelines Build Status](https://nunit.visualstudio.com/NUnit/_apis/build/status/NUnit%20Engine%20&%20Console/NUnit%20Console%20CI?branchName=master)](https://nunit.visualstudio.com/NUnit/_build/latest?definitionId=13&branchName=master) [![NuGet Version and Downloads count](https://buildstats.info/nuget/NUnit.ConsoleRunner)](https://www.nuget.org/packages/NUnit.ConsoleRunner) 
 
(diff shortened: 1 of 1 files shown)


ChrisMaddock:
Great - thanks Joseph!

CharliePoole:
It's done, but as a team member, I have to register my disagreement. Gitter is a platform lots of folks prefer. I'm one. If there had been discussion, I would have said so.

I think that changes, which modify how the team operates, should have a full discussion so we know there is consensus. In fact, that's what we said when we negotiated [this document](https://github.com/nunit/governance/blob/master/decisions.md)...

> For lazy consensus to be effective, we try to allow at least 72 hours before assuming that there are no objections to the proposal. This ensures that everyone is given enough time to read, digest and respond, regardless of their location and time commitments.

jnm2:
Are you talking about the engine team or the core team? The core team mainly uses github team discussions and I don't think I've seen a core team or engine team discussion on Gitter to date, but in fact, framework and engine team members are responsive on Slack.

This PR is not about changing how we operate. It reflects the way we are currently operating for the purpose of making it easier for community folks to get in touch with us. That's the only reason I'm submitting PRs; someone reached out to me on the C# community Discord channel in an effort to talk to someone.

jnm2:
This isn't to say that a discussion shouldn't happen, but as I didn't see a way that this PR modifies how a team operates, I would not be the one to start such a discussion.

jnm2:
This stuff would have been good for me to say in the original PR. Sorry about that, I'll spend more time next time.

CharliePoole:
@jnm2 Sorry not to be clear. I'm talking about the engine team, 'cause that's the nature of the issue we are on here.

I appreciate your noting that you could have explained more. We don't all see the same stuff. I'm not on Slack or Discord and I don't normally get notices for issues on projects where I am not a team member. Even so, based on what your PR __said__ I think it should have had the standard period for comment.

ChrisMaddock:
Hi both, 

This is a one line PR - it's very easy to revert or change if we want to. But I'll take on board the comments about allowing more time for discussion in future.

Let's have that discussion now. Personally I would prefer we start reducing the number of communication channels we use, and would prefer that Slack was a replacement for Gitter, rather than an addition. Whilst the duplication was never ideal, other contributors in the organisation are using Slack more actively than we've ever been on Gitter, and I'd like for this project to be a part of that.

My preference would be that the NUnit Gitter is eventually shut down, but if we want to retain both links side-by-side until that does or does not happen, I think that's a reasonable compromise.

Regardless of the eventual decision - thanks Joseph, for taking the time to make the project more accessible to newcomers. 🙂

CharliePoole:
@ChrisMaddock Based on a comment by @jnm2, my understanding is that the discussion already took place somewhere. Since I'm a team member and wasn't invited to that discussion, that p*ssed me off. Possibly, that was irrationally aggravated by delays in getting my own stuff merged.

Agree 100% with your last line.

Issue #919, opened 2021-03-15
Where should TestPackage be defined?


CharliePoole:
Question for the team for V4...

Currently, `TestPackage` is defined in the `nunit.engine.api` assembly. I'm working on issue #889 to remove some confusion about the constructor for `TestPackage` and simplify some other things I'm working on.

But it seems to me that there's a bigger question... Should `TestPackage` really be defined as a class in this assembly, which is made up almost entirely of interfaces?

So what are the alternatives? I can think of two:

1. Define an `ITestPackage` interface in the api assembly and define `TestPackage` itself in the engine, probably in `engine.core`. That would mean giving the user some new interface (maybe a service) to create test packages.
2. Define an `ITestPackage` interface that exposes all the functionality we need and have the user define it in their application.

Or we could keep doing it as we now are.

Throwing this out there for consideration, since we are planning on changing the engine interface anyway.

ChrisMaddock:
Interesting point. I'm not sure I can come up with a reason to feel particularly strongly either way. 

I agree its inconsistent, however TestPackage is such a fundamental class I almost feel like it's _allowed_ to be inconsistent? From my memory there's only two classes a runner should instantiate - TestPackage and TestEngine (Indirectly through TestEngineActivator).

Other than consistency, I'm not sure I can think of a significant advantage to putting TestPackage behind an Interface - I believe the exposed methods are quite limited, and I can't think of a particular need to mock it in any way. 

My only thinking is that changing this would be a breaking change for even the simplest of runners. No problem at all with making breaking changes in 4.0, but I'm less sure on making such a fundamental one unless we have a significant need to do so.

ChrisMaddock:
Here's a thought - will this be affected at all by our plans to split up the API assembly for extensions to help with the versioning? 

We talked about whether we'd need a nunit.engine.api.code.dll assembly. If we have a concrete TestPackage class, that would be big enough and change enough regularly that we definitely would, and also require all the associated extension.api.dll assemblies to be rereleased to reflect those changes. If we had a super simple ITestPackage interface, it wouldn't.

You might be onto something here, Charlie! 😄

CharliePoole:
@nunit/engine-team This is four years old and I'd like to get something done in the next beta - either close it without action or make a choice and implement it.

I prefer option 1 over 2. Requiring the user to define their own class seems a bit unfriendly. But I'm not adverse to leaving it alone either.

manfred-brands:
[...]

Issue #885, opened 2021-01-30
v4: Redesign API structure/Extension loading


ChrisMaddock:
In the v3 stream, there's been a long standing issue with API versioning that I think is best fixed with a breaking change as part of v4. Issue reference at https://github.com/nunit/nunit-console/issues/826.

Here's the changes I propose to fix this:

- We split the nunit.engine.api.dll assembly into smaller components.
- nunit.engine.api.dll becomes dedicated solely to the API's required by runners to access and utilise the engine. We start to increment the version of this assembly with every new version of the engine.
- We create an API assembly specific to each extension interface, versioned at 1.0, and only incremented when breaking changes to an extension interface need to be made. (Hopefully not until v5.0 - however this structure would allow us to make these changes causing minimum disruption if required.)
- There may be a need for a general extension api assembly to contain functionality shared by all known extension interfaces, also versioned as 1.0. I'm hoping we can architect that out, and avoid the need for it.
- Extensions should start to be packaged with the version of the API assemblies that the reference, rather than relying on the engine to be packaged with all versions.

CharliePoole:
Nice idea. I'll add this to my pluggable agent experimentation.

ChrisMaddock:
Plans to take this further now up for discussion in #948

CharliePoole:
As far as I can tell, everything mentioned here and in linked issues is now implemented so I'm closing with no further action needed.

PR #920, opened 2021-03-17
Eliminate string constructor for TestPackage

### LINKED ISSUE #889: v4: Create single constructor method/structure for a TestPackage
From memory, there are currently multiple constructors for a test package, which create different package sturctures which the agent must handle. We should reduce this to a single method.

CharliePoole:
Fixes #889 

This fix enables us to move forward without having to support multiple constructors for `TestPackage`, simplifying tests and new feature development. It's not intended as the last word on the interface provided by `TestPackage`, which we will need to rethink somewhat for the design of the final 4.0 release. This is covered in issues #885 and #919.

The old string constructor created a package with a file name but no subpackages. The list constructor created a package with no file name and subpackages. Named packages are only used internally by the engine, so that constructor has now been replaced by the `Named()` modifier. Addition of a `params` overload makes the TestPackage constructor work the same whether one or multiple test files are used.


CharliePoole:
Although this is only a refactoring toward 4.0, I've asked for review because of substantive changes to the tests. I'll be happy if someone just looks those over and agrees that what I've done is the way to go. In fact, the tests of what runner is used in various circumstances represent what is currently done. If that's not what we want done in some case, we should flag that as a separate issue.

src/NUnitEngine/nunit.engine.tests/Services/Fakes/FakeProjectService.cs
```diff
@@ -27,7 +27,7 @@ void IProjectService.ExpandProjectPackage(TestPackage package)
             if (_projects.ContainsKey(package.Name))
             {
                 foreach (string assembly in _projects[package.Name])
-                    package.AddSubPackage(new TestPackage(assembly));
+                    package.AddSubPackage(new TestPackage().Named(assembly));
```
ChrisMaddock, line 30: All these changes make me think we should have a `package.AddSubPackage(tests.dll)` method.

Actually, now I've written that - what would you think of that as a replacement public API for `TestPackage().Named()`? So project extensions would instead call that method, rather than creating individual packages separately and putting them onto the TestPackage?

We might need it to be `var subpackage = package.AddSubPackage("tests.dll")` to make it easy to set settings on the subpackages...

What do you think?

src/NUnitEngine/nunit.engine.core/Services/DomainManager.cs
```diff
@@ -296,13 +296,18 @@ public static string GetPrivateBinPath(string basePath, IList<TestPackage> packa
...
+
             List<string> dirList = new List<string>();
             StringBuilder sb = new StringBuilder(200);
 
             foreach( string assembly in assemblies )
             {
                 string dir = PathUtils.RelativePath(
-                    Path.GetFullPath(basePath),
+                    basePath,
                     Path.GetDirectoryName( Path.GetFullPath(assembly) ) );
```
ChrisMaddock, line 311: What's the reason for the changes in this file, sorry? 🙂

ChrisMaddock (commented):
Hi Charlie,

The changes look good - but I'm wondering if we can come up with a more explicit API than the `Named()` method - I feel like whether a package is named or not is a consequence of it's place in the structure, rather than key characteristic.

I think I might have stumbled across an alternative while reviewing this - let me know what you think.

src/NUnitEngine/nunit.engine.tests/Services/Fakes/FakeProjectService.cs
ChrisMaddock: One thought - I guess we would lose the ability to create a standalone subpackage. I don't think there's anywhere that needs to do that outside of NUnit's own tests though, is there? I guess it would force us to come up with a nicer solution for testing... 😅

src/NUnitEngine/nunit.engine.api/TestPackage.cs
```diff
@@ -26,39 +26,40 @@ namespace NUnit.Engine
...
-            ID = GetNextID();
-
-            if (filePath != null)
-            {
-                FullName = Path.GetFullPath(filePath);
-                Settings = new Dictionary<string,object>();
-                SubPackages = new List<TestPackage>();
-            }
+            InitializeSubPackages(testFiles);
         }
```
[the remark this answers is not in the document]
CharliePoole: No, we didn't because it would never have been used for a single string argument - the named constructor would have been used. I added it primarily for ease of use in tests, but then I thought users like ease of use as well. We could have it as the only constructor, but would have to change two or three places where we use the IList constructor. I can go any way you like on this.

src/NUnitEngine/nunit.engine.tests/Services/Fakes/FakeProjectService.cs
CharliePoole: I had `AddSubPackage(string)` briefly. But the main use for adding subpackages is in project loader extensions and they need to add settings as well. So if you just add the subpackage, then you have to use an indexer to get it back and add settings. It's easier if they can create the subpackage, add any needed settings and then add it to the original package.

src/NUnitEngine/nunit.engine.core/Services/DomainManager.cs
CharliePoole: The var `basepath` is already a full path before we get here.

CharliePoole:
I guess the alternative to `Named(string)` was `AddSubPackages(string)`. As I've explained, extension authors need to create full packages, including a name and settings. One option is to make `FullName` settable and I tried that for a while. But some sort of fluent modifier seemed more natural to me.

src/NUnitEngine/nunit.engine.tests/Services/Fakes/FakeProjectService.cs
ChrisMaddock: We could just have AddSubPackage return the subpackage, couldn't we? 

`var subpackage = package.AddSubPackage("tests.dll")`

ChrisMaddock:
> I guess the alternative to Named(string) was AddSubPackages(string). As I've explained, extension authors need to create full packages, including a name and settings. One option is to make FullName settable and I tried that for a while. But some sort of fluent modifier seemed more natural to me.

Interested in what anyone on @nunit/engine-team would think on this one! ☝️ I think we're into preference territory here - and it's a case where mine and your preferences differ, Charlie! 😄

src/NUnitEngine/nunit.engine.api/TestPackage.cs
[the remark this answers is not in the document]
ChrisMaddock: Personally, I prefer the clean API with a single constructor, over the convenience of both. 

It's not one I feel incredibly strongly on though - if you particularly disagree here. 🙂

src/NUnitEngine/nunit.engine.tests/Services/Fakes/FakeProjectService.cs
CharliePoole: Good idea. I'll do that!

package-tests.cake
```diff
@@ -293,19 +293,20 @@ public class ZipPackageTester : NetFXPackageTester
...
-            new ExpectedResult("Failed")
-            {
-                Total = 2 * 37,
-                Passed = 2 * 23,
-                Failed = 2 * 5,
-                Warnings = 0,
-                Inconclusive = 2 * 1,
-                Skipped = 2 * 7
-            }));
+        // TODO: Get extensions working - need new releases due to interface change
```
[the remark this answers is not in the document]
CharliePoole: I'll be doing at least the NUnit Project as a new release while we work on this. Don't want to do too much until the API is established.

src/NUnitEngine/nunit.engine.api/TestPackage.cs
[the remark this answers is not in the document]
CharliePoole: I don't have a strong preference in the abstract, but when I tried out each of the two alternatives separately, I was forced to write ugly code in one place or another. Runners and project loaders would usually have some kind of dynamic list and would be forced to construct an array from it. OTOH the 39 uses of the params constructor would have to add a new [] construct to their code.

package-tests.cake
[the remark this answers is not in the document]
CharliePoole: Actually, it may only be the NUnit Project loader that has a problem... it uses the old string constructor, which now looks like a params string[] constructor!

CharliePoole:
I've gone with the "AddSubPackage" suggestion in place of "Named".

CharliePoole:
Ready for re-review

ChrisMaddock:
Looks good, thanks!

PR #922, opened 2021-03-22
Forced cancellation when framework does not cancel completely

### LINKED ISSUE #642: Engine needs preemptive cancellation
I'm marking this as an idea for discussion, although I have pretty much convinced myself it's a necessary feature.

It's a problem I came upon in the GUI. Issue TestCentric/testcentric-gui#337 complained that the GUI was not stopping a test run immediately like NUnit V2 but only after any already running tests complete. The GUI uses a cooperative cancel rather than a forced cancel, so I decided to use force for version 1.0 to be compatible with the V2 GUI.

It turns out that force doesn't work. The framework stops the tests, but it appears not too. The GUI doesn't get the notifications it expects and therefore doesn't think the test has actually stopped. Of course, that's a framework issue, which I'll deal with separately.

Why this issue then?

Thinking about it, I realized that both options, cooperative and preemptive cancellation, are in some way cooperative. The true cooperative cancel requires the cooperation of the tests. The "forced" cancel requires the cooperation of the framework code.

If we consider that runner, engine and framework are independent layers and that drivers can be written for any framework in the future, it seems to me that we need a true preemptive cancel that doesn't depend on the framework at all.

If the test is running in a separate process, that's pretty easy. We could just kill the process or simply tell it to terminate. Telling it seems OK, because the engine is running in both processes.

Things get a bit harder if we are running in process. We might have to use some pretty low-level stuff to get to the executing threads and kill them. Or we might be able to do something at the level of the AppDomain.

As I said at first, this is an idea. I'd like to find out first if people agree with it, then we can worry about how it would be done.

CharliePoole:
Fixes #642

### DIFF
--- a/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
+++ b/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
@@ -185,7 +185,7 @@ namespace NUnit.Engine.Runners
 
                 if (!_workItemTracker.WaitForCompletion(WAIT_FOR_CANCEL_TO_COMPLETE))
                 {
-                    _workItemTracker.IssuePendingNotifications(_eventDispatcher);
+                    _workItemTracker.SendPendingTestCompletionEvents(_eventDispatcher);
 
                     // Indicate we are no longer running
                     IsTestRunning = false;
@@ -440,6 +440,10 @@ namespace NUnit.Engine.Runners
         /// <returns>A TestEngineResult giving the result of the test execution</returns>
         private TestEngineResult RunTests(ITestEventListener listener, TestFilter filter)
         {
+            _workItemTracker.Clear();
+            _eventDispatcher.Listeners.Clear();
+            _eventDispatcher.Listeners.Add(_workItemTracker);
+
             if (listener != null)
                 _eventDispatcher.Listeners.Add(listener);
 
--- a/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
+++ b/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
@@ -42,6 +42,11 @@ namespace NUnit.Engine.Runners
         private ITestRunnerFactory _testRunnerFactory;
         private bool _disposed;
 
+        private TestEventDispatcher _eventDispatcher = new TestEventDispatcher();
+        private WorkItemTracker _workItemTracker = new WorkItemTracker();
+
+        private const int WAIT_FOR_CANCEL_TO_COMPLETE = 5000;
+
         public MasterTestRunner(IServiceLocator services, TestPackage package)
         {
             if (services == null) throw new ArgumentNullException("services");
@@ -165,6 +170,37 @@ namespace NUnit.Engine.Runners
         public void StopRun(bool force)
         {
             _engineRunner.StopRun(force);
+
+            if (force)
+            {
+                // Frameworks should handle StopRun(true) by cancelling all tests and notifying
+                // us of the completion of any tests that were running. However, this feature
+                // may be absent in some frameworks or may be broken and we may not pass on the
+                // notifications needed by some runners. In fact, such a bug is present in the
+                // NUnit framework through release 3.12 and motivated the following code.
+                //
+                // We try to make up for the potential problem here by notifying the listeners 
+                // of the completion of every pending WorkItem, that is, one that started but
+                // never sent a completion event.
+
+                if (!_workItemTracker.WaitForCompletion(WAIT_FOR_CANCEL_TO_COMPLETE))
+                {
+                    _workItemTracker.IssuePendingNotifications(_eventDispatcher);
+
+                    // Indicate we are no longer running
+                    IsTestRunning = false;
+
+                    // Signal completion of the run
+                    _eventDispatcher.OnTestEvent($"<test-run id='{TestPackage.ID}' result='Failed' label='Cancelled' />");
+
+                    // Since we were not notified of the completion of some items, we can't trust
+                    // that they were actually stopped by the framework. To make sure nothing is
+                    // left running, we unload the tests. By unloading only the lower-level engine
+                    // runner and not the MasterTestRunner itself, we allow the tests to be loaded
+
+                    _engineRunner.Unload();
+                }
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
```diff
@@ -42,6 +42,11 @@ public class MasterTestRunner : ITestRunner
         private ITestRunnerFactory _testRunnerFactory;
         private bool _disposed;
 
+        private TestEventDispatcher _eventDispatcher = new TestEventDispatcher();
+        private WorkItemTracker _workItemTracker = new WorkItemTracker();
```
ChrisMaddock, line 46: I don't think this is plugged in, is it? As in, it's not been added been added as an event listener. They're not detected automatically or anything, are they?

src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
```diff
@@ -165,6 +170,37 @@ public ITestRun RunAsync(ITestEventListener listener, TestFilter filter)
...
+                // notifications needed by some runners. In fact, such a bug is present in the
+                // NUnit framework through release 3.12 and motivated the following code.
+                //
+                // We try to make up for the potential problem here by notifying the listeners 
+                // of the completion of every pending WorkItem, that is, one that started but
+                // never sent a completion event.
+
+                if (!_workItemTracker.WaitForCompletion(WAIT_FOR_CANCEL_TO_COMPLETE))
+                {
+                    _workItemTracker.IssuePendingNotifications(_eventDispatcher);
```
ChrisMaddock, line 188: Suggestion: this method name initially confused me - I thought it was sending notifications that "the following tests are still running" 😄 How about something like `ForceSendTestCompletionEvents` or something?

src/NUnitEngine/nunit.engine/Runners/WorkItemTracker.cs
```diff
@@ -0,0 +1,113 @@
...
+            XmlNode message = failure.AddElementWithCDataSection("message", "Test run cancelled by user");
+            return notification.OuterXml;
+        }
+
+        void ITestEventListener.OnTestEvent(string report)
+        {
+            XmlNode xmlNode = XmlHelper.CreateXmlNode(report);
+
+            lock (_trackerLock)
+            {
```
ChrisMaddock, line 81: This is going to be high pretty throughput code, I'm thinking. Would it be more efficient to use a multi-thread safe collection type, and just lock it when we give start to force the cancellation messages?

src/NUnitEngine/nunit.engine/Runners/WorkItemTracker.cs
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        private void RemoveItem(string id)
+        {
+            foreach (XmlNode item in _itemsInProcess)
+            {
+                if (item.GetAttribute("id") == id)
+                {
+                    _itemsInProcess.Remove(item);
+                    return;
```
ChrisMaddock, line 108: How about some sort of id-based hashset here? There's probably going to be single figure numbers of tests running at any one point, but still probably quite an intensive bit of code, I'd think?

src/NUnitEngine/nunit.engine/Runners/WorkItemTracker.cs
```diff
@@ -0,0 +1,113 @@
...
+                        _itemsInProcess.Add(xmlNode);
+                        break;
+
+                    case "test-case":
+                    case "test-suite":
+                        string id = xmlNode.GetAttribute("id");
+                        RemoveItem(id);
+
+                        if (_itemsInProcess.Count == 0)
+                            _allItemsComplete.Set();
```
ChrisMaddock, line 95: I'm not totally clear on how these ManualResetEvents work - but would this logic work for a test run where zero tests are actually run e.g. all filtered out?

Of course, you'd be hard-pressed to cancel it before it finished normally in that case! 😄

ChrisMaddock (commented):
Looks good Charlie. My only thought - this could be a bit of an overhead to solve a niche problem, that we currently only know to exist in old versions of the NUnit framework. Have left a couple of suggestions as to how we could maybe minimse that impact - not sure if they'll all work, but thought it was something we should think about here.

ChrisMaddock:
(Something's gone a bit wrong here - for some reason half my comments have become separated from the review! Hopefully you get 6, I think that was everything I typed... 🤔)

src/NUnitEngine/nunit.engine/Runners/WorkItemTracker.cs
CharliePoole: In fact, my thinking was that it's likely to not be very deep. I figured a hash or dictionary would have to be coupled with a list, as we send the notifications in reverse order. So that would end up being even more overhead.

The biggest overhead load will occur when someone has a very large number of very fast tests.

Also, I'm not very sure how niche this is. I have yet to see an NUnit framework that cancels properly. And typically, the largest number of users don't upgrade very fast.

src/NUnitEngine/nunit.engine/Runners/WorkItemTracker.cs
CharliePoole: Now that's niche! :smiley_cat:

src/NUnitEngine/nunit.engine/Runners/WorkItemTracker.cs
CharliePoole: In the GUI I followed my usual practice... assume no performance problem and only fix it if it shows up. Haven't seen one yet.

src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
CharliePoole: I remember adding it but I don't see it... could have removed it "temporarily" in order to test something. I'll check.

src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
CharliePoole: "Force" sounds odd, since the whole purpose of the class is to send notices. I like "Completion". How about "SendPendingTestCompletionEvents"?

My general problem with naming is that these reports (that's what they are actually called when received) are in two places: those received by the tracker and those sent by the tracker.

CharliePoole:
I think I saw all your comments. Still checking on one item I'm not sure of.

src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
CharliePoole: Good catch! It was in but I must have removed it for a test and forgot to put it back... see start of `MasterTestRunner.RunTests'

CharliePoole:
Ready for second review. I would prefer to wait to see if theirs a problem before tuning  the code for efficiency.

ChrisMaddock:
I wish we had some sort of performance tests - unfortunately I don't have time to check manually at the moment. Ah well - merging!

PR #927, opened 2021-03-24
Stop double reporting of assemblies when running after a reload

### LINKED ISSUE #609: Run after Reload reports assemblies multipletimes

CharliePoole:
This replaces PR #610 and fixes #609 directly, without making any of the discussed enhancements to Reload.

The reported "doubling" of assemblies after a reload was due to not clearing the list of drivers.

The bigger questions raised about how we do a reload are reserved for a future enhancement.

The old issue-609 branch can be deleted once this is merged, in addition to issue-609b.

### DIFF
--- a/src/NUnitEngine/nunit.engine.core/Runners/DirectTestRunner.cs
+++ b/src/NUnitEngine/nunit.engine.core/Runners/DirectTestRunner.cs
@@ -103,6 +103,8 @@ namespace NUnit.Engine.Runners
 
             var driverService = Services.GetService<IDriverService>();
 
+            _drivers.Clear();
+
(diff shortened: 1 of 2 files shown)


ChrisMaddock:
Looks good! 👍

PR #928, opened 2021-03-27
Remove unused dependencies from .NET Standard 2.0 builds

### LINKED ISSUE #908: Known Vulnerability in System.Xml.XPath.XmlDocument
`System.Xml.XPath.XmlDocument` 4.3.0 and other packages have a known security vulnerabilities and need to be updated to 4.3.1.

As reported by a user,

> We are using NUnit3TestAdapter package in some of our projects (.NET Core 3.1), which according to our latest security (whitesource) scans contains some vulnerabilities (System.Text.RegularExpressions 4.3.0). These policy vulnerabilities are coming from some packages that are used in your component, "System.Xml.XmlDocument 4.3.0", "System.Xml.XPath.XmlDocument 4.3.0" and so on. From what we know the solution will be to update the version of System.Text.RegularExpressions to 4.3.1.

ChrisMaddock:
Fixes #908.

When these dependencies were added, they were all added for both the .NET Standard 1.3 and 2.0 builds. I don't believe any of them are actually required, now we've removed .NET Standard 1.3.

### DIFF
--- a/nuget/engine/nunit.engine.api.nuspec
+++ b/nuget/engine/nunit.engine.api.nuspec
@@ -18,12 +18,6 @@
     <language>en-US</language>
     <tags>nunit test testing tdd engine</tags>
     <copyright>Copyright (c) 2021 Charlie Poole, Rob Prouse</copyright>
-    <dependencies>
-      <group targetFramework="net20" />
-      <group targetFramework="netstandard2.0">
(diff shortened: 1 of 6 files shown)


CharliePoole (approved):
Makes sense to me - and all the tests still pass! :smile:

PR #930, opened 2021-04-03
feat:  Allow more flexibility in extension naming

UnderwoodNC:
This change will allow developers to use a different base namespace and have those packages searched by default as well instead of updating a file within the NuGet package’s folder, which would be bad practice.  For example, it will match, SomeTeam.NUnit.Extensions.MyProjectLoader

### DIFF
--- a/nuget/runners/nunit.console.nuget.addins
+++ b/nuget/runners/nunit.console.nuget.addins
@@ -1,4 +1,4 @@
-../../NUnit.Extension.*/**/tools/        # nuget v2 layout
-../../../NUnit.Extension.*/**/tools/     # nuget v3 layout
-../../nunit.extension.*/**/tools/        # nuget v2 layout
-../../../nunit.extension.*/**/tools/     # nuget v3 layout
+../../*NUnit.Extension.*/**/tools/        # nuget v2 layout
+../../../*NUnit.Extension.*/**/tools/     # nuget v3 layout
+../../*nunit.extension.*/**/tools/        # nuget v2 layout
+../../../*nunit.extension.*/**/tools/     # nuget v3 layout


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/nunit/nunit-console?pullRequest=930) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/nunit/nunit-console?pullRequest=930) before we can accept your contribution.<br/><br/>:x: UnderwoodNC [sign now](https://cla.dotnetfoundation.org/nunit/nunit-console?pullRequest=930)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/nunit/nunit-console?pullRequest=930) it.</sub>

CharliePoole:
The existing file is there to point to the "official" NUnit addins, which we distribute.

The intended way to point to your own extensions is to either use the naming convention "NUnit.Extension.XXX" or to create additional `.addins` files in the same directory. Is there a reason you prefer this approach?

ChrisMaddock:
Hi @UnderwoodNC - thanks for the PR.

Afraid I agree with @CharliePoole here. The .addins files are intended to be edited or added to - and the items listed are in this particular version are aimed specifically to load the four NUnit-organisation-owned extensions that are distributed with the NUnit.Console NuGet package.

I do agree with you however that editing a file within the NuGet package is not a particularly nice approach. (Note you could also add a second addins file to the console directory, which is perhaps simpler.) For a long time I've wanted to add an `nunit3-console install-extension PATH` command to make this a bit nicer - but I'm afraid I think that's the direction we'd want to go in, rather than widening the wildcard search to allow a slightly larger set of package names to be picked up by default.

Based on that, afraid I'm going to close this PR - but thanks for taking the time regardless. If you are interested in digging in further, some help with new install/uninstall extension commands would be very welcome!

Issue #923, opened 2021-03-22
Agent communication layer


CharliePoole:
This issue is aimed at implementing a communications approach like that in the TestCentric engine, as we discussed in testcentric/testcentric-gui#592

1. Communications as a separate layer with possibility of multiple approaches.
2. Each agent specialized to one communications channel
3. Separate definitions of protocols, transports and messages

Once resolved, this will also close the above NUnit / TestCentric compatibility issue as well as enabling execution of .NET Core tests under the nunit3 console runner.

CharliePoole:
@ChrisMaddock Since this is non-breaking, I put it in the 3.13 milestone. Not sure if that will really work out alongside in-process execution, but let's try it and see.

Issue #702, opened 2019-10-25
Replacement for RuntimeFrameworkService


ChrisMaddock:
RuntimeFrameworkService suffers from a number of problems:

1. Lack of thread-safety (#173) - preventing improvements in other areas (#140)
2. Mix of functionality relating to runtimes and target frameworks, leading to confusion in code (e.g. use of `Runtime.Any`).
3. High-overhead in certain usecases - e.g. every test run currently calculates and discovers all available runtimes, when only a minority of runners require that information.
4. Limited testing, due to tight integration of assembly inspection/File system requirements, and logic.
5. Lot's of netfx specific functionality, which can't be compiled for the .NET Standard builds
6. Reliance on two different versions of Mono Cecil, due to no version of Cecil having support for both .NET 2.0 and .NET Standard.

After spending a couple of years teasing round the edges of this, and not making a lot of progress (https://github.com/nunit/nunit-console/pull/407, https://github.com/nunit/nunit-console/pull/402, https://github.com/nunit/nunit-console/pull/408, https://github.com/nunit/nunit-console/pull/670, #406) I'd like to instead propose a wholesale replacement for these classes. I'm marking this as high priority, as it's an area which is going to need lots of upcoming development for the .NET Standard improvements - and I'd like to create a better structure here, before we start building on top of it.

---------------

I'm proposing two new services to replace it:

1. **AssemblyTargetingService**: Responsible for everything around what an assembly was built to target. Would be used to inspect and annotate test packages, and inspect engine extensions.
2. **RuntimesService**: Responsible for all functionality around discovering available runtimes, and deciding on the runtime to be used for each agent. 

I'd then like the Obsolete the current RuntimeFrameworkService, and remove in v4.

Thoughts welcome. 🙂

CharliePoole:
@ChrisMaddock Regarding removal of RuntimeFrameworkService... it might be possible before v4, because it's not exposed outside the engine. The IRuntimeFrameworkService interface is not exposed but IAvailableRuntimes is. The latter must be implemented somewhere, but does not have to be in the same class.

FYI, I'm doing similar work in the GUI, but incrementally. I'm thinking about how to make you and the @nunit/engine-team members aware of each new thing I do... more on that to come.

CharliePoole:
[...]

Issue #406, opened 2018-04-22
Split RuntimeFramework into separate RuntimeFramework and TargetFramework classes


ChrisMaddock:
This has come out of https://github.com/nunit/nunit-console/issues/394.

The RuntimeFramework functionality is one of the [more legacy](https://github.com/nunit/nunit-console/issues/173) areas of our code base - and could use a refactor to better fit the NUnit 3 architecture. 

RuntimeFramework is currently used for two purposes, both to detect available runtimes (e.g. full framework, mono, CF), and to determine the target framework of an assembly (e.g net45, netcore20). There's something of a mismatch here as a net45 could of course run on .NET Framework or Mono, so the Runtime.Any enum exists. (Although it's not technically 'Any' I don't believe, as this bit of code still supports Silverlight and CF!)

With supporting .NET Standard and .NET Core for https://github.com/nunit/nunit-console/issues/394 and https://github.com/nunit/nunit-console/issues/271, this bit of codes going to need some additional complexity. Accordingly, I'd like to refactor and do the following:

1. Refactor for thread-safety and proper division of RuntimeFrameworkService and RuntimeFramework responsibilities (As #173)
2. Remove all support for Silverlight and CF. (I _think_ this class used to be shared with NUnit Framework? Or was it that the NUnit 2 console supported these platforms?)
3. Split out separate classes for 'runtime' (e.g. full framework, mono, .NET Core) and 'target' (.NET, .NET Standard etc.)

Any thoughts? 🙂

rprouse:
All good by me 😀

ChrisMaddock:
Superceeded by https://github.com/nunit/nunit-console/issues/702 - I think unfortunately the problem is a little bigger than this one class 😭

PR #937, opened 2021-04-13
Run .NET Core Tests from NUnit3 Console

CharliePoole:
Another step toward issue #923.

This PR adds TCP as a transport and uses it to run tests under .NET Core.

A .NET Core 3.1 agent has been added. We'll have to decide whether we want others as separate agents. A single package test has been added to exercise the agent. It is run for the NuGet, Chocolatey and Zip packages of the console runner.

The MSI package does not yet include the .NET Core agent. I made a start on this but backed out the changes when I saw that we need to make some decisions about the msi package. I think this will be easier to do if there's a separate PR for it. This one is pretty big as it is.


CharliePoole:
Unable to figure out why the Azure builds are failing.

Windows job can't find `src\NUnitEngine\nunit-agent\obj\Release\x86\netcoreapp3.1\apphost.exe`, which is present when I do a local build.

Linux and MacOs jobs say that the net-4.0 framework isn't available.

Not being an Azure guy, I don't even know where to look.

src/NUnitEngine/nunit.engine.core/RuntimeFramework.cs
```diff
@@ -135,6 +138,18 @@ private Version GetClrVersionForFramework(Version frameworkVersion)
                             return new Version(4, 0, 30319);
                     }
                     break;
+                case RuntimeType.NetCore:
+                    switch (frameworkVersion.Major)
+                    {
+                        case 1:
+                        case 2:
+                            return new Version(4, 0, 30319);
```
ChrisMaddock, line 146: Is this correct? The Azure error looks like it's trying to run the .NET 3.5 tests on .NET 4.X and not finding the framework. This line looks suspicious?

```
1) Error : 
System.ArgumentException : The net-4.0 framework is not available
Parameter name: framework
--ArgumentException
The net-4.0 framework is not available
```

ChrisMaddock:
Not sure if the CI has been re-run since your comment Charlie, but looks like the Windows build is ok. Looks to me like the Mono tests are all failing to start up an agent on Linux/OSX. Not sure what's in this PR that would affect that, however...

src/NUnitEngine/nunit.engine.core/RuntimeFramework.cs
ChrisMaddock: Actually, not sure that would make any sense. Still not sure this line is correct, though. 😅

CharliePoole: According to https://docs.microsoft.com/en-us/dotnet/api/system.environment.version?view=net-5.0, "for ... .NET Core versions before 3.0, [Environment.Version] has the form 4.0.30319.42000."

The current `RuntimeFramework` class assumes that each framework version has a corresponding CLR version. With .NET Core, that no longer makes sense and I think `RuntimeFramework` needs to be rewritten or removed. I didn't do that here, because this PR is already too big. Rather, I tried to avoid using the CLR version to make any decisions.

However, that net-4.0 framework is not available looks suspicious to me too. I'll check into it further.

CharliePoole:
@ChrisMaddock @jnm2 

The error in building windows continues to appear intermitently, i.e. failure to locate `apphost.exe`, which should be generated by the restore. I have noticed that when it does appear, there is nothing in the log about restoring `nunit-agent.csproj`. OTOH `nunit-agent-x86.csproj` __is__ restored correctly.

I'm working on the hypothesis that nuget is choking on the odd setup we have for the two agent projects: two projects in one directory with separate intermediate output folders, `obj` and `obj\x86`. We always knew this was odd and it was somewhat surprising that it actually worked. But I suspect it isn't working for .NET Core builds.

I'll run an experiment, separating the two projects into two different folders and replicating the code.

CharliePoole:
That worked! The latest commit builds under Windows on both AppVeyor and Azure. The Linux and MacOs builds are still failing on Azure, but not from the same cause.

CharliePoole:
Down to three tests failing on Linux and MacOs!

CharliePoole:
The Linux / MacOs problem was due to my porting changes  to TestAgency and RuntimeSelectionService from TestCentric, without also porting changes to RuntimeFramework. Rather than doing that in this PR, I have simply added some ugly procedural code to ensure that net-x.x is treated as compatible with mono-x.x. I'll do a refactoring change at some point after this is all merged.

As it now stands in the PR...
 * Nunit3-console is able to run .NET Core Tests
 * All the packages include this capability except for the msi, which doesn't yet have the dotnet core agent.
 * Each package (except the msi) has a single test running .NET Core 3.1

Remaining to do for issue #923...
 * Add the .NET Core runner to the msi package
 * Add more package tests... 2.1? 2.0? 1.6? a mixture?

Held over as a separate issue...
 * Refactor RuntimeFramework

ChrisMaddock:
Hi Charlie - this looks great! Will try and have a proper look through soon.

Just from a quick glance, I think you might have accidentally duplicated the agent projects, rather than moved them as you intended? There's a lot of files showing as new in the diff, which suggests to me the old ones are still there too. Do you mind taking a look at that first - it'll be easier to review once we can see the diff's properly.

CharliePoole:
@ChrisMaddock You're right! I started out moving them all to an agents sub-directory and then changed my mind. I removed the new copies from the solution but not from the repo. I think we will eventually want to do that if we create more agents, but for now it's not needed and the projects that are at the original level are the ones currently in the solution, so I'll drop those folders and files. That will make this a bit smaller as well.

src/NUnitEngine/nunit.engine.core/RuntimeFramework.cs
```diff
@@ -444,6 +466,23 @@ public bool Supports(RuntimeFramework target)
...
+            if (this.Runtime == RuntimeType.Any || targetRuntime == RuntimeType.Any)
+                return true;
+
+            if (this.Runtime == RuntimeType.Net && targetRuntime == RuntimeType.Mono)
+                return true;
+
+            if (this.Runtime == RuntimeType.Mono && targetRuntime == RuntimeType.Net)
+                return true;
+
+            return false;
```
ChrisMaddock, line 483: Good move expanding this clause to it's own method. I'm struggling to follow the logic here, but is this a slight behaviour change? What happens now if I use the `--framework=mono` flag when running the console on Windows .NET Framework, would my tests now end up running on netfx instead?

src/NUnitEngine/nunit.engine.core/RuntimeFramework.cs
```diff
@@ -641,6 +684,73 @@ private static void AddMonoFramework(Version frameworkVersion, string profile)
 
             _availableFrameworks.Add(framework);
         }
+
+        private static void FindDotNetCoreFrameworks()
+        {
+            const string WINDOWS_INSTALL_DIR = "C:\\Program Files\\dotnet\\";
```
ChrisMaddock, line 690: My understanding is that this is a default installation location, but not guaranteed, users can install where they please. 

As an alternative/additional implementation to this method, how about invoking `dotnet --list-runtimes`? The output from that includes paths to exe's.

I'm don't know where that command gets its info from. Although does seem like an issue we can then delegate to dotnet itself

src/NUnitEngine/nunit.engine.core/RuntimeFramework.cs
```diff
@@ -641,6 +684,73 @@ private static void AddMonoFramework(Version frameworkVersion, string profile)
...
+            string INSTALL_DIR = Path.DirectorySeparatorChar == '\\'
+                ? WINDOWS_INSTALL_DIR
+                : LINUX_INSTALL_DIR;
+
+            if (!Directory.Exists(INSTALL_DIR))
+                return;
+            if (!File.Exists(Path.Combine(INSTALL_DIR, "dotnet.exe")))
+                return;
+
+            string runtimeDir = Path.Combine(INSTALL_DIR, "shared\\Microsoft.NETCore.App");
```
ChrisMaddock, line 701: Niggle - should this be a forward slash for Linux? Not sure if Path.Combine corrects this for us?

src/NUnitEngine/nunit.engine.core/RuntimeFramework.cs
```diff
@@ -641,6 +684,73 @@ private static void AddMonoFramework(Version frameworkVersion, string profile)
...
+            var dirNames = new List<string>();
+            foreach (var dir in dirList)
+                dirNames.Add(dir.Name);
+            var runtimes = GetNetCoreRuntimesFromDirectoryNames(dirNames);
+
+            _availableFrameworks.AddRange(runtimes);
+        }
+
+        // Internal for testing
+        internal static IList<RuntimeFramework> GetNetCoreRuntimesFromDirectoryNames(IEnumerable<string> dirNames)
```
ChrisMaddock, line 715: This method looks exciting! 😂 Would you mind clarifying what's going on here? All the directory names in my install are simple version numbers, but the logic here makes me think there are other options?

src/NUnitEngine/nunit.engine.core/nunit.engine.core.csproj
```diff
@@ -9,6 +9,13 @@
     <Reference Include="System.Runtime.Remoting" />
     <Reference Include="System.Web" />
   </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)'=='netstandard2.0' or '$(TargetFramework)'=='netcoreapp3.1'">
+    <PackageReference Include="Microsoft.Win32.Registry" Version="4.3.0" />
+    <PackageReference Include="System.ComponentModel.TypeConverter" Version="4.3.0" />
+    <PackageReference Include="System.Reflection" Version="4.3.0" />
+    <PackageReference Include="System.Diagnostics.Process" Version="4.3.0" />
+    <PackageReference Include="System.Xml.XPath.XmlDocument" Version="4.3.0" />
+  </ItemGroup>
```
ChrisMaddock, line 18: Is this a rebase issue? Think these are the ones I removed just recently. 🙂

src/NUnitEngine/nunit.engine/Communication/Transports/Tcp/TcpServer.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.Net;
+using System.Net.Sockets;
+using System.Threading;
+using NUnit.Engine.Internal;
+
+namespace NUnit.Engine.Communication.Transports.Tcp
+{
+    public class TcpServer
+    {
+        private static readonly Logger log = InternalTrace.GetLogger(typeof(TestAgentTcpTransport));
```
ChrisMaddock, line 13:
```suggestion
        private static readonly Logger log = InternalTrace.GetLogger(typeof(TcpServer));
```

Presumably should be this?

src/NUnitEngine/nunit.engine/Communication/Transports/Tcp/TestAgencyTcpTransport.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.Net;
+using System.Threading;
+using NUnit.Common;
+using NUnit.Engine.Internal;
+
+namespace NUnit.Engine.Communication.Transports.Tcp
+{
+    /// <summary>
+    /// Summary description for TestAgencyTcpTransport.
+    /// </summary>
```
ChrisMaddock, line 13: Would you mind removing the empty summary, please? 🙂

src/NUnitEngine/nunit.engine/Communication/Transports/Tcp/TestAgentTcpProxy.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        public TestEngineResult Explore(TestFilter filter)
+        {
+            SendCommandMessage("Explore", filter);
+            return CommandResult<TestEngineResult>();
+        }
+
+        public void Dispose()
+        {
+            throw new NotImplementedException();
```
ChrisMaddock, line 97: Should this be empty, rather than not implemented exception? Or is there something here that does need disposing? I seem to remember we had a need to dispose runners at some point...can't remember why it was though however.

src/NUnitEngine/nunit.engine/Services/AgentProcess.cs
```diff
@@ -88,16 +96,34 @@ public static string GetTestAgentExePath(RuntimeFramework targetRuntime, bool re
...
+                    agentExtension = ".exe";
+                    break;
+                case RuntimeType.NetCore:
+                    runtimeDir = "netcoreapp3.1";
+                    agentExtension = ".dll";
+                    break;
+                default:
+                    log.Error($"Unknown runtime type: {targetRuntime.Runtime}");
+                    return null;
+            }
```
ChrisMaddock, line 124: I think this code needs to match a decision on an netcoreapp31 x86 agent, right? Right now, if I ran a .NET Core 3.1 tests with the --x86 flag, I think we'd end up with an agent exe not being found?

src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs
```diff
@@ -72,6 +75,23 @@ private static bool FrameworksMatch(RuntimeFramework f1, RuntimeFramework f2)
...
+                return true;
+
+            if (rt1 == RuntimeType.Any || rt2 == RuntimeType.Any)
+                return true;
+
+            if (rt1 == RuntimeType.Net && rt2 == RuntimeType.Mono)
+                return true;
+
+            if (rt1 == RuntimeType.Mono && rt2 == RuntimeType.Net)
+                return true;
```
ChrisMaddock, line 90: Can you clarify what this method is doing, sorry? It makes me a little uneasy that we have a method called `RuntimesMatch` which returns true for two different runtimes: netfx and mono. Maybe its just a naming issue?

src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs
```diff
@@ -109,21 +129,45 @@ private static RuntimeFramework SelectRuntimeFrameworkInner(TestPackage package)
...
+                        break;
+                    case ".NETCoreApp":
+                        targetRuntime = RuntimeType.NetCore;
+                        break;
+                    default:
+                        throw new NUnitEngineException("Unsupported Target Framework: " + imageTargetFrameworkNameSetting);
+                }
 
-            if (targetVersion == RuntimeFramework.DefaultVersion)
-                targetVersion = package.GetSetting(InternalEnginePackageSettings.ImageRuntimeVersion, currentFramework.FrameworkVersion);
```
ChrisMaddock, line 126: Please could you also clarify what the changes in this method are doing too? Sorry - the RuntimeFramework stuff has always been so convoluted, struggling to work out which stage in the process this is affecting here. :-(

ChrisMaddock (commented):
Hi Charlie - this looks great - really big step for the engine! 😄 

I've had a look through and left a load of comments. Afraid I've mostly skimmed to TCP code - that's really beyond my technical expertise to understand - would be great if someone with more experience was able to take a look at that.

I think we need to sort the msi and x86 agent before merging this - just to keep master in a releasable state. Hopefully that's not too much more to add however. 🙂

CharliePoole:
@ChrisMaddock I agree about the need to have this reviewed by someone with TCP experience. I've asked for review from @jnm2, but if he isn't able to get to it, I don't know who else we have. Do you?

CharliePoole:
@ChrisMaddock "Releasable" is a mostly a technical question, whereas the decision to actually release is a business decision... one that you have to make in this case. What you're asking for is a level of consistency in the availability of the new feature across different packages. That's reasonable as a business choice but I don't think it needs to affect how we break things into PRs.

I don't think this is just a semantic quibble. The bigger our PRs get, the slower we move. It's like a law of nature. :smile: This one is too big for my comfort already but I didn't know how to make it smaller and still have it make sense to the reviewers.

On the specifics... I believe the MSI needs to be redone to incorporate the .NET Core functionality and that will be substantial work, but it's work that's easily separable from the work here. I'll address the x86 functionality in responding to the detail comments.

package-tests.cake
```diff
@@ -160,6 +166,23 @@ public abstract class NetFXPackageTester : PackageTester
...
                 Skipped = 2 * 7
             }));
+
+        // TODO: Remove this check when msi package is
+        // updated to include .net core assemblies.
+        if (GetType().Name != "MsiPackageTester")
+        PackageTests.Add(new PackageTest(
+            "netcoreapp3.1",
+            "Run mock-assembly.dll under .NET Core 3.1",
+            "netcoreapp3.1/mock-assembly.dll --trace:Debug",
```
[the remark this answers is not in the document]
CharliePoole: Yes... left in by mistake...

nuget/engine/nunit.engine.nuspec
```diff
@@ -34,37 +34,52 @@
...
     <file src="bin/agents/net40/nunit.engine.api.dll" target="contentFiles/any/agents/net40" />
     <file src="bin/agents/net40/testcentric.engine.metadata.dll" target="contentFiles/any/agents/net40" />
+    
+    <file src="bin/agents/netcoreapp3.1/nunit-agent.dll" target="contentFiles/any/agents/netcoreapp3.1" />
+    <file src="bin/agents/netcoreapp3.1/nunit-agent.dll.config" target="contentFiles/any/agents/netcoreapp3.1" />
+    <file src="bin/agents/netcoreapp3.1/nunit-agent.runtimeconfig.json" target="tools/agents/netcoreapp3.1" />
+    <file src="bin/agents/netcoreapp3.1/nunit.engine.core.dll" target="contentFiles/any/agents/netcoreapp3.1" />
+    <file src="bin/agents/netcoreapp3.1/nunit.engine.api.dll" target="contentFiles/any/agents/netcoreapp3.1" />
+    <file src="bin/agents/netcoreapp3.1/testcentric.engine.metadata.dll" target="contentFiles/any/agents/netcoreapp3.1" />
+
```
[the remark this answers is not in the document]
CharliePoole: Not silly, I hope, because I originally thought so as well. However, I have read that the proper approach is to use the x86 version of the dotnet cli, so I removed those builds.

src/NUnitEngine/nunit-agent-x86/app.manifest
```diff
@@ -0,0 +1,41 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<asmv1:assembly manifestVersion="1.0"
+                xmlns="urn:schemas-microsoft-com:asm.v1"
+                xmlns:asmv1="urn:schemas-microsoft-com:asm.v1"
+                xmlns:asmv2="urn:schemas-microsoft-com:asm.v2"
+                xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <!-- We should update the version number if we remember, but it is not required -->
+  <assemblyIdentity version="3.0.0.0" name="nunit-agent"/>
```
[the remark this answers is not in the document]
CharliePoole: Me neither... but I didn't make any changes to this file, just moved it.

src/NUnitEngine/nunit.engine/Communication/Transports/Tcp/TestAgentTcpProxy.cs
CharliePoole: I'll check this one.

nuget/engine/nunit.engine.nuspec
```diff
@@ -34,37 +34,52 @@
...
     <file src="bin/agents/net40/nunit-agent.exe.config" target="contentFiles/any/agents/net40" />
     <file src="bin/agents/net40/nunit-agent-x86.exe" target="contentFiles/any/agents/net40" />
     <file src="bin/agents/net40/nunit-agent-x86.exe.config" target="contentFiles/any/agents/net40" />
     <file src="bin/agents/net40/nunit.engine.core.dll" target="contentFiles/any/agents/net40" />
     <file src="bin/agents/net40/nunit.engine.api.dll" target="contentFiles/any/agents/net40" />
     <file src="bin/agents/net40/testcentric.engine.metadata.dll" target="contentFiles/any/agents/net40" />
+    
+    <file src="bin/agents/netcoreapp3.1/nunit-agent.dll" target="contentFiles/any/agents/netcoreapp3.1" />
+    <file src="bin/agents/netcoreapp3.1/nunit-agent.dll.config" target="contentFiles/any/agents/netcoreapp3.1" />
+    <file src="bin/agents/netcoreapp3.1/nunit-agent.runtimeconfig.json" target="tools/agents/netcoreapp3.1" />
```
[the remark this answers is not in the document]
CharliePoole: It should probably match the others... although the package is working anyway. I don't understand why we need the whole business with `contentFiles/any` - my GUI build of the engine doesn't use it at all.

src/NUnitEngine/nunit-agent/Program.cs
```diff
@@ -63,16 +64,18 @@ public static void Main(string[] args)
...
 
             LocateAgencyProcess(agencyPid);
 
-            log.Info("Agent process {0} starting", pid);
-
+#if NETFRAMEWORK
             log.Info("Running under version {0}, {1}",
                 Environment.Version,
                 RuntimeFramework.CurrentFramework.DisplayName);
+#endif
```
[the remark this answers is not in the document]
CharliePoole: Good idea.

src/NUnitEngine/nunit.engine.core/RuntimeFramework.cs
CharliePoole: I could be wrong, but I don't think it works currently and I don't think it will start working after this either. :confused: 

As I mentioned, I think we have a load of  work to do with RuntimeFramework, which I don't want to do in this PR. One of the big parts of it is to figure out how we handle mono which is more like an alternate implementation of the .NET Framework than a different runtime.

BTW, this is a general method and doesn't specifically relate to the command-line option... i.e. targetFramework may come from the `TargetFrameoworkAttribute`, which never mentions Mono.

src/NUnitEngine/nunit.engine.core/RuntimeFramework.cs
CharliePoole: I think you're right.

src/NUnitEngine/nunit.engine/Services/AgentProcess.cs
CharliePoole: Yes... I didn't come back and change this after I discovered we didn't need an x86 agent for .NET Core. I'll need to change TestAgency as well.

src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs
CharliePoole: I totally agree. I left it alone because it's not new and my additions were in the same spirit as the original... where "match" didn't mean "equals" IMO we need to get back to this soon and redo the whole set of runtime classes...

However, I can only work on one thing at a time. :confused:

src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs
CharliePoole: Sorry. What I've done is take a disparate set of runtime checks, which were being made somewhat later in the process, and centralized them here. For example, we used to get to the point of creating a process and then discover that we either didn't know what runtime to use or that the requested runtime wasn't available. I haven't totally eliminated the late discoveries, but I think I've got 90% of it.

ChrisMaddock:
@rprouse - would you have some time to look through just the TCP communications classes in this PR? It's not something I'm familiar enough with to do any sort of meaningful review on. 

I'm comfortable I can review all the other changes well enough, it's just the actual comms structure we'd appreciate a second pair of eyes on. 🙂

ChrisMaddock:
> What you're asking for is a level of consistency in the availability of the new feature across different packages. 

I think I'd be less concerned if we made it an unimplemented feature in the msi that was handled, but at the moment I think we'd just get a crash, wouldn't we? As the engine code is set up to pass things to the netcoreapp31 agent, but the actual assembly would be missing in the msi packaging?

I had expected it just to be a few lines change to the msi however to include the new agent - sounds like you think there's more to do? Happy to leave it for a separate PR if you think that's best - just want to be sure it's done before the next release. Would you mind creating an issue with what needs to be done? Thanks!

ChrisMaddock:
Meta-question: Should the msi survive into 4.X? Maybe it's finally time... 😢

nuget/engine/nunit.engine.nuspec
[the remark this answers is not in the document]
ChrisMaddock: Ah, cool! So how does this actually work for x86 assemblies then, sorry? I can see the code just calls `dotnet` based on the PATH var - does dotnet itself launch an x86 process if required?

src/NUnitEngine/nunit.engine.core/RuntimeFramework.cs
ChrisMaddock: Definitely agree about refactoring, this area of the code is a nightmare. #702 and #406 track that - one more issue I've never got round to. 😞 

Happy to go with this if you're confident we're not breaking any existing scenarios. I just noticed the logic change, and didn't want us to inadvertantly break anything which is working currently.

src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs
ChrisMaddock: Again, if we're happy this doesn't break any existing scenarios, then let's go with it and get this all in. 🙂 

The `RuntimesMatch` method name is new from what I can see, however. Could we call this something else? Is it another `RuntimeSupports` maybe?

src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs
ChrisMaddock: Ahh, cool. Thanks!

CharliePoole:
@ChrisMaddock I agree that the msi package needs to recover gracefully, and I thought it was doing that. I'll check to make sure that moving many checks early has left in assumptions about what agents are available.

Regarding dropping the msi... that's now above my pay grade. :smile: You might recall that I wanted to drop it years ago, leading to your taking it over as a separate task for a period of time. Back then, of course, it included the framework too.

That said, in your position, I'd check the downloads and ask for feedback from the user community. IMO there's no technical need for it. It used to be a user preference for many folks who took all their advice from Microsoft, but I strongly suspect that has changed.

nuget/engine/nunit.engine.nuspec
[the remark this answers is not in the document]
CharliePoole: That's what I was referring to re TestAgency. It may need to add something to the command-line.

[Actually, I should have said AgentProcess rather than TestAgency. Thinking about two different engines is confusing. :confused: ]

src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs
```diff
@@ -109,21 +129,45 @@ private static RuntimeFramework SelectRuntimeFrameworkInner(TestPackage package)
...
+            RuntimeType targetRuntime;
+            Version targetVersion;
+
+            if (string.IsNullOrEmpty(imageTargetFrameworkNameSetting))
             {
-                requestedFramework = new RuntimeFramework(RuntimeType.Any, RuntimeFramework.DefaultVersion);
-                log.Debug($"No specific framework requested for {package.Name}");
+                // Assume .NET Framework
+                targetRuntime = RuntimeType.Net;
+                targetVersion = package.GetSetting(InternalEnginePackageSettings.ImageRuntimeVersion, new Version(2, 0));
```
ChrisMaddock, line 151: Niggle - this looks to be a logic change when ImageTargetFrameworkName is unset - previously we defaulted to the current framework, now we default to net20.

Not totally averse to the change if there's a particular reason for it, but I feel like the old behaviour was preferable. We know the current framework exists, and if the console has launched under it, then it's likely to be the "default" framework on this machine for other .NET executables. It seems to make sense to run the tests on the default framework, if we have no better information to go off.

Afraid I don't know how often ImageTargetFrameworkName would be unset, to judge how much of an edge-case this is that we're talking about here!

ChrisMaddock:
11.5k downloads for the MSI for v3.11.1, which was latest release for just under a year.

That seems pretty significant, especially assuming the msi would be more likely to be a "one time" download, vs the zip or nuget packages which might be pulled by CI systems for every build.

I guess there's still a market there...

src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs
CharliePoole: I believe it doesn't break anything... and all the same tests are still there!

Longer story... in the GUI fork, Runtime is a class, which knows how to compare itself to other runtimes. In this engine, it's still just an enum. So created this method to match the same functionality, but in a procedural way so as not to add unrelated work (redesigning RuntimeFramework) to the PR. I've done that a lot here, and I suppose it's a little confusing if you don't know where it comes from.

RuntimeSupports captures the meaning perfectly.

src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs
CharliePoole: I made the change because I've only seen it unset under pre-4.0 frameworks, using old versions of VS. CurrentFramework is safer, especially if we eventually add this logic to a .NET Core runner. I'll try it and we can see what it does to the tests.

CharliePoole:
In that case, I'll check for a quick fix now and we can do a second PR later...

I think a full rewrite is needed if / when you add the dotnet runner.

src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs
ChrisMaddock: > especially if we eventually add this logic to a .NET Core runner

That hadn't crossed my mind, and is a really good point! Thanks. 🙂

src/NUnitEngine/nunit.engine/Communication/Transports/Tcp/TestAgencyTcpTransport.cs
CharliePoole: I added a good comment here and for a number of other communication classes.

src/NUnitEngine/nunit.engine.core/RuntimeFramework.cs
CharliePoole: I used this approach because I didn't want to run the dotnet command without first knowing that dotnet was installed. I'll look into this further. At the time I wrote it I did a bunch of research, but I've forgotten the details. I'll check further.

src/NUnitEngine/nunit.engine.core/nunit.engine.core.csproj
CharliePoole: I wrote and modified this as bugs arose around preview and other pre-release installations. I added a comment.

src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs
CharliePoole: OK... this is a can of works and you should review the changes to this file again.

I left the name the same and changed the names of the arguments. As used, the first is always the requested runtime and the second is the available runtime. The available one can never actually be Any, so I eliminated that. So far so good.

This method is called by the FrameworksMatch method so I made the same change to that one. But that method also compares versions, using CLR version for some things, framework version for others. That's because, back in the day, we supported requesting a _specific_ CLR version, i.e. pre-releases of .NET 2.0 and .NET 4.0. We don't need that any longer and I'd like to get away from making much use of CLR version if possible, so I changed the method so that all checks now use FrameworkVersion. All tests still pass, so I feel safe, but you should still look at it.

As I reviewed the code, I realized that this support check is actually implemented in both `RuntimeFramework` (`Supports` method) and in the `RuntimeFrameworkService`. It's possible that I was in the process of moving it from `RuntimeFramework` to the service at the point in time when I had to stop working on the engine. I vaguely remember that was my intent anyway. We should get back to this.

CharliePoole:
@ChrisMaddock The latest commit takes care of all the  issues raised in your review except for those I need to do a bit of work on... that is...

* Fixing the MSI
* Handling x86 runs of .NET Core
* Deciding if we need to Dispose `TestAgentTcpProxy`

src/NUnitEngine/nunit.engine/Communication/Transports/Tcp/TestAgentTcpProxy.cs
CharliePoole: I put in the fix you suggested, 'cause it works! :smile: 

However, I went down a bit of a rabbit hole before settling on this and I think we ought to revisit it after this is merged, as a part of reviewing out API interfaces.

Specifically, `TestAgentTcpProxy` needs `Dispose()` because it implements `ITestEngineRunner`, which in turn derives from `IDisposable`. I spent a lot of time pondering this and I now think the latter derivation is a mistake.

When a _class_ implements `IDisposable`, it means that the class holds resources that need to be released, either directly or indirectly. But when an _interface_ derives from `IDisposable`, it says that _every_ class implementing it will have resources that need to be released. After a lot of thought, I believe that's almost never true.

This is a great example. The proxy implements `ITestAgentRunner` in order to return itself _as a runner_. But of course it only _looks like_ a runner, which is exactly what a proxy does. it has nothing to dispose.

All that said, for version 3.x we can only do what you have suggested. For version 4.0, I think we should look at cases where any interface derives from `IDisposable` very critically and consider removing it in favor of having each class that needs to implement it. There will be tradeoffs in each case, of course.

I could create an issue for this @ChrisMaddock but I'm not sure if it should be a broad one (Review our API) or a narrower one as described here. What do you think?

src/NUnitEngine/nunit.engine/Services/AgentProcess.cs
CharliePoole: Done, but continues to incorporate a hard-coded path. We'll need to revisit for dotnet installations at a non-standard location.

CharliePoole:
The latest commit runs an x86 test under .NET Core by using the 32-bit version of `dotnet.exe`. While this fixes the problem, it still requires `dotnet` to be installed in the standard location. However, since our entire detection of .NET Core depends on the use of the standard location, it's not making things any worse. I'd like to finish this PR and then come back to the problem of non-standard install locations by experimenting on a machine that has no other versions installed.

CharliePoole:
Latest Commit is failing on CI but works on my machine. It's possible that the x98 version of `dotnet.exe` is not present on the CI machine. I'll run some tests to find out.

CharliePoole:
The result is that AppVeyor doesn't seem to have the X86 version of dotnet.exe installed. We could probably install it in our script, but I'd rather save that for later. Instead I make a check for it's availability before trying to run the test, which is more robust.

After this last commit, the only remaining issue is the msi build.

CharliePoole:
@ChrisMaddock OK, last bit is done and this is ready for final review.

src/NUnitEngine/nunit.engine/Communication/Transports/Tcp/TestAgencyTcpTransport.cs
ChrisMaddock: Even better! 😁

src/NUnitConsole/nunit3-console.tests/App.config
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8" ?> 
+<configuration>
+  <appSettings>
+    <add key="test.setting" value="54321"/>
+  </appSettings>
+</configuration>
```
ChrisMaddock, line 6: This file is included within the "add agent to msi" commit - is it included accidentally? I'm not sure what the purpose of this is.

src/NUnitEngine/nunit.engine.tests/nunit.engine.tests.csproj
```diff
@@ -14,6 +14,9 @@
     <PackageReference Include="NUnit" Version="3.13.0" />
     <PackageReference Include="NSubstitute" Version="2.0.3" />
   </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)'=='netcoreapp2.1' OR '$(TargetFramework)'=='netcoreapp3.1'">
+    <Reference Include="System.Configuration" />
+  </ItemGroup>
```
ChrisMaddock, line 19: This change is also in the msi commit, and I presume also accidental?

msi/nunit/engine-files.wxi
```diff
@@ -87,5 +88,34 @@
                 <File Id="nunit.agent.net40.addins" Source="$(var.InstallImage)nunit.agent.addins" />
             </Component>
         </ComponentGroup>
+        <ComponentGroup Id="NETCORE31_AGENT" Directory="NETCORE31_AGENT_DIR">
+            <Component Id="NUNIT_AGENT_NETCORE31" Location="local" Guid="21E39B4C-07E4-4C38-8F91-7030516127E2">
+                <File Id="nunit_agent_netcore31.exe"
+                      ProcessorArchitecture="msil"
+                      Source="$(var.InstallImage)bin/agents/netcoreapp3.1/nunit-agent.dll" />
+                <File Id="nunit_agent.runtimeconfig.json"
+                      Source="$(var.InstallImage)bin/agents/netcoreapp3.1/nunit-agent.runtimeconfig.json" />
```
ChrisMaddock, line 97: I'm not totally clear on what is and isn't required for .NET Core, but I _think_ this should also include nunit-agent.deps.json. 

I was also wondering about nunit-agent.dll.config, but I think is actually netfx specific, so it should probably be removed entirely from the netcoreapp agent? I'm not sure how to do that off the top of my head, will give it a go.

msi/nunit/engine-files.wxi
```diff
@@ -87,5 +88,34 @@
...
+                      Source="$(var.InstallImage)bin/netcoreapp3.1/nunit.engine.core.dll" />
+            </Component>
+            <Component Id="NUNIT_AGENT_NETCORE31_ENGINE_METADATA" Location="local" Guid="43798734-A674-4D13-A1F5-B1EBB22B165A">
+                <File Id="nunit.agent.netcore31.engine.metadata.dll"
+                      ProcessorArchitecture="msil"
+                      Source="$(var.InstallImage)bin/netcoreapp3.1/testcentric.engine.metadata.dll" />
+            </Component>
+            <Component Id="NUNIT_AGENT_ADDINS_NETCORE31" Location="local" Guid="21D2D9FB-A57D-4B02-92A0-6697674FB869">
+                <File Id="nunit.agent.netcore31.addins" Source="$(var.InstallImage)nunit.agent.addins" />
+            </Component>
```
ChrisMaddock, line 118:
```suggestion
```

I think this addins file points to a dir of netfx extensions, none of which can be loaded by the netcoreapp31 agent. Shall we just leave this out for now? We might want to package netcoreapp-compatible extensions with the engine in future, but they'll probably need to be in a different subdir anyway.

nuget/engine/nunit.engine.nuspec
ChrisMaddock, line 61:
```suggestion
    <file src="bin/agents/netcoreapp3.1/nunit-agent.runtimeconfig.json" target="contentFiles/agents/netcoreapp3.1" />
```

I think this one's been missed, Charlie. (I also don't know why we use contentFiles instead of tools here, but keeping it consistent seems safest! 😅)

src/NUnitEngine/mock-assembly-x86/mock-assembly-x86.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <AssemblyOriginatorKeyFile>..\..\nunit.snk</AssemblyOriginatorKeyFile>
+    <PlatformTarget>x86</PlatformTarget>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="NUnit" Version="3.13.0" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="..\EngineVersion.cs" LinkBase="Properties" />
+    <Compile Include="..\mock-assembly\AccessesCurrentTestContextDuringDiscovery.cs" Link="AccessesCurrentTestContextDuringDiscovery.cs" />
+    <Compile Include="..\mock-assembly\MockAssembly.cs" Link="MockAssembly.cs" />
```
ChrisMaddock, line 15:
```suggestion
    <Compile Include="..\mock-assembly\*.cs" />
```

Maintenance suggestion: I believe the above will mean we don't need to add any potential new files to this list in future. 🙂

src/NUnitEngine/nunit-agent-x86/nunit-agent-x86.csproj
```diff
@@ -15,6 +14,8 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="..\EngineVersion.cs" LinkBase="Properties" />
+    <Compile Include="..\nunit-agent\AgentExitCodes.cs" Link="AgentExitCodes.cs" />
+    <Compile Include="..\nunit-agent\Program.cs" Link="Program.cs" />
```
ChrisMaddock, line 18:
```suggestion
    <Compile Include="..\nunit-agent\*.cs" />
```

As above.

src/NUnitEngine/nunit.engine.core/nunit.engine.core.csproj
ChrisMaddock: @CharliePoole - I think this one's been missed too. If I'm understanding right, your comment above refers to the logic in RuntimeFramework.cs. (And is very helpful, thank you! 😆)

My original comment on this file still stands.

ChrisMaddock (commented):
Great to see this all coming together Charlie! Just a few niggles at this point. 🙂 

Just checkpointing the review here as GitHub's gone a bit funny - still a bit more to go through...

src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs
ChrisMaddock: I think this change is still outstanding too, Charlie.

nuget/engine/nunit.engine.nuspec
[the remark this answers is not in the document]
ChrisMaddock: This change is still to do - "GitHub code suggestion" added in second review below, which hasn't linked up to this convo...

src/NUnitEngine/nunit.engine.core/RuntimeFramework.cs
ChrisMaddock: Comment at https://github.com/nunit/nunit-console/pull/937#discussion_r619679210 refers - resolving this one.

src/NUnitEngine/nunit.engine.core/RuntimeFramework.cs
ChrisMaddock: I think https://github.com/nunit/nunit-console/pull/937#discussion_r623511925 refers to this thread. Agree about separating this out from this PR, I'll create a new thread here: https://github.com/nunit/nunit-console/pull/937#discussion_r623511925

src/NUnitEngine/nunit.engine/Communication/Transports/Tcp/TestAgentTcpProxy.cs
ChrisMaddock: I think we should tackle this for 4.X. Would you be ok to create this issue? I think you have a clearer understanding of the core issue here than I do right now.

src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs
```diff
@@ -50,26 +53,37 @@ public bool IsAvailable(string name)
...
+
+            if (requestedVersion == AnyVersion)
+                return true;
+
+            return requestedVersion.Major == availableVersion.Major &&
+                   requestedVersion.Minor == availableVersion.Minor &&
+                   (requestedVersion.Build < 0 || availableVersion.Build < 0 || requestedVersion.Build == availableVersion.Build) &&
+                   (requestedVersion.Revision < 0 || availableVersion.Revision < 0 || requestedVersion.Revision == availableVersion.Revision) &&
+                   requestedVersion.Major == availableVersion.Major &&
+                   requestedVersion.Minor == availableVersion.Minor;
```
ChrisMaddock, line 72:
```suggestion
                   (requestedVersion.Revision < 0 || availableVersion.Revision < 0 || requestedVersion.Revision == availableVersion.Revision);
```

If I'm reading right (I may well not be, in this class! 😂), I think the last two checks have been included twice here?

src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs
ChrisMaddock: Thanks Charlie. This is indeed a can of worms. 😄 

The changes look good, and your explanation makes sense. I think there are two duplicate lines that are no longer needed - I've commented on those. 

The only thing that still makes me uneasy is the name "RuntimesMatch" for a method which returns true for two different runtimes. Likewise for "FrameworksMatch" to be honest. That said, this whole area really needs work, so if you'd prefer to leave this for now, I'm happy to go with that.

ChrisMaddock:
Ok - I _think_ I've covered everything here. 

I've resolved a few discussions that I think are sorted, and unresolved a couple that I think were missed. I've added a few new comments, almost entirely re: the new changes. I think I've also responded to all still-open conversations.

The x86 changes look good - glad finally we've got some tests around that too! In future, I'd like to add tests to mock-assembly which only pass if they are running on the targeted framework/architecture, so we can smoke test that functionality too.

Let me know if there's anything I've missed - let's get this merged! 😁

src/NUnitConsole/nunit3-console.tests/App.config
CharliePoole: I had to add it to make all tests pass. There's a test that reads it. In fact, this should eventually be cleaned up and we should stop using `.config` files entirely for .NET Core as a part of reworking RuntimeFramework. `.runtimeconfig.json` is where this should end up. 

Hard to figure out the boundaries of a PR... i.e. where do you stop and create a new issue. Personally, I'd prefer not to deal with this one while we still allow `--inprocess` because that makes RuntimeFramework more complicated. I do think that RuntimeFramework is a good place for me to exercise my preference for tiny PRs. :smiley:

src/NUnitEngine/nunit.engine.tests/nunit.engine.tests.csproj
CharliePoole: I had to add this in a few places before all the tests passed. I'll test removing them one at a time.

Alternatively, we could remove that single test from our .NET Core builds for the time being.

msi/nunit/engine-files.wxi
CharliePoole: I believe that `nunit-agent.deps.json` is only for use in building the project. It causes `nunit-agent.runtimeconfig.json` to be generated. If that's wrong, I can add it. Do you have different info? At this point all the tests pass using all the packages.

msi/nunit/engine-files.wxi
CharliePoole: Good point. I'll take a look. In fact, this directory should __only__ used for the .NET Core 3.1 agent, so I may be able to just change the content of either the `addins` file or the directory itself.

nuget/engine/nunit.engine.nuspec
CharliePoole: I initially tried to keep it consistent, but I got failures when I did that. However, I can now think of another reason for that, so I'll try again. If this were for 4.0, I'd say just drop the whole `contentFiles` thing unless somebody can say why.

src/NUnitEngine/mock-assembly-x86/mock-assembly-x86.csproj
CharliePoole: That's OK if we end up seeing the files in SolutionExplorer! I'll give it a try. That said, I think the first file, which is an actual test, is a misuse of mock-assembly, which was only ever intended as fodder for the engine to use in testing loading, running and XML generation... i.e. it's data!

CharliePoole:
I answered some questions and resolved a couple of items but I'll go
through all the items again. It's really hard to track all this with so
many changes, but I'm keeping a page open where I've marked all the ones I
no longer want to see as "viewed".

WRT testing that it really runs under the targeted runtime and
architecture...

I'd like to do that too, but when I took a casual look at the idea, it
seemed as if I would have to add more info to TestResult.xml to do that.
How about if I do whatever can be done without changing the XML format and
then trying to do more when we incorporate this in the 4.0 branch?

ChrisMaddock:
Thanks Charlie - and it is a lot to keep track of. Hopefully we'll catch most things between us!

> I'd like to do that too, but when I took a casual look at the idea, it
seemed as if I would have to add more info to TestResult.xml to do that.

Ah - I was thinking more simply a test which fails when it's run on e.g. .NET 2.0, and using pre-compilation symbols to make sure the right version is included in the right assembly. That's what I do in my manual tests before releases. 🙂

CharliePoole:
If I understand you, that would work if the test can run, e.g. net20 on net40. But if the test can't run at all, the package tests typically need to examine the TestResult file to find out __why__ it failed. That's because we keep running subsequent tests rather than throwing immediately. I may be able to figure out a way to recognize that the test failed because the agent couldn't load it though.

msi/nunit/engine-files.wxi
ChrisMaddock: Hmm, ok. I always thought it should be as it details where to load dependencies from at runtime. I'm struggling to find any formal documentation however, or any clear answer elsewhere. 

This page seems to be the closest there is. From my reading of that, for a framework-dependent deployment as ours is, the file should be included. What do you think?

https://github.com/dotnet/cli/blob/master/Documentation/specs/runtime-configuration-file.md

CharliePoole:
@ChrisMaddock This PR and reviews and comments on reviews etc. is getting a bit hard to follow. So to checkpoint where we are, here's a list of items you brought up, which I believe still need to be dealt with...

1. Figure out whether we need to package `nunit-agent.deps.json`. I have tests passing without it, but it won't hurt to add. We should either include it in all four packages or leave it out of all.

2. Decide about `.addins` files for agents. Given that we currently have no extensions of our own for .NET Core, we could just leave out the file. Or we could point to a directory that users can put things in... or... or... or... It would be nice if we didn't need to re-release the first time there's a .NET Core extension, although that's possible as well.

3. Decide whether to use `contentFiles` or tools for the `.json` files. My memory is that using `contentFiles` didn't work, but I'll retest that.

4. For `mock-assembly`, I'll test out your idea of just including the files in the parent dir without a link.

5. Check whether various references and package references are really needed for the tests to pass. I'm a little vague here because you hadn't followed up to some of my responses. Basically, however, I think we need to leave in any references if the tests start to fail when I remove them. It may be that this works differently if I include `.deps.json`, so I'll try with and without. That said, at some future point it would help if you could write something about __why__ those package refs and refs ought to be removed. Maybe that should be a 4.0 issue.

6. Remove extra checks from `RuntimeFrameworkService.FrameworksMatch`

7. Re-examine default behavior in `RuntimeFrameworkService.SelectRuntime` when no `RuntimeTargetAttribute` is present.

Let me know if I've missed something. :smiley:

ChrisMaddock:
That looks about right to me! 👍

2. I'm in favour of taking the easy way out for the msi, and just removing the file until we have extensions to go with it. 
4. I did try myself this before suggesting, and it seemed to work for me. 🙂 (Same suggestion for the new nunit-agent projects, too!)
5. Can you highlight to me which responses I've missed? I think this covers two separate issues:
5a. The references added to nunit.engine.core.csproj, which I believe are just a rebase error as I'd just removed those from master recently. 
5b. The reference added to nunit.engine.tests.csproj, and the new app.config added to the nunit-3console.tests project. Both of these - I'm unclear on why the tests have been passing previously (I think!), and why they're now needed. Would appreciate your thoughts there.

CharliePoole:
@ChrisMaddock Regarding (5) I probably didn't make myself clear. It's not a rebase error. I saw that you had removed them and I had to put them back in to the package tests for .NET Core to pass. There was no failure before because we didn't have such a test or even a .NET Core agent to fail!

That said, I'm taking on trying again without them and seeing if this interacts with the presence or absence of `nunit-agent.deps.json'. I think this is something we really need to understand better.

src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs
CharliePoole: Yes, I guess it still makes sense to use Current framework in 3.x, since `--inprocess` is still supported.

src/NUnitEngine/nunit.engine.core/nunit.engine.core.csproj
CharliePoole: It turns out that only the Microsoft.Win32.Registry reference was needed.

CharliePoole:
Should be ready to go now. I think we have a number of packaging decisions to make for 4.0, however.

ChrisMaddock:
Great, thanks for getting those sorted! Afraid there's a couple of things in the last commit I still think need looking at.

1. The latest commit has added **nunit-agent.runtimeconfig.dev.json** to the packages, rather than **nunit-agent.deps.json**. I presume that was unintentional, as you'd put deps.json in your list above? The "dev" file I believe is for development only.

1. Sorry, we've miscommunicated around the .addins files. By my suggestion, I was only meaning removing it from the msi, as we know that there are no .NET Core compatible extension assemblies installed with the msi. I see the last commit also removes it from the NuGet packages. I think the NuGet version needs to stay I'm afraid, I know that's currently used at least for the TeamCity extension.

Let me know if there's anything we need to discuss further - otherwise please go ahead and merge once the changes are in. Thanks for all your work getting it sorted! 😄

CharliePoole:
I'm afraid I thought I was doing what you wanted without really understanding why you wanted it. I'll take the blame for that choice. In future, I think it would help to have failing scenarios that force us to add something to the packages. After all, that's why we have package tests.

I'll make the changes in the morning.

ChrisMaddock:
Great - thanks for this Charlie! One step closer to where we want to be! 😁

Issue #939, opened 2021-04-27
NUnit.Engine 3.12.0 .NET Core and Binary serialization


bouchraRekhadda:
Hello,

When using **NUnit.Engine 3.12.0** to run tests (targeting **.NET Core 3.1**) doing binary serialization, tests fail with `System.Runtime.Serialization.SerializationException`. 

### How to reproduce
Sample code:
```
[Serializable]
public class MyCoolClass
{
    public MyCoolClass(string name)
    {
        Name = name;
    }
    public string Name { get; }
    public override bool Equals(object? obj)
    {
        return obj is MyCoolClass;
    }
    public override int GetHashCode()
    {
        return Name.GetHashCode();
    }
}
private static T SerializeAndDeserialize<T>(T t)
{
    var formatter = new BinaryFormatter();
    byte[] array;
    using (var ms = new MemoryStream(16 * 1084))
    {
        formatter.Serialize(ms, t);
        array = ms.ToArray();
    }
    using (var ms = new MemoryStream(array))
    {
        return (T)formatter.Deserialize(ms);
    }
}
[Test] public void TestSerialization()
{
    var instance = new MyCoolClass("Class1");
    var deserialize = SerializeAndDeserialize(instance);
    Assert.AreEqual(instance, deserialize);
}
```
Then run test above using [NUnit.ConsoleRunner.NetCore](https://www.nuget.org/packages/NUnit.ConsoleRunner.NetCore/) (which is using NUnit.Engine 3.12.0). For my case it was:

`%userprofile%\.nuget\packages\nunit.consolerunner.netcore\3.12.0-beta1\tools\nunit3-console.exe D:\ForNUnit\bin\Debug\netcoreapp3.1\UnitTests.dll`

### Expected behavior
![image](https://user-images.githubusercontent.com/14876573/116263052-9731e900-a768-11eb-9f10-8961eb8edd91.png)

### Actual behavior
>Error : UnitTests.Test.TestSerialization
System.Runtime.Serialization.SerializationException : Unable to find assembly 'UnitTests, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'.
   at System.Runtime.Serialization.Formatters.Binary.BinaryAssemblyInfo.GetAssembly()
   at System.Runtime.Serialization.Formatters.Binary.ObjectReader.GetType(BinaryAssemblyInfo assemblyInfo, String name)
[...]

Issue #916, opened 2021-03-09
NUnit.Engine 3.12.0 .NET Core and TypeDescriptor.GetConverter([...])


Learfin:
In .NET Core, NUnit.Engine 3.12.0 causes some problems when the tested code calls [TypeDescriptor.GetConverter([...])](https://docs.microsoft.com/fr-fr/dotnet/api/system.componentmodel.typedescriptor.getconverter?view=netcore-3.1).
For info, TypeDescriptor.GetConverter([...]) is used by [JSonConvert.SerializeObject([...])](https://www.newtonsoft.com/json/help/html/Overload_Newtonsoft_Json_JsonConvert_SerializeObject.htm) (from [NewtonSoft.Json](https://www.newtonsoft.com/json/)).
 
I use the following code for my tests:
```
using System;
using System.ComponentModel;
using System.Globalization;
using NUnit.Framework;

namespace MyNUnitTestProject
{
    [TestFixture]
    public class Tests
    {
        private sealed class MyClassTypeConverter : TypeConverter
        {
            public override object ConvertTo(ITypeDescriptorContext context, CultureInfo culture, object value, Type destinationType)
            {
                if (destinationType == typeof(string))
                {
                    var toSerialize = (MyClass)value;
                    return toSerialize.Value;
                }
                throw new NotImplementedException();
            }
        }
        [TypeConverter(typeof(MyClassTypeConverter))]
        private sealed class MyClass
        {
            public MyClass()
            {
                Value = Guid.NewGuid().ToString();
            }
            public string Value { get; }
        }
        [Test] public void ShouldPass()
        {
            var myClass = new MyClass();
            var converted = TypeDescriptor.GetConverter(typeof(MyClass)).ConvertTo(null, null, myClass, typeof(string));
            Assert.AreEqual(myClass.Value, converted as string);
        }
    }
}
```

### Problem 1 - Failure in NUnit.ConsoleRunner.NetCore (version 3.12.0-beta2)
For this test, the code above is compiled with the following csproj (created with dotnet test):
```
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
   <TargetFramework>netcoreapp3.1</TargetFramework>
   <IsPackable>false</IsPackable>
  </PropertyGroup>
  <ItemGroup>
   <PackageReference Include="NUnit" Version="3.12.0" />
   <PackageReference Include="NUnit3TestAdapter" Version="3.17.0" />
[...]

Issue #843, opened 2020-12-17
NUnit.ConsoleRunner.NetCore cannot find extensions


NikolayPianikov:
Latest build cannot find any extensions. We should add the hint "../../../../../NUnit.Extension.*/**/tools/" to the file ".\tools\netcoreapp3.1\any\nunit.console.nuget.addins"

[InternalTrace.27768.log](https://github.com/nunit/nunit-console/files/5709742/InternalTrace.27768.log)

CharliePoole:
@NikolayPianikov 
Since you use NUnit.ConsoleRunner.NetCore I'd like to ask for your comments on #1045

NikolayPianikov:
@CharliePoole 

> @NikolayPianikov Since you use NUnit.ConsoleRunner.NetCore I'd like to ask for your comments on #1045

done

CharliePoole:
@NikolayPianikov Are you sure this is an issue of finding extensions or simply that there are no .NET Core extensions avaialble?

CharliePoole:
@NikolayPianikov Ping!

NikolayPianikov:
@CharliePoole Maybe you are right, I will try reproduce it again

NikolayPianikov:
I just tried it and it looks like you are right about .NET Standard. I close this ticket.

CharliePoole:
We can't support .NET Standard for test assemblies so think the same limitation was built into extensions. Possibly, we should allow .NET Standard for extensions but it would require a bit of rewriting. Let me know if the non-support becomes a problem.

PR #835, opened 2020-11-22
[CI] Test on .NET 5.0

ChrisMaddock:
Fixes #830 and also fixes #511 

This runs the tests on .NET 5.0, where available. I thought this was important to at least test on .NET 5.0, as it's a platform I suspect a large number of our users will shortly be running on. I haven't yet seen any reason to add a .NET 5.0 build though - so this instead runs tests for the .NET Core 3.1 Engine/Console on .NET 5.0. I also didn't see a need to add a requirement for users to have .NET 5.0 installed to contribute, so if unavailable, the default is to provide a warning - assuming the tests will be run on all platforms in CI instead.

It also improves the CI's handling of .NET Core platforms, as per #511. The new behaviour: 
- The Cake script will now locate the most appropriate runtime for the request - that being the highest major/minor version installed for the runtime requested. e.g. on my machine, the 3.1 tests will chose to run on 3.1.5 over 3.1.2.
- If a runtime isn't available, e.g. .NET 5.0:
     - When building locally through the Cake scripts, if a certain runtime isn't available the script will provide a warning.
     - When building in CI and a certain runtime isn't available, the script will error. This is with various known exceptions built in, such as the .NET Core 1.1 test running on Azure.

One day I hope the Engine itself will do all this boilerplate stuff for us...we'll get there! 😄

PR #942, opened 2021-05-02
Use EnterContextualReflection to address .NET Core dependency loading issues

### LINKED ISSUE #828: Reported issues with dependency loading in .NET Core Console
Users have reported issues with assembly loading in the .NET Core Console. Ref: https://github.com/nunit/nunit-console/issues/806#issue-671998124 and https://github.com/nunit/nunit-console/issues/806#issuecomment-674191820.

Workaround: add `<CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>` to the test project csproj.

Contributions very welcome to fix this - and if anyone is able to provide a small minimum repro project for this, that would also be very helpful.

ChrisMaddock:
Hopefully fixes #828 and #916 

This uses the new EnterContextualReflection API introduced in netcore30, which will hopefully solve our .NET Core dependency loading issues. I wasn't aware of this previously as I don't think Microsoft's docs have yet been updated to include this - however there's some design documentation at the link below. (Thanks @Learfin for mentioning it in your write-up!) From what I understand, the API was added to tackle issues with xUnit that seem to be very similar to the problems we've been hitting.

https://github.com/dotnet/coreclr/blob/master/Documentation/design-docs/AssemblyLoadContext.ContextualReflection.md

I wanted to add a test for this, but unfortunately it's not really possible our current build structure. It's another example where a proper suite of acceptance test assembly's would be really handy - hopefully we'll get round to that one day.

Note the only change is in the final commit - this is just branched off #923 as they will interact. Once #923 is merged, I'll rebase this.

-------

@Learfin, @joeltankam, @VoltanFr, @bouchraRekhadda, @runehalfdan, @webMan1 - I gather you've all encountered this issue in one form or another. I've tested against the repro's provided in #916 and #939, but would really appreciate anyone who could test against their particular usecase before merging this. You should be able to download updated packages from the CI build here: https://ci.appveyor.com/project/CharliePoole/nunit-console/builds/38972209/artifacts

### DIFF
--- a/src/NUnitEngine/nunit.engine.core/Drivers/NUnitNetCore31Driver.cs
+++ b/src/NUnitEngine/nunit.engine.core/Drivers/NUnitNetCore31Driver.cs
@@ -38,6 +38,7 @@ namespace NUnit.Engine.Drivers
         Assembly _frameworkAssembly;
         object _frameworkController;
         Type _frameworkControllerType;
+        CustomAssemblyLoadContext _assemblyLoadContext;
 
         /// <summary>
         /// An id prefix that will be passed to the test framework and used as part of the
@@ -56,10 +57,11 @@ namespace NUnit.Engine.Drivers
             var idPrefix = string.IsNullOrEmpty(ID) ? "" : ID + "-";
 
             assemblyPath = Path.GetFullPath(assemblyPath);  //AssemblyLoadContext requires an absolute path
-            var assemblyLoadContext = new CustomAssemblyLoadContext(assemblyPath);
+            _assemblyLoadContext = new CustomAssemblyLoadContext(assemblyPath);
+
             try
             {
-                _testAssembly = assemblyLoadContext.LoadFromAssemblyPath(assemblyPath);
+                _testAssembly = _assemblyLoadContext.LoadFromAssemblyPath(assemblyPath);
             }
             catch (Exception e)
             {
@@ -67,11 +69,11 @@ namespace NUnit.Engine.Drivers
             }
 
             var nunitRef = _testAssembly.GetReferencedAssemblies().FirstOrDefault(reference => reference.Name.Equals("nunit.framework", StringComparison.OrdinalIgnoreCase));
-            
+
             if (nunitRef == null)
                 throw new NUnitEngineException(FAILED_TO_LOAD_NUNIT);
 
-            _frameworkAssembly = assemblyLoadContext.LoadFromAssemblyName(nunitRef);
+            _frameworkAssembly = _assemblyLoadContext.LoadFromAssemblyName(nunitRef);
             if (_frameworkAssembly == null)
                 throw new NUnitEngineException(FAILED_TO_LOAD_NUNIT);
 
@@ -112,7 +114,7 @@ namespace NUnit.Engine.Drivers
         }
 
         /// <summary>
-        /// Executes the tests in an assembly asyncronously.
+        /// Executes the tests in an assembly asynchronously.
         /// </summary>
         /// <param name="callback">A callback that receives XML progress notices</param>
         /// <param name="filter">A filter that controls which tests are executed</param>
@@ -179,7 +181,11 @@ namespace NUnit.Engine.Drivers
             {
                 throw new NUnitEngineException(INVALID_FRAMEWORK_MESSAGE);
             }
-            return method.Invoke(_frameworkController, args);
+
+            using (_assemblyLoadContext.EnterContextualReflection())
+            {
+                return method.Invoke(_frameworkController, args);
+            }
         }
     }
 }


ChrisMaddock:
CI is very confusing. Looks like Travis is failing due to the changes in #835 - and for some reason has just started running again, after not running for the last 7 months?

No idea what's caused that. I think the issue is just that net50 and netcoreapp21 aren't installed. I think this is probably our cue to remove Travis, given we're now maintaining 3 separate CI systems. As far as I'm aware, everything tested on Travis is also covered by Azure.

CharliePoole:
How come this PR is including my changes on the other PR?

CharliePoole:
OK, never mind. I see you have a comment about that in the issue.

bouchraRekhadda:
Hello @ChrisMaddock, thank you for the fix.
I've tried to run tests (similar to ones described in #939 ) through the package [3.13.0](https://ci.appveyor.com/project/CharliePoole/nunit-console/builds/38972209/artifacts) you have provided us, unfortunately it crashed before launching any test:
>NUnit.Engine.NUnitEngineException : The NUnit 3 driver encountered an error while executing reflected code.
  ----> System.InvalidCastException : Unable to cast transparent proxy to type 'System.Web.UI.ICallbackEventHandler'.
--NUnitEngineException
The NUnit 3 driver encountered an error while executing reflected code.
Server stack trace:
   at NUnit.Engine.Drivers.NUnit3FrameworkDriver.CreateObject(String typeName, Object[] args)
   at NUnit.Engine.Drivers.NUnit3FrameworkDriver.Load(String testAssemblyPath, IDictionary`2 settings)
   at NUnit.Engine.Runners.DirectTestRunner.LoadDriver(IFrameworkDriver driver, String testFile, TestPackage subPackage)
   at NUnit.Engine.Runners.DirectTestRunner.LoadPackage()
   at NUnit.Engine.Runners.DirectTestRunner.EnsurePackageIsLoaded()
   at NUnit.Engine.Runners.DirectTestRunner.RunTests(ITestEventListener listener, TestFilter filter)
   at System.Runtime.Remoting.Messaging.StackBuilderSink._PrivateProcessMessage(IntPtr md, Object[] args, Object server, Object[]& outArgs)
   at System.Runtime.Remoting.Messaging.StackBuilderSink.SyncProcessMessage(IMessage msg)
Exception rethrown at [0]:
   at NUnit.Engine.ITestEngineRunner.Run(ITestEventListener listener, TestFilter filter)
   at NUnit.Engine.Runners.ProcessRunner.RunTests(ITestEventListener listener, TestFilter filter)
--
InvalidCastException
Unable to cast transparent proxy to type 'System.Web.UI.ICallbackEventHandler'.
   at NUnit.Framework.Api.FrameworkController.LoadTestsAction..ctor(FrameworkController controller, Object handler) in C:\src\nunit\nunit\src\NUnitFramework\framework\Api\FrameworkController.cs:line 504
Test Run Summary
  Overall result: Failed
  Test Count: 0, Passed: 0, Failed: 0, Warnings: 0, Inconclusive: 0, Skipped: 0
 
 
 One last thing, I've noticed that it claims the environment is .NET Framework instead of .NET Core 3.1 
image
As opposed to [Nunit.ConsoleRunner.NetCore 3.12.0-beta2](https://www.nuget.org/packages/NUnit.ConsoleRunner.NetCore/):
image.

ChrisMaddock:
Thanks for trying it out @bouchraRekhadda! Just to check, that's definitely with the NUnit.ConsoleRunner.NetCore package, and not the normal one?

Could you come up with a minimal example that reproduces this, please? Unfortunately there's not enough info in the stack trace to work out what's going wrong.

bouchraRekhadda:
Hello @ChrisMaddock,
>Could you come up with a minimal example that reproduces this, please? Unfortunately there's not enough info in the stack trace to work out what's going wrong

I'll try to provide tomorrow a reproduction code for binary serialization particular case.

Regarding other use cases with assembly load issues, we have one more where we use code compilation at runtime. Please find more details in [nunit-sample](https://github.com/bouchraRekhadda/nunit-sample) repository with a little description of how to reproduce in the README (I hope this will be helpful).

runehalfdan:
I haven't managed to catch up with all issues that is fixed in this version of the console runner, so could be I have tested things that is known not to work

But I downloaded it and gave it a run. Had to add <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies> to all test-projects to avoid FileNotFoundException: 
`System.IO.FileNotFoundException : Could not load file or assembly 'nunit.framework, Version=3.13.2.0,`

Also still cannot run tests out-of-process. (Since we build all our test-assemblies to separate outputs, some shared dlls need to have some static data initialized, and the runtime gets confused about this; first test-assembly works, the seconds don't initialize the static, but don't find the data either)

Also, for Bamboo reasons, we need to have NUnit2 output. 
And Nunit project files would be nice, but any methods that allows use to list all test-assemblies in a file is ok.

bouchraRekhadda:
Hello @ChrisMaddock 

Regarding the [above stack-trace](https://github.com/nunit/nunit-console/pull/942#issuecomment-831322270), and after pulling my hair out, I finally understood why it occurred: we have disabled the Target  Framework attribute generation through `<GenerateTargetFrameworkAttribute>False</GenerateTargetFrameworkAttribute>`, which I believe confused NUnit.Console into thinking it is a .NET Framework assembly instead of .NET Core.

So after enabling the attribute generation, I can finally confirm that the current version **3.13.0** fixes the serialization issue we had (cf. #939 ) 🎉.

CharliePoole:
Wow! Just read this and realized we have another way we should test the engine... disabling generation of TargetFrameworkAttribute! Thanks for that @bouchraRekhadda.

Yes, The engine currently assumes the current framework. But the change you are testing was based on an earlier version of my unmerged PR #937 which changes that to assume .NET Framework, since I've only ever seen a missing attribute for netfx.

Really, we shouldn't assume anything but when the attribute is not there, we should go further in trying to figure out what the assembly actually targets. :wink:

bouchraRekhadda:
>Regarding other use cases with assembly load issues, we have one more where we use code compilation at runtime. Please find more details in [nunit-sample](https://github.com/bouchraRekhadda/nunit-sample) repository with a little description of how to reproduce in the README (I hope this will be helpful).

Regarding this scenario (code emission at runtime), unfortunately I still reproduce the failures with **3.13.0**.
In [UnitTests.EmitDemo](https://github.com/bouchraRekhadda/nunit-sample/blob/master/UnitTests/EmitDemo.cs) I basically compile a VB code that's referencing a .NET Core 3.1 library project class, I do an [`Assembly.LoadFrom()`](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.loadfrom?view=net-5.0) that ends up adding my compiled VB assembly in the `AssemblyLoadContext.Default`, when I try to get assembly types it returns `null`.

@ChrisMaddock, @CharliePoole what do you advise for this case ?

ChrisMaddock:
@bouchraRekhadda Great news we've now got the normal case up and running! 🎉 Your compilation at run-time case, afraid I have no idea how to get that working. 😅 I would have thought `EnterContextualReflection()` would have encapsulated something like that - but it sounds like it doesn't. I think what we have here is an improvement, so lets merge this and move the new case to a new issue - would appreciate any ideas you or your team have to get that example working too!

-----

@runehalfdan Hi - good to hear from you again! 👋 

Please could you elaborate on "Had to add true to all test-projects"? It would be really helpful if you could add a small repo for this problem (in a new issue, perhaps!). As things stand, we don't currently have any more reproducible examples which are not working, other than @bouchraRekhadda's code-compliation-at-runtime, which I expect is a very different problem!

On the other points you mention:
- Out of process running is now that bit closer, thanks to Charlie's work in #937.
- nunit2 and project files I believe are just awaiting an interested contributor to convert the extensions. Although @CharliePoole may know more about where those are at - he currently maintains those extension.

-----

@CharliePoole - What are the chances that the first user to try this would encounter the TargetFrameworkAttribute issue we'd just been talking about, eh? 😄

ChrisMaddock:
I'm going to merge this, as it seems to be at least one step closer to where we want to be. Thanks to those who've been able to try it out. 🙌

runehalfdan:
> @runehalfdan Hi - good to hear from you again! 👋 
> 
> 
> 
> Please could you elaborate on "Had to add true to all test-projects"? It would be really helpful if you could add a small repo for this problem (in a new issue, perhaps!). As things stand, we don't currently have any more reproducible examples which are not working, other than @bouchraRekhadda's code-compliation-at-runtime, which I expect is a very different problem!
> 
> 
> 
> On the other points you mention:
> 
> - Out of process running is now that bit closer, thanks to Charlie's work in #937.
> 
> - nunit2 and project files I believe are just awaiting an interested contributor to convert the extensions. Although @CharliePoole may know more about where those are at - he currently maintains those extension.

Edited original;markdown problems. Was CopyLocalLockFileAssemblies that must be true.

bouchraRekhadda:
Hello @ChrisMaddock, first of all thank you for the fix, I can't wait to update our Test Runner with this new version.

Regarding this: 
>Your compilation at run-time case, afraid I have no idea how to get that working. 😅 I would have thought EnterContextualReflection() would have encapsulated something like that - but it sounds like it doesn't. I think what we have here is an improvement, so lets merge this and move the new case to a new issue - would appreciate any ideas you or your team have to get that example working too!

Always using version **3.13.0** provided through this pull request, we have tried enforcing the current reflection context by simply doing
`var assembly = (AssemblyLoadContext.CurrentContextualReflectionContext ?? AssemblyLoadContext.Default).LoadFromAssemblyPath(assemblyPath);` instead of [hard-coding the NUnit ALC](https://github.com/bouchraRekhadda/nunit-sample/blob/973fe78613c58992f19ee637e5d6121a4f3c4ca0/UnitTests/EmitDemo.cs#L75), and it worked 🙆‍♀️ (the change is available [here](https://github.com/bouchraRekhadda/nunit-sample/commit/99705bf56ea00be9c225c0460fa38e0a770ab508#diff-24900bcb5ac9fc913cb9b368d658b03d907dc770f142f52acc68cd4c183305b1R70)).

bouchraRekhadda:
Hello, @ChrisMaddock, @CharliePoole,
Do you know when will **3.13.0** be publicly released ? we could use a beta as well if you think the official release will take time.

Thanks

ChrisMaddock:
Hi @bouchraRekhadda - probably not imminently I'm afraid, there's a few more issue we need to look into first like #956 and #843. You can get pre-release builds from our MyGet feed though, that should cover what you need? 🙂 

https://www.myget.org/feed/nunit/package/nuget/NUnit.ConsoleRunner.NetCore

bouchraRekhadda:
>You can get pre-release builds from our MyGet feed though, that should cover what you need?

We were hoping to get a beta release like [`3.12.0.beta2`](https://www.nuget.org/packages/NUnit.ConsoleRunner.NetCore/).

To be more precise, we only need [NUnit.Engine](https://www.nuget.org/packages/NUnit.engine/) including the fix from this pull request; we could re-package internally the `nunit.engine.dll` from [NUnit.ConsoleRunner.NetCore 3.13.0-dev-05299](https://www.myget.org/feed/nunit/package/nuget/NUnit.ConsoleRunner.NetCore) and use it instead in our custom Test Runner, but we prefer to avoid this kind of hacks if possible.

bouchraRekhadda:
Hello,
Any update on my [request above](https://github.com/nunit/nunit-console/pull/942#issuecomment-842451617) please ?

Thanks

Issue #946, opened 2021-05-09
Allow test engine to be created inside a running test on .NET Core


ChrisMaddock:
This is a useful feature when creating a runner, to be able to do full end-to-end tests.

There is a test for this named `NUnit.Engine.Tests.Acceptance.InstantiateEngineInsideTest.CanInstantiateEngineInsideTest` which is currently disabled for .NET Core. 

The issue is that the TestEngineActivator is unable to be cast the engine to an ITestEngine instance - likely due to incorrect types being in scope somewhere. Unfortunately I've had to break this out of my current work as a separate issue.

ChrisMaddock:
Just found that `NUnit.Engine.Tests.WorkingDirectoryTests.EngineCanBeCreatedFromAnyWorkingDirectory` is also affected. (Test was previously silently returning a null.)

CharliePoole:
The workaround we have used in the past is to provide a unique name to the TestAgency when it is created. However, this is only possible if your test is creating the engine, not the SUT.

JamesChristie:
Thanks for the reply!

Currently we have our main project in the solution that defines our custom runner and some test cases itself for verifying some system behavior elsewhere. We have a unit test project alongside it that has an integration test case where we try and create our entry point and run a set of tests. That's where everything breaks, since we're trying to test a project that is, itself, utilizing NUnit for our purposes.

Would you be able to expand a bit on the `TestAgency` thing you mentioned? (That or link me to the possibly obvious documentation I may have missed. 😑 )

JamesChristie:
Thank you so much for the detailed response! I'll mess with this a bit on my end and see if I can't get our test working again. 🙂

ChrisMaddock:
Just to add to this - are you using the .NET Core engine @JamesChristie, or .NET Framework? 

The .NET Framework version has the workaround Charlie describes. It should be fixed by #947, that will be in the next release, or you can use the pre-release version from our MyGet feed (see the README for details there) 

If it's the .NET Core version of the engine, then afraid I'm not sure why that doesn't work at the moment, so don't have a workaround for it. I think it's a separate problem to the TestAgency one though, as the .NET Core version doesn't create the remoting URI which causes the failure on .NET Framework.

PR #947, opened 2021-05-09
Make default TestAgency URI unique per run (v2)

### LINKED ISSUE #764: Error: Found two different objects associated with the same URI, /xxxxx/TestAgency
I am using Nunit API in my application to run test cases. 
When more than one thread of execution reaches at this point
runner = engine.GetRunner(package);
I get this error: Found two different objects associated with the same URI, /xxxxx/TestAgency

Looks like engine is not able to provide more than one runner to run testcases.

ChrisMaddock:
Replacement for #931. Fixes #764 

This change allows an engine to be created within a running NUnit Test, on .NET FX. Unfortunately I had to cut my losses on supporting the same in .NET Core after several hours of reflection woes - have broken that out into #946.

This is a useful feature for acceptance style testing of runners - we have various examples in our own console tests where we've had to interact with non-public engine classes to test, which we ideally want not to be necessary.

The first commit adds a failing test, which is fixed in the second commit. The third commit removes a couple of workarounds we've previously built in to get around this problem.


ChrisMaddock:
Thanks Charlie! I've resolved conflicts and set this to auto-merge when CI passes.

PR #952, opened 2021-05-09
Dispose classes as required in Tests to prevent failed app domain unload

### LINKED ISSUE #949: nunit.engine.tests are failing to unload on master branch
The nunit.engine.tests are currently failing to unload incorrectly on the master branch. I haven't had the time yet to bisect and find out where this crept in - if anyone else has the time to take a look, please feel free!

```
NUnit.Engine.NUnitEngineUnloadException : Agent Process was terminated successfully after error.
  ----> NUnit.Engine.NUnitEngineUnloadException : Exception encountered unloading application domain
  ----> NUnit.Engine.NUnitEngineException : Exception encountered unloading application domain: Error while unloading appdomain. (Exception from HRESULT: 0x80131015)
Application domain name: domain-
Application domain BaseDirectory: D:\a\1\s\bin\Release\net35\
Error trying to read application domain details: Could not load file or assembly 'nunit.engine, Version=3.13.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb' or one of its dependencies. The system cannot find the file specified.
```

ChrisMaddock:
Fixes #949 

I believe the new TCP functionality in #937 now cause the test app domain to fail to unload if certain classes aren't Disposed. I don't think that's a problem - all the classes in question already implemented IDisposable, so to my mind should have be disposed of properly in tests anyway.

### DIFF
--- a/src/NUnitConsole/nunit3-console.tests/ConsoleRunnerTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/ConsoleRunnerTests.cs
@@ -17,16 +17,17 @@ namespace NUnit.ConsoleRunner.Tests
         [Test]
         public void ThrowsNUnitEngineExceptionWhenTestResultsAreNotWriteable()
         {
-            var testEngine = new TestEngine();
+            using (var testEngine = new TestEngine())
+            {
+                testEngine.Services.Add(new FakeResultService());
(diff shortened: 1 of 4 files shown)


mikkelbu (approved):
LGTM

PR #954, opened 2021-05-09
Update TestCentric and NUnit packages

### LINKED ISSUE #933: Update TestCentric MetaData dependency
To version 1.7.0 or later, to resolve an issue with System.Func being exposed to runners targeting .NET 3.5.

ChrisMaddock:
Fixes #933...while I'm doing a bit of janitoring anyway. 😄


mikkelbu (approved):
Only a comment, but I think this is fine as it is.

ChrisMaddock:
Thanks Mikkel - good idea! Done. 🙂

PR #958, opened 2021-05-18
Fix counting of warnings in aggregation of multiple assemblies

### LINKED ISSUE #904: The test-run element is missing a count of warnings
The error is in an overload of `Resulthelper.Aggregate` found at https://github.com/nunit/nunit-console/blob/master/src/NUnitEngine/nunit.engine.core/Internal/ResultHelper.cs#L134

The accumulation of counters doesn't look at the `warnings` attribute at all.

Once this is fixed, a workaround in `test-results.cake` script can be removed.

Based on a review of the code, this would appear to affect the totals for any `test-suite` elements with `type=Project` as well.

CharliePoole:
Fixes #904

### DIFF
--- a/package-tests.cake
+++ b/package-tests.cake
@@ -90,6 +90,8 @@ public abstract class PackageTester
                 var report = new TestReport(packageTest, result);
                 reporter.AddReport(report);
 
+                Console.WriteLine($"Ran tests in {_outputDir}");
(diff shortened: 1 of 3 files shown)


ChrisMaddock (approved):
Looks good! 👍

PR #1013, opened 2021-09-05
Fix paths in addins file for NUnit.ConsoleRunner.NetCore nuget package

k15tfu:
cc @ChrisMaddock


nuget/runners/nunit.console.nuget.addins
```diff
@@ -1,4 +1,6 @@
 ../../NUnit.Extension.*/**/tools/        # nuget v2 layout
 ../../../NUnit.Extension.*/**/tools/     # nuget v3 layout
+../../../../NUnit.Extension.*/**/tools/
```
k15tfu, line 3: @ChrisMaddock Hi! BTW, in another file nunit.engine.nuget.addins this line also has _nuget v3 layout_ comment. Should I add it here?

rprouse (approved):

Issue #915, opened 2021-03-07
StackOverflowException in console-runner when addins-file contains "./" or ".\"


x789:
# Observed Behavior
The console-runner crashes due to a StackOverflowException when an addins-file in the executable's directory contains a reference to its directory (i.e. `./`or `.\`).

I only checked the .NET 2.0 and .NET Core 3.1 builds on Microsoft Windows but suspect that all target platforms and frameworks are affected. I compiled the code from the `master` branch at commit c533b4e.
```text
# stackoverflow.addins
./
```

Outputs:
```text
bin\Release\net20> .\nunit3-console.exe --trace=Debug .\nunit.engine.tests.dll
NUnit Console Runner 3.13.0 (.NET 2.0)
Copyright (c) 2021 Charlie Poole, Rob Prouse
Sunday, March 7, 2021 1:11:55 PM


Process is terminated due to StackOverflowException.
```

```text
bin\Release\netcoreapp3.1> .\nunit3-console.exe --trace=Debug .\nunit.engine.tests.dll
NUnit Console Runner 3.13.0
Copyright (c) 2021 Charlie Poole, Rob Prouse
Sunday, March 7, 2021 4:31:45 PM

Stack overflow.
```

# Expected Behavior
The console-runner does not crash but parses the current directory.

x789:
I think I can fix the issue when someone assigns it to me.

mikkelbu:
I've assigned the issue to you @x789. Please ask if you have any questions

CharliePoole:
@x789 @mikkelbu I think this is a bug but I don't agree with the stated expected behavior.

IMO having an addins file that points to the directory containing it is a configuration error - as is any combination of directories and addin files, which lead to a loop.

The processing of a directory for finding addins is defined as
* If the directory contains one or more addins files, process them
* If not, process all dlls in the directory

I do agree that we should handle this better and not have a stack overflow. We already ignore directories we visit more than once but only __after__ the visit is complete. We could fix that code so that the same directory is ignored __while__ it's the directory we are parsing.

Hypothetically, we could instead give an error message, but I believe that would be breaking behavior. Also, use of wild-cards means that we are often looking at the name of a directory more than once.

@x789 Are you attempting to process both dlls and addins files in the same directory? If you want to do that, I think you should create another addins file that points to those assemblies. In fact, I often package extensions themselves with an addins file so that the engine doesn't try to parse referenced files packaged in the same directory.

x789:
[...]

PR #1024, opened 2021-10-20
Fixed a stack overflow when an entry of an addins-file points to an already visited directory.

x789:
This fixes issue #915.

The `ExtensionService` now adds all visited directories to the list of visited items and checks this list before a directory is processed. That avoids an infinite loop and the resulting `StackOverflowException`.
The already existing `_visited` list was re-used but adapted to keep track if a directory was visited as part of the 'wildcard' or 'no-wildcard' scenario.

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/Services/ExtensionServiceTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Services/ExtensionServiceTests.cs
@@ -635,7 +635,8 @@ namespace NUnit.Engine.Services.Tests
         }
 
         [Test]
-        public void ProcessAddinsFile_Issue915_AddinsFilePointsToContainingDirectory()
+        [Platform("win")]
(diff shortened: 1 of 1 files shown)


CharliePoole:
Continuing to see the MacOs failure at line 687. The linux failure looks like a configuration error... probably our problem to fix.

x789:
I forgot that there are two different control-flows inside `ExtensionService` because the logic treats `/` as a 'magic' character. Therefore I added a separate test-case for non-Windows systems (that use `/` as directory separator).

macOS and Windows builds are working now. I also believe that the Linux build is failing because of a problem with the pipeline configuration. Unfortunately, I guess I can't support at this point.

CharliePoole:
Surprisingly, I'm not able to merge this without the final check passing. @ChrisMaddock You have this set up so that nobody, not even admins can override the failed checks. That's a change as far as I know. I don't want to modify what you have decided to set up, so I'll leave it as is. However, not having any way out seems like a bad strategy. :-)

PR #1025, opened 2021-10-21
chore(pipeline): Use ubuntu-latest

mikkelbu:
Try to use the same pipeline setup as for MacOS


CharliePoole:
Looks like you fixed the pipeline but we have some test failures under Linux. If you don't want to work on them, they could just be skipped under linux for now.

PR #1032, opened 2021-11-08
Restore MyGet push url to its original value

### LINKED ISSUE #1031: Fix incorrect MyGet push URL
The last two merges failed to push the artifacts to MyGet because of an incorrect push URL. We should continue to push to https://www.myget.org/F/nunit/api/v2/package 

The changes from those two failed pushes are already in master, so fixing this should bring the latest myget dev release up to date.

CharliePoole:
Fixes #1031

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,7 +13,7 @@ artifacts:
 
 deploy:
   - provider: NuGet
-    server: https://www.myget.org/F/nunit/api/v3/index.json
+    server: https://www.myget.org/F/nunit/api/v2
     api_key:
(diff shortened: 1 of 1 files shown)


CharliePoole:
@rprouse @ChrisMaddock @mikkelbu @jnm2 

No more merges will succeed in pushing to myget as dev builds until this PR merges. In fact, we won't even know if this works until it merges! I wish there were a better way to test this kind of infrastructure work!

rprouse (approved):

rprouse:
I'm not sure about AppVeyor, but if we switch some of our builds to GitHub Actions, there is a tool that you can run locally to test your changes. It is an issue though. I know I've merged a few similar PRs in my day to follow on with a fix.

PR #1036, opened 2021-11-16
Run assemblies from a project file in parallel

### LINKED ISSUE #382: Only 1 agent running with multiple projects and --process=Multiple

CharliePoole:
Fixes #382

### DIFF
--- a/src/NUnitEngine/nunit.engine.core/Runners/AggregatingTestRunner.cs
+++ b/src/NUnitEngine/nunit.engine.core/Runners/AggregatingTestRunner.cs
@@ -56,7 +56,6 @@ namespace NUnit.Engine.Runners
                 {
                     _runners = new List<ITestEngineRunner>();
                     foreach (var subPackage in TestPackage.Select(p => !p.HasSubPackages()))
-                    //foreach (var subPackage in TestPackage.SubPackages)
                     {
                         _runners.Add(CreateRunner(subPackage));
                     }
--- a/src/NUnitEngine/nunit.engine.core/Runners/AggregatingTestRunner.cs
+++ b/src/NUnitEngine/nunit.engine.core/Runners/AggregatingTestRunner.cs
@@ -55,7 +55,8 @@ namespace NUnit.Engine.Runners
                 if (_runners == null)
                 {
                     _runners = new List<ITestEngineRunner>();
-                    foreach (var subPackage in TestPackage.SubPackages)
+                    foreach (var subPackage in TestPackage.Select(p => !p.HasSubPackages()))
+                    //foreach (var subPackage in TestPackage.SubPackages)
                     {
                         _runners.Add(CreateRunner(subPackage));
                     }
--- a/src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/TestCases/NetStandardTestCases.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/TestCases/NetStandardTestCases.cs
@@ -79,17 +79,10 @@ namespace NUnit.Engine.Tests.Services.TestRunnerFactoryTests.TestCases
                     new RunnerResult
                     {
                         TestRunner = typeof(AggregatingTestRunner),
-                        SubRunners = new List<RunnerResult>
+                        SubRunners = new []
                         {
-                            new RunnerResult
-                            {
-                                TestRunner = typeof(AggregatingTestRunner),
-                                SubRunners = new List<RunnerResult>
-                                {
-                                    RunnerResult.LocalTestRunner,
-                                    RunnerResult.LocalTestRunner
-                                }
-                            },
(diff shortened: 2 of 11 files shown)


src/NUnitEngine/nunit.engine.core/Runners/AggregatingTestRunner.cs
```diff
@@ -55,7 +55,8 @@ public IList<ITestEngineRunner> Runners
                 if (_runners == null)
                 {
                     _runners = new List<ITestEngineRunner>();
-                    foreach (var subPackage in TestPackage.SubPackages)
+                    foreach (var subPackage in TestPackage.Select(p => !p.HasSubPackages()))
+                    //foreach (var subPackage in TestPackage.SubPackages)
```
jnm2, line 59: Was the comment meant to remain?

src/NUnitEngine/nunit.engine.core/Runners/AggregatingTestRunner.cs
```diff
@@ -55,7 +55,8 @@ public IList<ITestEngineRunner> Runners
                 if (_runners == null)
                 {
                     _runners = new List<ITestEngineRunner>();
-                    foreach (var subPackage in TestPackage.SubPackages)
+                    foreach (var subPackage in TestPackage.Select(p => !p.HasSubPackages()))
```
jnm2, line 58: I'm confused by this expression. It looks like we're selecting a `bool` value, meaning that `subPackage` is a boolean. I'm guessing this is not the System.Linq `Select` but rather something more like System.Linq's `Where` with ` `.SubPackages.Where`?

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/TestCases/NetStandardTestCases.cs
```diff
@@ -79,17 +79,10 @@ public static IEnumerable<TestCaseData> TestCases
                     new RunnerResult
                     {
                         TestRunner = typeof(AggregatingTestRunner),
-                        SubRunners = new List<RunnerResult>
+                        SubRunners = new []
```
jnm2, line 82: `new []` is an inconsistent style compared to existing and added lines. (Several occurrences)

jnm2 (approved):
Nothing blocking

src/NUnitEngine/nunit.engine.tests/Services/TestRunnerFactoryTests/TestCases/NetStandardTestCases.cs
CharliePoole: Yeah, it's used more consistently in the cases and results on the .NET Framework side. I like [] better but I probably shouldn't have changed it arbitrarily. That said, I strongly suspect we will be reworking all these things - and they may have already been changed in the dev-4.0 branch because I made lots of changes there.

src/NUnitEngine/nunit.engine.core/Runners/AggregatingTestRunner.cs
CharliePoole: It's an extension I wrote to TestPackage a while back, which selects packages recursively out of the tree. IIRC I used Select to avoid confusion (with Linq Where) but I guess I didn't succeed. :-) It's used in a number of places.

src/NUnitEngine/nunit.engine.core/Runners/AggregatingTestRunner.cs
CharliePoole: Nope. I tested by alternating between the two expressions in several places. Forgot this one. I'll delete it.

PR #1038, opened 2021-11-17
Eliminate end of life checks on runtimes

### LINKED ISSUE #1037: Eliminate End of Life Check in builds
While we will sooner or later discontinue building for runtimes designated as out of support by Microsoft, we continue to do so for now.

We're not likely to forget which runtimes are out of support, so it would be nice to not get all the warning messages.

CharliePoole:
Fixes #1037

### DIFF
--- a/src/NUnitEngine/mock-assembly-x86/mock-assembly-x86.csproj
+++ b/src/NUnitEngine/mock-assembly-x86/mock-assembly-x86.csproj
@@ -5,6 +5,7 @@
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\..\nunit.snk</AssemblyOriginatorKeyFile>
     <PlatformTarget>x86</PlatformTarget>
+    <CheckEolTargetFramework>false</CheckEolTargetFramework>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="NUnit" Version="3.13.2" />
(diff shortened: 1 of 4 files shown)


src/NUnitEngine/mock-assembly-x86/mock-assembly-x86.csproj
```diff
@@ -5,6 +5,7 @@
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\..\nunit.snk</AssemblyOriginatorKeyFile>
     <PlatformTarget>x86</PlatformTarget>
+    <CheckEolTargetFramework>false</CheckEolTargetFramework>
```
jnm2, line 8: If desired, this could be added in https://github.com/nunit/nunit-console/blob/master/src/Directory.Build.props rather than in each csproj.

jnm2 (approved):

src/NUnitEngine/mock-assembly-x86/mock-assembly-x86.csproj
CharliePoole: I felt like having it here gave better visibility. In fact, I'm not in love with Directory.Build.props but I haven't wanted to change it because some people like that better for things that are always the same - or which we hope will remain so. :-)

PR #1041, opened 2021-11-21
Consider later NUnit versions as a valid framework.

manfred-brands:
This goes towards fixing https://github.com/nunit/nunit/issues/3867

### DIFF
--- a/src/NUnitEngine/nunit.engine.core/Drivers/NUnit3DriverFactory.cs
+++ b/src/NUnitEngine/nunit.engine.core/Drivers/NUnit3DriverFactory.cs
@@ -9,7 +9,7 @@ namespace NUnit.Engine.Drivers
 {
     public class NUnit3DriverFactory : IDriverFactory
     {
-        private const string NUNIT_FRAMEWORK = "nunit.framework";
+        public const string NUNIT_FRAMEWORK = "nunit.framework";
 
         /// <summary>
         /// Gets a flag indicating whether a given assembly name and version
--- a/src/NUnitEngine/nunit.engine.core/Drivers/NUnit3DriverFactory.cs
+++ b/src/NUnitEngine/nunit.engine.core/Drivers/NUnit3DriverFactory.cs
@@ -18,7 +18,7 @@ namespace NUnit.Engine.Drivers
         /// <param name="reference">An AssemblyName referring to the possible test framework.</param>
         public bool IsSupportedTestFramework(AssemblyName reference)
         {
-            return NUNIT_FRAMEWORK.Equals(reference.Name, StringComparison.OrdinalIgnoreCase) && reference.Version.Major == 3;
+            return NUNIT_FRAMEWORK.Equals(reference.Name, StringComparison.OrdinalIgnoreCase) && reference.Version.Major >= 3;
         }
 
 #if NETFRAMEWORK
--- a/src/NUnitEngine/nunit.engine.tests/Drivers/NUnit3DriverFactoryTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Drivers/NUnit3DriverFactoryTests.cs
@@ -1,0 +1,42 @@
+﻿using NUnit.Engine.Drivers;
+using NUnit.Framework;
+using System;
+using System.Reflection;
+
+namespace NUnit.Engine.Tests.Drivers
+{
+    public class NUnit3DriverFactoryTests
+    {
+        private NUnit3DriverFactory _factory;
+
+        [SetUp]
+        public void SetUp()
+        {
+            _factory = new NUnit3DriverFactory();
+        }
+
+        [TestCase(2, ExpectedResult = false)]
+        [TestCase(3, ExpectedResult = true)]
+        [TestCase(4, ExpectedResult = true)]
+        public bool SupportsNetFramework(int majorVersion)
+        {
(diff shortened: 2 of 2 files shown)


CharliePoole (changes requested):
See my general comment. This single token change basically tries to re-purpose the NUnit 3 driver factory to handle NUnit 4, which we may not want to do. For now, we'll keep it as a reminder of a possible simple fix but we'd want tests before merging it.

CharliePoole:
All the tests pass because this change is untested. That's because the latest release of `nunit.framework` is 3.13.2. Version 4 is under development and will eventually be supported by the engine, although possibly in a future  __NUnit4DriverFactory__ rather than __NUnit3DriverFactory__. That depends on whether the final NUnit 4 API is similar to the NUnit 3 API or different.

If you accompanied this with a test, I'd be willing to merge it into the dev-4.0 branch rather than master. I'd also be willing to accept a PR against master, for inclusion in the 3.13 release, giving a cleaner error message - if you're interested.

For now, I'm marking this PR as Blocked until we know what will be happening with NUnit 4 and when.

manfred-brands:
@CharliePoole I agree with your comments, but shouldn't this have been considered when nunit 4 developement was started?
If nunit4 is not ready for main, why is it not in an 4.0.0-dev branch with main still pointing to 3.13.2?
Now we cannot run tests from within VS on the main branch of nunit, resulting in various issues raised against nunit and slowing down developing PRs there. 

I'm happy for moving this into a 4.0-dev branch which is referenced in nunit3testadapter (5.0.0-dev?branch) to be used in nunit4.

Regarding a PR in master for a better error message, I'm not sure if we can. Yes we can improve the message in the `SkippedAssemblyFrameworkDriver` but that doesn't seem to bubble up to the user.

CharliePoole:
@manfred-brands A few years back, we decided that the NUnit Project was too big and split it into individual key projects, like NUnit framework and NUnit Console. (The NUnit test adapters had already been separate for a long time.)

This allows us to move ahead in each project without necessarily worrying about the others - at least most of the time. :-)

In this case, both the NUnit Framework and the Engine/Console Runner are at the point of building toward a major release at the same time. That's kind of awkward, obviously, and each project made the best decision it could based on the info available to it. The original issue, which triggered the PR, seems to be more about the adapter than anything else. You should be aware that any change to the engine will have no impact on the adapter until a new adapter version actually uses that new engine version.

I no longer work on the framework, so I'm not aware of how the framework team actually runs the framework tests. My guess is that they are not using the engine, because somebody would have complained by now if they were. :-)

If you can think of something I can do in version 3.13 to make things easier for you, let me know. We are not currently releasing dev builds of the console runner or engine from the dev-4.0 branch, but we could conceivably do that if it would help.

Not sure what you mean by a 5.0.0 branch. That's farther ahead than I can even think right now. :-)

manfred-brands:
> The original issue, which triggered the PR, seems to be more about the adapter than anything else.

I know, that is where I started searching, but then it called into nunit.engine to find tests and hence I got here.
Locally indeed, I had to make changes to 3 repositories to get the tests showing up again.

> how the framework team actually runs the framework tests
They use nunitlite console runner.

> Not sure what you mean by a 5.0.0 branch. That's farther ahead than I can even think right now. :-)

That was referring to NUnit3TestAdapter which is currently at version 4.1.0. If this has to going into a new main branch it would have to be 5.0.0

If NUnit 4 is compatible with the NUnit3 engine at least for now, then all we need is my change (plus a new unit test), followed by updated references in the NUnit3testadapter and nunit repositories. They could be called a fix iso a feature.
I would need to get input from @rprouse to see what the way forward is. Is this an accidental oversight of increasing the version to 4.0.0 or deliberate due to possible incompatibilities.

As an alternative is to tell everyone to make nunit PRs against 3.13.2 and future port it to nunit4 iso backport.

There are some issues in nunit that also affect my developments and if I can I wouldn't mind assisting in fixing some if I can.
But being able to step through some unfamiliar code in tests is handier than adding `Debugger.Launch()` at places.

CharliePoole:
AFAIK the change to 4.0 in master branch is deliberate. That's a decision the team and team lead need to make, of course. It basically has to do with what you want your focus to be. Whichever is not the main branch tends to get less attention.

WRT the adapter, adding a new feature (running 4.0) isn't a breaking change, so I assume Terje would just go to 4.2 or 4.3 when upgrading the adapter, unless the 4.0 engine were actually missing some features he uses.

CharliePoole:
I thought I would be able to edit the PR to change the merge target, but apparently not. See if GitHub will allow you to do it as the author. The branch we want is dev-4.0. Otherwise, if you delete this and start over, we lose all the discussion.

manfred-brands:
Re-directed the branch to dev-4.0 and added unit test for the `NUnit3DriverFactory.IsSupportedTestFramework` method.

CharliePoole (approved):
This looks good for the dev-4.0 branch. We may, run into further glitches down the road, but support for NUnit 4 seems like a must-have feature, so we'll fix them as they happen. Thanks for your work.

CharliePoole:
@manfred-brands Because the dev-4.0 branch has drifted quite a bit from main, this can't just be merged into main. So I'll be replicating your changes separately. See #1102.

manfred-brands:
Thanks @CharliePoole As I don't know the state of nunit 4, I do my fixes on nunit 3.13 where I can run the tests properly, then cherry pick to 4.0

PR #1051, opened 2021-12-16
Remove Rebracer.xml and NUnitConsole.sln.DotSettings

### LINKED ISSUE #361: Remove Rebracer file in favour of .editorconfig
@ChrisMaddock commented on [Thu Jan 25 2018](https://github.com/nunit/nunit/issues/2685)

We currently have three different files providing formatting guidelines - which I don't believe are all maintained.

The most recent work has been done in .editorconfig, I believe. Do we have any need to maintain all three files, or could we remove the Rebracer and ReSharper options?

I'd like to apply the same change to both the nunit and nunit-console repo's, if we can make a decision.

---

@rprouse commented on [Thu Jan 25 2018](https://github.com/nunit/nunit/issues/2685#issuecomment-360473003)

Agreed. Rebracer isn't needed in VS 2017 which is now required to build so can be dropped. We also haven't updated Resharper in years other than your automatic changes.

---

@jnm2 commented on [Thu Jan 25 2018](https://github.com/nunit/nunit/issues/2685#issuecomment-360480875)

I was just looking at that. I think we should remove them but be open to adding *minimal* ReSharper settings back in to quiet down ReSharper's default complaints in our codebase for things that .editorconfig doesn't control yet.

---

@ChrisMaddock commented on [Thu Jan 25 2018](https://github.com/nunit/nunit/issues/2685#issuecomment-360484066)

That sounds like a good idea to me. 😄 

---

@zastrowm commented on [Fri Jan 26 2018](https://github.com/nunit/nunit/issues/2685#issuecomment-360641703)

Note that ReSharper also has support for configuring it via .editorconfig: https://www.jetbrains.com/help/resharper/EditorConfig_Index.html

---

@ChrisMaddock commented on [Fri Jan 26 2018](https://github.com/nunit/nunit/issues/2685#issuecomment-360732649)

Thanks @zastrowm!

Looks like we've made a decision. Should anyone new to contributing fancy picking this up, to clarify, the task is to remove the Rebracer.xml and NUnit.sln.DotSettings files from the repository. Pull Requests Welcome!

CharliePoole:
Fixes #361


CharliePoole:
I'm removing both files as that seems to be the consensus of the discussion on issue #361. I'm not a ReSharper user, so I'll have to rely on others for their judgement.

mikkelbu (approved):
LGTM

PR #1052, opened 2021-12-18
Remove ConsoleRunner.NetCore and .NET Core 3.1 engine build

### LINKED ISSUE #1045: Should we continue to build NUnit.ConsoleRunner.NetCore?
Now that NUnit.ConsoleRunner is able to run .NET Core tests, it may be useful to focus our efforts there and stop producing a console runner with reduced capabilities, which runs under .NET Core 3.1.

That's not to say we shouldn't look toward a future in which the console runs under a more modern runtime, like .NET 6.0 for example. But keeping the 3.1 build working and having to deal with issues about it's limitations seems to only slow down our progress.

Let's discuss this and make a decision and plan for .NET 4.0.

CharliePoole:
Fixes #1045


CharliePoole:
Removed NUnit.ConsoleRunner.NetCore build as well as the .NET Core 3.1 build of the engine. The .NET Standard engine build remains, at least for now.

mikkelbu (approved):
LGTM

CharliePoole:
@nunit/engine-team Let's discuss this further on #1045, in particular considering @k15tfu 's comment.

PR #1053, opened 2021-12-20
Update to Cake 1.3.0

### LINKED ISSUE #1047: Upgrade Cake to version 1.3
We are using Cake version 0.37. Cake 1.0, 1.1, 1.2, 1.3 and 2.0 have now come out. We're missing a lot of features and even more bug fixes.

Since 2.0 has breaking changes, the most significant being that the .NET Framework runner is no longer supported, so we would have to use the dotnet tool runner. There are others we will need to review as well.

To avoid the breaking changes, I'd like to first upgrade us to 1.3. Note that this requires more than just updating the version in packages.config. New bootstrap scripts need to be downloaded as well and any changes that have been made re-applied. GitHub should be able to tell us if any updates were made to the scripts since updating to 0.37.

CharliePoole:
Fixes #1047

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -37,7 +37,7 @@ https://cakebuild.net
 
 [CmdletBinding()]
 Param(
-    [string]$Script = "build.cake",
+    [string]$Script,
     [string]$Target,
     [string]$Configuration,
     [ValidateSet("Quiet", "Minimal", "Normal", "Verbose", "Diagnostic")]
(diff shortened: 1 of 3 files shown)


mikkelbu (approved):
LGTM. I was going to say that we probably needed to update `appveyor.yml` as we did in the NUnit project when we changed to the dotnet tool version of Cake - see below - but it seems like this is still working on AppVeyor, so we don't have to do this change.

https://github.com/nunit/nunit/commit/f2c174c176205f2e4ee428a620c84150b2b03bc0#diff-92ab9a36df5d8e9f7076f2fdec59492d1ac2d9cf27ea046767a7fc4d542ef3dc

mikkelbu:
Ps. I'll take a look at the other PR within the next hour.

CharliePoole:
@mikkelbu That's because it's still the .NET Framework runner. I wanted to change the version first and switch to the dotnet tool once we know it's all working.

PR #1054, opened 2021-12-27
Refactor cake scripts

CharliePoole:
One of my bigger priorities is to extend the existing scripts so that the entire release process can be done automatically.

That's not what this is. :-) This is pure refactoring. The scripts continue to produce the same outputs and all the unit tests and package tests continue to pass.

There is a mix of large and small refactoring here. The most significant are probably the separation of package tests from the package tester and the introduction of a class to represent a package. Since we produce eight primary packages as well as five source packages, I think that such a representation will be important in the future.

Next stage in improving our build process, will be to outline what we would like the build to do. I'll create a team discussion for that purpose.


rprouse (approved):
Pretty massive PR to hold in your head for a code review but looks good to me.

CharliePoole:
Sorry about the size. I got carried away. :-)

PR #1056, opened 2021-12-27
Add VERSIONING document

### LINKED ISSUE #1055: Create a statement of how we version each of our packages
Users need to know what it means when we make a major, minor or patch release.

So do we. :-)

I'll create a VERSIONING.md in the project root document for review.

CharliePoole:
Fixes #1055

@nunit/engine-team 

I think this is super-important so I have asked each of you to do a review. It can be quick, but please just make sure that the versioning policy makes sense to you. 

Additionally, after the notes, I've asked a question for which I don't have the answer. :-)

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -1,0 +1,71 @@
+# NUnit Console Runner and Engine Versioning
+
+Both **Console** and **Engine** packages use a version number in the form MAJOR.MINOR.PATCH. We change the
+
+- MAJOR version when we have made incompatible API changes
+- MINOR version when adding functionality in a backwards-compatible manner
+- PATCH version when you making backwards-compatible bug fixes
+
+This versioning approach is, of course, what is known as "Semantic Versioning." Taken by itself, however, the above description leaves out some rather important stuff...
+
+* What exactly are the APIs we are protecting from incompatible changes?
+* What degree of change is needed in each of those APIs for it to be considered a breaking change?
+
+The remainder of this document attempts to answer those questions.
+
+## Console Runner API
+
+First, let's note that the **Console Runner** is not designed to be used as a library, so there is no ABI or library-like API to worry about. However, the runner does present a command-line interface, which is used both interactively and through user-created scripts. The command-line API is defined by our documentation, including the help message the runner displays and the [published documentation](https://docs.nunit.org/articles/nunit/running-tests/Console-Command-Line.html) on our website.
+
+### Examples of MAJOR Command-Line Changes
+
+- Complete removal of an option.
+- Changing the behavior of an option so that it no longer does what it was defined to do.
+- Removing the ability to run under a particular platform.
+
+### Examples of MINOR Command-Line Changes
+
+- Addition of a new option
+- Modifying the behavior of an option so it does the same thing better or more effectively or has added sub-options.
+- Adding support for running under an new platform
+
+### Examples of PATCH Command-Line changes
+
+- Fixing a bug in the behavior of an option.
+- Restoring an option to its original defined behavior.
+- Correcting spelling errors.
+
+## Engine APIs
+
+The **NUnit Engine** is intended to be used by third-party runners as well as our own **Console Runner**. Usage is through the defined API, specified by interfaces included in the `nunit.engine.api` assembly. It's important to note that public types and methods included in other assemblies are not considered a part of the published API. Direct creation and use of `nunit.engine`, `nunit.engine.core` or any other assemblies included in our packages is discouraged for that reason. See below for more about this.
+
+### Examples of MAJOR Api Changes
+
+- Removing a published interface.
+- Modifying an interface so that it breaks existing usage.
+
+### Examples of MINOR Api Changes
+
+- Adding a new published interface and implementation.
+- Modifying an interface in a way which does __not__ break existing usage.
+- Modifying the implementation of an interface so it does the same thing better or more effectively.
+- Restoring an API or it's implementation to the original behavior following an erroneous MINOR release.
+
+### Examples of PATCH Api Changes
+
+- Fixing bugs in the implementation of an API.
+- Restoring an API or it's implementation to the original behavior following an erroneous PATCH release.
+
+## Notes
+
+1. Access to Types and Methods through reflection, while possible, is not part of our supported API.
+
+2. As explained above, the existing `nunit.engine.api` assembly specifies the _entire_ published API for the **NUnit Engine**. Other guaranteed APIs may be added in the future, using the same `.api` naming convention.
+
+3. Historically, NUnit has used public visibility for many Types and Methods, which were not actually part of the API. We are in the process of changing this approach to a more modern focus on limiting visibility. That said, even where a Type or Method is publicly accessible, it is __not__ automatically a part of our API and may change from release to release.
+
+4. Even in the case of non-breaking changes, as defined above, we will make a reasonable  effort to avoid negative impact on users.
+
+## Unresolved (Attention Reviewers!)
+
+* How do we consider a change to the Engine API, which requires recompilation of a runner using it. IOW, do we want source or binary compatibility for the Engine API?
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -42,7 +42,7 @@ The command-line is a _published interface_ as defined above. It is specified in
 
 - Adding a new option
 - Modifying the behavior of an option so it does the same thing better or more effectively or has added sub-options.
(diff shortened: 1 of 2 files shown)


mikkelbu:
I looks reasonable - I only have some comments/questions (some of my original comments/questions were due to misreading on my part in the initial reading of the documents).
* Is it possible to "Modifying an interface in a way which does __not__ break existing usage." ? Most/all changes to an interface I can think of can break the users. Perhaps adding a default is a non-breaking change?
* For the "Unresolved" I lean towards binary compatibility 
* I think we should have some tooling to help us with this - as it is quite easy to break people (@jnm2 added https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md for the NUnit project in https://github.com/nunit/nunit/pull/3516, but I'm not sure if this is still the best practice in .NET). Or perhaps we can use Microsoft.DotNet.PackageValidation - https://devblogs.microsoft.com/dotnet/package-validation/

mikkelbu (approved):

CharliePoole:
@jnm2 Any comments?

CharliePoole:
@mikkelbu We will have less need of tooling if we limit the definition of breaking to changes to the API assembly. Essentially, anything that changes it has the potential of being breaking and nothing that changes other assemblies can be breaking... that's providing we can all agree to that principle.

VERSIONING.md
```diff
@@ -0,0 +1,71 @@
+# NUnit Console Runner and Engine Versioning
+
+Both **Console** and **Engine** packages use a version number in the form MAJOR.MINOR.PATCH. We change the
+
+- MAJOR version when we have made incompatible API changes
+- MINOR version when adding functionality in a backwards-compatible manner
+- PATCH version when you making backwards-compatible bug fixes
```
jnm2, line 7:
```suggestion
- PATCH version when making backwards-compatible bug fixes
```

VERSIONING.md
```diff
@@ -0,0 +1,71 @@
...
+The remainder of this document attempts to answer those questions.
+
+## Console Runner API
+
+First, let's note that the **Console Runner** is not designed to be used as a library, so there is no ABI or library-like API to worry about. However, the runner does present a command-line interface, which is used both interactively and through user-created scripts. The command-line API is defined by our documentation, including the help message the runner displays and the [published documentation](https://docs.nunit.org/articles/nunit/running-tests/Console-Command-Line.html) on our website.
+
+### Examples of MAJOR Command-Line Changes
+
+- Complete removal of an option.
+- Changing the behavior of an option so that it no longer does what it was defined to do.
```
jnm2, line 23:
```suggestion
- Changing the behavior of an option so that it no longer does what it was previously defined to do.
```

VERSIONING.md
```diff
@@ -0,0 +1,71 @@
...
+- Removing the ability to run under a particular platform.
+
+### Examples of MINOR Command-Line Changes
+
+- Addition of a new option
+- Modifying the behavior of an option so it does the same thing better or more effectively or has added sub-options.
+- Adding support for running under an new platform
+
+### Examples of PATCH Command-Line changes
```
jnm2, line 32:
```suggestion
### Examples of PATCH Command-Line Changes
```

VERSIONING.md
```diff
@@ -0,0 +1,71 @@
...
+
+- Fixing a bug in the behavior of an option.
+- Restoring an option to its original defined behavior.
```
jnm2, line 35: How does this differ from 'fixing a bug'?

VERSIONING.md
```diff
@@ -0,0 +1,71 @@
...
+
+- Removing a published interface.
+- Modifying an interface so that it breaks existing usage.
+
+### Examples of MINOR Api Changes
+
+- Adding a new published interface and implementation.
+- Modifying an interface in a way which does __not__ break existing usage.
+- Modifying the implementation of an interface so it does the same thing better or more effectively.
+- Restoring an API or it's implementation to the original behavior following an erroneous MINOR release.
```
jnm2, line 52:
```suggestion
- Restoring an API or its implementation to the original behavior following an erroneous MINOR release.
```

VERSIONING.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Examples of PATCH Api Changes
+
+- Fixing bugs in the implementation of an API.
+- Restoring an API or it's implementation to the original behavior following an erroneous PATCH release.
```
jnm2, line 57:
```suggestion
- Restoring an API or its implementation to the original behavior following an erroneous PATCH release.
```

VERSIONING.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Notes
+
+1. Access to Types and Methods through reflection, while possible, is not part of our supported API.
+
+2. As explained above, the existing `nunit.engine.api` assembly specifies the _entire_ published API for the **NUnit Engine**. Other guaranteed APIs may be added in the future, using the same `.api` naming convention.
+
+3. Historically, NUnit has used public visibility for many Types and Methods, which were not actually part of the API. We are in the process of changing this approach to a more modern focus on limiting visibility. That said, even where a Type or Method is publicly accessible, it is __not__ automatically a part of our API and may change from release to release.
```
jnm2, line 65: Even in the `.api` assembly? Since that's the only assembly where the visibility of types could matter?

VERSIONING.md
```diff
@@ -0,0 +1,71 @@
...
+- Correcting spelling errors.
+
+## Engine APIs
+
+The **NUnit Engine** is intended to be used by third-party runners as well as our own **Console Runner**. Usage is through the defined API, specified by interfaces included in the `nunit.engine.api` assembly. It's important to note that public types and methods included in other assemblies are not considered a part of the published API. Direct creation and use of `nunit.engine`, `nunit.engine.core` or any other assemblies included in our packages is discouraged for that reason. See below for more about this.
+
+### Examples of MAJOR Api Changes
+
+- Removing a published interface.
+- Modifying an interface so that it breaks existing usage.
```
jnm2, line 45: It may be worth linking https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/breaking-change-rules.md for our own ease of reference to define what we mean by "breaks existing usage."

jnm2:
> Is it possible to "Modifying an interface in a way which does not break existing usage." ? Most/all changes to an interface I can think of can break the users. Perhaps adding a default is a non-breaking change?

@mikkelbu The only changes to an interface which do not break it that I can think of:
- adding new static members
- adding new nested types
- adding new instance members which have default implementations (requiring the runtime to be .NET Core 3.0+)

On the unresolved question, why not consider all three major types of break? Behavioral, binary, and source? I don't see why making a decision about one kind of break would restrict the other kinds of breaks we might want to prevent.

CharliePoole:
Lot's to think about here... I'll do another draft. I think I need to be a bit more explicit about how I'm using "interface" to mean C# interface some of the time and just as a dictionary word other times.

VERSIONING.md
CharliePoole: I combined the two. I wanted to be clear that the __defined__ behavior is what counts, not the actual (broken) behavior. In practice, of course, we would probably be less rigid about this if the bad behavior had gone unnoticed over multiple releases.

VERSIONING.md
CharliePoole: I'll look at the wording.

In the case of the API assembly, I think we would be hard pressed to argue that making something public in an assembly called "API" isn't "publishing" it. So we need to eliminate any public stuff we don't want to support. A great example of this is the public constructor of TestPackage that takes a string. It was never intended to be used except in a specific situation internally and causes a lot of work for us to maintain tests using both forms of constructor.

However, I did want to restate this for any users who use public methods out of other assemblies. Belt and suspenders, I guess. :-)

CharliePoole:
@nunit/engine-team Second draft out for review

mikkelbu (approved):

VERSIONING.md
```diff
@@ -0,0 +1,81 @@
...
+
+- Complete removal of an option.
+- Changing the behavior of an option so that it no longer does what it was previously defined to do.
+- Removing the ability to run under a particular platform.
+
+### Examples of MINOR Command-Line Changes
+
+- Adding a new option
+- Modifying the behavior of an option so it does the same thing better or more effectively or has added sub-options.
+- Adding support for running under an new platform
```
jnm2, line 45:
```suggestion
- Adding support for running under a new platform
```

jnm2 (approved):

PR #1060, opened 2021-12-29
Upgrade cake to version 2.0.0

### LINKED ISSUE #1059: Upgrade Cake to 2.0.0 and use Cake.Tool so we can build on all platforms
With 2.0.0, Cake.Tool is the only available runner so both changes will be made simultaneously.

CharliePoole:
Fixes #1059


jnm2 (approved):

mikkelbu (commented):
We should also update azure-pipelines.yml to use the new format for the pipeline to pass. 

BUILDING.md also need updating (and perhaps more places, but I'm on my phone, so searching is not optimal. I can make a better search tomorrow)

CharliePoole:
Looks like I may have broken something by changing Package so it depends on Build. The Windows job on Azure first builds and tests, then publishes test results and finally re-runs the script to do packaging. I may need to add a "PackageOnly" target.

CharliePoole:
The Package target continues to depend on Build, to prevent accidentally building the wrong stuff, particularly when used locally. To avoid a build, you have to use the PackageExistingBuild target.

Issue #1057, opened 2021-12-28
Automate all package publication and release


CharliePoole:
This issue is an Epic tying together improvements to the build process, which I want to make. As such, it's a good place for discussing our overall CI and build automation strategy. The individual changes to the build will be made using separate issues and/or PRs.

### Overall Goal

To be able to take the code at a given state and create a new release within an hour or so, including running automated package tests, publishing to NuGet and Chocolatey, creating the release notes, releasing on github and sending out notifications. This is admittedly ambitious but I'm pretty close to that target in some other projects.

### Individual Issues or PRs

- [x] #1059
- [x] #1061
- [x] #1064
- [x] #1065
      - Should allow all publishing steps to continue even if one fails due to communication or password issues
      - We should be able to use a local build target for retrying the failed steps.
- [x] #1070
- [x] #1072

### Notes

1. I have existing code for all of the above items except for the last one. I'm using it in five or six projects. It will need to be tailored to our own project, of course.
2. At some point in the sequence of additions, the build script will most likely be refactored to make use of shared state rather than global constants for most inter-task communication.
3. When we start using GtiReleaseManager, we will want to change at least some of our GitHub labels, since they are used as headings in the release notes, e.g.: is:feature => Feature.

@nunit/engine-team Anything else we should throw in here?

CharliePoole:
@nunit/engine-team I created a draft document describing the process as it works so far... https://github.com/nunit/nunit-console/wiki/Creating-a-New-Release

CharliePoole:
I removed #1073 from this Epic. It will be pursued separately once there's some feedback on it.

CharliePoole:
Updated the wiki page to reflect running CreateDraftRelease locally.

PR #1062, opened 2021-12-30
Let PackageDefinition build the package itself

CharliePoole:
This is a refactoring change in support of #1057.

My last bit of script refactoring was much too big, so I'm just doing little pieces now. The advantage is they are easier to review. The disadvantage is it may not be obvious why I'm making a particular change.

In this case, I'm eliminating a bunch of procedural code and allowing the package definition class to do the building through an abstract method, which is overridden for each of the four package types: NuGet, Chocolatey, Msi and Zip. This works together with the representation of all our packages in package-definition.cake and will eventually lead to the packages publishing themselves as well.


mikkelbu (approved):
LGTM 👍

PR #1063, opened 2021-12-31
Use GitVersion for versioning

### LINKED ISSUE #1061: Use GitVersion to determine package version to be produced
Currently, we always build the version coded in the build.cake file, like 3.13.0, except on AppVeyor where we generate a unique pre-release suffix. That means the versioning of a given build is different on the local machine and any other CI servers from what it is in appveyor. By using GitVersion to determine the next version to build, we will be building the same version for a given commit on all platforms.

CharliePoole:
Fixes #1061

Once this merges, I'll need to delete any existing 3.14.0-devxxxxx entries with a larger suffix from MyGet.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -1,3 +1,7 @@
+static string Target; Target = Argument("target", "Default");
+static string Configuration; Configuration = Argument("configuration", "Release");
+
+#load cake/constants.cake
 #load cake/ci.cake
 #load cake/header-check.cake
 #load cake/package-checks.cake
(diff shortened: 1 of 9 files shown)


CharliePoole:
Closed this PR and will create a new one. The issue-1061 branch already has the same commits as master plus 2 more commits. The branch build is passing on AppVeyor but the PR fails. Let's see if recreating it helps!

build.cake
```diff
@@ -1,135 +1,45 @@
+static string Target; Target = Argument("target", "Default");
+static string Configuration; Configuration = Argument("configuration", "Release");
```
jnm2, line 2: What happens if you try `static readonly string Configuration = Argument("configuration", "Release");`?

cake/versioning.cake
```diff
@@ -0,0 +1,141 @@
...
+
+public class BuildVersion
+{
+    private ISetupContext _context;
+    private GitVersion _gitVersion;
+
+    // NOTE: This is complicated because (1) the user may have specified 
+    // the package version on the command-line and (2) GitVersion may
+    // or may not be available. We'll work on solving (2) by getting
+    // GitVersion to run for us on Linux, but (1) will alwas remain.
```
jnm2, line 11:
```suggestion
    // GitVersion to run for us on Linux, but (1) will always remain.
```

cake/versioning.cake
```diff
@@ -0,0 +1,141 @@
...
+    //
+    // We simplify things a by figuring out the full package version and
```
jnm2, line 13:
```suggestion
    // We simplify things by figuring out the full package version and
```

jnm2 (approved):

CharliePoole:
No help there! Suspecting some code, which only executes for a PR.

cake/versioning.cake
CharliePoole: Actually, I should take most of this out. I got it running on Linux!

build.cake
CharliePoole: That's fair. Some things can't be readonly, of course, but I got carried away.

CharliePoole:
Thanks @jnm2. Now if I can just get the PR to build!

build.cake
jnm2: For those things, just a shortening of the syntax then, like:

```suggestion
static string Configuration = Argument("configuration", "Release");
```

CharliePoole:
Diagnostic gives me...

> GitVersion.BugException: GitVersion has a bug, Your HEAD has moved after repo normalisation.
> 
> To disable this error set an environment variable called IGNORE_NORMALISATION_GIT_HEAD_MOVE to 1.

Trying that next.

jnm2:
I wonder if it has anything to do with the fact that the PR build (not the branch build) is building a dynamically created merge commit back to the main branch.

CharliePoole:
Yes, I took a quick look at the code and that seems to be what it is. When I get a moment (I'm taking down our tree now) I'll try the suggested workaround in AppVeyor.yml. If that fails, I'm pretty sure I can merge locally and just push to master, but I'd prefer something that will keep working for future PRs.

CharliePoole:
Well, that makes it build by suppressing the error but it results in an improper PackageVersion for the PR. Not that important, since we don't do anything with artifacts from a PR build anyway. I'll do one more iteration cleaning out all my debug stuff.

BTW, this is issue GitTools/GitVersion#2554

CharliePoole:
A comment on the issue indicates that GitVersion 5.6.3 doesn't have the problem, so I'm trying that.

CharliePoole:
This seems to be as good as it gets for the moment. With GitVersion 5.0.0 (the version I have used in my other projects) I get failures in Azure. With 5.8.1 (latest) the aforementioned bug shows up. Everything passes in 5.6.3 but we no longer get a special naming pattern with a "pr" label and the PR number. However, since we don't publish them it doesn't matter that much. When I get a chance I'll follow up on this with a little test project.

PR #1066, opened 2022-01-01
Use Cake script to upload to MyGet

### LINKED ISSUE #1064: Publish `dev` builds to MyGet using the Cake script
By doing this we can choose where to do the publishing rather than having to do it only on AppVeyor. In particular, it  enables doing it on the local dev machine when necessary and with the right permissions.

CharliePoole:
Fixes #1064

Once PR this passes, I'll go ahead and merge. That's the only way to actually see it work, i.e. by pushing to myget. Fortunately, it's relatively easy to fix myget issues. The harder part is when we do nuget and chocolatey.


CharliePoole:
Well it merged but the publishing step was skipped. Working on it...

PR #1074, opened 2022-01-02
Code to perform a production release on GitHub

### LINKED ISSUE #1072: Automatically upload assets and perform the release on GitHub using GitReleaseManager.
This will be triggered whenever a production (non-prerelease, alpha, beta or rc) release is being done based on the tag.

CharliePoole:
Fixes #1072

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -484,7 +484,7 @@ Task("PublishToMyGet")
     .Description("Publish packages to MyGet")
     .Does(() =>
     {
-        if (!ShouldPublishToMyGet())
+        if (!ShouldPublishToMyGet)
             Information("Nothing to publish to MyGet from this run.");
         else
         {
@@ -516,7 +516,7 @@ Task("PublishToNuGet")
     .Description("Publish packages to NuGet")
     .Does(() =>
     {
-        if (!ShouldPublishToNuGet())
+        if (!ShouldPublishToNuGet)
             Information("Nothing to publish to NuGet from this run.");
         else
         {
@@ -541,7 +541,7 @@ Task("PublishToChocolatey")
     .Description("Publish packages to Chocolatey")
     .Does(() =>
     {
-        if (!ShouldPublishToChocolatey())
+        if (!ShouldPublishToChocolatey)
             Information("Nothing to publish to Chocolatey from this run.");
         else
         {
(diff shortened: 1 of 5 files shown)


CharliePoole:
I've been moving pretty fast on the build changes, but this one is the point where it really gets to impact users, so some more eyes on it would be very good! This is the point where packages will be pushed and released... not immediately after merge, but when I do a release branch.

jnm2 (approved):
Looks good to me.

CharliePoole:
OK then... here goes! First the merge and then I'll start getting the 3.13.1 release ready to fire.

CharliePoole:
So the merge ran correctly, publishing a dev release of all 13 packages on MyGet. Next actions will be on the release branch.

CharliePoole:
AppVeyor failed running CreateDraftRelease. I just ran it locally and it succeeded. Possibly a security failure. Fortunately, CreateDraftRelease can be run repeatedly.

CharliePoole:
Back at it... after analyzing the results and re-running some builds, I decided that this is still part of the earlier problem with how newer releases of GRM handle PRs. We may need some code to detect PRs on appveyor and work around it or maybe there is a solution in the GRM project. I'll document this in a separate issue.

So... I used my admin privileges to force the merge, which is the next step of the release process. That succeeded.

I then went into the Draft Release online and published it.  That tags master and triggers the final build, which actually makes the production release. That failed due to an error in the script code. The nuget upload worked correctly. But the chocolatey upload failed. That means I'll have to fix it up by running targets directly on master on my local machine. That recovery is part of the process, intended for communications and security failures but also useful for newly found bugs. :-)

@nunit/engine-team I'm documenting all the steps I take in doing the release here. To understand what's done and what needs doing, read starting with the comment before this one. I'll take a break now before continuing.

CharliePoole:
On my local machine, I pulled the updated master and rebuilt and packaged release 3.13.1.

I then ran `build --target=CreateProductionRelease` which successfully created the production release. I'll send out notices manually in the morning.

You'll note that the result of this recovery is that the packages on NuGet and chocolatey were built on AppVeyor while those on GitHub were built locally. If we decided this were a problem we could actually run the entire process locally in the future. I'm not sure that's necessary, although it might have some advantages.

Now that 3.13.1 is released, I'll create an issue to fix the problems we saw in this release so that the next one is smoother.

PR #1086, opened 2022-01-08
Release notes: Update hash section formatting

mikkelbu:


CharliePoole:
Looks good... in my comment about this suggestion by @jnm2 (some other thread) I mistakenly thought this was built into GRM. Since it's in the yaml, I'm all for fixing it.

PR #1088, opened 2022-01-11
Add net5.0 agent

### LINKED ISSUE #1048: Agent for Net 5.0
This is a lack that has been referred to in many issues, most notably #858. I'm creating this as a separate issue in order to spell out clearly that we intend to have a .NET 5.0 agent.

I'm pretty sure that this will require some preliminary steps, which I'll add to this issue as they come up.

CharliePoole:
Fixes #1048


CharliePoole:
@nunit/engine-team In particular, please take a look at how I restructured the build, changing the logic under linux.

global.json
```diff
@@ -0,0 +1,5 @@
+{
+  "sdk": {
+    "version": "5.0.404"
+  }
```
rprouse, line 4: Add `"rollForward": "patch"` so this doesn't break for minor updates on build machines?

azure-pipelines.yml
```diff
@@ -6,6 +6,7 @@ trigger:
 jobs:
 
 - job: Windows
+  condition: not(startsWith(variables['Build.SourceBranchName'], 'azure-linux-'))
```
rprouse, line 9: What's up with this branch here and in MacOS?

src/NUnitEngine/nunit.engine.core/nunit.engine.core.csproj
```diff
@@ -1,7 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <RootNamespace>NUnit.Engine</RootNamespace>
-    <TargetFrameworks>net20;netstandard2.0;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>net20;netstandard2.0;netcoreapp3.1;net5.0</TargetFrameworks>
+    <NoWarn>$(NoWarn);SYSLIB0011;SYSLIB0012</NoWarn><!-- TODO: Get rid of obsolete stuff -->
```
rprouse, line 5: How much is obsolete? Should we use local pragmas to disable warnings so we can find them or are there too many?

rprouse (approved):
Minor comments but looks good. You mentioned doing a release, should we do an alpha or beta first?

azure-pipelines.yml
CharliePoole: It's a "feature" I added to make it easier to fix the linux build.
* Any branch starting "azure-" is not built on AppVeyor now.
* Any branch starting "azure-linux-" only runs the linux job on Azure.

This speeds things up quite a bit when working on the scripts themselves. I should probably finish it up by adding similar logic for "azure-windows-" and "azure-macOS-" but I only put in what I needed immediately.

BTW I tried to do it by setting variables at the start of the script, before the jobs, and then referring to them in each job. But the subtleties of how static vs dynamic variables are defined in the yml made it too much trouble, so I used copy and paste. If you know of an example somewhere, that would be helpful.

src/NUnitEngine/nunit.engine.core/nunit.engine.core.csproj
CharliePoole: I started to do pragmas but it kept popping up in new places. Once we get rid of the net20 build, I think we can get rid of this. That implies (1) making the console runner use some flavor of .NET 4 and (2) using a pluggable agent to run under .NET 2.0. I see both of those as still being a few months down the road though.

CharliePoole:
Regarding doing an alpha or beta... I'd lean toward just releasing in this case. It's an add-on feature and it seems to be fairly separate. That is, there's no use of .NET 5.0 anywhere except in the 5.0 agent and in the build script itself. When we get to big changes (i.e. 4.0) I think we want to have one or more betas (and maybe even alphas) but we're committed to being pretty conservative in the 3.x series.

@jnm2 @mikkel What do you think about this?

azure-pipelines.yml
CharliePoole: I set it up to use `azure-windows-` and  `azure-macOS-` prefixes the same as `azure-linux-`. Ran tests of all three to verify they work.

build.cake
```diff
@@ -104,39 +104,96 @@ Task("Build")
...
+
+    MSBuild(ENGINE_PROJECT, CreateMSBuildSettings("Publish")
+       .WithProperty("TargetFramework", "netstandard2.0")
+       .WithProperty("PublishDir", BIN_DIR + "netstandard2.0"));
+    CopyFileToDirectory(
+       BIN_DIR + "netstandard2.0/testcentric.engine.metadata.dll",
+       BIN_DIR + "netcoreapp2.1");
+    MSBuild(ENGINE_TESTS_PROJECT, CreateMSBuildSettings("Publish")
+       .WithProperty("TargetFramework", "netcoreapp2.1")
+       .WithProperty("PublishDir", BIN_DIR + "netcoreapp2.1"));
```
mikkelbu, line 172: How come these are only done for some of the targets?

E.g. in `BuildUsingMSBuildOnWindowsOrMacOS` we build both `"netstandard2.0"` and `"netcoreapp3.1"` for `ENGINE_PROJECT`, but in `BuildUsingMSBuildOnWindowsOrMacOS` we only build `"netstandard2.0"` ?

Also we don't build `ENGINE_API_PROJECT` on Linux ?

build.cake
```diff
@@ -104,39 +104,96 @@ Task("Build")
...
+    BuildProject(ENGINE_PROJECT, "net20", "netstandard2.0", "netcoreapp3.1");
+    BuildProject(CONSOLE_PROJECT, "net20", "netcoreapp3.1");
+    BuildProject(AGENT_PROJECT, "net20", "net40", "netcoreapp3.1", "net5.0");
+    BuildProject(AGENT_X86_PROJECT, "net20", "net40");
+
+    BuildProject(ENGINE_TESTS_PROJECT, "net35", "netcoreapp2.1", "netcoreapp3.1", "net5.0");
+    BuildProject(CONSOLE_TESTS_PROJECT, "net35", "netcoreapp3.1");
+
+    BuildProject(MOCK_ASSEMBLY_X86_PROJECT, "net35", "net40", "netcoreapp2.1", "netcoreapp3.1");
+    BuildProject(NOTEST_PROJECT, "net35", "netcoreapp2.1", "netcoreapp3.1");
```
mikkelbu, line 161: How come I cannot see this for windows/macOS ?

mikkelbu (commented):
I've just added some questions. Otherwise the changes look sound.

cake/package-definitions.cake
```diff
@@ -94,7 +97,9 @@ public void InitializePackageDefinitions(ICakeContext context)
             checks: new PackageCheck[] {
                 HasDirectory("tools").WithFiles("LICENSE.txt", "NOTICES.txt", "VERIFICATION.txt").AndFiles(CONSOLE_FILES).AndFiles(ENGINE_FILES).AndFile("nunit.choco.addins"),
                 HasDirectory("tools/agents/net20").WithFiles(AGENT_FILES).AndFile("nunit.agent.addins"),
-                HasDirectory("tools/agents/net40").WithFiles(AGENT_FILES).AndFile("nunit.agent.addins")
+                HasDirectory("tools/agents/net40").WithFiles(AGENT_FILES).AndFile("nunit.agent.addins"),
+                HasDirectory("tools/agents/netcoreapp3.1").WithFiles(AGENT_FILES_NETCORE).AndFile("nunit.agent.addins"),
```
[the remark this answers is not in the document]
CharliePoole: Yes.

build.cake
CharliePoole: In the new (i.e. linux) code, I actually experimented to determine what was copied correctly without publishing. I actually ascertained that 3.1 and 5.0 didn't need it at all. In the case of netcoreapp2.1, Publish failed to copy the the metadata assembly reference so I resorted to copying it from netstandard2.0.

ENGINE_API_PROJECT is built implicitly when we build ENGINE_PROJECT.

I strongly suspect that we are doing more than we need to with Publish in the Windows build, but I preferred not to mess with it right now.

build.cake
CharliePoole: Because we build the solution for Windows and MacOS, which we cannot do for Linux. I spent a lot of time trying to make building the solution work, but was defeated by the net35 projects. In fact, I think fewer stuff will work under a .NET Core build as time goes on and other runtimes go out of support.

jnm2:
I'm good with that.

PR #1092, opened 2022-01-14
Separate nunit.engine.core.tests from nunit.engine.tests

### LINKED ISSUE #1075: Separate nunit.engine.core.tests from nunit.engine.tests
The `nunit.engine` and `nunit.engine.core` assemblies are built for different sets of runtimes. The tests should match these rather than relying on conditional compilation to include the correct test assemblies.

Getting this done will make other work with nunit.engine.core much simpler.

mikkelbu:
Fixes #1075


CharliePoole:
If I can get the net5.0 tests to run in the morning, I'll add that before merging. Either way, the release is ready to go tomorrow.. well, today actually. :-)

PR #1093, opened 2022-01-15
Set roll-forward policy to Major for nunit3-console

### LINKED ISSUE #1012: Automatically roll-forward if no .NET Core 3.1 runtime available
Hi!

What do you think about adding RollForward Major policy in nunit3-console.csproj to allow it to run on .NET 5 without additional options? Currently I can run it without any problem using .NET Core 3.1 runtime, but on .NET 5 it fails with:
```
$ .../dotnet-runtime-5.0.9-linux-x64/dotnet exec .../nunit.consolerunner.netcore.3.12.0-beta2/tools/netcoreapp3.1/any/nunit3-console.dll
It was not possible to find any compatible framework version
The framework 'Microsoft.NETCore.App', version '3.1.0' was not found.
  - The following frameworks were found:
      5.0.9 at [.../dotnet-runtime-5.0.9-linux-x64/shared/Microsoft.NETCore.App]

You can resolve the problem by installing the specified framework and/or SDK.

The specified framework can be found at:
  - https://aka.ms/dotnet-core-applaunch?framework=Microsoft.NETCore.App&framework_version=3.1.0&arch=x64&rid=ubuntu.18.04-x64
```

And I need to use `--fx-version 5.0.9` or `--roll-forward Major` options to get it to run successfully. It will also help switching to .NET 6 after release. I can make a PR for this.

Linked issues: #858, #932

cc @ChrisMaddock

k15tfu:
Set roll-forward policy to Major for nunit3-console

Fixes #1012

### DIFF
--- a/src/NUnitConsole/nunit3-console/nunit3-console.csproj
+++ b/src/NUnitConsole/nunit3-console/nunit3-console.csproj
@@ -4,6 +4,7 @@
     <RootNamespace>NUnit.ConsoleRunner</RootNamespace>
     <AssemblyName>nunit3-console</AssemblyName>
     <TargetFrameworks>net20;netcoreapp3.1</TargetFrameworks>
+    <RollForward>Major</RollForward>
   </PropertyGroup>
   <PropertyGroup>
     <ApplicationIcon>..\..\..\nunit.ico</ApplicationIcon>


mikkelbu (approved):
LGTM, but I cannot test this, as I'm only on my phone

CharliePoole:
I thought about adding a test package test, but it would require running the test on a separate machine from the one that built it. Since Ilia has been using the fix in his work, we know it works, and the CI shows it doesn't break anything else, so I'm cool with it for the release so long as the merge is successful.

PR #1109, opened 2022-01-23
Generate assemblyInfo from csproj files

### LINKED ISSUE #1108: Generate AssemblyInfo files from csproj
Currently, we use hard-coded AssemblyInfo files but edit them in a build step prior to building. Ideally, we should generate them using MSBuild so they don't need to be checked in and we don't need to edit them.

CharliePoole:
Fixes #1108


CharliePoole:
Review: Tests all pass but I'm primarily looking for comments on using this approach  rather than manipulating assemblyinfo files. Seems like a better way to me, but are there any drawbacks?

mikkelbu:
I prefer this solution. The only downside - I can see - is that building outside of the cake script will not set this information, but I don't think that this is an issue.

CharliePoole:
That surprises me, but I checked and you're right. I removed `<GenerateAssemblyInfo>false</GenerateAssemblyInfo>` from the `Directory.Build.props` file, which caused the CI build to generate it by default. Maybe the default is different in the IDE. I'll add it back with `true` and see what happens.

CharliePoole:
@mikkelbu Well, actually, I see that it __is__ being generated. It's only the version numbers that are not being updated.

I experimented and I can force a particular version in `Directory.Build.props` conditioned on the `Version` property not being set by the CI build. However, that means the file would have to be updated after each release.

I agree with you that it's not a big issue but I'd rather not see version 1.0.0 each time I build in the IDE. Maybe some fixed number > any we plan to use for 3.x? Or maybe an environment variable? What do you think?

CharliePoole:
See new PR #1110 - no rush to review it but I think it's nice to have a 3.x anyway. At some point I'll try to figure out how to make GitVersion interact with the IDE so we always get the same versions.

PR #1110, opened 2022-01-24
Set version to 3.99 when building in IDE

CharliePoole:
Fix to earlier PR #1109 which left version at 1.0.0.0 when building in the IDE. This code sets it to 3.99.0.0 instead and uses suffix -VSIDE for informational version.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -58,27 +58,28 @@ Task("Clean")
     .Description("Cleans directories.")
     .Does(() =>
     {
-        CleanDirectory(PROJECT_DIR + "bin/");
-        CleanDirectory(PROJECT_DIR + "packages/");
+        CleanDirectory(BIN_DIR);
+        CleanDirectory(PACKAGE_DIR);
         CleanDirectory(IMAGE_DIR);
         CleanDirectory(EXTENSIONS_DIR);
         CleanDirectory(PACKAGE_DIR);
     });
 
-Task("DeleteObjectDirectories")
-    .WithCriteria(BuildSystem.IsLocalBuild)
+Task("CleanAll")
+    .Description("Cleans both Debug and Release Directories followed by deleting object directories")
     .Does(() =>
     {
-        Information("Deleting object directories");
+        Information("Cleaning both Debug and Release");
+        CleanDirectory(PROJECT_DIR + "bin");
+        CleanDirectory(PACKAGE_DIR);
+        CleanDirectory(IMAGE_DIR);
+        CleanDirectory(EXTENSIONS_DIR);
+        CleanDirectory(PACKAGE_DIR);
 
+        Information("Deleting object directories");
         foreach (var dir in GetDirectories("src/**/obj/"))
             DeleteDirectory(dir, new DeleteDirectorySettings() { Recursive = true });
     });
-
-Task("CleanAll")
-    .Description("Perform standard 'Clean' followed by deleting object directories")
-    .IsDependentOn("Clean")
-    .IsDependentOn("DeleteObjectDirectories");
 
 //////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


src/NUnitEngine/nunit.engine.api/nunit.engine.api.csproj
```diff
@@ -15,7 +15,7 @@
     <AssemblyTitle>NUnit Engine API ($(TargetFramework))</AssemblyTitle>
     <Description>Defines the interfaces used to access the NUnit Engine</Description>
     <AssemblyVersion>3.0.0.0</AssemblyVersion>
-    <FileVersion>$(ApiFileVersion)</FileVersion>
+    <FileVersion>3.99.0.0</FileVersion>
```
mikkelbu, line 18: I'm probably overlooking something, but why is it that this file has hardcoded `AssemblyVersion`/`FileVersion`? Is it because when we make a change here then we will manually update these versions?

build.cake
```diff
@@ -58,28 +58,29 @@ Task("Clean")
     .Description("Cleans directories.")
     .Does(() =>
     {
-        CleanDirectory(PROJECT_DIR + "bin/");
-        CleanDirectory(PROJECT_DIR + "packages/");
+        CleanDirectory(BIN_DIR);
+        CleanDirectory(PACKAGE_DIR);
```
mikkelbu, line 62: Where these wrong before. `BIN_DIR` is defined as `PROJECT_DIR + "bin/" + Configuration + "/"`
and `PACKAGE_DIR` is `Argument("artifact-dir", PROJECT_DIR + "package") + "/"`, so we are cleaning different folders now

build.cake
```diff
@@ -58,28 +58,29 @@ Task("Clean")
...
 
-Task("DeleteObjectDirectories")
-    .WithCriteria(BuildSystem.IsLocalBuild)
+Task("CleanAll")
+    .Description("Cleans both Debug and Release Directories followed by deleting object directories")
     .Does(() =>
     {
-        Information("Deleting object directories");
+        Information("Cleaning both Debug and Release");
+        CleanDirectory(PROJECT_DIR + "bin");
```
mikkelbu, line 73: So "Clean" only cleans the current configuration, but "CleanAll" cleans both?

mikkelbu (approved):
Just a couple of questions, otherwise it looks good

src/NUnitEngine/nunit.engine.api/nunit.engine.api.csproj
CharliePoole: No. It's because we set AssemblyVersion to 3.0.0.0. If we set AssemblyVersion explicitly, then MSBuild uses that value for the FileVersion rather than the value we passed in the Version property. We can stop doing this in 4.0 if we can come up with a breakdown of API such that the version can be updated without breaking runners or extensions.

build.cake
CharliePoole: Yes. I was comparing the Debug and Release builds using both command-line and the IDE and I was surprised to see that building Release in the command-line also cleaned my Debug folder. That seemed unexpected so I fixed it without creating a new issue - which in theory would have been the right thing to do. :-(

build.cake
CharliePoole: Exactly.

CharliePoole:
I'll merge this but I do want to come back to it at some point.

I have a question out to the GitVersion folks. I know that you can use GitVersion from within an IDE build, which would give us the same versions no matter how we build. I'd like that! I know how to do it but what I don't know is how to do it without installing two different GitVersion packages.

Anyway... I think 3.9.0.0-VSIDE is pretty clear for now and better than 1.0.0.0. :-)

Issue #1118, opened 2022-01-28
Reorganize build output (bin) directories


CharliePoole:
Currently we build into a set of bin directories according to the target runtime... net20, netcoreapp3.1, etc. Then we Publish the projects into the _same directories_. This can be very confusing to debug because it's not always obvious whether a problem has arisen in the build step or the publish step.

Since we do now publish all the binaries to BIN_DIR, I'm proposing that we go back to the naive build approach of having a bin directory under each project. We can then continue to publish them to the main bin directory (BIN_DIR) and that published structure does not have to match the existing one. Rather it can be made to look pretty much like the structure we want in our packages allowing further simplification.

@rprouse @mikkelbu @jnm2 
I'd appreciate comments on this idea before I do it.

mikkelbu:
Given the confusion in #1044 I think it is fine to separate them

CharliePoole:
Changes in the 3.15.0 release have simplified the publishing step substantially. I'm moving this to 4.0 for further investigation.

PR #1119, opened 2022-01-29
Issue 1044

### LINKED ISSUE #1044: Agent for .NET 6.0
@nunit/engine-team 

It's certain we will want one eventually. Do we want one for the 3.13 release? 

I think it's easy to add ... pretty much a copy of the 5.0 launcher code. Drawback is everyone who works on the engine needs to have it.

One way to lessen the developer impact would be to include it conditionally and only define the condition in our CI build.

Thoughts?

CharliePoole:
Fixes #1044 

This PR gets the net6.0 agent working while minimizing structural changes. The latter will wait till 4.0.


CharliePoole:
@mikkelbu @jnm2 This new feature is significant enough that I would feel better having one of you review it before I commit. TIA

build.cake
```diff
@@ -91,17 +91,17 @@ Task("Build")
...
-public void BuildUsingMSBuildOnWindowsOrMacOS()
+public void BuildSolution()
 {
     MSBuild(SOLUTION_FILE, CreateMSBuildSettings("Build").WithRestore());
 
-    Information("Publishing .NET Core & Standard projects so that dependencies are present...");
+    DisplayBanner("Publish .NET Core & Standard projects");
 
     foreach (var framework in new[] { "netstandard2.0", "netcoreapp3.1" })
         MSBuild(ENGINE_PROJECT, CreateMSBuildSettings("Publish")
```
mikkelbu, line 107: I cannot comment on the line below (lines 111-114). But shouldn't we add "net6.0" to building of agent project below? 

I.e. in
```
    foreach (var framework in new[] { "netcoreapp3.1", "net5.0" })
        MSBuild(AGENT_PROJECT, CreateMSBuildSettings("Publish")
            .WithProperty("PublishDir", BIN_DIR + "agents/" + framework));
```

build.cake
```diff
@@ -267,6 +269,19 @@ Task("TestNet50EngineCore")
...
+    .IsDependentOn("Build")
+    .OnError(exception => { UnreportedErrors.Add(exception.Message); })
+    .Does(() =>
+    {
+        RunDotnetNUnitLiteTests(NETCORE_ENGINE_CORE_TESTS, "net6.0");
+    });
+
 //////////////////////////////////////////////////////////////////////
 // TEST .NET 2.0 ENGINE
 //////////////////////////////////////////////////////////////////////
```
mikkelbu, line 287: I cannot comment on the line, but I was going to ask why we didn't add "net6.0" in `TestNetCore31Engine`, but now I can see that we actually don't use `runtime` in the loop, so we are just running `RunDotnetNUnitLiteTests(NETCORE_ENGINE_TESTS, "netcoreapp3.1");` two times.

```
        var runtimes = new[] { "netcoreapp3.1", "net5.0" };

        foreach (var runtime in runtimes)
        {
            RunDotnetNUnitLiteTests(NETCORE_ENGINE_TESTS, "netcoreapp3.1");
        }
    });
```

mikkelbu (approved):
LGTM. Mostly a lot of comments, questions, and suggestions

azure-pipelines.yml
```diff
@@ -105,9 +115,14 @@ jobs:
 - job: macOS
   condition: not(or(startsWith(variables['Build.SourceBranchName'], 'azure-windows-'),startsWith(variables['Build.SourceBranchName'], 'azure-linux-')))
   pool:
-    vmImage: macOS-latest
+    vmImage: macOS-11
```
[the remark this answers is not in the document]
CharliePoole: Yes... I did it to be sure I knew what image we were using. Good point about Ubuntu - I'll fix.

src/NUnitEngine/nunit.engine.core/nunit.engine.core.csproj
```diff
@@ -21,7 +21,7 @@
     <Reference Include="System.Web" />
   </ItemGroup>
   
-  <ItemGroup Condition="'$(TargetFramework)'=='netstandard2.0' or '$(TargetFramework)'=='netcoreapp3.1' or '$(TargetFramework)'=='net5.0'">
```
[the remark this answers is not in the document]
CharliePoole: No, I appear to have added it unnecessarily when I did net5.0.

build.cake
CharliePoole: I'll check this after I make the other changes. I tried to only publish where I determined experimentally that I needed to and it's possible that I didn't need to for net6.0.

build.cake
CharliePoole: Also odd. I'll check this after I make the other changes.

build.cake
```diff
@@ -267,6 +269,19 @@ Task("TestNet50EngineCore")
         RunDotnetNUnitLiteTests(NETCORE_ENGINE_CORE_TESTS, "net5.0");
     });
 
+//////////////////////////////////////////////////////////////////////
+// TEST NET 5.0 ENGINE CORE
+//////////////////////////////////////////////////////////////////////
+
+Task("TestNet60EngineCore")
```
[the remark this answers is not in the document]
CharliePoole: Good catch!

cake/package-definitions.cake
```diff
@@ -26,6 +27,7 @@ public void InitializePackageDefinitions(ICakeContext context)
         NetCore21Test,
         NetCore31Test,
         Net50Test,
+        Net60Test,
         NetCore21PlusNetCore31PlusNet50Test
```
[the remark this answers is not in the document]
CharliePoole: Good Idea. I just changed that one to include all four. Also added a test of running .NET  Framework 4.0 and .NET 6.0 together, since we didn't have an FX / Core combination test.

build.cake
CharliePoole: Well this is a mystery. I absolutely know we needed to do the Publish step at one time. Now I can comment out all that code and everything still works. It may be because I'm now using .NET  6.0 to host the build.

For now, I'll leave it as is and revisit in another issue.

build.cake
CharliePoole: Yes... the second intention was to run the engine tests twice, under each runtime. I'll think about whether that's really necessary now.

build.cake
CharliePoole: This reminds me about issue #1118. It seems to me that publishing in the same directory we use for compile output leads to confusion about where a particular file came from. What do you think?

build.cake
CharliePoole: I decided this is no longer necessary, since we actually have .NET 5.0 and 6 .0 builds, which we test separately.

build.cake
mikkelbu: I think it is fine to separate them, as you suggest in #1118

PR #1121, opened 2022-02-01
Split ExtensionService into two classes

### LINKED ISSUE #1120: Split ExtensionService into two classes: the service itself and an implementation class
Part of #1049
.
ExtensionService will aim specifically at handling extensions to the engine making use of a new class, which I'm tentatively calling ExtensionManager. ExtensionManager will provide a lower-level interface allowing the caller to specify which assemblies are to be examined for extension points and where the extensions are to be looked for.

CharliePoole:
Fixes #1120


mikkelbu (approved):
LGTM

PR #1136, opened 2022-02-11
Eliminate in-process execution and all process options

### LINKED ISSUE #861: Eliminate In-Process Execution
Allowing the user to select in-process execution gives no specific benefit although it has the side-effect of making debugging easier. We have discussed removing it in version 4.0. This issue is created as a place for the @nunit/engine-team to discuss the two options I see for this change. I've listed some notes on each option.

Option 1 - Just eliminate In-process execution

1. This is a test runner, which has to deal with code that may behave badly. It's really not reasonable to expect the runner to survive and report all bad behavior by the tests if the tests are running in-process.
2. The test itself benefits by being isolated in a separate process.
3. The change would simplify the engine because it no longer requires verifying that a particular assembly is capable of running in the same process as the runner, based on its target runtime.
4. The main use of in-process is to allow easier debugging, so loosing the option will make debugging something we need to focus on. There is already a way to debug the tests inside the agent process, which we should work to improve.

Option 2 - Eliminate the ProcessModel setting entirely

1. If we drop in-process execution, then all tests run in separate assemblies. The only remaining issue is whether multiple assemblies run in the same process or in separate processes. It's arguable that there is no reason for running in one process. That setting is only available for historical reasons: it was developed before we developed multiple process execution.
2. As with AppDomains, the user has no need to control how the runner allocates assemblies to processes.
3. Multiple Process execution (one process per assembly) is now the default. It's how the engine actually works best.
4. We could simply present Agents (rather than processes) to the user as the main conceptual entity used in running tests. The runner can decide in many cases what those agents are - processes, separate machines, etc.

CharliePoole:
Fixes #861


CharliePoole:
Azure Devops Pipelines are down. Merging anyway.

PR #1156, opened 2022-02-22
Issue1139 add nunit4 tests

### LINKED ISSUE #1139: Add a package test that uses nunit framework 4.0
That way we'll know when and if the api changes!

manfred-brands:
Fixes #1139 

First commit is to find a preview version of VS2022 when VS2019 is installed.
The code should preferably throw if it can't find MSBuild 17.0 as the .NET 6 sdk complains it it is run with MSBuild 15.0
Also update the minimum requirement in BUILDING.md

Second commit adds mock-assembly-nunit4 and associated tests.

### DIFF
--- a/src/NUnitEngine/mock-assembly-nunit4/mock-assembly-nunit4.csproj
+++ b/src/NUnitEngine/mock-assembly-nunit4/mock-assembly-nunit4.csproj
@@ -1,0 +1,27 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <RootNamespace>NUnit.Tests</RootNamespace>
+    <TargetFrameworks>net45;netcoreapp3.1;net5.0;net6.0</TargetFrameworks>
+    <SignAssembly>true</SignAssembly>
+    <AssemblyOriginatorKeyFile>..\..\nunit.snk</AssemblyOriginatorKeyFile>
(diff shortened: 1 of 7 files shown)


src/NUnitEngine/mock-assembly-nunit4/mock-assembly-nunit4.csproj
```diff
@@ -0,0 +1,27 @@
...
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <Product>NUnit Engine</Product>
+    <AssemblyTitle>NUnit4 Mock Assembly ($(TargetFramework))</AssemblyTitle>
+    <Description>Assembly used in testing the engine</Description>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="NUnit" Version="4.0.0-dev-07273" />
```
CharliePoole, line 18: I like that you specify an exact build here, to avoid errors due to changes in the framework project. It would be nice to use a property, based no a constant we define more prominently at the top level, but that's probably best for a latter improvement.

CharliePoole (approved):
Nice and clean! Thanks for doing this. If you're willing, I can definitely find more for you to do. :+1:

CharliePoole:
@manfred-brands I sent you an invitation to join the engine team. If you have questions, feel free to contact me offline.

PR #1167, opened 2022-02-28
#1166 use app manifest

### LINKED ISSUE #1166: nunit3-console.exe does not have a manifest
### Steps to reproduce:
1. Download the NUnit Console Runner executable (e.g. Download NuGet package and extract it)
2. Open a Visual Studio Developer Command Prompt
3. Extract the app manifest by running the following command: `mt.exe -inputresource:"C:\path\to\nunit3-console.exe";#1 -out:;#1 -out:"C:\path\to\extracted.manifest"`
4. Inspect the extracted manifest

Expected: manifest is identical to https://github.com/nunit/nunit-console/blob/main/src/NUnitConsole/nunit-console/app.manifest
Actual: manifest is a default manifest

### Implications: 
* test code that uses `Environment.OSVersion.Version` (and friends) is not working as expected
* message on application start is misleading:
```
NUnit Console 3.15.0 (Release)
Copyright (c) 2022 Charlie Poole, Rob Prouse
Montag, 28. Februar 2022 15:13:16

Runtime Environment
   OS Version: Microsoft Windows NT 6.2.9200.0 ********** this is printed when running on Windows 10
   Runtime: .NET Framework CLR v4.0.30319.42000
```

johannes-schmitt:
fixes #1166

### DIFF
--- a/src/NUnitConsole/nunit-console/nunit-console.csproj
+++ b/src/NUnitConsole/nunit-console/nunit-console.csproj
@@ -16,6 +16,7 @@
 
   <PropertyGroup>
     <ApplicationIcon>..\..\..\nunit.ico</ApplicationIcon>
+    <ApplicationManifest>app.manifest</ApplicationManifest>
   </PropertyGroup>
 
   <ItemGroup>


CharliePoole:
Chocolatey is choking on the version string `4.0.0--1166-add-applicati`. The pre-release suffix is constructed from your branch name, which starts with `#` and is replaced by `-`. I suspect the double `-` is causing it, but I'm not sure. Easiest way to find out would be to change your branch name and do a new PR. Once I know the cause, I may be able to avoid the problem in the future by modifying the scripts.

johannes-schmitt:
@CharliePoole, recreated the PR with a new branch name, see https://github.com/nunit/nunit-console/pull/1169

Issue #1193, opened 2022-06-28
Prevent Crash under .NET 7.0 in 3.15.1 build


CharliePoole:
This was fixed for 4.0 in PR #1175. The fix needs to be back-ported to the 3.15.1 release.

See issue nunit/nunit3-vs-adapter#987

CharliePoole:
Fix is available from MyGet as version 3.15.1-dev00003

CharliePoole:
It turns out we already had an unlisted 3.15.1 release, identical to 3.15.0, on Nuget.org. This was due to an error in the build when 3.15 was released. In spite of being unlisted, it has had 11000 downloads!

Consequently, this issue and its release will now be rebuilt as 3.15.2.

PR #1197, opened 2022-06-30
Recognize .NET 7 when installed

CharliePoole:
Re-merging the fix for #1193 into the new version3 branch

### DIFF
--- a/src/NUnitEngine/nunit.engine.core/RuntimeFramework.cs
+++ b/src/NUnitEngine/nunit.engine.core/RuntimeFramework.cs
@@ -201,6 +201,8 @@ namespace NUnit.Engine
                             return new Version(5, 0, 1);
                         case 6:
                             return new Version(6, 0, 0);
+                        case 7:
+                            return new Version(7, 0, 0);
                     }
                     break;
             }
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
     "version": "6.0.101",
-    "rollForward": "patch"
+    "rollForward": "feature"
   }
 }


src/NUnitEngine/nunit.engine.core/RuntimeFramework.cs
```diff
@@ -201,6 +201,8 @@ private Version GetClrVersionForFramework(Version frameworkVersion)
                             return new Version(5, 0, 1);
                         case 6:
                             return new Version(6, 0, 0);
+                        case 7:
+                            return new Version(7, 0, 0);
```
AbhitejJohn, line 205: @CharliePoole : I was curious if this would be a problem once .NET 8 (and future versions of .NET) get released...
/cc: @OsirisTerje

CharliePoole: Yes, it would, at least for anyone continuing to use 3.15.2 at that point. However, subsequent changes, to be released in 3.16, eliminated throwing of an exception when the runtime is not recognized. In addition, version 4.0 code eliminates this call entirely, dropping the need for the CLR version entirely. Both of the above are superior, in my opinion, to the often suggested option of using `<MajorVersion>.0.0` because that involves predicting Microsoft's future actions.

AbhitejJohn: That makes a lot of sense. Glad the newer versions have this fixed up.

madelson: @CharliePoole I'm hitting this again with .NET 8 preview installed. How do I get VS/dotnet test to use 3.16?

CharliePoole: That's based on the version of the NUnit VS adapter you are using. Check with @OsirisTerje on that project.

OsirisTerje: @madelson We need a new version of the engine, 3.16.3.  It should be released any day now, and then version 4.40 of the adapter will have this embedded.  It will be released immediately after the 3.16.3

OsirisTerje:
@AbhitejJohn Version 3.16.3 of the engine/console is out now, and it accepts all higher versions, and the adapter 4.4.0 is out (beta.1 now, but the non-beta will be released this weekend,) with the 3.16.3 engine, so there will be no further issues with that one either.  See https://docs.nunit.org/articles/vs-test-adapter/AdapterV4-Release-Notes.html

PR #1241, opened 2022-10-12
Restructure build output directories

### LINKED ISSUE #1118: Reorganize build output (bin) directories

CharliePoole:
Fixes #1118


CharliePoole:
The Azure build is now completely broken because .net 7.0 RC 2 has replaced RC 1.

PR #1285, opened 2023-01-01
Resolve WindowsDesktop and AspNetCore tests correctly

### LINKED ISSUE #1274: NUnit Console won't run unless .NET Desktop and ASP.NET Runtimes are installed
This was a side effect of the fixes for #1203 and #1208 for the 3.16 release. The agent now requires additional runtimes to be installed even though not all projects under test will require them. Many users will already have the runtimes installed because commonly used approaches to installing .NET bring them both in. However, that won't be the case for all users and some CI environments may not be set up this way. The additional runtimes should not be required unless testing an assembly, which uses them.

CharliePoole:
Fixes #1274

As noted, this fix backs out the changes, which resulted in issue #1274 and adds code to the resolver to directly locate the two additional frameworks, Microsoft.WindowsDesktop.App and Microsoft.AspNetCore.App, when the test assembly depends on one of them.

This is a major change and should have a code review if someone has the time. I'll leave it for a few days before merging.


mikkelbu:
I'll take a look now

src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyResolver.cs
```diff
@@ -64,8 +77,74 @@ private Assembly OnResolving(AssemblyLoadContext context, AssemblyName name)
...
+                        return _loadContext.LoadFromAssemblyPath(candidate);
+                }
+            }
+
             return null;
         }
+
+        private static string GetDotNetInstallDirectory()
+        {
+            if (Path.DirectorySeparatorChar == '\\')
```
mikkelbu, line 96: Can we use `RuntimeInformation.IsOSPlatform(OSPlatform.Windows)` instead to check whether we are on windows?

src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyResolver.cs
```diff
@@ -64,8 +77,74 @@ private Assembly OnResolving(AssemblyLoadContext context, AssemblyName name)
                 }
             }
 
+            foreach(string frameworkDirectory in AdditionalFrameworkDirectories)
```
mikkelbu, line 80:
```suggestion
            foreach (string frameworkDirectory in AdditionalFrameworkDirectories)
```

src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyResolver.cs
```diff
@@ -64,8 +77,74 @@ private Assembly OnResolving(AssemblyLoadContext context, AssemblyName name)
...
+
             return null;
         }
+
+        private static string GetDotNetInstallDirectory()
+        {
+            if (Path.DirectorySeparatorChar == '\\')
+            {
+                // Running on Windows so use registry
+                RegistryKey key = Registry.LocalMachine.OpenSubKey(@"Software\dotnet\SetUp\InstalledVersions\x64\sharedHost\");
```
mikkelbu, line 99: Is this regkey always the one to use?

src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyResolver.cs
```diff
@@ -64,8 +77,74 @@ private Assembly OnResolving(AssemblyLoadContext context, AssemblyName name)
...
+                return (string)key?.GetValue("Path");
+            }
+            else
+                return "/usr/shared/dotnet/";
```
mikkelbu, line 103: What if the user have installed dotnet in another location

mikkelbu (approved):
I don't know much about this area of the codebase, but as far as I can tell the changes look sound. My main worry is whether the regkey always can be used (in windows) and whether the path is always the correct one in Linux (or we should look it up - using e.g. `which dotnet` or `whereis dotnet`).

CharliePoole:
@mikkelbu Thanks for reviewing. A general answer to your general comment... I'm too suspect that non-standard installs could break this, particularly on Linux. However it wouldn't be a new issue - the same key and linux path is what we already use to figure out what versions of the runtime are installed.

In general, I think non-standard install locations are intended to be hidden from general use. This is similar to when you install a tool to a special path... dotnet cli can't find the tool unless you tell it the path in some way.

src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyResolver.cs
CharliePoole: AFAIK - and it's what we use elsewhere

src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyResolver.cs
CharliePoole: In that case, it can't be found. Short of a global search of the hard drive, there seems to be no built-in way of finding installations. I believe this is actually a feature, allowing you to install a test copy of the runtime without polluting the version in the standard location.

Issue #1307, opened 2023-01-13
NUnit3TestAdapter integration:  Exception when using NUnit.Engine 3.16.2


OsirisTerje:
See the following branch in the adapter repo:  https://github.com/nunit/nunit3-vs-adapter/tree/engine316-2

The adapter test project has to be set to run only netcore3.1 (but that is not this issue, just to be able to run it at all).
The test failing is: ThatDiscovererNUnitEngineAdapterIsInitialized

The exception happens here: 
![image](https://user-images.githubusercontent.com/203432/212388532-2e35f0f4-1a1a-4500-b5fc-e0e25c244200.png)
![image](https://user-images.githubusercontent.com/203432/212388566-b381692f-5223-4c5c-bae1-332ac4685ed4.png)

So it fails at NUnit.Engine.Internal.TestAsemblyResolver.FindBestVersionDir

The runner is MasterTestRunner
![image](https://user-images.githubusercontent.com/203432/212389113-58919031-6a8d-479a-bcd7-afde1ea1865c.png)

The filter contains the method being run, and that is that.

CharliePoole:
You'll want to figure out what subdirectory name it is trying to process when it fails. It's likely your .NET 8.0 sdk directory.

OsirisTerje:
No, it is not, I have not installed net 8.   Going to debug it to find out :-)

OsirisTerje:
The error starts here:
![image](https://user-images.githubusercontent.com/203432/212495616-51a90071-a964-4ff6-b9bd-9fd275ee6e24.png)
and fails here
![image](https://user-images.githubusercontent.com/203432/212495638-77c4d645-a4e9-4779-9e06-d59418db2023.png)

Now, I can't see MonoTouch anywhere in the folders for the Sdks.

CharliePoole:
Variable name is the `AssemblyName` of the assembly we are trying to resolve. It should be a test assembly or a dependency of a test assembly. It's apparently possible that some dependencies may be references without a version. I've never seen that before so I'd like to know how it happens. It may be an error we should report or perhaps we should accept any version.

CharliePoole:
Also note that we wouldn't in `FindBestVersionDir` unless "monotouch" were found somewhere in scanning the `AdditionalFrameworkDirectories`, which are initialized in the static constructor of `TestAssemblyResolver`.

It may be useful to add some WriteLines in the loop that calls `FindBestVersionDir`.

OsirisTerje:
[...]

PR #1308, opened 2023-01-15
Fix null version

OsirisTerje:
Fix for #1307 : NUnit3TestAdapter integration: Exception when using NUnit.Engine 3.16.2

Removed the line that caused the nullref exception, as the resulting variable was not used in the method.


CharliePoole:
Ha! I've done this a few times myself.

PR #1309, opened 2023-01-15
Fix for #1307

OsirisTerje:
Fix for #1307 : NUnit3TestAdapter integration: Exception when using NUnit.Engine 3.16.2

Removed the line that caused the nullref exception, as the resulting variable was not used in the method.

### DIFF
--- a/src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyResolver.cs
+++ b/src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyResolver.cs
@@ -108,6 +108,8 @@ namespace NUnit.Engine.Internal
 
         private static string FindBestVersionDir(string libraryDir, Version targetVersion)
         {
+            if (targetVersion == null)
+                return null;
             Version bestVersion = new Version(0,0);
             foreach (var subdir in Directory.GetDirectories(libraryDir))
             {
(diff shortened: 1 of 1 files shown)


src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyResolver.cs
```diff
@@ -108,7 +108,8 @@ private static string GetDotNetInstallDirectory()
 
         private static string FindBestVersionDir(string libraryDir, Version targetVersion)
         {
-            string target = targetVersion.ToString();
+            if (targetVersion == null)
```
CharliePoole, line 111: It may be clearer to put the null check in the calling code and never call this method with a null version. Since it's a private method and is only called in one place, I don't think a guard is really needed, but you could add one here in addition to the external check.

CharliePoole (approved):
That change seems to do it, and you may want to merge and release in order to fix the problem with the adapter. However, see my inline comment and my general comment as well.

CharliePoole:
Here's another possibility - a bit more work maybe but could eliminate future problems.

We started having problems after I put code in the engine to load .NET Core dependencies,
eventually using `Microsoft.Extensions.DependencyModel`. I had to do that for the console
and GUI but the adapter never actually needed it.

Should we consider bypassing that code when running under the adapter? The engine could probably detect
the environment in which it is used, but it might be cleaner to just introduce a setting, which would eliminate
all or part of the Reload event handling I introduced.

CharliePoole:
BTW, I am trying to strike a balance here between following procedures as they are written and common sense.

As I'm no longer part of the Engine Team, I should only comment rather than approve a PR. Nevertheless, I did use the Approved selection to indicate that I think it's a good change. I also added some alternative suggestions.

On the other hand, I'll now leave it to someone else to actually merge any change.

PR #1317, opened 2023-02-12
fix assembly dependencies resolver for netcore

svg2003:
based on discussion from here: https://github.com/nunit/nunit-console/issues/1311


OsirisTerje:
Merging this. Seems this fix will unblock other changes (net 8 fix).

danielmarbach:
We are trying to figure out if this change might have broken some of our tests. The scenario is that we have an assembly that dual targets net472 and net6.0. It references System.Text.Json 7.0.2. When we execute the tests with the new engine containing those fixes it seems to load System.Text.Json of .NET 6 instead of 7.0.2 (the explicit reference). This issue only occurs on NET6 on Linux but not on Windows. // cc @Mauroservienti

PR #1320, opened 2023-02-14
Net8

OsirisTerje:
- Updated to not crash when .net 8 is installed
- Updated building information

### DIFF
--- a/src/NUnitEngine/nunit.engine.core/Internal/RuntimeFrameworks/NetCoreFrameworkLocator.cs
+++ b/src/NUnitEngine/nunit.engine.core/Internal/RuntimeFrameworks/NetCoreFrameworkLocator.cs
@@ -20,16 +20,10 @@ namespace NUnit.Engine.Internal.RuntimeFrameworks
 
             foreach (string dirName in GetRuntimeDirectories())
             {
-                Version newVersion;
-                if (TryGetVersionFromString(dirName, out newVersion) && !alreadyFound.Contains(newVersion))
+                if (TryGetVersionFromString(dirName, out var newVersion) && !alreadyFound.Contains(newVersion))
                 {
                     alreadyFound.Add(newVersion);
-                    // HACK: Avoid Exception for an unknown version - see issue #1223
-                    // Requires change in RuntimeFramework.GetClrVersionForFramework()
-                    if (newVersion.Major <= 8)
-                        yield return new RuntimeFramework(RuntimeType.NetCore, newVersion);
-                    else
-                        log.Error($"Found .NET {newVersion.ToString(2)}, which is not yet supported.");
+                    yield return new RuntimeFramework(RuntimeType.NetCore, newVersion);
                 }
             }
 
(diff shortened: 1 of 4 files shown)


src/NUnitEngine/nunit.engine.core/RuntimeFramework.cs
```diff
@@ -204,6 +204,8 @@ private Version GetClrVersionForFramework(Version frameworkVersion)
                             return new Version(6, 0, 0);
                         case 7:
                             return new Version(7, 0, 0);
+                        case 8:
+                            return new Version(8, 0, 0);
```
manfred-brands, line 208: Instead of doing this every year, can we not have a default case where version is passed through.

src/NUnitEngine/nunit.engine.core/Internal/RuntimeFrameworks/NetCoreFrameworkLocator.cs
```diff
@@ -26,7 +26,7 @@ public static IEnumerable<RuntimeFramework> FindDotNetCoreFrameworks()
                     alreadyFound.Add(newVersion);
                     // HACK: Avoid Exception for an unknown version - see issue #1223
                     // Requires change in RuntimeFramework.GetClrVersionForFramework()
-                    if (newVersion.Major <= 7)
+                    if (newVersion.Major <= 8)
```
manfred-brands, line 29: As nothing is really done with this, we could as well make this the default case for all Major > 5

manfred-brands (changes requested):
As the runner doesn't seem to really do anything with this, is it possible to cherry pick my commit to main from 11 months ago where the dependency was removed completely?

See:
Author:		Manfred Brands <manfred-brands@users.noreply.github.com>
Date:		11 months ago (31/03/2022 12:15:58)
Commit hash:	d5648b93cdf8e755050b80a3ac12a416a32e5a18
Child:		100b4340
Parent:		0d5ea360

Removed ClrVersion from RuntimeFramework

Decision logic moved to Runtime class.

src/NUnitEngine/nunit.engine.core/RuntimeFramework.cs
rprouse: I'd agree with this. It seems silly to keep adding a new version and passing in the case value.

src/NUnitEngine/nunit.engine.core/Internal/RuntimeFrameworks/NetCoreFrameworkLocator.cs
rprouse: This also sounds like a good idea.

src/NUnitEngine/nunit.engine.core/RuntimeFramework.cs
OsirisTerje: Agree, I'll fix

src/NUnitEngine/nunit.engine.core/Internal/RuntimeFrameworks/NetCoreFrameworkLocator.cs
OsirisTerje: Agree, will fix too.

OsirisTerje:
@manfred-brands I wasn't really aware of this, @CharliePoole made some comments about it, but I didn't know it had been done.  Will do this.

rprouse (approved):

Issue #1330, opened 2023-04-05
Wrong assembly resolving logic when several runtimes are installed and assembly is missing in deps.json


svg2003:
Suppose we have Test.dll, targeted for net6 that is using some net472 targeted assembly (RealCode.dll).
RealCode.dll is using some assembly, that is NOT in deps.json file for Test.dll. Let it be System.Windows.Forms

When TestAssemblyResolver is trying to resolve this assembly, it will get: 

_private Assembly OnResolving(AssemblyLoadContext context, AssemblyName name)
        {
            DebugLog("OnResolving {0} version {1}", name.Name, name.Version); =>  [TestAssemblyResolver][OnResolving]: OnResolving System.Windows.Forms version 4.0.0.0_

Now, if test computer has several Runtimes installed, Shared folder would have several folders for every installed runtime

 _Directory of C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App

01/28/2023  11:53 AM    <DIR>          .
01/28/2023  11:53 AM    <DIR>          ..
06/18/2022  01:16 AM    <DIR>          3.1.26
06/27/2022  10:03 PM    <DIR>          5.0.17
10/21/2022  01:27 AM    <DIR>          6.0.10
11/23/2022  02:21 PM    <DIR>          6.0.11
01/28/2023  11:53 AM    <DIR>          6.0.13
01/28/2023  03:48 AM    <DIR>          7.0.2_

So, when TestAssemblyResolver would start enumerate those folders, it will load wrong version (5.0.17) instead of some one from 6.0.x, that is in reality using agent:

https://github.com/nunit/nunit-console/blob/main/src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyResolver.cs#L86

 foreach(string frameworkDirectory in AdditionalFrameworkDirectories)
            {
                **DebugLog("AdditionalDir: {0}", frameworkDirectory); => [TestAssemblyResolver][OnResolving]: AdditionalDir: C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App**
                var versionDir = FindBestVersionDir(frameworkDirectory, name.Version);
                **DebugLog("VerDir: {0}", versionDir); => [TestAssemblyResolver][OnResolving]: VerDir: 5.0.17**
                if (versionDir != null)
                {
                    string candidate = Path.Combine(frameworkDirectory, versionDir, name.Name + ".dll");
                    if (File.Exists(candidate))
[...]

PR #1332, opened 2023-04-06
Wrong assembly resolving logic when several runtimes are installed and assembly is missing in deps.json

svg2003:
Based on https://github.com/nunit/nunit-console/issues/1330
Improved assembly resolving logic

### DIFF
--- a/src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyLoadContext.cs
+++ b/src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyLoadContext.cs
@@ -12,6 +12,8 @@ namespace NUnit.Engine.Internal
 {
     internal sealed class TestAssemblyLoadContext : AssemblyLoadContext
     {
+        private static readonly Logger _log = InternalTrace.GetLogger(typeof(TestAssemblyLoadContext));
+
         private readonly string _testAssemblyPath;
         private readonly string _basePath;
         private readonly TestAssemblyResolver _resolver;
@@ -27,18 +29,15 @@ namespace NUnit.Engine.Internal
 
         protected override Assembly Load(AssemblyName name)
         {
-            var assemblies = AppDomain.CurrentDomain.GetAssemblies();
-            var loadedAssembly = assemblies.FirstOrDefault(x => x.GetName().Name == name.Name);
+            _log.Debug("Loading {0} assembly", name);
+
+            var loadedAssembly = base.Load(name);
             if (loadedAssembly != null)
             {
+                _log.Info("Assembly {0} ({1}) is loaded using default base.Load()", name, GetAssemblyLocationInfo(loadedAssembly));
                 return loadedAssembly;
             }
 
-            loadedAssembly = base.Load(name);
-            if (loadedAssembly != null)
-            {
-                return loadedAssembly;
-            }
 
             var runtimeResolverPath = _runtimeResolver.ResolveAssemblyToPath(name);
             if (string.IsNullOrEmpty(runtimeResolverPath) == false &&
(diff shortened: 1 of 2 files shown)


svg2003:
@dotnet-policy-service agree

svg2003:
Could you please merge this PR ?

src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyLoadContext.cs
```diff
@@ -12,6 +12,8 @@ namespace NUnit.Engine.Internal
 {
     internal sealed class TestAssemblyLoadContext : AssemblyLoadContext
     {
+        private static readonly Logger _log = InternalTrace.GetLogger(typeof(TestAssemblyLoadContext));
```
CharliePoole, line 15: Virtually all the other code uses `log` rather than `_log` for this even though it's technically not in accordance with the naming standards. To avoid surprise to other developers, I would use `log`.

CharliePoole (commented):
I like the extra logging you added.

I'd like to see a package test added which fails without the change and passes with it. Otherwise, all knowledge of what a change actually accomplished tends to get lost and regressions are not detected until the users report them.

CharliePoole:
A team member has to review and approve this in order for it to be merged. As I'm no longer a team member, I made my review for comment only.

src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyLoadContext.cs
svg2003: I selected name, looking on other class members. Seems you're using '_' prefix for them.

svg2003:
Overall, I agree, that it's the pity that "units test package" does not use units tests itself.
Would be good, if some experienced contributor, who knows whole internal architecture would prepare some "framework" to add such tests. But I think it should be the separate PR for that.

CharliePoole:
Package Tests are, in fact, not unit tests. They are run in the build script after the build, unit tests, package creation and package content verification. Assuming you are running the script locally you should see that happening... otherwise it's visible in the output from AppVeyor.

Repeating here my comments to @OsirisTerje in #1330 

> Yes, the script consists of `.cake files`, which are (except for `build.cake`) in the `cake` directory. This is something I had experimented with for a while, so it's at different stages in each project I worked on and the version3 and version 4 engine code is different as well.

> In the version3 branch, `cake/build-settings.cake` has the definitions of the tests and the logic for running them is in `cake/package-definitions`. The individual packages we create are defined in `cake/packages.cake`. Note that this code is more advanced than the main branch, which uses an earlier version of the script. Ideally, the script should be a package itself, referenced by projects that use it. That's the direction I had planned to take with the engine and have now done with TestCentric.

> It sounds as if you are on the right path @OsirisTerje and I hope you'll find this useful. To be frank, I couldn't have put out releases of the engine over the past year without having these tests to lean on.

Provided you have a test assembly demonstrating what didn't work before, it's fairly trivial to add a package test. Unfortunately, the script was still a work in progress when I left, so there's no how-to documentation.

src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyLoadContext.cs
CharliePoole: Yes, that's technically the standard... but it's log in virtually all other files.

svg2003: I frankly don't see the difference. It's the same class member variable, like any other members, so I'd expect consistency in naming there. 
But if my assumptions are wrong - that's fine. If it's the requirements, but not suggestion - I can change the name from '_log' to 'log'

CharliePoole: It's not a choice between consistency and inconsistency but between two consistencies: (1) The observed practice that member variables start with '_' and (2) the use of `log.Debug`, `log.Info`etc. throughout the other files of the application. Beyond that, I don't believe there are any requirements except doing what makes sense to you.

As I already said, my comments are only comments. Only follow them if they make sense. :-)

svg2003: Well, might be "requirement" sounds too hard, but, it's always better when some person/or group of people maintains architectural consistency. 
Anyway, I checked other places, and there are "log" everywhere, so I renamed my loggers as well, and pushed a commit. Sorry for inattention.

CharliePoole:
@nunit/engine-team Ping!

OsirisTerje:
Would like to see some more testing here, but merging this, it may solve the issues at hand.  
I'll see if I can have some package tests added.

PR #1345, opened 2023-05-22
Detect the 'partition' token during filter element parsing

Plasma:
This adds support for a new PartitionFilter at https://github.com/nunit/nunit/pull/4392


OsirisTerje:
@Plasma Thanks for this.  In order to get this into the adapter, you have to create similar PRs tom branch `version315`.  Further, another PR is needed for 'master' which handles the 4.0.   I am not sure if we will continue the 3.16 series, due to the critical loading issues we have there.  Can be better to jump straight to 4.X.

Plasma:
Thanks @OsirisTerje I've opened PR #1346 and #1347

PR #1370, opened 2023-12-15
Load unmanaged assemblies the same way as managed assemblies.

### LINKED ISSUE #1253: Problem loading unmanaged assemblies
Issue #1208 became a sort of anthology of somewhat related problems, leading (me) to a certain amount of confusion. All issues are solved, except the one described in [this comment](https://github.com/nunit/nunit-console/issues/1208#issuecomment-1255019266) by @cw397. I'm making this problem a separate issue. The extract below gives the original description of the problem.

---

Our issue is with an unmanaged/native library. We are using a NuGet package (Esri.ArcGISRuntime) that has some unmanaged dependencies. As seems to be the convention, these get copied to a runtimes/{rid}/native folder within our test project bin/Debug directory, e.g. runtimes/win-x64/native. As with all the other issues, tests that use this library run fine in Visual Studio but fail with the error "Could not load ArcGIS Runtime (RuntimeCoreNet100_15.dll) or one of its dependencies" when run with NUnit Console.

I can workaround the issue in a couple of ways:

    Copy the relevant files out of the runtimes/{rid}/native folder and put them directly in the bin/Debug directory.
    Copy the entire runtimes folder into the .nuget\packages\nunit.consolerunner\3.15.2\tools\agents\net6.0 folder and also copy the relevant items from our test project's .deps.json file into the nunit-agent.deps.json file.

I have found some explanation about how native assemblies are probed for (https://learn.microsoft.com/en-us/dotnet/core/dependency-loading/default-probing), and potential ways of configuring it (https://github.com/dotnet/cli/blob/master/Documentation/specs/runtime-configuration-file.md), however I haven't been able to figure out any better workarounds than the two listed above.

veleek:
Loading of unmanaged assemblies suffers from the same issues as loading managed assemblies in that the wrong deps.json is used by default.  This updates the TestAssemblyLoadContext to use the exact same pattern for loading unmanaged assemblies as managed assemblies.
1. Use the default assembly loading logic (I'm unsure if we actually need this since, in both managed and unmanaged, the base AssemblyLoadContext logic is a no-op and it's actually the VM that has some logic for loading assemblies).
2. Use an AssemblyDependencyResolver for the test assembly.
3. Check in the same folder as the test assembly (in case the dependencies are not fully specified in deps.json).

Resolves #1253

See #1318 for the original fix for managed assemblies.

### DIFF
--- a/src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyLoadContext.cs
+++ b/src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyLoadContext.cs
@@ -3,6 +3,7 @@
 #if NETCOREAPP3_1_OR_GREATER
 
 using System.Reflection;
+using System.Runtime.InteropServices;
 using System.Runtime.Loader;
 using System.IO;
 using System;
@@ -56,7 +57,7 @@ namespace NUnit.Engine.Internal
             if (loadedAssembly != null)
             {
                 log.Info("Assembly {0} ({1}) is loaded using the TestAssembliesResolver", name, GetAssemblyLocationInfo(loadedAssembly));
-                
+
                 return loadedAssembly;
             }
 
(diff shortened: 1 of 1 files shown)


veleek:
This was verified locally on a project that had the original issue and was working around it by providing a `AssemblyLoadContext.ResolvingUnmanagedDll` callback to do the resolution (the workaround mentioned in the issue).

OsirisTerje:
Awesome work!   

Any chance of getting a test in here too ? Or are there existing tests for this?

veleek:
I don't believe there are any existing tests for this, and I'm not sure the best way to go about testing this anyway.  We need some sort of shim native dll that we can reference (maybe using a [DllImport] or something like that) and then trick it into not being in the right places?  

Our existing code that ran into this bug was uing a NuGet package, that referenced another NuGet package, that copied some DLLs into some places which is how we ran into it.  But I'm not even positive how it referenced the native DLL.

veleek:
I didn't see any tests for the managed code version, so I was just following that pattern, but if we want a test I can see if I can make one.

OsirisTerje:
I'm not on my computer now, but I do think we have a test for it in the adapter with a native dll.  Could try to grad that.

veleek:
@OsirisTerje - Finally got back to take a look through this.  Is the test with a native dll that you're referring to the one in `mock-cpp-clr`?  If so, then I'm not sure that this covers this scenario.  This isn't a problem loading native _test assemblies_, it's a problem when the test assembly loads a native _dependency_ (e.g. a test that uses `[DllImport]` to reference a function in a native dependency).  

While I think I could probably figure something out, I'm a little hesitant to start pulling in extra arbitrary NuGet packages just to get one with a native dependency to verify this.  A _proper_ test would require a test assembly with a native dependency that gets deployed to a different folder than the test assembly (can be a sub-folder) which will cause the default assembly resolution logic to fail because it won't use the test assemblies deps.json for resolving the path to the native dependency.

I spent a few minutes this morning looking into hackier options like creating an `AssemblyLoadContext` that derives from `TestAssemblyLoadContext` and then overrides `LoadUnmanagedDllFromPath` to verify that it's called with the expected paths in various scenarios.  But because the logic that calls that depends on the behavior of `AssemblyDependencyResolver` to actually resolve the dependency path (and it's non-trival for us to inject something in there) I think that the best we could do easily would be a test that verifies that when everything else fails we eventually call `LoadUnmanagedDllFromPath` with the an assembly path using `_basePath`, which isn't really the interesting this this code is intended to solve.  

In the end, I think my opinion is that this code only ever gets called if the dependency resolution has already failed.  As such, we're unlikely to break any existing code paths so it's at least safe that way.  So I think it's probably safe to accept my manual test verification and have a thorough review of the code (but I'm obviously biased 😁).  Happy to entertain other opinions or ideas on a good way to test this.

OsirisTerje:
Thanks @veleek !  

This merges into the main, which is the 4.X version.  It may be a while before that is released, it is nowhere near being ready.  Can you do another PR with the same changes and raise a PR for the 3.15 series ?  (That is the version315 branch)

veleek:
Will do!

veleek:
Do I also need it in the version316 branch?  Or is there some automated process to rebase/cherry-pick stuff?

OsirisTerje:
You don't need to merge to version316, that branch will be abandoned, and the commits there will be moved to the 315 branch, except the ones that have caused crash issues in 316.

Issue #1374, opened 2023-12-22
Add .NET 8 build for nunit-agent


mlessmann:
We are currently migrating to .NET 8 and are having trouble executing NUnit 3.13 tests with the NUnit Console 3.16.3. There are some errors at test runtime that occur because the test is run on the .NET 7 build of nunit-agent. Specifically, the test depends on a particular version of a DLL that is only present in .NET 8.

Would it be possible to create a .NET 8 build of the nunit-agent?

I have already created branches from [main](https://github.com/mlessmann/nunit-console/tree/DotNet8Agent-3.15) and from the [3.16.3 tag](https://github.com/mlessmann/nunit-console/tree/DotNet8Agent-3.16) and changed code and project files. I haven't touched the pipelines.

ilCosmico:
After updating my UnitTesting project from .NET6/.NET7 to .NET8 I get the following errors. 

![image](https://github.com/nunit/nunit-console/assets/8678145/ce8fa260-9828-4457-a574-3549cddc88c7)

Is this the same issue as yours?

I'm using NUnit 3.13.3 and console runner 3.16.3 too.

OsirisTerje:
Can you please try to run version 3.15.5 instead.

mlessmann:
Using 3.15.5 leads to the same error messages:
```
Unable to load one or more of the requested types.
Could not load file or assembly 'System.Runtime, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.
```

ilCosmico:
I confirm I get the same `System.Runtime` error when using 3.15.5.

OsirisTerje:
Thanks @ilCosmico  and @mlessmann .

I am on travel now, and have small opportunities to do changes myself.  I can review, merge and get stuff out if you provide the code for this.

The engine should be good for .net 8, that was done some time ago, and works for the adapter.  So there is then the console app that remains.

@mlessmann It seems you have found the reasons, and also, providing an agent for .net 8 should be trivial.  But, need you to do the changes on the version315 branch.  

The process for the console/engine release is not trivial however, but having the code in there in a PR would certainly make it easier for us to get it out.

mlessmann:
I have created a PR that adds .NET 8 builds for the console and the agent. My .NET 8 tests run successfully with this version locally.
I have tried to adjust the cake scripts, but I'm unfamiliar with them and would appreciate feedback.

[...]

PR #1378, opened 2024-01-10
.NET 8 Console and Agent

mlessmann:
- Use .NET 8 SDK in global.json
- Add .NET 8 target to nunit-console and nunit-engine-core
- Add .NET 8 and 7 target to nunit-agent
- Selectively disable NuGet audit to suppress errors when building for .NET Core 2.1
- Adjusted cake scripts

See #1374

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,8 +9,18 @@ jobs:
   condition: not(or(startsWith(variables['Build.SourceBranchName'], 'azure-linux-'),startsWith(variables['Build.SourceBranchName'], 'azure-macOS-')))
   pool:
     vmImage: windows-2022
-  
+
   steps:
+
+  - task: UseDotNet@2
+    displayName: 'Install .NET 8.0'
+    inputs:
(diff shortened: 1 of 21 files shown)


mlessmann:
@dotnet-policy-service agree company="dSPACE GmbH"

OsirisTerje (approved):
Looks good to me.  I can't see anything that seems out of place.

CharliePoole:
@OsirisTerje @mlessmann 
I'm a bit confused, as I was about to add a .NET 8.0 agent for 3.18.

For some reason, this PR was merged into branch 3.15 and is also now in main, but I can't figure out how it got there.
I think this also applies to other work that @mlessmann did and I'd like to straighten it out. At a minimum, I'll create a "retroactive issue" for 3.18, so that the release notes show it as being added and @mlessmann is credited.

@mlessmann I think this also applies to some other commits you made, which I'm investigating. Can we set up a channel to communicate without using issue comments? My email is on my GH profile.

Issue #1178, opened 2022-03-26
Running tests with nunit3-console version 3.15.0 generates empty log files


ciconq:
Hello,

When running tests through nunit3-console version 3.15.0 for example the following command:

`nunit3-console.exe Tests.dll`

I am observing a generation of empty log files like internal-trace and nunit-agent after each execution. Which is adding those types of files with random numbers concatenated to them, resulting in a lot of empty log files.

I was able to disable the generation of the internal-trace log files with the `--trace=off` option but the nunit-agent empty log files keep getting generated after each execution.

I had a question about this on stackoverflow and was given the suggestion to add a bug report here which might help fixing this.

@CharliePoole answered that the version of the new engine version 3.15.0 has introduced a new internal feature allowing code to change the level of debugging dynamically which is not yet exposed to users. 
And the side effect is that empty log files are being created.
Also that there is a fix for this in the development code for version 4.0 which is not ported back to version 3 code.

The only fix for the moment is to go back to version 3.14.0 but I am using .Net 6.0 and can't really go back to version 3.14.0.

Would be really nice if it can be fixed in version 3 code. 

Thank you for your time and attention, really appreciate it.

P.S. @CharliePoole Sorry if I am not supposed to tag you but thought I wouldn't want to discredit that you were the one that answered my question on stackoverflow. :)

CharliePoole:
I put this in the 3.15.1 milestone with one other non-critical bug. There will be a 3.15.1 release only if a critical bug comes along to trigger it.

CharliePoole:
No longer failing in 3.15.1, so I'm closing. If it pops up again, we can reactivate the bug and take a closer look.

cw397:
We are still getting this issue in 3.15.2. I created a new solution in VS2022 with a single NUnit Test Project targeting .NET 6. When calling NUnit Console from the command line on the dll produced by that project I get the two empty log files mentioned above in the working directory of the console.

- NUnit: 3.13.3
- NUnit Console: 3.15.2

It isn't a big issue, but after a few runs they soon stack up!

TimothyDJewell:
Same report, but the "workaround" I've found is to run with `--trace=Off` as well as `--inprocess`.  It's not a real solution, but each command takes out the internal trace and nunit agent logs respectively.

CharliePoole:
OK... I see I missed something in the original report. Let me clarify what was "fixed" first:

The BUG...
[...]

PR #1394, opened 2024-04-02
Fix for issue #1178 based on fix in PR #1214 for version 3.17++

OsirisTerje:


src/NUnitConsole/nunit3-console/Program.cs
```diff
@@ -106,8 +106,12 @@ public static int Main(string[] args)
...
                         engine.WorkDirectory = Options.WorkDirectory;
 
-                    if (Options.InternalTraceLevel != null)
-                        engine.InternalTraceLevel = (InternalTraceLevel)Enum.Parse(typeof(InternalTraceLevel), Options.InternalTraceLevel);
+                    //if (Options.InternalTraceLevel != null)
+                    //    engine.InternalTraceLevel = (InternalTraceLevel)Enum.Parse(typeof(InternalTraceLevel), Options.InternalTraceLevel);
+                    // See PR 1114  https://github.com/nunit/nunit-console/pull/1214/files  
+                    engine.InternalTraceLevel = Options.InternalTraceLevel != null
+                        ? (InternalTraceLevel)Enum.Parse(typeof(InternalTraceLevel), Options.InternalTraceLevel)
+                        : InternalTraceLevel.Off;
```
manfred-brands, line 114: If we want the default to be `InternalTraceLevel.Off` why doesn't the `TestEngineActivator.CreateInstance` ensure that?
Now it explicitly requires switching off.
Would this not affect other places where an engine is created (e.g. the adapter)

In the engine the level is set as: `public InternalTraceLevel InternalTraceLevel { get; set; } = InternalTraceLevel.Default;`
Where `Default` < `Off`.

However in its `Initalize` method it only tests on `Off`.
Instead of `Default` being its own state, should we not have `Default = Off` or drop `Default` completely?

```
            if(InternalTraceLevel != InternalTraceLevel.Off && !InternalTrace.Initialized)
            {
                var logName = string.Format("InternalTrace.{0}.log", Process.GetCurrentProcess().Id);
                InternalTrace.Initialize(Path.Combine(WorkDirectory, logName), InternalTraceLevel);
            }
```

OsirisTerje: Ideally yes, but I am very shaky on everything here, so I don't dare doing too much now.  We have three sort of main branches, and they are very diverged.  So this is just part of it, and I try to do as little as possible to make them even more diverged. 

I am also very uncertain about what to do with the v4 (main) here, since it is not working, and is not finished, and I don't have enough information to know what is missing or not.

OsirisTerje:
@stevenaw That piece of code has been ported all over the different branches inexactly that way.  As said above (answer to Manfred), the branches are a mess, and I just want to keep the code changes "as is" for now.  
Once (hopefully) we get this cleaned up, stuff like this should go.

PR #1405, opened 2024-05-28
Make v4 build

CharliePoole:
I had already done a fair bit of work in main before @OsirisTerje suggested I work in the version3X branch, so here is a PR for that work.

Basically this does little more than getting the CI build to work. I dropped all .NET Core 2.1 builds and fixed a problem with the loading of NUnit projects. Also did a few minor changes to the build scripts.


CharliePoole:
@OsirisTerje @manfred-brands or anyone else
I'd definitely like a review before merging this as it's the basis for the next major release.

build.cake
```diff
@@ -677,13 +677,13 @@ Task("TestConsole")
...
     .IsDependentOn("TestNet20EngineCore")
-    .IsDependentOn("TestNetStandard20EngineCore")
+    //.IsDependentOn("TestNetStandard20EngineCore")
     .IsDependentOn("TestNetCore31EngineCore");
 
 Task("TestEngine")
     .Description("Builds and tests the engine assembly")
     .IsDependentOn("TestNetFxEngine")
-    .IsDependentOn("TestNetStandard20Engine")
+    //.IsDependentOn("TestNetStandard20Engine")
```
veleek, line 686: Before checking in I think we should just remove this outright unless we expect that this is something we'll want to add back in the future.

src/NUnitEngine/nunit.engine.core.tests/Services/ExtensionManagerTests.cs
```diff
@@ -153,7 +153,7 @@ public void SkipsGracefullyLoadingOtherFrameworkExtensionAssembly()
             var assemblyName = Path.GetFullPath("../net35/nunit.engine.core.tests.exe");
 #else
             // Attempt to load the .NET Core 2.1 version of the extensions from the .NET 3.5 tests
```
veleek, line 155: Update comment

veleek (commented):
Approved, but I don't have permission to actually approve.

CharliePoole:
> Approved, but I don't have permission to actually approve.

That's OK, I can approve it myself but I still wanted eyes on it.

build.cake
CharliePoole: Leaving as a reminder to discuss this. I don't see the need for a .NET Standard engine but I wanted confirmation from others. I'll add a  TODO comment to make that clear.

manfred-brands: A library can be written against the standard.
Actual testing has to be done on a real platform.
If we need different versions of the engine for the different platforms, then a netstandard engine has no benefit as it won't be used.

cake/build-settings.cake
```diff
@@ -68,16 +63,6 @@ public class BuildSettings
             "Run mock-assembly-x86.dll under .NET 4.6.2",
             $"src/TestData/mock-assembly-x86/bin/{Configuration}/net462/mock-assembly-x86.dll",
             MockAssemblyExpectedResult(1));
-        NetCore31X86Test = new PackageTest(
-            "NetCore31X86Test",
-            "Run mock-assembly-x86.dll under .NET Core 3.1",
-            $"src/TestData/mock-assembly-x86/bin/{Configuration}/netcoreapp3.1/mock-assembly-x86.dll",
-            MockAssemblyExpectedResult(1));
```
manfred-brands, line 75: You dropped a net core 3.1 test, was that deliberate?

cake/build-settings.cake
```diff
@@ -152,17 +136,10 @@ public class BuildSettings
         if (IsRunningOnWindows)
         {
             StandardRunnerTests.Add(Net60WindowsFormsTest);
-
-            if (!IsRunningOnAppVeyor)
-            {
-                StandardRunnerTests.Add(NetCore21X86Test);
-                StandardRunnerTests.Add(NetCore31X86Test);
```
manfred-brands, line 159: I see it is remove here, but why would a .net core test be limited to Windows?
The `Net462X86Test` is in the standard list.

global.json
manfred-brands, line 1: Not sure if I like this deleted.
It means that build will move to newer major versions when they come out.
New version might come with new rules that break the build.
Is this not a process we would like to control manually?

manfred-brands (approved):
I had a few questions, but refer to you having more knowledge about this than I have.

build.cake
CharliePoole: I'm pretty sure we won't want a .NET Standard engine in the future. Even if we abandoned NETFX for the console runner, I think we would want to use .NET Core, probably 8.0, for the engine.

In any case, dropping or keeping it is a policy question and needs a bit more discussion at some point.

global.json
CharliePoole: Deleted in error. Will fix.

src/NUnitEngine/nunit.engine.core.tests/Services/ExtensionManagerTests.cs
CharliePoole: Done

cake/build-settings.cake
CharliePoole: Currently, we only support X86 under Windows. AgentProcess.cs makes a check and has a //TODO to remove the limitation. I removed this because I couldn't run it on my own machine and forgot to come back to it. There should really be a check for dotnet x86 being installed in the script, just as there is in our code.

cake/build-settings.cake
CharliePoole: Unintentional... see following comment.

cake/build-settings.cake
CharliePoole: Added a property IsDotNetX86Installed and used it here. Property currently checks for Windows, which we can remove at some future point. I'll push and see if this works on AppVeyor as well. Thanks for pointing this out.

PR #1376, opened 2024-01-09
Integrate TestAssemblyLoadContext changes to version315

veleek:
As per discussion in #1370, this change integrates all of the updates to TestAssembyLoadContext (nee CustomAssemblyLoadContext) from the `main` to the `version315` branch.  Unfortunately, there have been some changes to some of the packaging bits and pieces in order to add in the extra library used by TestAssemblyResolver, so I had to get those changes all merged as well.

Here is the set of commits that touched `TestAssemblyLoadContext.cs` and `TestAssemblyResolver.cs` (from oldest to newest) and what I did to resolve them:

3f854b20e4ea7e42ba7ae0d7bf81af902a6bc8fd - Manual resolution needed to deal with packaging changes due to new dependency.
f11499a418d5b5bb7ecdd5c91b2686fff4e32f74 - Do we need this change?  I merged the modifications to TestAssemblyLoadContext but just ignored everything else related to project changes for extra tests.
485153dd0b414f71e2432aadc1028afce114e7cf - Clean
b01ad3e6b61ad48640a90d8dbabbcc1dbb14b578 - Clean
38d141eaa8cf71c1095aeb5050b9a5641c71452d - Clean
462d9d25da21d4fa9558f5ab9db8f2c3d1dbd4b8 - Clean

> Note: Merge commits c7566820da5a444659e61469fe21afd262ba445d, b92df46ec6a8ca62963c9d5057c3dbe1746820a7, b92df46ec6a8ca62963c9d5057c3dbe1746820a7, 8f472b0ad29404593f3084622dac88ba6e5c5191, and 887516ae976a4b1b2dcc7704fe110080f9ec1a4b were just ignored.

PR #1406, opened 2024-05-29
Fix path for Microsoft.Extensions.DependencyModel in nuspec

veleek:
As part of PR #1376, 3f854b20e4ea7e42ba7ae0d7bf81af902a6bc8fd was integrated into the `version3X` branch.  The `main` branch only includes net6.0 in the package so it copies all dependencies into the root folder and references them from there (including Microsoft.Extensions.DependencyModel.dll).

In `version3X` the package includes both net6.0 and net8.0 dependencies so the binaries are referenced from sub-folders, so when the change was merged, it was referencing `Microsoft.Extensions.DependencyModel.dll` when it should have been looking in `net6.0/Microsoft.Extensions.DependencyModel.dll`.

> This also includes two whitespace changes that were automatically fixed by VSCode when I saved the file.  They're harmless and match the style of the rest of the file, so I figured I'd just keep them.

### DIFF
--- a/nuget/runners/nunit.console-runner.netcore.nuspec
+++ b/nuget/runners/nunit.console-runner.netcore.nuspec
@@ -23,7 +23,7 @@
     <tags>nunit test testing tdd runner</tags>
     <copyright>Copyright (c) 2021 Charlie Poole, Rob Prouse</copyright>
     <packageTypes>
-        <packageType name="DotnetTool" />
+      <packageType name="DotnetTool" />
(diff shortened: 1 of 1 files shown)


CharliePoole:
@veleek This looks good to me but the change is also incorporated in my PR #1407 with a number of other changes needed to make the version3X head build. Could you review that PR? Also #1405, which applies to main.

veleek:
Closing as it's fixed elsewhere.

PR #1407, opened 2024-05-29
Make the version3X build correctly

CharliePoole:
As with PR #1405, which applies to branch main, this PR merely does what is needed so that the branch builds correctly in CI. As part of this, all .NET Core 2.1 builds have been removed.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -114,11 +114,6 @@ public void BuildSolution()
         .WithProperty("TargetFramework", "netstandard2.0")
         .WithProperty("PublishDir", BIN_DIR + "netstandard2.0"));
 
-    DisplayBanner("Publishing ENGINE TESTS Project for NETCOREAPP2.1");
-    MSBuild(ENGINE_TESTS_PROJECT, CreateMSBuildSettings("Publish")
-        .WithProperty("TargetFramework", "netcoreapp2.1")
-        .WithProperty("PublishDir", BIN_DIR + "netcoreapp2.1"));
-
     // TODO: May not be needed
     foreach (var framework in new[] { "netcoreapp3.1", "net5.0" })
     {
@@ -129,6 +124,7 @@ public void BuildSolution()
     }
 }
 
+// TODO: Test this on linux to see if changes are needed
 private void BuildEachProjectSeparately()
 {
     DotNetRestore(SOLUTION_FILE);
@@ -237,7 +233,7 @@ Task("TestNetStandard20EngineCore")
     .OnError(exception => { UnreportedErrors.Add(exception.Message); })
     .Does(() =>
     {
-        RunDotnetNUnitLiteTests(NETCORE_ENGINE_CORE_TESTS, "netcoreapp2.1");
+        RunDotnetNUnitLiteTests(NETCORE_ENGINE_CORE_TESTS, "netcoreapp3.1");
     });
 
 //////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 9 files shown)


CharliePoole:
@OsirisTerje @manfred-brands or anyone else
I'd definitely like a review before merging this. The version3X branch will be spawning future 3.X releases, so it's important to get it right.

build.cake
```diff
@@ -237,7 +233,7 @@ Task("TestNetStandard20EngineCore")
     .OnError(exception => { UnreportedErrors.Add(exception.Message); })
     .Does(() =>
     {
-        RunDotnetNUnitLiteTests(NETCORE_ENGINE_CORE_TESTS, "netcoreapp2.1");
+        RunDotnetNUnitLiteTests(NETCORE_ENGINE_CORE_TESTS, "netcoreapp3.1");
```
veleek, line 236: Update Task name and comments above.

build.cake
```diff
@@ -302,7 +298,7 @@ Task("TestNetStandard20Engine")
     .OnError(exception => { UnreportedErrors.Add(exception.Message); })
     .Does(() =>
     {
-        RunDotnetNUnitLiteTests(NETCORE_ENGINE_TESTS, "netcoreapp2.1");
+        RunDotnetNUnitLiteTests(NETCORE_ENGINE_TESTS, "netcoreapp3.1");
```
veleek, line 301: Ditto

src/NUnitEngine/nunit.engine.core.tests/Services/ExtensionManagerTests.cs
```diff
@@ -153,7 +153,7 @@ public void SkipsGracefullyLoadingOtherFrameworkExtensionAssembly()
             var assemblyName = Path.Combine(GetSiblingDirectory("net35"), "nunit.engine.core.tests.exe");
 #else
             // Attempt to load the .NET Core 2.1 version of the extensions from the .NET 3.5 tests
-            var assemblyName = Path.Combine(GetSiblingDirectory("netcoreapp2.1"), "nunit.engine.core.tests.dll");
+            var assemblyName = Path.Combine(GetSiblingDirectory("netcoreapp3.1"), "nunit.engine.core.tests.dll");
```
veleek, line 156: Ditto

veleek (commented):
Approved

build.cake
CharliePoole: Actually, that's a really good catch! The Task name is what it's supposed to do, but it's no longer doing it. When running tests under .NET Core 2.1, the SUT was picked up as .NET Standard, but when running under 3.1 it picks up the 3.1 build of the engine core. So the two tasks TestNetStandard20EngineCore and TestNetCore31EngineCore actually do the same thing now!!! I'll have to fix that by making sure the tests pull in the correct SUT.

build.cake
CharliePoole: Same as above... it's supposed to test the netstandard engine but isn't doing it.

src/NUnitEngine/nunit.engine.core.tests/Services/ExtensionManagerTests.cs
CharliePoole: This one is just a typo. I'll fix.

CharliePoole:
It turns out that the two commentsfrom @veleek about the engine and engine core .NET standard 2.0 tests is both important and non-trivial to fix if we continue to test using NUnitLite. We might need a few new projects for the .NET Standard stuff. I think I'll postpone this and create separate issues.

I'll continue to wait for any more reviews before merging, however.

build.cake
CharliePoole: Postponing this.

build.cake
CharliePoole: Postponing this.

PR #1408, opened 2024-06-01
Make v4 build

CharliePoole:
@OsirisTerje @manfred-brands @veleek
This is an entirely new PR. The first "Make v4 build" PR #1405 was a merge from my CharliePoole account. While we were reviewing it, @OsirisTerje added me as a maintainer and I ended up a bit confused with multiple local repos. So my changes based on earlier code reviews went into the NUnit organization account, using the same branch name. Sorry for any confusion. No need to repeat your earlier comments if they were answered satisfactorily on the earlier PR.

This PR's main change is that we don't try to run any dotnet core x86 tests without first checking that an x86 version of dotnet is installed. As with the first PR, I haven't changed lots of things we will probably want to change eventually. The goal is to get this building cleanly without adding any new problems.

I will now proceed to delete all my forks of NUnit so this doesn't happen again. :-)


cake/build-settings.cake
```diff
@@ -211,6 +194,7 @@ public class BuildSettings
 	public bool IsRunningOnUnix => _context.IsRunningOnUnix();
 	public bool IsRunningOnWindows => _context.IsRunningOnWindows();
 	public bool IsRunningOnAppVeyor => _buildSystem.AppVeyor.IsRunningOnAppVeyor;
+    public bool IsDotNetX86Installed => IsRunningOnWindows && System.IO.File.Exists(@"C:\ProgramFiles )x86)\dotnet\dotnet.exe");
```
manfred-brands, line 197:
```suggestion
    public bool IsDotNetX86Installed => IsRunningOnWindows && System.IO.File.Exists(@"C:\Program Files (x86)\dotnet\dotnet.exe");
```

NUnitConsole.sln
```diff
@@ -20,7 +20,6 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 		src\Directory.Build.props = src\Directory.Build.props
 		GitReleaseManager.yaml = GitReleaseManager.yaml
 		GitVersion.yml = GitVersion.yml
-		global.json = global.json
```
manfred-brands, line 23: Suggest to restore this line so `global.json` can be viewed/edited from the solution.

manfred-brands (changes requested):
Some typos in the path for dotnet X86

manfred-brands (approved):
Changes look good, not sure why the build fails.

CharliePoole:
That .NET Core X86 test is now trying to run and is failing on both my own machine and appveyor. Actually, it's good that it fails locally as well, so I can debug it. I'm going to comment out that test for now and merge this, since it fixes a number of other outstanding issues. I'll create a new issue just for those X86 builds.

PR #1421, opened 2024-06-12
Create platform-support.md

### LINKED ISSUE #1414: Define console / engine runtime platform support
Platform support from Microsoft changes rapidly, so we need to look at this periodically. I think it's time to do it again, since .NET 7.0 went out of support this month and 6.0 has six months left.

It might be helpful to have a statement that describes our support, to include

* What runtime are needed to run the console
* What runtimes we support for execution of tests.
* What agents we supply - this is different from the preceding point since, for example, a .NET 8.0 agent can run .NET 5.0 tests.
* If possible, how we will change the above in the future as Microsoft drops support and adds new runtimes.

NOTE: The above may differ between versions 3 and 4.

I'll start working on a draft document to include in the root

@OsirisTerje @manfred-brands @anyone Feel free to comment right away and/or after there is a doc to review

CharliePoole:
Fixes #1414

Note that I cancelled the builds, which make no sense in this case. Just using the PR as a way to review the document.

### DIFF
--- a/platform-support.md
+++ b/platform-support.md
@@ -29,17 +29,17 @@ official end of life. This is intended to support continued testing of legacy ap
 process of upgrade. However, agents for runtimes which have been declared a security risk may be removed immediately.
 
 Based on that policy and the planned end-of-life dates for runtimes, we expect to retire agents on or after the
-dates listed in the following table.
+dates listed in the following table. 
 
-| Runtime              | Agent Supported<br>At Least Until |
-| -------------------- | --------------- |
-| .NET Framework 2.0   |                 |
-| .NET Framework 4.6.2 |                 |
-| .NET Core 3.1        |                 |
-| .NET 5.0             | July, 2024      |
-| .NET 6.0             | May, 2025       |
-| .NET 7.0             | November, 2024  |
-| .NET 8.0             | May, 2027       |
+| Runtime              | Microsoft<br>End of Support | Agent Retirement | Notes |
+| -------------------- | --------------- | --------------------- | --- |
+| .NET Framework 2.0   | July, 2011      | July, 2024            | Will be removed in version 3.18.0
+| .NET Framework 4.6.2 | January, 2027   | after July, 2027      | May be upgraded to 4.8.1 before retirement date |
+| .NET Core 3.1        | December, 2022  | after December, 2024  | 
+| .NET 5.0             | May, 2022       | July, 2024            | Will be removed in version 3.18.0
+| .NET 6.0             | November, 2024  | after May, 2025       |
+| .NET 7.0             | May, 2024       | after November, 2024  |
+| .NET 8.0             | November, 2027  | after May, 2027       |
 
 **NOTES:**
 1. Some discussion is needed before we can determine when the .NET Framework agents will be removed and how they will be replaced.
--- a/platform-support.md
+++ b/platform-support.md
@@ -18,7 +18,6 @@ runtime available.
 We currently supply a fairly large selection of agents with the console runner:
 * .NET Framework 2.0
 * .NET Framework 4.6.2
-* .NET Core 2,1
 * ,NET Core 3.1
 * .NET 5.0
 * .NET 6.0
@@ -36,7 +35,6 @@ dates listed in the following table.
 | -------------------- | --------------- |
 | .NET Framework 2.0   | TBD             |
 | .NET Framework 4.6.2 | TBD             |
-| .NET Core 2.1        | July, 2024      |
 | .NET Core 3.1        | July, 2024      |
 | .NET 5.0             | July, 2024      |
 | .NET 6.0             | May, 2025       |
--- a/platform-support.md
+++ b/platform-support.md
@@ -1,32 +1,55 @@
 # Platform Support Lifecycle
 
 "Platform Support" for this project includes three somewhat different things:
-1. The target platforms for which tests may be written.
-2. The platforms under which those tests will run, which is equivalent to the list of agents we supply.
+1. Test assembly target platforms, that is, the target platforms for which tests may be written.
+2. The target platforms under which those tests will actually run, which is equivalent to the list of agents we provide.
 3. The minimum platforms required to execute the runner and engine themselves, without regard to the tests being run.
 
 ## Test Assembly Target Runtimes
 
-We will continue to support tests targeting runtimes which are out of support from Microsoft's perspective, so long
-as we are able to do so without significant additional effort or security risk. If no agent is available for a runtime,
-the tests will be run on the closest higher runtime available.
-
-> _We currently support execution of tests written to target any version of the .NET Framework >= 2.0 and any version
-> of .NET Core >= 3.1, including .NET 5.0 and higher._
+We currently support execution of tests written to target any version of the .NET Framework >= 2.0 and any version
+of .NET Core >= 3.1, including .NET 5.0 and higher. We will continue to support tests targeting runtimes which are
+out of support from Microsoft's perspective, so long as we are able to do so without significant additional effort
+and without security risk. If no agent is available for a runtime, the tests will be run on the closest higher 
+runtime available.
 
 ## Agents Provided
 
-We will continue to provide agents for any Microsoft runtime for at least six months after its end of life. This is 
-intended to support continued testing of legacy applications while users are in the process of upgrade. However, agents
-for runtimes which have been declared a security risk may be removed immediately.
+We currently supply a fairly large selection of agents with the console runner:
+* .NET Framework 2.0
+* .NET Framework 4.6.2
+* .NET Core 2,1
+* ,NET Core 3.1
+* .NET 5.0
+* .NET 6.0
+* .NET 7.0
+* .NET 8.0 (coming in version 3.18.0)
 
-> _We currently supply agents for .NET Famework 2.0 and 4.6.2. .NET Core 2,1 and 3.1 and .NET 5.0, 6.0 and 7.0. The .NET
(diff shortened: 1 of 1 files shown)


manfred-brands (approved):
Looks good

platform-support.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Agents Provided
+
+We will continue to provide agents for any Microsoft runtime for at least six months after its end of life. This is 
+intended to support continued testing of legacy applications while users are in the process of upgrade. However, agents
+for runtimes which have been declared a security risk may be removed immediately.
+
+> _We currently supply agents for .NET Famework 2.0 and 4.6.2. .NET Core 2,1 and 3.1 and .NET 5.0, 6.0 and 7.0. The .NET
+> Core 2.1 runner is being removed in version 3.18.0 and .NET 8.0 will probably be added before this document is finalized.
+> Once this document is approved, we will begin to phase out agents according to the above criteria.
```
mikkelbu, line 25: Note: That this is currently not in italics due to the missing `_` in the end

platform-support.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Required Runtimes
+
+The console runner and engine target .NET 4.6.2, so that runtime or greater is required to execute any tests at all
+without loss of features. In individual cases, it may be possible to run using a lesser version of .NET 4.x.
+
+> **Note:** Editorial comments in italics will be removed in the final version of this document
```
mikkelbu, line 32: What is the reason for removing the comments in the final version? I know that having this information in the document will require us to update it now and then, but my personal feeling is that the more explicit we can be the better

mikkelbu (approved):
Looks good. I've just a couple of comments

stevenaw (approved):
Changes look good, great to see you again @CharliePoole 

As a question for future changes. Do we see us eventually listing an explicit "support" matrix with "EOL" dates of specific runtimes, or is this intended to be more general to describe our high-level approach?

CharliePoole:
@stevenaw A matrix is a good idea. I'll have to think about how to express firm vs soft dates if we want to retain some flexibility. This may also relate to whether we have pluggable agents in v4.

CharliePoole:
Updated based on your comments and ready for final review.

manfred-brands (approved):
One small typo

CharliePoole:
@OsirisTerje Comments? (Particularly with regard to removing agents)

OsirisTerje:
Looking at download numbers for NUnit 4, which dropped some older platform version, but also had some other breaking changes that may have affected the numbers,: 

The last 6 weeks numbers are
 
| Version | Downloads | Breaking changes
|---|---|---1
| 3.11.0 | 0.3 mill | |
|3.12.0| 1.2 mill. | Drop framework 2.0
|3.13.2| 1.1 mill. | None |
|3.13.3| 3.8 mill.| Dropped .net core 2.1 |
|3.14.0| 1.3 mill.| None |
|4.0.1| 0.7 mill.| Platforms >= 4.6.2 and .net 6 , moved classic asserts to new namespace |
| 4.1.0 | 2.1 mill | None |

So dropping framework 2.0 seems to be a no-brainer. Dropping  net core 2.1 also doesnt seem to be any issue, only 2.6 mill downloads on earlier version.

Why the 3.13.3 didn't upgrade to 3.14.0 - don't understand, but can be they don't upgrade once included.

The 4.0.1 version was obviously an issue for people.  
But it picked up with 4.1, but there are still many more people using earlier versions and not upgrading. 

So, based on this, I do believe people react to not supporting older versions of the dotnet platform.  On the other side, it also looks like many people simply don't upgrade their packages.  

So if we want to get people to use the NUnit tools, then I am starting to believe we need to cater for those lagging behind wrt to what framework they're running on.  Just moving up may not be for everyone.

I don't think NUnit has enough leverage to make people upgrade their frameworks, they would then much more just not upgrade.  So perhaps we should be more careful, and keeping version for longer, even if the package gets bigger ?
Afaik no one has complained about the package size yet.

So setting fixed dates for when we retire them may be too much trying to predict the future. 

We do lack metrics however, and I am not at all sure about this.

CharliePoole:
I didn't want fixed dates, so I said we would maintain them at least for six months - at your suggestion.

But maybe the inclusion of a table makes the dates seem more fixed. Maybe we should remove the table or state the dates in a "softer" way.

What is currently blocking me is the question of dropping .NET Core 3.1 and .NET 5.0 agents. See #1427 and #1428.

OsirisTerje:
I believe the table is fine, but instead of "Retirement" it could be changed to something stating the "Supported at least until".  

Dropping net 5.0 I think is just fine.  People who upgraded that far, will go further.   
Based on the numbers I showed above, I am unsure about dropping net core 3.1.  Personally I would have dropped it, but when I see the NUnit 4 downloads, I get a bit vary about it.  Could perhaps try to find some general info on how much 3.1 is actually used.

CharliePoole:
I'll do that and cancel my PR for dropping 3.1. 

My own thoughts: people need a reason to upgrade, some feature they want to use. What would you see as the big motivating feature to cause people to want to use NUnit 4? What was the logic behind using .NET 6.0 rather than .NET Standard 2.0? I haven't been following it too closely, but my impression is that dropping platforms and classic asserts are the two main changes. That's OK as preparation for some other leap forward but is there a leap coming? FWIW, I mostly use 3.14.

OsirisTerje:
The main reason for dropping netstandard 2.0/1 and net core 3.1 is maintainability of the source code.   Max C# version for netstandard is C# 8, and for net core C# 10.  There are also numerous other changes in the framework.  If we were to support those version it would be even more #ifdefs around in the code, and much higher probability for errors.  Staying at C# 8 didnt seem like a good idea either.
Secondary, increasing package size, with more versions.

CharliePoole:
I'm aware of all those reasons - that's why I would drop 3.1 and 5.0 agents as well. And I'm not saying that the NUnit framework team made a wrong decision either. BUT... if you are going to measure your success by downloads, you have to have something to attract downloads and convenience of maintenance doesn't do that. That's why v2 of the engine and console were not released... Chris and I removed things and were in the process of adding attractive features before doing a release. Unfortunately, we never got there. :-(

I wasn't aware of the C# version limitation for net standard, however. That's unfortunate.

OsirisTerje:
It is enough if we can keep things going.  Easy maintenance means higher possibility for new inventions in the field.  This is a delicate balance.  If we focus too much of what can be done, and loose focus on keeping older stuff, that would be sad. If we do opposite, we will be stuck in just old code, and not able to invent. 
I think it is just about finding the right balance, at this moment in time.  :-)

CharliePoole:
@OsirisTerje So I made that change but we still have three items in the table without dates. That will look odd, I think. Also, I'm a bit unsure, based on your comments, if we are or are not removing .NET 5.0 now.

Separately from this document, I need to know whether to scrap #1427 and #1428.

CharliePoole:
Regarding the delicate balance... one thing that was going to help us with that was the planned addition of pluggable agents in V4. I couldn't see how to implement it while users were able to completely control how we used processes and application domains, which is what made it a V4 item. The agents become completely self-contained and for TestCentric, all I'm doing is archiving the ones that reach eol, but they are still available for users who want them.

OsirisTerje:
That sounds great!.

platform-support.md
```diff
@@ -0,0 +1,53 @@
...
+official end of life. This is intended to support continued testing of legacy applications while users are in the 
+process of upgrade. However, agents for runtimes which have been declared a security risk may be removed immediately.
+
+Based on that policy and the planned end-of-life dates for runtimes, we expect to retire agents on or after the
+dates listed in the following table.
+
+| Runtime              | Agent Supported<br>At Least Until |
+| -------------------- | --------------- |
+| .NET Framework 2.0   |                 |
+| .NET Framework 4.6.2 |                 |
```
OsirisTerje, line 37: Jan. 2027  (same as MS for now, since it says "least until".

platform-support.md
```diff
@@ -0,0 +1,53 @@
...
+| .NET Core 3.1        |                 |
```
OsirisTerje, line 38: This is a kind of promise, so hard to say.  MS ended Dec 2022, so, tend to say "Until further notice"

platform-support.md
```diff
@@ -0,0 +1,53 @@
...
+As a general policy, we will continue to provide agents for any Microsoft runtime for at least six months after its
+official end of life. This is intended to support continued testing of legacy applications while users are in the 
+process of upgrade. However, agents for runtimes which have been declared a security risk may be removed immediately.
+
+Based on that policy and the planned end-of-life dates for runtimes, we expect to retire agents on or after the
+dates listed in the following table.
+
+| Runtime              | Agent Supported<br>At Least Until |
+| -------------------- | --------------- |
+| .NET Framework 2.0   |                 |
```
OsirisTerje, line 36: Should this be supported at all anymore,  support ended in 2011.

OsirisTerje (changes requested):

CharliePoole:
@OsirisTerje We can keep or remove .NET 5.0 and .NET Core 3.1 independently and I don't think package size is an issue. Unfortunately, I got ahead of myself and co-mingled the removal of the two but I can redo that.

platform-support.md
CharliePoole: OK... I revised the table layout substantially and I hope I put in dates that suit. @OsirisTerje I used "after December 2024" rather than "until further notice" since it at least gives a hint that we will reconsider in some time frame. I think the use of "after" tells folks which dates are "soft" and which are "hard".

OsirisTerje (approved):

PR #1452, opened 2024-08-04
Switch cake to build with DotNetBuild instead of MSBuild

veleek:
# Switch to using DotNetBuild action instead of MSBuild.
- NoRestore is specified because we have an explicit step to restore already, this should speed up builds a bit.
- Removed unused parameters from BuildSettings.Initialize (set defaults on BuildSettings directly, they can be overridden if necessary.
- Move .WithProperty("Version"...) inside BuildSettings property because it's recreating the settiings object on each call anyway, so it's unnecessary to have the caller do it.

# Enable binary logs
Binary logs contain more information and are significantly smaller than corresponding text logs.  This is especially helpful for CI builds where the reduced output makes it easier to figure out what's going on when looking at live logs, but still enables detailed debugging after the fact.
- Verbosity is set to Minimial for Build
- Verbosity is set to Quiet for Restore

### DIFF
--- a/.github/workflows/NUnitConsoleAndEngine.CI.yml
+++ b/.github/workflows/NUnitConsoleAndEngine.CI.yml
@@ -1,118 +1,84 @@
 ﻿name: NUnitConsoleAndEngine.CI
 
 on:
-  push:
-    branches:
-      - main
-      - release
-      - version3x
   pull_request:
-  
+    branches-ignore:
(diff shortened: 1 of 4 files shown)


veleek:
Make this on nunit-console before realizing it was changed to use shared recipe.  See NUnit.Cake.Recipe#5 instead.

PR #1453, opened 2024-08-04
Improve Github Actions CI workflow

veleek:
- Switch to using a Matrix so we don't need to duplicate steps.
- Install .NET 6 SDK instead of 5 (this seemed like it was causing a build fix.
- Upload binary log as an artifact.
- Upload test results as an artifact.

### DIFF
--- a/.github/workflows/NUnitConsoleAndEngine.CI.yml
+++ b/.github/workflows/NUnitConsoleAndEngine.CI.yml
@@ -62,7 +62,7 @@ jobs:
         with:
           name: NUnitConsoleLogs
           # This path is defined in build-settings.cake
-          path: "build/*.binlog"
+          path: "build-results/*.binlog"
           # if-no-files-found: error
 
       # Wait with this until package errors (tests following packaging) are all fixed
--- a/.github/workflows/NUnitConsoleAndEngine.CI.yml
+++ b/.github/workflows/NUnitConsoleAndEngine.CI.yml
@@ -44,7 +44,6 @@ jobs:
         with:
           # global-json-file: global.json
           dotnet-version: |
-            2.1.x
             3.1.x
             6.0.x
             8.0.100
@@ -64,7 +63,7 @@ jobs:
           name: NUnitConsoleLogs
           # This path is defined in build-settings.cake
           path: "build/*.binlog"
-          if-no-files-found: error
+          # if-no-files-found: error
 
       # Wait with this until package errors (tests following packaging) are all fixed
       # - name: 📦 Package
--- a/.github/workflows/NUnitConsoleAndEngine.CI.yml
+++ b/.github/workflows/NUnitConsoleAndEngine.CI.yml
@@ -9,8 +9,8 @@ on:
       - "*.md"
 
 env:
-  DOTNET_NOLOGO: true                     # Disable the .NET logo
-  DOTNET_CLI_TELEMETRY_OPTOUT: true       # Disable sending .NET CLI telemetry  
+  DOTNET_NOLOGO: true # Disable the .NET logo
+  DOTNET_CLI_TELEMETRY_OPTOUT: true # Disable sending .NET CLI telemetry
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


.github/workflows/NUnitConsoleAndEngine.CI.yml
```diff
@@ -1,118 +1,84 @@
...
 
 on:
-  push:
-    branches:
-      - main
-      - release
-      - version3x
   pull_request:
-  
+    branches-ignore:
```
veleek, line 5: This is just to match the appveyor config, I can revert this if necessary (but the paths-ignore part can probably stay).

.github/workflows/NUnitConsoleAndEngine.CI.yml
```diff
@@ -1,118 +1,84 @@
...
+        # If you need to get more verbose logging, add the following to the dotnet-cake above:  --verbosity=diagnostic
+        run: dotnet cake --target=Test --test-run-name=${{ matrix.testRunName }} --configuration=Release
+
+      - name: 🪵 Upload build logs
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: NUnitConsoleLogs
+          # This path is defined in build-settings.cake
+          path: "build/*.binlog"
```
veleek, line 66: This file will not exist (and the step will error) until binlogs are enabled via nunit/NUnit.Cake.Recipe#5.

veleek:
Note that all three runs of this action fail at the moment.  MacOS and Linux fail due to some assembly reference issues (I think that nunit/NUnit.Cake.Recipe#5 will fix this), and Windows succeeds, but fail because there are no binlogs to upload yet.

.github/workflows/NUnitConsoleAndEngine.CI.yml
```diff
@@ -1,118 +1,84 @@
...
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: 🛠️ Setup .NET
+        uses: actions/setup-dotnet@v4
+        with:
+          # global-json-file: global.json
+          dotnet-version: |
+            2.1.x
```
CharliePoole, line 47: I thought I had gotten rid of all the .Net 2.1 stuff.

CharliePoole:
Let's wait till nunit/NUnit.Cake.Recipe#5 is merged and see how this runs.

.github/workflows/NUnitConsoleAndEngine.CI.yml
```diff
@@ -1,118 +1,84 @@
...
+    strategy:
+      matrix:
+        include:
+          - os: windows-latest
+            testRunName: Windows
+          - os: ubuntu-latest
+            testRunName: Linux
+          - os: macos-14
+            testRunName: Linux
+      fail-fast: false
```
CharliePoole, line 28: I'm thinking we may want one or more jobs to run after all the builds under the matrix. 

The matrix builds could do all compilation, testing and package creation and testing.

If all (or possibly selected) builds succeeded, a publish job could actually publish the results.

If all that passed, a Release job could perform a production release.

.github/workflows/NUnitConsoleAndEngine.CI.yml
veleek: I didn't try it without .NET 2.1 so I'll remove this and give it a shot.

.github/workflows/NUnitConsoleAndEngine.CI.yml
veleek: Yeah, I think there are a few options to do that.  Either have a portion of the job that only runs on selected branches to publish after merges, or have something that triggers a separate workflow to do the publish.

But either way, unless we go whole hog into Actions or make some major adjustments to the cake recipe it's not as nice because we won't be able to reuse the results from the build: e.g. if we run `dotnet cake --target=Package` and check if it succeeds and then run `dotnet cake --target=Publish` [or whatever it actually is], then cake will re-run the build/test/package steps again.  If everything was in Actions then we would run the package stuff, save the build artifacts, and then separately run the publish step which would just reference them.

veleek:
I have disabled the Mac and Ubuntu builds for now because they fail during compile (https://github.com/nunit/nunit-console/actions/runs/10303084006/job/28518240557) but the Windows build and test succeeds.  I'd like to get this checked in and then work and fixing the other bits after.

PR #1455, opened 2024-08-09
Probe assemblies with .exe extension when loading from the test assembly's folder in TestAssemblyLoadContext

k15tfu:
cc @CharliePoole


CharliePoole (approved):
Looks good. Thanks!

PR #1483, opened 2024-09-16
Reorganize assembly resolution; fix error loading WindowsBase.dll

### LINKED ISSUE #1466: System.TypeLoadException : Could not load type 'System.Windows.IWeakEventListener' from assembly 'WindowsBase'
This problem happens in nunit console 3.18.1 but not when I run in Resharper test runner. 

Here's my main code

```
using System;
using System.ComponentModel;
using System.Linq;
using System.Text;
using System.Threading.Tasks;
using System.Windows;
using System.Windows.Controls;

namespace Net80WindowsLib
{

    public class Person : INotifyPropertyChanged
    {
        private string _name;

        public string Name
        {
            get { return _name; }
            set
            {
                if (_name != value)
                {
                    _name = value;
                    OnPropertyChanged("Name");
                }
            }
        }

        public event PropertyChangedEventHandler PropertyChanged;

        protected virtual void OnPropertyChanged(string propertyName)
        {
            PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
        }
    }

    public class MyWeakEventListener : IWeakEventListener
    {
        public bool ReceiveWeakEvent(Type managerType, object sender, EventArgs e)
        {
            if (managerType == typeof(PropertyChangedEventManager))
            {
                OnPropertyChanged(sender, (PropertyChangedEventArgs)e);
                return true;
            }
            return false;
        }

        private void OnPropertyChanged(object sender, PropertyChangedEventArgs e)
        {
            if (e.PropertyName == "Name")
            {
                var person = sender as Person;
                Console.WriteLine("Person's name changed to: " + person?.Name);
            }
        }
    }

}

```

And here's my csproject

```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net8.0-windows</TargetFramework>
	<UseWPF>True</UseWPF>
  </PropertyGroup>

</Project>

```

here's my test code:

```
using System.ComponentModel;
using Net80WindowsLib;
using NUnit.Framework;
using NUnit.Framework.Legacy;

namespace Net80WindowsTest
{
    [TestFixture]
    public class Net80TestClass
    {
        [Test]
        public void RunTest()
        {
            var person = new Person();
            var listener = new MyWeakEventListener();

            // Use PropertyChangedEventManager to add a weak event listener
            PropertyChangedEventManager.AddListener(person, listener, "Name");

            person.Name = "Alice";  // Triggers the listener
            person.Name = "Bob";    // Triggers the listener again
            ClassicAssert.AreEqual("Bob", person.Name);
        }
    }
}

```

And the test csproject

```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net8.0-windows</TargetFramework>
	<UseWPF>True</UseWPF>
	<OutputType>Library</OutputType>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="NUnit" Version="4.2.1" />
	<PackageReference Include="Microsoft.Net.Test.Sdk" Version="17.10.0" />
	<PackageReference Include="NUnit.ConsoleRunner" Version="3.18.1" />
	<PackageReference Include="NUnit.Extension.VSProjectLoader" Version="3.9.0" />
  </ItemGroup>

  <ItemGroup>
    <ProjectReference Include="..\Net80WindowsLib\Net80WindowsLib.csproj" />
  </ItemGroup>

</Project>

```

And if I run the test in Resharper test runner, the test passes!

But if I run it in nunit console

```
set NunitConsole=C:\Users\USER\.nuget\packages\nunit.consolerunner\3.18.1\tools
call  "%NunitConsole%\nunit3-console.exe" "%slnFile%" --skipnontestassemblies --result="%testresultxml%"  --timeout=900000 > "%testresulttxt%"
```

And I will get the following error:

```
1) Error : Net80WindowsTest.Net80TestClass.RunTest
System.TypeLoadException : Could not load type 'System.Windows.IWeakEventListener' from assembly 'WindowsBase, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'.
   at Net80WindowsTest.Net80TestClass.RunTest()
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)

```

Seems like a bug but I don't rule out that it could be some config issue also. What could be the error?

CharliePoole:
This PR fixes #1466, which is caused by loading the wrong copy of `WindowsBase.dll`, the copy that's part of `Microsoft.NetCore.App` rather than the one from `Microsoft.WindowsDesktop.App`.

In order to resolve it, I did a great deal of refactoring in `TestAssemblyResolver`. We were using four main strategies to locate assemblies in a fixed order. I've now changed the structure as follows:

1. The strategies are now formal, using the Strategy pattern.
2. I added some heuristics to decide on the order to apply the strategies to each test assembly. For example, if the test assembly directly references either a Windows Forms or a WPF assembly, we check the `Microsoft.WindowsDesktop.App` directory first. 

I view this as a start. In the future, we may want to get self-diagnostic reporting of how often each strategy is used. We may  want to subdivide the strategies further or add new ones. We'll almost certainly want to look at efficiency.

So I'd like to get comments on this PR about the usefulness of the new structure for those future purposes in addition to the normal review of the bug fix. I've asked @veleek to do the primary review as he has worked in this area but anyone on the @nunit team is welcome to review as well.

### DIFF
--- a/src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyResolver.cs
+++ b/src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyResolver.cs
@@ -2,6 +2,9 @@
 
 #if NETCOREAPP3_1_OR_GREATER
 
+using Microsoft.Extensions.DependencyModel;
+using Microsoft.Extensions.DependencyModel.Resolution;
+using Microsoft.Win32;
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -9,9 +12,7 @@ using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Runtime.Loader;
-using Microsoft.Extensions.DependencyModel;
-using Microsoft.Extensions.DependencyModel.Resolution;
-using Microsoft.Win32;
+using TestCentric.Metadata;
 
 namespace NUnit.Engine.Internal
 {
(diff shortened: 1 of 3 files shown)


src/TestData/wpf-test/WpfTest.cs
```diff
@@ -1,26 +1,32 @@
...
+using System.Windows;
 using System.Windows.Controls;
 using NUnit.Framework;
 
 // Test which resolves issue #1203
 namespace Test1
 {
     [TestFixture]
-    public class WPFTest
+    public class WPFTest : IWeakEventListener
```
veleek, line 12: Is this just to force these tests to reference an assembly that causes it to be treated like a WPF app?

veleek (approved):

src/TestData/wpf-test/WpfTest.cs
veleek: NM, just read the original issue. :)

src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyResolver.cs
```diff
@@ -2,171 +2,251 @@
...
+                    return loadedAssembly;
+
+            log.Info("Cannot resolve assembly '{0}'", assemblyName);
+            return null;
         }
 
-        private Assembly OnResolving(AssemblyLoadContext context, AssemblyName name)
+        #region Nested ResolutionStrategy Classes
+
+        public abstract class ResolutionStrategy
```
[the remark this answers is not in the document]
CharliePoole: I started out with a common constuctor, since removed, and I imagined there would be some shared implementation as well. Turns out that the approaches have very little in common right now but I decided to keep the class because I plan to add collection of stats there.

From debugging, I know that the RuntimeLibraries approach originally found most things but the addition of the trusted platforms section changed that and now that finds most things but not the Windows, WPF or AspNetCore code. It would be great if we could eventually understand this better and simplify it.

PR #1484, opened 2024-09-17
Upgrade to use C# 11

### LINKED ISSUE #1479: Update C# language version used for console and engine
NUnit Framework uses C#11 while the engine and console use C# 7. That's too great a difference and is confusing for developers. Ideally we would also use C#11.

The first step in this change will be to determine whether to make this change in one jump or in stages. If possible, I'd like to upgrade the language version alone, without making immediate use of the new features in existing code. We will also need coding standards to indicate whether certain features are used in production code.

Given the above, it's likely that multiple PRs will be needed for this issue.

CharliePoole:
Fixes #1479 by upgrading the C# language level to 11 without making use of any new features yet.

### DIFF
--- a/src/NUnitConsole/nunit3-console.tests/nunit3-console.tests.csproj
+++ b/src/NUnitConsole/nunit3-console.tests/nunit3-console.tests.csproj
@@ -23,8 +23,6 @@
 
   <ItemGroup Condition="'$(TargetFramework)'!='net462'">
     <PackageReference Include="System.ComponentModel.TypeConverter" Version="4.3.0" />
-    <!--<PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.3.0" />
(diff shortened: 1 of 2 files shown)


src/NUnitConsole/nunit3-console.tests/nunit3-console.tests.csproj
```diff
@@ -23,8 +23,6 @@
 
   <ItemGroup Condition="'$(TargetFramework)'!='net462'">
     <PackageReference Include="System.ComponentModel.TypeConverter" Version="4.3.0" />
-    <!--<PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.3.0" />
```
manfred-brands, line 26: Not really related to language version.

manfred-brands (approved):
No real issues here.

PR #1495, opened 2024-10-05
Honor DOTNET_ROOT environment variables

### LINKED ISSUE #1381: Engine should honor the DOTNET_ROOT environment variable if set
Add a parameter to NUnit Console for .NET installation/binaries directory path that, when defined, will omit or extend `NUnit.Engine.Internal.TestAssemblyResolver.GetDotNetInstallDirectory()`.

The motivation behind this feature is to enable the NUnit Console to work in Windows environments where .NET is not installed but only .NET binaries are downloaded, e.g. during pipeline execution.

CharliePoole:
Fixes #1381

This PR also consolidates the code for locating the dotnet install directory in one place.

### DIFF
--- a/src/NUnitEngine/nunit.engine.core/DotNetHelper.cs
+++ b/src/NUnitEngine/nunit.engine.core/DotNetHelper.cs
@@ -1,0 +1,66 @@
+﻿// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+using Microsoft.Win32;
+using System;
+using System.IO;
+using System.Runtime.InteropServices;
+
+namespace NUnit.Engine
+{
+    public static class DotNet
+    {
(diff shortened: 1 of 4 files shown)


veleek:
@CharliePoole, what's the process for getting changes like this migrated over to v4?  I just noticed that this isn't in the most recent v4 tag.

CharliePoole:
@veleek It's listed as done in #1547.

I used these kinds of issues to document what needed to be ported but I didn't document what _didn't_ need porting, like bug fixes for features removed in V4. For now, I'm just trusting my past self's judgement. :-)

In this case, you'll find that there is a file DotNet.cs in nunit.common, which is identical to DotNetHelper.cs. So if there's any bug in V4, it's a new bug.

I created the V4 tag after the fact as a convenient way of tying all V4 stuff together but I may not have captured PRs for ported changes for which no separate V4 issue was created.

AFAIK the porting issue for 3.20 is the only one with incomplete work.

Issue #1510, opened 2024-11-06
exception occurred while invoking executor Unexpected token '(' at position 100 in selection expression.


livebe01:
I have the following test

[Test]
[TestCase(ThisEnum.Option1, "ABC")]
[TestCase(ThisEnum.Option2, "DEF")]
[TestCase(ThisEnum.Option3, "GHI")]
public void TestSomething(ThisEnum thisEnum, string expected)
{
// do something
}

I can run the tests individually, and as a group in Rider w/o issue. But, for some use cases, it'd be really useful to be able to run parameterized tests individually, from the CLI.

I'm developing on a Mac using Rider. My project is using "NUnit" Version="3.13.3" and "NUnit3TestAdapter" Version="4.3.1".

I run the following from the Mac terminal:
`dotnet test -- NUnit.Where="test==Some.Name.Space.SomeClass.TestSomething(EnumValue,\"ABC\")"`

and get this error:
```
Starting test execution, please wait...
A total of 1 test files matched the specified pattern.
An exception occurred while invoking executor 'executor://nunit3testexecutor/': Unexpected token '(' at position 100 in selection expression.
Stack trace:
   at NUnit.Engine.TestSelectionParser.Expect(Token[] valid)
   at NUnit.Engine.TestSelectionParser.Parse(String input)
   at NUnit.Engine.TestFilterBuilder.GetFilter()
   at NUnit.VisualStudio.TestAdapter.NUnitTestFilterBuilder.FilterByWhere(String where) in C:\repos\nunit\nunit3-vs-adapter\src\NUnitTestAdapter\NUnitTestFilterBuilder.cs:line 93
   at NUnit.VisualStudio.TestAdapter.NUnit3TestExecutor.RunTests(IEnumerable`1 sources, IRunContext runContext, IFrameworkHandle frameworkHandle) in C:\repos\nunit\nunit3-vs-adapter\src\NUnitTestAdapter\NUnit3TestExecutor.cs:line 129
   at Microsoft.VisualStudio.TestPlatform.Common.ExtensionDecorators.SerialTestRunDecorator.RunTests(IEnumerable`1 sources, IRunContext runContext, IFrameworkHandle frameworkHandle)
   at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Execution.RunTestsWithSources.InvokeExecutor(LazyExtension`2 executor, Tuple`2 executorUriExtensionTuple, RunContext runContext, IFrameworkHandle frameworkHandle)
[...]

PR #1513, opened 2024-11-09
Add a few tests for an issue under open investigation

stevenaw:
Relates to #1510

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/Services/TestSelectionParserTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Services/TestSelectionParserTests.cs
@@ -1,10 +1,8 @@
 ﻿// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
-using System;
-using System.Collections.Generic;
-using System.Text;
-using System.Xml;
 using NUnit.Framework;
+using System;
+using System.Xml;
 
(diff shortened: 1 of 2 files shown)


CharliePoole:
@stevenaw I'm back from my class and looking this PR over. Do I understand correctly that the first commit is what _should_ work, while the second simply changes the tests to reflect current behavior?

stevenaw:
Welcome back @CharliePoole 

Admittedly, my investigation here was partly self-exploration. The first commit was to try what had been attempted in the original reporter's repro, and then how the tokenizer interpreted it. The second commit updates it to be enclosed in commas after I noticed other places where you had recommended the same. This, of course, changed the tokenizer output and so I updated the test accordingly.

My own lack of knowledge about this part of the codebase prevents me from saying which of these two is correct, though the fact that the engine could parse the comma-enclosed value but the framework couldn't resolve it to a test had me was beginning to wonder if it were a framework issue.

I'd welcome your investigation here if you are able. If you did, please do feel free to close this PR or to build upon it depending on the direction it may go.

CharliePoole:
Thanks, that gives me the context I wanted. I'll adopt this branch and see what I can figure out. It really seems that we need tests to verify the XML content of each example as well as tests that exercise the framework using the same filters.

CharliePoole:
Replaced by PR #1519

PR #1518, opened 2024-11-27
Implement new API for finding standard extensions

### LINKED ISSUE #1504: Use a programmed search to locate extensions in place of default addins files.

CharliePoole:
Fixes #1504 Replaces PR #1506 Part of #488

@manfred-brands I'd appreciate your re-review based on the changes made since you reviewed the original PR (second commit)


src/NUnitEngine/nunit.engine.core/Services/ExtensionManager.cs
```diff
@@ -131,6 +135,24 @@ public void FindExtensions(string startDir)
...
         }
 
+        /// <summary>
+        /// Find and install standard extensions for a host assembly using
+        /// a built-in algorithm that searches in certain known locations.
+        /// </summary>
+        /// <param name="hostAssembly">An assembly that supports extensions.</param>
+        public void FindStandardExtensions(Assembly hostAssembly)
+        {
+            bool isChocolateyPackage = System.IO.File.Exists(Path.Combine(hostAssembly.Location, "VERIFICATION.txt"));
```
manfred-brands, line 145: Should this use `_fileSystem.GetDirectory`

src/NUnitEngine/nunit.engine.core/Services/ExtensionManager.cs
```diff
@@ -416,18 +459,29 @@ internal void FindExtensionsInAssembly(ExtensionAssembly assembly)
...
-                object versionArg = extensionAttr.GetNamedArgument("EngineVersion");
-                if (versionArg != null && new Version((string)versionArg) > ENGINE_VERSION)
-                    continue;
+                // TODO: This is a remnant of older code. In principle, this should be generalized
+                // to something like "HostVersion". However, this can safely remain until
+                // we separate ExtensionManager into its own assembly.
+                string versionArg = extensionAttr.GetNamedArgument("EngineVersion") as string;
+                if (versionArg != null)
+                {
+                    Assembly THIS_ASSEMBLY = Assembly.GetExecutingAssembly();
```
manfred-brands, line 475: This is constant for this run, but obtained for every attribute found. Could be set before searching anything.

src/NUnitEngine/nunit.engine.core/Services/ExtensionManager.cs
```diff
@@ -131,6 +135,24 @@ public void FindExtensions(string startDir)
...
+        /// </summary>
+        /// <param name="hostAssembly">An assembly that supports extensions.</param>
+        public void FindStandardExtensions(Assembly hostAssembly)
+        {
+            bool isChocolateyPackage = System.IO.File.Exists(Path.Combine(hostAssembly.Location, "VERIFICATION.txt"));
+            string[] extensionPatterns = isChocolateyPackage
+                ? new[] { "nunit-extension-*/**/tools/", "nunit-extension-*/**/tools/*/" }
+                : new[] { "NUnit.Extension.*/**/tools/", "NUnit.Extension.*/**/tools/*/" };
+
+            FindExtensionAssemblies(hostAssembly, extensionPatterns);
```
manfred-brands, line 150: Personally I prefer methods returning the set of found assemblies instead of setting a field. 
But the approach is pre-existing.

src/NUnitEngine/nunit.engine.core/Services/ExtensionManager.cs
```diff
@@ -233,14 +255,35 @@ private ExtensionPoint DeduceExtensionPointFromType(TypeReference typeRef)
...
+        /// <param name="initialDirectory">Path to the initial directory.</param>
+        private void FindExtensionAssemblies(IDirectory initialDirectory)
         {
-            // First check the directory itself
-            ProcessAddinsFiles(startDir, false);
+            // Since no patterns are provided, look for addins files
+            ProcessAddinsFiles(initialDirectory, false);
+        }
+
+        /// <summary>
```
manfred-brands, line 268: Should we not use `<inheritdoc />` to prevent risking interface and implementation documentation getting out of sync?

manfred-brands (approved):
@CharliePoole fundamentally code looks good.
A few small remarks/opinions.

src/NUnitEngine/nunit.engine.core/Services/ExtensionManager.cs
CharliePoole: I did that for the FindExtensions method, but in this case it seemed unnecessarily complicated. That is, I would convert a path to an IDirectory, get an IFile from that and then convert them back to a path. But since the calling assembly must always exist, there's always a real path to start with.

src/NUnitEngine/nunit.engine.core/Services/ExtensionManager.cs
CharliePoole: In my own code I have always avoided it but I know others use it. For me it's a question of compatibility. If somebody wants to use a different compiler or ide, do you know if will `<inhreitdoc>` always work? In any case, if we decided to always use it, then I'd do one big PR rather than changing things one at a time. What do you think?

CharliePoole:
@manfred-brands I implemented one of your three suggestions and merged so as to be able to work some issues that depend on this change. Feel free to comment further, particularly wrt `<inheritdoc>` which may need to be a separate refactoring issue.

src/NUnitEngine/nunit.engine.core/Services/ExtensionManager.cs
manfred-brands: The compiler does the copying to the XML file.
See [inheritdoc](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags#inheritdoc)

src/NUnitEngine/nunit.engine.core/Services/ExtensionManager.cs
manfred-brands: Would that argument not also cover `_startDir` which comes from the same host assembly?

CharliePoole: I see your point. But of course `startDir` is defined in a different method and reused multiple times, so it seems worth the trouble. Probably, it would be better to combine those two methods and just get startDir once. I'll add a TODO to my local copy to remind me to do that next time I'm in it.

src/NUnitEngine/nunit.engine.core/Services/ExtensionManager.cs
CharliePoole: The _Microsoft_ compilers do. But maybe I'm beating a dead horse. Currently, I can't find any viable alternatives to the MS compiler, such as we used to have a long time ago. And of course we're using it to build the engine anyway.

I don't like to combine unrelated changes in the same PR, so I'll create an issue.

PR #1522, opened 2024-12-07
Add support for additional extension directories

### LINKED ISSUE #488: Simplify locating of addins

CharliePoole:
Fixes #488 (Simplify Location of Addins)

Barring bugs that arise, I'm looking at this as the final version 3 feature aimed at the loading of extensions. As such, I'd appreciate one or more reviews that focus on the functionality (is it enough?) and the api itself (is it clear enough?) as well as the more detailed aspects of the code. 

I have updated [this comment](https://github.com/nunit/nunit-console/issues/488#issuecomment-2515504076) to reflect the actual implementation.


src/NUnitEngine/nunit.engine.core/Services/ExtensionManager.cs
```diff
@@ -118,13 +120,19 @@ public virtual void FindExtensionPoints(params Assembly[] targetAssemblies)
...
-            // Check each assembly to see if it contains extensions
-            foreach (var candidate in _assemblies)
-                FindExtensionsInAssembly(candidate);
+                // Create the list of possible extension assemblies,
+                // eliminating duplicates, start in the provided directory.
+                FindExtensionAssemblies(_fileSystem.GetDirectory(startDir));
+
+                // Check each assembly to see if it contains extensions
+                foreach (var candidate in _assemblies)
+                    FindExtensionsInAssembly(candidate);
```
manfred-brands, line 134: `_assemblies` is appended to in each call to `FindExtensionAssemblies`.
This means that one the 2nd call to `FindExtensions` for a 2nd directory, we will revisit the assemblies found in the 1st call again. This could mean that the same extension is loaded multiple times.
In addition the assemblies loaded by `FindStandardExtensions` will be visited again.

If `FindExtensionsAssemblies` would return only the list of new assemblies, only those would be searched.

Alternative split the finding of assemblies from the finding of extension in those assemblies.
Delete this method and make `FindExtensionAssemblies` public.
Then after all assemblies are found call a new method `FindExtensionsInAssemblies()` which iterates over `_assemblies` like here.
In addition remove the iteration from `FindStandardExtensions`.

manfred-brands (changes requested):
The same assemblies will be visited multiple times looking for extensions.
In my comments I suggested 2 alternatives for fixing this.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -48,28 +51,40 @@ public class ConsoleRunner
...
+            if (!string.IsNullOrEmpty(extensionPath))
+                foreach (string extensionDirectory in extensionPath.Split(new[] { Path.PathSeparator }, StringSplitOptions.RemoveEmptyEntries))
+                    _extensionService.FindExtensions(extensionDirectory);
 
-            // TODO: Exit with error if any of the services are not found
+            foreach (string extensionDirectory in _options.ExtensionDirectories)
+               _extensionService.FindExtensions(extensionDirectory);
+
+            _workDirectory = options.WorkDirectory ?? Directory.GetCurrentDirectory();
+            if (!Directory.Exists(_workDirectory))
```
stevenaw, line 76: nitpick: I'm unsure if it was a stylistic choice, but I think the `if` here isn't needed. I believe `Directory.CreateDirectory()` will instead do nothing if the directory already exists.

stevenaw (commented):
Thanks for including me as a reviewer here @CharliePoole . Looks great. I've left one nitpick as a comment.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
CharliePoole: Not actually new, but moved from elsewhere. I don't think it was needed even under .NET Framework 1.1.

src/NUnitConsole/nunit3-console/ConsoleOptions.cs
```diff
@@ -41,14 +41,19 @@ internal ConsoleOptions(IDefaultOptionsProvider defaultOptionsProvider, IFileSys
...
         public bool Explore { get; private set; }
 
         public bool ShowHelp { get; private set; }
 
         public bool ShowVersion { get; private set; }
 
+        public bool ListExtensions { get; private set; }
+
+        // Additional directories to be used to search for user extensions
+        public IList<string> ExtensionDirectories { get; } = new List<string>();
```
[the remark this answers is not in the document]
CharliePoole: I agree with you in general and I considered using a `List` here. But I saw that there were six other properties using `IList` and opted for consistency.  I usually don't mix global changes of any sort with actual new work, but it wouldn't be such a bad thing in this case.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -48,28 +51,40 @@ public class ConsoleRunner
...
             _filterService = _engine.Services.GetService<ITestFilterService>();
+            Guard.OperationValid(_filterService != null, "Internal Error: TestFilterService was not found");
+
             _extensionService = _engine.Services.GetService<IExtensionService>();
+            Guard.OperationValid(_extensionService != null, "Internal Error: ExtensionService was not found");
+
+            var extensionPath = Environment.GetEnvironmentVariable(NUNIT_EXTENSION_DIRECTORIES);
+            if (!string.IsNullOrEmpty(extensionPath))
+                foreach (string extensionDirectory in extensionPath.Split(new[] { Path.PathSeparator }, StringSplitOptions.RemoveEmptyEntries))
+                    _extensionService.FindExtensions(extensionDirectory);
```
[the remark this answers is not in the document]
CharliePoole: I dislike a single statement within braces. Personal style I guess. :-)

People used to talk about the risk that someone would add an extra indented statement, expecting it to be controlled by the `if`, but I've never seen it happen.

src/NUnitEngine/nunit.engine.core/Services/ExtensionManager.cs
CharliePoole: We keep a dictionary of all assembly paths which have already been seen so that we don't re-process them. This was needed before we changed to this three-stage process of finding extensions, because the user could reference the same directory or assembly in multiple `.addins` files. You can see this in the `ProcessDirectory` and `ProcessCandidateAssembly` methods.

Please look at that code to confirm that it works as intended. Meanwhile, I'll try to construct a test to prove it and prevent regression.

By the way, the new check on _extensionDirectories wasn't strictly necessary, but I wanted to have a clean, non-duplicated list of extension directories for use in this issue. In the end, I didn't make use of it as intended, but I have kept it for a future enhancement.

CharliePoole:
@manfred-brands @stevenaw Thanks for reviewing. I must say that GitHub's file compare leaves something to be desired when code has been moved around.

@manfred-brands I don't think there can be duplication of extensions but I think the code could make that clearer. Please review my comment and see what you think. Clearly there should be a test that demonstrates there's no duplication.

src/NUnitEngine/nunit.engine.core/Services/ExtensionManager.cs
manfred-brands: Indeed you don't visit the same directory twice (_existingDirectories), 
don't process the same directory multiple times (_visited) and 
don't add the same assembly multiple times (IsDuplicateOf),
However, unless I missed some other guard, I still think you load extensions multiple times.

Say we have two directories on the search lists: A and B, each of them has 1 assembly with 1 extension.
When calling FindExtensions(A) it adds A/Assembly to `_assemblies` and calls 'FindExtensionsInAssembly(A/Assemby)'
When calling FindExtensions(B) it adds B/Assembly to `_assemblies` and calls both FindExtensionsInAssembly(A/Assemby) and FindExtensionsInAssembly(A/Assemby)
There is no protection in FindExtensionsInAssembly.

If you have a setup with 2 extension, but them in separate directories and check the logs.
There is no test on `FindExtensions`, it would require creating extensions on disk and not the Fake ones.

CharliePoole: What we need to demonstrate is as follows.
1. There is only one place where an assembly can be added to _assemblies: `ProcessCandidateAssembly`.
2. That method could be called for a duplicate assembly if...
    2.1 The same asssembly path is specified in two .addins files
    2.2 The same assembly is found in different directories.
3. In either case, it should not add the duplicate to the list...
    3.1 The same path should be ruled out by `WasVistited`.
    3.2 The same AssemblyName should be ruled out by `IsDuplicateOf`.

At this point, I'm not convinced it's happening as it should, so I'll set up extensions in two directories and check the logs as you suggest. But really, all of the above needs tests. I think we can create unit tests without anything on disk because we aren't dealing with a list of assemblies here, but a list of `ExtensionAssembly` and that class can be faked.

Interesting problem. :-)

src/NUnitConsole/nunit3-console/ConsoleOptions.cs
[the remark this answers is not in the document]
CharliePoole: Changed all IList properties to List.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -310,6 +325,14 @@ private static string GetOSVersion()
...
+            {
+                _outWriter.WriteLine(ColorStyle.SectionHeader, "User Extension Directories");
+                foreach (var dir in _options.ExtensionDirectories)
+                    _outWriter.WriteLine($"  {Path.GetFullPath(dir)}");
+                _outWriter.WriteLine();
+            }
+
             _outWriter.WriteLine(ColorStyle.SectionHeader, "Installed Extensions");
 
             foreach (var ep in _extensionService?.ExtensionPoints ?? new IExtensionPoint[0])
```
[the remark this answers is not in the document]
CharliePoole: Although it's not part of the PR, I do find it surprising for a different reason... There are always extension points, although there may not be extensions! I believe this goes way back to when we first developed the extensibility infrastructure but had not added any extension points. At the present time this would be some kind of internal error, so I think I should just throw.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
CharliePoole: I changed it a bit but left the test because I really can't bear to call `Directory.CreateDirectory` on the current directory. :-)

src/NUnitEngine/nunit.engine.core/Services/ExtensionManager.cs
CharliePoole: I'm fairly convinced we are avoiding actual duplicates but we're definitely doing too much re-processing of the assembly list. I'm going to use your first suggestion of returning only the list of new assemblies. The second approach means exposing FindExtensionAssemblies, which should really be called FindPossibleExtensionAssemblies or something like that. I don't think that's a good public method.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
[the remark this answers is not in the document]
CharliePoole: I now throw in  the class constructor if there are no extension points.

CharliePoole:
The second commit includes the smaller changes requested in review as well as some refactoring. Additional logging is performed and the `IExtensionManager` interface is removed. The interface was no longer serving any purpose as `ExtensionManager` is now only used internally.

The next commit will deal with eliminating the possiblility of duplicate loading of extensions. The key issue is that all candidate assemblies must be identified before any of them are examined for extensions. This is necessary because the best version of a given assembly may end up replacing one that was found earlier.

Currently, this is not happening because the initial call to `FindStandardExtensions` is made by `ExtensionService` during it's initialization, while the `FindExtension` calls are made after initialization by the `ConsoleRunner`. This will need to be changed so that `ExtensionPoints` are identified in initialization but all calls to find extensions are made by the runner.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
stevenaw: That's a good point about calling it on the current directory! 😃 I hadn't considered that. Thanks for having updated this.

CharliePoole:
In the third commit, I have followed up on @manfred-brands comments about the potential for duplication. That was a good catch and working to fix it turned up an additional problem.

Before this change, the `ExtensionManager` located all potential extension assemblies _before_ searching through them for the extensions themselves. This is necessary because the same assembly may exist in different paths, possibly in different versions and targeting different runtimes, and we need to select the best version for our purpose before going any deeper. This was broken in the original code of this PR, with assemblies being found and scanned in distinct groups.

I looked at various ways to fix the problem and settled on the following:
1. FindExtensions methods are replaced with FindExtensionAssemblies.
2. Scanning the assemblies for extensions happens in the ExtensionManager as soon as some call requires it.

CharliePoole:
@manfred-brands @stevenaw 
I'm doing some cleanup and refactoring today for another commit, after which the PR will be ready for re-review.

I'd like to add both of you to the engine team but I wanted to ask first. That's handy even if you don't have time to write code, since GitHub will let me set up automatic assignment of reviews on a round-robin basis. So... would that be OK with you?

manfred-brands:
@CharliePoole It could be a way for me to get insight in the engine so we can give it a try.

CharliePoole:
The fourth commit does some refactoring and adds tests. A utility class is now used to track extension assemblies and maintain indexes to them so that it is no longer possible to add an assembly without also updating the indexes. A new internal constructor allows creation of `ExtensionAssembly` instances without an existing assembly file on disk and is used in testing that class. Some name changes have been made.

@manfred-brands @stevenaw This is now ready to be reviewed again

stevenaw:
> I'd like to add both of you to the engine team but I wanted to ask first

@CharliePoole Thanks for asking. I'd be happy to give it a try as well on the team.

src/NUnitEngine/nunit.engine.core/Services/ExtensionManager.cs
```diff
@@ -107,84 +132,129 @@ public virtual void FindExtensionPoints(params Assembly[] targetAssemblies)
...
         }
 
-        /// <inheritdoc/>
+        /// <summary>
+        /// Get an ExtensionPoint based on its unique identifying path.
+        /// </summary>
         public IExtensionPoint GetExtensionPoint(string path)
         {
-            return _pathIndex.ContainsKey(path) ? _pathIndex[path] : null;
+            return _extensionPointIndex.ContainsKey(path) ? _extensionPointIndex[path] : null;
```
manfred-brands, line 196: This is a doing a double lookup.
```suggestion
            _extensionPointIndex.TryGetValue(path, out ExtensionPoint ep);
            return ep;
```

src/NUnitEngine/nunit.engine.core/Services/ExtensionManager.cs
```diff
@@ -23,11 +24,24 @@ public class ExtensionManager : IExtensionManager
...
 
-        private readonly List<ExtensionNode> _extensions = new List<ExtensionNode>();
-        private readonly List<ExtensionAssembly> _assemblies = new List<ExtensionAssembly>();
+        // Index to ExtensionPoints based on the Path
+        private readonly Dictionary<string, ExtensionPoint> _extensionPointIndex = new Dictionary<string, ExtensionPoint>();
+
+        // List of ExtensionNodes for all extensions discovered.
+        private List<ExtensionNode> _extensions = new List<ExtensionNode>();
+
+        private bool _extensionsAreLoaded;
```
manfred-brands, line 36: The down side of initializing this here is that you don't get exceptions if you access it from a method not protected by `EnsureExtensionsAreLoaded`. But behave as if no extension are loaded.

Alternatively you could use `Lazy<T>` and then use `_extensions.Value`.

src/NUnitEngine/nunit.engine.core/Services/ExtensionManager.cs
```diff
@@ -363,32 +412,29 @@ private void ProcessAddinsFile(IFile addinsFile, bool fromWildCard)
...
                     return;
-
-                for (var i = 0; i < _assemblies.Count; i++)
+                
+                // Do we already have a copy of the same assembly at a diffrent path?
+                if (_assemblies.ByName.ContainsKey(assemblyName))
                 {
-                    var assembly = _assemblies[i];
+                    if (candidateAssembly.IsBetterVersionOf(_assemblies.ByName[assemblyName]))
+                        _assemblies.ByName[assemblyName] = candidateAssembly;
```
manfred-brands, line 432: Could use `TryGetValue` to prevent need for double lookup.
```csharp
                if (_assemblies.ByName.TryGetValue(assemblyName, out ExtensionAssembly existing))
                {
                    if (candidateAssembly.IsBetterVersionOf(existing))
```

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -48,28 +52,47 @@ public class ConsoleRunner
...
+            var extensionPath = Environment.GetEnvironmentVariable(NUNIT_EXTENSION_DIRECTORIES);
+            if (!string.IsNullOrEmpty(extensionPath))
+                foreach (string extensionDirectory in extensionPath.Split(new[] { Path.PathSeparator }, StringSplitOptions.RemoveEmptyEntries))
+                    _extensionService.FindExtensionAssemblies(extensionDirectory);
+
+            foreach (string extensionDirectory in _options.ExtensionDirectories)
+               _extensionService.FindExtensionAssemblies(extensionDirectory);
+
+            // Trigger lazy loading of extensions
+            var dummy = _extensionService.Extensions;
```
manfred-brands, line 78: Is this really needed? The service ensures these are loaded when needed.
If you want to keep it, I much prefer: `_extensionService.EnsureExtensionsAreLoaded()` over this assignment.

manfred-brands (changes requested):
Mostly fine. A few comments.

src/NUnitConsole/nunit3-console/ConsoleOptions.cs
```diff
@@ -99,13 +103,15 @@ public string WorkDirectory
...
         private readonly List<OutputSpecification> resultOutputSpecifications = new List<OutputSpecification>();
-        public IList<OutputSpecification> ResultOutputSpecifications
+        public List<OutputSpecification> ResultOutputSpecifications
         {
             get
             {
-                if (noresult)
-                    return new OutputSpecification[0];
+                //if (noresult)
+                //    return new List<OutputSpecification>();
```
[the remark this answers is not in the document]
CharliePoole: Nope!

src/NUnitEngine/nunit.engine.core/Services/ExtensionManager.cs
CharliePoole: Good catch! I preferred to keep the null test for clarity even though it's not needed. Of course, clarity is in the eye of the beholder. :-)
```
return _extensionPointIndex.TryGetValue(path, out ExtensionPoint ep) ? ep : null;
``

src/NUnitEngine/nunit.engine.core/Services/ExtensionManager.cs
CharliePoole: Yes, it's a tradeoff. I ran tests for several days and I think all methods are now protected. If an error comes up in future due to this, I think the cause should be fairly clear. I'll leave this at least for now.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
CharliePoole: Left over from earlier testing, before `EnsureExtensionsAreLoaded` existed.

src/NUnitEngine/nunit.engine.core/Services/ExtensionManager.cs
CharliePoole: Nice. I really ought to wrap that logic in the ExtensionAssemblyTracker class. Later though. :-)

manfred-brands (approved):
Thanks @CharliePoole, nothing further.

Issue #1539, opened 2024-12-15
V4 Codebase Should be Null-Enabled


CharliePoole:
I'm treating this as a prerequisite for re-booting work on version 4. I observed that null-enabling the framework code brought up many issues, so it may be desirable to do this one assembly at a time, in separate PRs.

This depends on updating the language level, which I'm doing in a separate issue.

manfred-brands:
@CharliePoole Converting existing code to nullable is not simple if done right.

It heavily depends on the code at hand.
If a class has various fields that are set in certain  methods and used in others it could result in a lot of tests on null.

Removing the need for this might mean splitting functionality in separate classes.

How far do you want to take this?

I can have a look and do an initial investigation.

What project do you think I should start with?

mikkelbu:
I was considering starting with determining how the different project relates and then start with the "bottom" (the project that only has no/external dependencies).

I think we should just start with the quick solutions - e.g. marking certain areas/methods as ignored via `nullable disable` and similar and then improve on this later on. Otherwise, I fear that this could take a very long time.

I had originally considered volunteering for giving a hand on this, but I lack the time at the moment (or at least proper time chunks and not just 15-25 minute blocks now and then).

CharliePoole:
@manfred-brands Thanks! I'm glad to have you working on this. I'll help by asking naive questions. :-)

Seriously, I've experimented only a little with nullable types, so I prefer not to do this myself in the engine. One bit of general guidance: I suggest _not_ trying to get each project "done" all at once. If there are particularly troublesome areas, @mikkelbu 's suggestion of ignoring them for a time makes sense to me. The only risk is that we forget to return and fix it, so let's not do that... we can leave behind some cleanup issues as we go or scan the entire repo and make a list at the end.

The dependencies will change as development continues, so I'll show both the current ones and those planned.

### Current Dependencies

```
nunit4-console
  => nunit.engine.api
  => nunit.engine (1)

nunit.engine
  => nunit.engine.api
  => nunit.engine.core  =>  nunit.engine.api

nunit-agents (2)
  => nunit.engine.api
  => nunit.engine.core
```

### Planned Dependencies

```
nunit4-console
  => nunit.engine.api

nunit.engine
  => nunit.engine.api
  => nunit.extensibility (3)

pluggable agents
  => nunit.engine.api
  => nunit.agent.core (4)
  => nunit.extensibility (3)
```

### Notes
[...]

PR #1579, opened 2024-12-30
WIP: Version4 nullable

manfred-brands:
Progress towards #1539 

nunit.engine.api
nunit.engine.core
nunit.engine.core.tests


manfred-brands:
Wrong target branchh.

PR #1583, opened 2024-12-31
Remove old scripts; drop zip package; update docs

### LINKED ISSUE #1581: Remove unused scripts and targets and document.
The `cake` directory contains scripts that were previously included by `build.cake` and used to build the runner. They are no longer used and should be removed. In addition, `build.cake` includes extra targets, which depend on those scripts but no longer work. All building is now done using `NUnit.Cake.Recipe`.

Additionally, although the script help target explains how to use the recipe, it would be helpful to have visible information somewhere that tells the user how to display the help information.

### LINKED ISSUE #1582: Eliminate zip package in version 4
### Discussed in https://github.com/nunit/nunit-console/discussions/1567

<div type='discussions-op-text'>

<sup>Originally posted by **CharliePoole** December 27, 2024</sup>
@nunit/engine-team The zip package doesn't seem to contribute much and it seems that those who use it could use other packages just as easily. I'll lay out my reasoning here and I'd appreciate your feedback.

Those who use the package typically do so because they want to create a centralized installation on their systems but either don't want to use Chocolatey or are unable to do so on their platforms.

Problems with this package...
1. It's basically a dump of everything we build - even our tests - so that anyone using it has to pick the correct directories.
2. It adds about 25% to the package test time for our builds. I suspect it adds even more to the "thinking time" required to maintain our packaging system, although this is purely anecdotal.
3. A global installation can easily be created from either of the console runner nuget packages, by use of appropriate command-line options. Such an installation would then contain exactly what is needed to execute the runner. Alternatively, of course, a user unfamiliar with the nuget command-line could simpy rename the `.nupkg` file to `.zip`.

If folks see reasons to keep the zip, then I would prefer to tailor it to contain _only_ what is needed to execute the console runners.</div>

CharliePoole:
Fixes #1581 Fixes #1582


manfred-brands:
@CharliePoole FYI: You accidentally push a WIP commit for the changes to `TestEngineActivator` with this PR.

CharliePoole:
Well spotted. :-) I saw it too late to squash it. Although it's only a dev build, the myget releases are releses and I prefer not to overwrite them, even though myget allows it.

PR #1588, opened 2025-01-02
Ensure Loggers use late bound `TraceLevel` and `Writer`.

### LINKED ISSUE #1586: Loggers don't use latest writer
The `InternalTrace` class uses a `static` _writer_ which is set in a call to `Initialize`.
Any caller that gets a `ILogger` before that call to `Initialize` is made gets a `null` _writer_.

`InternalTrace.Initialize` is called from `TestEngine.Initialize`.
Before a user can get there, the initializer of `TestEngine` has called `new ServiceContext()` which calls `new ServiceManager()` which calls `InternalTrace.GetLogger()`. This logger is called with the `InternalTraceLevel.Default` meaning it will never log anything.

If the user added any services before that call to `TestEngine.Initialize` none of them would every log either.

To solve this, the `InternalTrace.GetLogger` calls should not take the _writer_ and `DefaultTraceLevel` when called, but the ones that are active.

manfred-brands:
Fixes #1586

### DIFF
--- a/src/NUnitEngine/nunit.engine.core.tests/Internal/Logging/LoggingTests.cs
+++ b/src/NUnitEngine/nunit.engine.core.tests/Internal/Logging/LoggingTests.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
 using System.IO;
-using NUnit.Engine.Internal;
 using NUnit.Framework;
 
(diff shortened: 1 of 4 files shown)


CharliePoole (approved):
Well that's a pretty clever fix! 

I think the compiler will handle this pretty efficiently, but if not we can deal with the  problem then.

CharliePoole:
I see you were caught by the default branch merge. It still happens to me as well. :-)

PR #1597, opened 2025-01-05
Combine two NUnit3 Drivers

### LINKED ISSUE #1596: Combine nunit3 drivers and switch to new interface
The `NUnit3FrameworkDriver` is still using the original framework API we created for NUnit3. Back in 2018, @rprouse created an improved API for use with the .NET Core targets but we have continued to use the older API for .NET Framework targets. 

Until recently, we were supporting three different NUnit3 drivers. The removal of the .NET Standard driver, which has not been used for a while, brought us to two.

In this issue, I'm proposing to combine those two, `NUnit3FrameworkDriver` and `NUnitNetCore31Driver` into a single driver combining the  code for both .NET Framework and .NET Core and allowing the driver to decide which API to use depending on the target runtime and NUnit framework version in use. This can give a number of benefits.

1. A cleaner API for NUnit 4 and newer builds of NUnit 3.
2. Simpler code in the driver factory
3. Fewer tests to maintain
4. In my opinion, code that makes it easier to understand the differences between the two APIs and how we use them.

Tasks

- [x] Refactor the two drivers so they "line up" cleanly for comparison.
- [x] Create an initial build of the new driver, using conditional code, which is functionally equivalent to what we have now, and switch over to using it for all targets.
- [ ] Switch dynamically to use the new API for NUnit versions that support it.
- [x] Remove the two old drivers.

CharliePoole:
Fixes #1596


CharliePoole:
This is still in progress, but comments are helpful anyway, especially about direction I'm taking.

The first two commits result in a combined driver using conditional compilation. The driver has the working name of `NUnitCombinedFrameworkDriver` but I hope to find a better one, possibly just `NUnitFrameworkDriver` since it works for both v3 and v4.

Currently the conditional code is scattered around the file, but that will change in the next commit.

CharliePoole:
The third commit wraps the two APIs in individual classes and implements a compile-time strategy pattern using API2009 for .NET Framework and API2018 for .NET Core.

CharliePoole:
Fourth commit does some refactoring and removes the two classes being replaced. Tne new driver is called `NUnitFrameworkDriver`, eliminating the '3' in the name.

At this point, the choice of interface is still determined at compile time. I've tried two different approaches for making the decision at runtime and neither worked. See #1596 for discussion of problems.

CharliePoole:
The seventh commit finally reaches the goal of running .NET Framework tests using the 2018 API. It's a pretty big change, doubling the number of files affected and, unfortunately, making a number of unrelated changes that were needed.

A key problem was that a newly created driver didn't have an ID set in it's constructor. Rather, it was set later after construction - usually immediately after. There were various alternative ways to handle this, all a bit disruptive. I chose to modify the constructor for the driver as well as the `IDriverFactory` interface. This allows the constructor to complete initialization of the ID, which is a must-not-be-null property.

Remaining to do:
* Parameterize the driver tests so that both APIs are tested under .NET Framework
* Select the exact version cutoff for using each API. My first guess would be 3.10+ for Api2018, but I'll try a range of them.

@manfred-brands This is still WIP but please take a look. Let me know if you'd rather see this as several PRs. Now that I know it works, I can easily create a new PR for the first six commits and make the rest of it a second PR.

CharliePoole:
The eighth commit refactors the tests in order to ensure that both APIs we support under .NET Framework are tested. An internal constructor is added to `NUnitFrameworkDriver` to support the tests. 

The driver now switches between the 2009 and the 2018 APIs based on the version of NUnit in use. I was happily surprised to discover that the new API works for versions 3.2 and higher. Very rudimentary package tests have been added to ensure that the switching takes place correctly, but for most cases we now use the newer API. If problems arise, it would be possible to add some sort of switch (e.g. an environment variable) to force use of the older API. I don't see much point in exposing it as a command-line option.

This is now ready for review. @manfred-brands See my earlier comment about splitting it up if it seems desirable.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2018.cs
```diff
@@ -0,0 +1,293 @@
...
+        const string INVALID_FRAMEWORK_MESSAGE = "Running tests against this version of the framework using this driver is not supported. Please update NUnit.Framework to the latest version.";
+        const string FAILED_TO_LOAD_ASSEMBLY = "Failed to load assembly ";
+        const string FAILED_TO_LOAD_NUNIT = "Failed to load the NUnit Framework in the test assembly";
+
+        const string CONTROLLER_TYPE = "NUnit.Framework.Api.FrameworkController";
+
+        private string _driverId;
+
+        AssemblyName _nunitRef;
+        string? _testAssemblyPath;
```
manfred-brands, line 40: Please mark constructor set fields as `readonly` and assign `scope`.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2018.cs
```diff
@@ -0,0 +1,293 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Reflection;
+using System.Runtime.Remoting;
+using System.Runtime.Serialization;
+using NUnit.Common;
+using NUnit.Engine.Extensibility;
+using NUnit.Engine.Internal;
+using TestCentric.Metadata;
```
manfred-brands, line 12: Some of the using are not needed.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2018.cs
```diff
@@ -0,0 +1,293 @@
...
+#else
+            _nunitRef = nunitRef;
+#endif
+        }
+
+#if NETCOREAPP
+        TestAssemblyLoadContext? _assemblyLoadContext;
+        Assembly? _testAssembly;
+        Assembly? _frameworkAssembly;
+#endif
```
manfred-brands, line 71: Suggest to move all fields declarations at the top.

manfred-brands (changes requested):
I didn't get through it completely. Will continue this afternoon.
The global change looks great, there are some knitpicking things.

src/TestData/NUnit3.10/NUnit3.10.csproj
```diff
@@ -0,0 +1,13 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <RootNamespace>NUnit.TestData</RootNamespace>
+    <TargetFramework>net462</TargetFramework>
+    <OutputPath>../../../bin/$(Configuration)/testdata/nunit3.0</OutputPath>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="NUnit" Version="3.0.0" />
```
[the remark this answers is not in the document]
CharliePoole: Huh. I know I had these correct locally but somehow the change didn't get pushed.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        /// <summary>
+        /// Executes the tests in an assembly.
+        /// </summary>
+        /// <param name="listener">An ITestEventHandler that receives progress notices</param>
+        /// <param name="filter">A XML string representing the filter that controls which tests are executed</param>
+        /// <returns>An Xml string representing the result</returns>
+        string Run(ITestEventListener? listener, string filter);
+
+        void RunAsync(Action<string> callback, string filter);
```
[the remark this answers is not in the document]
CharliePoole: Indeed.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2009.cs
```diff
@@ -0,0 +1,157 @@
...
+            const string INVALID_FRAMEWORK_MESSAGE = "Running tests against this version of the framework using this driver is not supported. Please update NUnit.Framework to the latest version.";
+            const string FAILED_TO_LOAD_ASSEMBLY = "Failed to load assembly ";
+            const string FAILED_TO_LOAD_NUNIT = "Failed to load the NUnit Framework in the test assembly";
+
+            const string CONTROLLER_TYPE = "NUnit.Framework.Api.FrameworkController";
+
+            string _driverId;
+
+            AppDomain _testDomain;
+            AssemblyName _nunitRef;
```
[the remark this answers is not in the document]
CharliePoole: We need that analyzer. :-)

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2018.cs
```diff
@@ -0,0 +1,293 @@
...
+            var idPrefix = string.IsNullOrEmpty(_driverId) ? "" : _driverId + "-";
+
+#if NETFRAMEWORK
+            // Normally, the caller should check for an invalid requested runtime, but we make sure here
+            var requestedRuntime = settings.ContainsKey(EnginePackageSettings.RequestedRuntimeFramework)
+                ? settings[EnginePackageSettings.RequestedRuntimeFramework] : null;
+
+            try
+            {
+                _frameworkController = AppDomain.CurrentDomain.CreateInstanceAndUnwrap(
```
manfred-brands, line 90: As the Api itself is loaded in the _testdomain_, is there now any reason to now not to use the same code as for the .NET Core version?

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2018.cs
```diff
@@ -0,0 +1,293 @@
...
+
+            _frameworkControllerType = _frameworkController?.GetType();
+            log.Debug($"Created FrameworkController {_frameworkControllerType?.Name}");
+
+            var controllerAssembly = _frameworkControllerType?.Assembly?.GetName();
+            log.Debug($"Controller assembly is {controllerAssembly}");
+#else
+            _assemblyLoadContext = new TestAssemblyLoadContext(testAssemblyPath);
+
+            _testAssembly = LoadAssembly(testAssemblyPath);
```
manfred-brands, line 122: The framework version doesn't seem to load the _testAssembly_. 
I don't think this one needs to either.
It is loaded by the Framework Controller when you call the _LOAD_METHOD_ at the end of this function,

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2018.cs
```diff
@@ -0,0 +1,293 @@
...
+            Guard.ArgumentNotNull(settings, nameof(settings));
+            Guard.ArgumentValid(File.Exists(testAssemblyPath), "Framework driver called with a file name that doesn't exist.", "testAssemblyPath");
+            log.Info($"Loading {testAssemblyPath} - see separate log file");
+
+            _testAssemblyPath = Path.GetFullPath(testAssemblyPath);
+            var idPrefix = string.IsNullOrEmpty(_driverId) ? "" : _driverId + "-";
+
+#if NETFRAMEWORK
+            // Normally, the caller should check for an invalid requested runtime, but we make sure here
+            var requestedRuntime = settings.ContainsKey(EnginePackageSettings.RequestedRuntimeFramework)
```
manfred-brands, line 85: _requestedRuntime_ is not used here. I would think it is too late then, it should be selected to before to run this function.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2018.cs
```diff
@@ -0,0 +1,293 @@
...
+        public void RunAsync(Action<string> callback, string filter)
+        {
+            CheckLoadWasCalled();
+            log.Info("Running {0} - see separate log file", Path.GetFileName(_testAssemblyPath.ShouldNotBeNull()));
+            ExecuteMethod(RUN_ASYNC_METHOD, new[] { typeof(Action<string>), typeof(string) }, callback, filter);
+        }
+
+        public void StopRun(bool force)
+        {
+            ExecuteMethod(STOP_RUN_METHOD, force);
```
manfred-brands, line 164: No call to `CheckLoadWasCalled`. Preventable by moving the check to the `Execute` method and deleting it from all those places that call it just before calling `Execute`

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2018.cs
```diff
@@ -0,0 +1,293 @@
...
+            //    {
+            //        return _testDomain.CreateInstance(
+            //            _nunitRef.FullName, typeName, false, 0, null, args, null, null)!;
+            //    }
+            //    catch (TargetInvocationException ex)
+            //    {
+            //        throw new NUnitEngineException("The NUnit 3 driver encountered an error while executing reflected code.", ex.InnerException);
+            //    }
+            //}
+#else
```
manfred-brands, line 193: Not needed if we can harmonize the calls as outlined above.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2018.cs
manfred-brands: Ok, I see that the .NET Core version uses `AssemblyLoadContext` and .NET Framework would use `AppDomain`.
Note sure if either of them here have any effect.
It is used here to load the `FrameworkController`, but that controller directly calls `Assembly.Load` itself without doing any `AssemblyLoadContext` or `AppDomain` distinction.

If there is an effect, it could be abstracted in a `TestAssemblyLoader` class to hide the distinctions between the .NET Core and Framework.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2009.cs
```diff
@@ -0,0 +1,157 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Reflection;
+using System.Runtime.Serialization;
+using NUnit.Common;
+using NUnit.Engine.Internal;
+
+namespace NUnit.Engine.Drivers
+{
+#if NETFRAMEWORK
```
manfred-brands, line 13: Suggest to move the conditional to the top of the file, excluding the whole file.
Now for non-framework builds the `using` statements are marked as unnecessary.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2009.cs
```diff
@@ -0,0 +1,157 @@
...
+            public string Run(ITestEventListener? listener, string filter)
+            {
+                CheckLoadWasCalled();
+                log.Info("Running {0} - see separate log file", Path.GetFileName(_testAssemblyPath.ShouldNotBeNull()));
+                return ExecuteAction(RUN_ACTION, listener, filter);
+            }
+
+            public void RunAsync(Action<string> callback, string filter) => throw new NotImplementedException();
+
+            public void StopRun(bool force) => ExecuteAction(STOP_RUN_ACTION, force);
```
manfred-brands, line 99: No check on `CheckLoadWasCalled`. Same comment as for 2018 driver. Move that check into the `ExecuteAction` method.

manfred-brands (changes requested):
A few more coments/questions.

src/NUnitEngine/nunit.engine/Runners/WorkItemTracker.cs
```diff
@@ -63,6 +65,8 @@ public int CompareTo(InProgressItem? other)
             }
         }
 
+        private static readonly ILogger log = InternalTrace.GetLogger(nameof(InProgressItem));
+
```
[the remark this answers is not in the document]
CharliePoole: It was added for temporary log entries to figure out a problem. I usually keep these once added so logging statements can be added quickly.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkDriver.cs
```diff
@@ -0,0 +1,132 @@
...
+        readonly NUnitFrameworkApi _api;
+
+#if NETFRAMEWORK
+        /// <summary>
+        /// Construct an NUnitFrameworkDriver
+        /// </summary>
+        /// <param name="testDomain">The application domain in which to create the FrameworkController</param>
+        /// <param name="nunitRef">An AssemblyName referring to the test framework.</param>
+        public NUnitFrameworkDriver(AppDomain testDomain, string id, AssemblyName nunitRef)
+            : this(testDomain, nunitRef.Version >= new Version(3,2,0) ? "2018" : "2009", id, nunitRef) { }
```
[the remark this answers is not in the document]
CharliePoole: Sure. But they aren't really numbers conceptually, just tags, so it bothered me conceptually. I think I'll go back to the original version where I had two entirely distinct constructors rather than using `this` in the public constructor.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkDriver.cs
```diff
@@ -0,0 +1,132 @@
...
+            Guard.ArgumentValid(api == "2009" || api == "2018", $"Invalid API specified: {api}", nameof(api));
+            Guard.ArgumentNotNullOrEmpty(id, nameof(id));
+            Guard.ArgumentNotNull(nunitRef, nameof(nunitRef));
+
+            ID = id;
+
+            _api = api == "2018"
+                ? (NUnitFrameworkApi)testDomain.CreateInstanceFromAndUnwrap(
+                    Assembly.GetExecutingAssembly().Location,
+                    $"NUnit.Engine.Drivers.NUnitFrameworkApi{api}",
```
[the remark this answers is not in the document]
CharliePoole: Of course!

src/NUnitEngine/nunit.engine.core.tests/Services/DriverServiceTests.cs
```diff
@@ -23,36 +23,26 @@ public void CreateDriverFactory()
...
-#else
-            new TestCaseData("mock-assembly.dll", false, typeof(NUnitNetCore31Driver)),
-            new TestCaseData("mock-assembly.dll", true, typeof(NUnitNetCore31Driver)),
-            //new TestCaseData("notest-assembly.dll", false, typeof(NUnitNetCore31Driver)),
-#endif
-// Invalid cases should work with all target runtimes
+            new TestCaseData("mock-assembly.dll", false, typeof(NUnitFrameworkDriver)),
+            new TestCaseData("mock-assembly.dll", true, typeof(NUnitFrameworkDriver)),
+            //new TestCaseData("notest-assembly.dll", false, typeof(NUnitFrameworkDriver)),
+            //new TestCaseData"notest-assembly.dll", true, typeof(SkippedAssemblyFrameworkDriver))
```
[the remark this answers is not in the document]
CharliePoole: I used Ignore instead for visibility. This area of tests needs some rework.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2018.cs
CharliePoole: The sole purpose is to ensure that the error message thrown makes sense to the user. That is, if the user requested a particular runtime, it's a user error. If it's due to the engine picking the wrong runtime, it's an internal error. I'll finish some others and come back to this.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2018.cs
CharliePoole: It was present long ago, but StopRun does nothing if no test is running. So CheckLoadWasCalled would sometimes give a false error if the run had already terminated.

src/TestData/NUnit3.10/NUnit3.10.csproj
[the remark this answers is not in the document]
CharliePoole: This and the others were a great catch because when I fixed them the tests started to fail.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2018.cs
CharliePoole: Removing this.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2018.cs
CharliePoole: I think abstracting it is a good idea, but I'd rather get it working with separate code first. "To remove duplication, you must first create it."

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2018.cs
CharliePoole: It does need it and I just re-verified it by trying to remove it.

Background... I tried several alternate approaches to make this work, hoping that the code could be more similar between the two APIs. It didn't work because we have no common interface shared with the framework. 

In the end I returned to this very old technique used in V2, injecting a class into the same domain as the tests. The sequence of events is this...
1. We load the test assembly, which causes the framework assembly to also be loaded.
2. We "load" (it's already in memory) the framework assembly to get a reference to it.
3. We create an instance of the controller.
4. When we finally call LOAD_METHOD of the framework API, everything is already in memory. The loading simply lets the controller accesses what's already loaded.

This is how NUnit worked approximately its first decade. 

Writing this, I realize it's rather obscure and maybe deserves a comment in the code.

CharliePoole: What isn't needed is a member field to hold the test assembly reference. I made that a local variable.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkDriver.cs
```diff
@@ -0,0 +1,157 @@
...
+        /// <summary>
+        /// String naming the API in use, for use by tests
+        /// </summary>
+        internal string API { get; } = string.Empty;
+
+        /// <summary>
+        /// An id prefix that will be passed to the test framework and used as part of the
+        /// test ids created.
+        /// </summary>
+        public string ID { get; } = string.Empty;
```
manfred-brands, line 111: Suggest to drop the ` = string.Empty;` so any future change will still pass in `id` in the constructor.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2018.cs
```diff
@@ -0,0 +1,253 @@
...
+        {
+            if (method == null)
+                throw new NUnitEngineException(INVALID_FRAMEWORK_MESSAGE);
+
+            log.Debug($"Executing {method.DeclaringType}.{method.Name}");
+
+#if NETFRAMEWORK
+            return method.Invoke(_frameworkController, args).ShouldNotBeNull();
+#else
+            using (_assemblyLoadContext.ShouldNotBeNull().EnterContextualReflection())
```
manfred-brands, line 239: If I interpret the [documentation](https://github.com/dotnet/coreclr/blob/master/Documentation/design-docs/AssemblyLoadContext.ContextualReflection.md) correctly this is not needed.
The context needs setting when the instance is created, not when one of it's methods is called.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2018.cs
```diff
@@ -0,0 +1,253 @@
...
+        {
+            if (_frameworkController == null)
+                throw new InvalidOperationException(LOAD_MESSAGE);
+        }
+
+#if NETCOREAPP
+        private object CreateInstance(string typeName, params object?[]? args)
+        {
+            var type = _frameworkAssembly.ShouldNotBeNull().GetType(typeName, throwOnError: true)!;
+            return Activator.CreateInstance(type, args)!;
```
manfred-brands, line 157: This is where the `using (_assemblyLoadContext.ShouldNotBeNull().EnterContextualReflection())`
should be called as dependencies are loaded when calling `GetType` or `CreateInstance`.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2018.cs
```diff
@@ -0,0 +1,253 @@
...
+
+        public string Load(string testAssemblyPath, IDictionary<string, object> settings)
+        {
+            Guard.ArgumentNotNull(testAssemblyPath, nameof(testAssemblyPath));
+            Guard.ArgumentNotNull(settings, nameof(settings));
+            Guard.ArgumentValid(File.Exists(testAssemblyPath), "Framework driver called with a file name that doesn't exist.", "testAssemblyPath");
+            log.Info($"Loading {testAssemblyPath} - see separate log file");
+
+            _testAssemblyPath = Path.GetFullPath(testAssemblyPath);
+            var idPrefix = string.IsNullOrEmpty(_driverId) ? "" : _driverId + "-";
```
manfred-brands, line 66: `_driverId` should never be `null` and with a change to `Guard.IsNotNullOrEmpty` in the constructor it should not be empty either. We can then remove this conditional here.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2018.cs
```diff
@@ -0,0 +1,253 @@
...
+        private readonly string _driverId;
+
+        private readonly AssemblyName _nunitRef;
+
+        private object? _frameworkController;
+        private Type? _frameworkControllerType;
+
+#if NETCOREAPP
+        private TestAssemblyLoadContext? _assemblyLoadContext;
+        //private Assembly? _testAssembly;
```
manfred-brands, line 43: Remove commented out declaration.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2009.cs
```diff
@@ -0,0 +1,167 @@
...
+                {
+                _frameworkController = _testDomain.CreateInstanceAndUnwrap(
+                    _nunitRef.FullName,
+                    CONTROLLER_TYPE,
+                    false,
+                    0,
+                    null,
+                    new object[] { _testAssemblyPath, idPrefix, settings },
+                    null,
+                    null).ShouldNotBeNull();
```
manfred-brands, line 75: Indent.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2009.cs
```diff
@@ -0,0 +1,167 @@
...
+            }
+
+            private object CreateObject(string typeName, params object?[]? args)
+            {
+                try
+                {
+                return _testDomain.CreateInstanceAndUnwrap(
+                    _nunitRef.FullName, typeName, false, 0, null, args, null, null)!;
+                //return _testDomain.CreateInstanceAndUnwrap(
+                //        _nunitRef.FullName, typeName, false, 0, null, args, null, null)!;
```
manfred-brands, line 158: Indent and remove the commented out copy.

src/NUnitEngine/nunit.engine.core/Drivers/DriverService.cs
```diff
@@ -70,9 +70,9 @@ public IFrameworkDriver GetDriver(AppDomain domain, string assemblyPath, string?
 
                 if (platform == "Silverlight" || platform == ".NETPortable" || platform == ".NETStandard" || platform == ".NETCompactFramework")
                     if (skipNonTestAssemblies)
-                        return new SkippedAssemblyFrameworkDriver(assemblyPath);
+                        return new SkippedAssemblyFrameworkDriver(assemblyPath, "X");
```
manfred-brands, line 73: Suggest to use `package.ID` here as well as otherwise the `ID` will not be unique.

src/NUnitEngine/nunit.engine.core.tests/Services/DriverServiceTests.cs
```diff
@@ -7,7 +7,7 @@
 using NUnit.Engine.Drivers;
 using NUnit.Engine.Extensibility;
 
-namespace NUnit.Engine.Services.Tests
+namespace NUnit.Engine.Services
```
manfred-brands, line 10: Not sure what your _namespace rule_ is as the namespaces in the test assembly are inconsistent. Some contain _Tests_ others don't.

src/NUnitEngine/nunit.engine.core.tests/Drivers/NUnitFrameworkDriverTests.cs
```diff
@@ -13,22 +12,39 @@
 namespace NUnit.Engine.Drivers
 {
     // Functional tests of the NUnitFrameworkDriver calling into the framework.
-    public class NUnit3FrameworkDriverTests
+#if NETFRAMEWORK
+    [TestFixture("2009")]
+    [TestFixture("2018")]
+#endif
```
manfred-brands, line 18:
```suggestion
#if NETFRAMEWORK
    [TestFixture("2009")]
#endif
    [TestFixture("2018")]
```

src/NUnitEngine/nunit.engine.core.tests/Drivers/NUnitFrameworkDriverTests.cs
```diff
@@ -13,22 +12,39 @@
...
+        private NUnitFrameworkDriver _driver;
         private string _mockAssemblyPath;
 
+#if NETFRAMEWORK
+        private string _whichApi;
+        public NUnitFrameworkDriverTests(string whichApi)
+        {
+            _whichApi = whichApi;
+        }
+#endif
```
manfred-brands, line 35: If you only put the `2009` Test case inside `#if NETFRAMEWORK` you can make this code unconditional.
This then allows you to move `RunTestsAction_WithInvalidFilterElement_ThrowsException` outside `#if FRAMEWORK` so that you also test invalid elements for .NET Core.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2018.cs
CharliePoole: This is code from the original driver that I simply moved to the api class. To be honest, I didn't understand why it was needed but I hesitated to change it. I'll try removing and leave it out if nothing breaks.

CharliePoole: You're right. Not needed. Doesn't seem to be needed when creating the instance either.

manfred-brands: I hadn't heard of `AssemblyLoadContext` before.
If I interpreted the referenced documentation correctly, you can remove it here as it wouldn't be used.
That is _unless_ the test (controller) loads something indirect via reflection, e.g. creating a test fixture instance.

I think it might be important when the runner/engine and the testassembly both depend on a common assembly but use different versions. The test should load the version the testassembly wants, not the one the runner uses.
I will see if I can make some completely separate test code to see if it matters.

src/NUnitEngine/nunit.engine.core/Drivers/DriverService.cs
CharliePoole: An experiment I ran before creating the final form of the driver. :-(

src/NUnitEngine/nunit.engine.core.tests/Services/DriverServiceTests.cs
CharliePoole: Up to now my personal rule has been NOT to use `.Tests` but I don't go out of my way to change the code so it's still inconsistent. I'd like to eventually create a simple minimalistic coding standard, but it hasn't been a priority. It's also a chore I somewhat dread because it can bring out big arguments about very small things.

src/NUnitEngine/nunit.engine.core.tests/Drivers/NUnitFrameworkDriverTests.cs
CharliePoole: Nice!

src/NUnitEngine/nunit.engine.core.tests/Services/DriverServiceTests.cs
CharliePoole: Nevertheless, we should address this at some point. :-)

CharliePoole:
@manfred-brands Great review. I think the code is improved a lot. Sorry it was such a slog. I think I'll try to split big things into smaller PRs for review, even if they don't represent a releasable increment.

You'll need to either re-approve or merge the PR. It appears that I've lost my ability to force a merge, likely as a side effect of turning on the automatic reviewer selection feature.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2018.cs
CharliePoole: In case it helps, here's history... @rprouse wrote this for .NET Standard, where there isn't any `AssemblyLoadContext`. @ChrisMaddock  (I believe) enhanced it for .NET 3.1, which introduced `AssemblyLoadContext` and made it easier to take charge of how tests were loaded in a specific context. I don't think it has had much change since then.

In any case, I'd like to get this in so we can look at some other stuff. If you come up with something different, let's just do a bug or refactoring issue.

src/NUnitEngine/nunit.engine.core/Drivers/NUnitFrameworkApi2018.cs
CharliePoole: Not needed in our tests, but it's certainly possible it may be needed in some situation. Maybe try a test that depends on a different version of `TestCentric.Metadata` than the one we use. It's supposed to work as a result of our creating a special context for the tests - equivalent to the NetFramework approach of using a separate AppDomain.

manfred-brands:
> You'll need to either re-approve or merge the PR

I'm happy to leave the `AssemblyContext` issues to later.
However, you marked some issues I raised as resolved, but I don't see a new commit.

manfred-brands (approved):
Approved. Assuming you have a pending commit with the small changes.

CharliePoole:
I'm not sure what didn't get in. With such a large set of changes, I use "resolved" to keep track of what I need to pay attention to, resolving each item after I have changed the code or made a comment that may question the suggestion or disagree with it. Anyway, I'll double-check.

PR #1603, opened 2025-01-18
Drop agents for .NET 7.0 and .NET Core 3.1

### LINKED ISSUE #1601: Drop .NET Core 3.1 and .NET 7.0 agents in preparation for 3.20 release
I'd like to drop these agents from the working code now, to simplify the work. The change would be published with version 4, to which I expect to add a .NET 9.0 agent. I feel we should keep the 6.0 agent for a while longer.

I believe that dropping the agents will also allow me to drop the corresponding builds of nunit.engine.core, but I'll need to check that further.

We'll continue to build files like mock-assembly for those targets in order to use them for testing.

@nunit/engine-team @nunit/core-team What do you think?

CharliePoole:
Fixes #1601

@nunit/engine-team I have enabled automatic assignment of reviews for this team. If I have done it right, only those who have expressed a willingness to review will be included. Let's see what happens when I push the button!

UPDATE: Github picked @stevenaw ! So far so good. :-)

### DIFF
--- a/.github/workflows/NUnitConsoleAndEngine.CI.yml
+++ b/.github/workflows/NUnitConsoleAndEngine.CI.yml
@@ -31,12 +31,14 @@ jobs:
         with:
           fetch-depth: 0
 
+        # Ensure all required runtimes are available
       - name: 🛠️ Setup .NET
         uses: actions/setup-dotnet@v4
         with:
           dotnet-version: |
             3.1.x
             6.0.x
+            7.0.x
             8.0.100
 
       - name: 🔧 Install dotnet tools
--- a/build.cake
+++ b/build.cake
@@ -63,17 +63,13 @@ BuildSettings.Packages.AddRange(new PackageDefinition[] {
             HasFiles("LICENSE.txt", "NOTICES.txt"),
             HasDirectory("tools").WithFiles("nunit3-console.exe", "nunit3-console.exe.config").AndFiles(ENGINE_FILES),
             HasDirectory("tools/agents/net462").WithFiles(AGENT_FILES),
-            HasDirectory("tools/agents/netcoreapp3.1").WithFiles(AGENT_FILES_NETCORE),
             HasDirectory("tools/agents/net6.0").WithFiles(AGENT_FILES_NETCORE),
-            HasDirectory("tools/agents/net7.0").WithFiles(AGENT_FILES_NETCORE),
             HasDirectory("tools/agents/net8.0").WithFiles(AGENT_FILES_NETCORE)
         },
         symbols: new PackageCheck[] {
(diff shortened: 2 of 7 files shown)


.github/workflows/NUnitConsoleAndEngine.CI.yml
```diff
@@ -31,12 +31,14 @@ jobs:
...
           fetch-depth: 0
 
+        # Ensure all required runtimes are available
       - name: 🛠️ Setup .NET
         uses: actions/setup-dotnet@v4
         with:
           dotnet-version: |
             3.1.x
             6.0.x
+            7.0.x
```
veleek, line 41: Why are we adding the 7.0 runtime to remove support for it?

veleek (approved):
I know I wasn't assigned to this, but I had a few minutes.  :D

.github/workflows/NUnitConsoleAndEngine.CI.yml
stevenaw: A bit of a similar question from me @CharliePoole . I assume these are needed for tests?

stevenaw (approved):
Thanks @CharliePoole . LGTM, though I left a small question just for my own understanding.

OsirisTerje:
The adapter runs under .net core 3.1.  Not sure yet when we can step up that one.

.github/workflows/NUnitConsoleAndEngine.CI.yml
CharliePoole: An earlier build failed because I removed it by mistake, so I added it back. Yes, we need both 3.1 and 7.0 for the test data assemblies, like mock-assembly.dll.

CharliePoole:
@OsirisTerje The adapter, however doesn't use agents. Even so, I'll take a look at the latest code to be sure before I merge this.

CharliePoole:
@OsirisTerje I reviewed the code for the adapter and I see it references the engine package without specifying a version. The `NUnit.Engine` package targeted .NET 4.6.2 and .NET Standard 2.0 through version 3.18.2. In 3.18.3, .NET 6.0 was added experimentally. In 3.19.0 both .NET Standard 2.0 and .NET 6.0 were dropped in favor of .NET 8.0, so we now only have builds for .NET 4.6.3 and .NET 8.0.

So... if the adapter has continued to build through all those changes, I would think we're OK. In this PR I have not changed the engine  itself although the agents present in the package have changed.

OTOH, I'm surprised if you are able to call .NET 8.0 dependencies from .NET Core 3.1. In retrospect, I guess I ought to know more about what you need from the engine.

I'll go ahead and merge this because it seems that anything broken has been broken by the 3.19 release as well. Before we  get to a release point, I think one of us should experiment by forcing the adapter to use the latest engine alpha build, so we can be know whether some further fix is needed.

PR #1605, opened 2025-01-19
Eliminate built-in agents for .NET Core 3.1, .NET 6.0 and .NET 7.0

CharliePoole:
@OsirisTerje I've asked for your review because what platforms we support is important for all projects.


CharliePoole:
This is passing and the limited set of agents is expected to be enhanced before the first beta, so I'm going ahead and merging.

PR #1609, opened 2025-01-22
WIP: Create new nunit.common assembly

### LINKED ISSUE #1608: Separate assembly for InternalTrace and other common classes
InternalTrace is implemented in nunit.engine.core, which we plan to eliminate. We would then need to replicate it in nunit.engine, nunit.agent.core and NUnit.Extensibility. What's more, the framework has its own copy of InternalTrace.

In addition, a number of other useful classes are currently shared across projects by links to the source code. This is slightly messy and can only get worse.

I propose to create a single ~~`NUnit.InternalTrace`~~ `nunit.common` assembly to be referenced by each project that needs to have it. In fact, I'll be _re_-creating it, since we had one in the past.

This is an experimental issue - what we used to call a "spike." If it seems viable, I'll continue it. If not, I'll try to abandon it quickly, with minimal time invested. UPDATE: The strikeout above definitely show the experimental nature of this issue. :-)

Since I have already implented InternalTrace centrally for TestCentric I'll start by stealing that code. Of course, nunit-console is organized quite differently to TestCentric, with multiple projects under one solution, so changes will probably be needed.

I'll report back here when I either remove this issue or remove the experimental label.

@nunit/engine-team At this point, it would be good to put in your suggestions, including whether or not you think this is a good idea. :-)

CharliePoole:
Fixes #1608 

Throwing this out for review. I added only the InternalTrace code plus some classes that were using cross-project source links and so obviously belong in a common assembly.

I eliminated the NUnit.Common namespace and put all these things in NUnit, so they can be used without adding a using statement.

I also removed ILogging and ILogger interfaces, which didn't seem to be doing much for us.

I haven't moved the tests to a separate assembly yet but I think I should.


CharliePoole:
Getting this reviewed and merged will enable a number of things. I'm pretty sure we will find other things to move here as we go along, but I see that as being a file or two at a time, as demanded by other issues.

src/NUnitEngine/nunit.common/nunit.common.csproj
```diff
@@ -0,0 +1,20 @@
...
+    <AssemblyOriginatorKeyFile>..\..\nunit.snk</AssemblyOriginatorKeyFile>
+    <OutputPath>..\..\bin\$(Configuration)\</OutputPath>
+    <DebugType>portable</DebugType>
+    <DebugSymbols>true</DebugSymbols>
+  </PropertyGroup>
+
+  <Import Project="$(MSBuildThisFileDirectory)/../../Nullable.props" />
+
+  <ItemGroup>
+    <ProjectReference Include="..\nunit.engine.api\nunit.engine.api.csproj" />
```
manfred-brands, line 17: If this library is (in the future) to be shared by other nunit components, then it shouldn't depend on 'engine.api'.
It then becomes where to define `InternalTraceLevel`. Maybe it should move to common instead.

src/NUnitEngine/nunit.engine.core.tests/nunit.engine.core.tests.csproj
```diff
@@ -65,4 +65,8 @@
     <None Include="App.config" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="Internal\Logging\" />
+  </ItemGroup>
+
```
manfred-brands, line 71: Probably left over as the folder doesn't exist.

src/NUnitEngine/nunit.common.tests/Program.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    class Program
+    {
+        static int Main(string[] args)
+        {
+#if NETFRAMEWORK
+            return new TextRunner(typeof(Program).Assembly).Execute(args);
+#else
+            return new TextRunner(typeof(Program).GetTypeInfo().Assembly).Execute(args);
+#endif
```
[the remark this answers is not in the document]
CharliePoole: I'm happy to get rid of this horrible thing. :-)

src/NUnitEngine/nunit.common/AgentExitCodes.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+namespace NUnit
+{
+    internal static class AgentExitCodes
```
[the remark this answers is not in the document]
CharliePoole: I missed that. Will fix.

src/NUnitEngine/nunit.common/nunit.common.csproj
CharliePoole: I agree it needs to go but I left it for now because `ITestEngine` has an `InternalTraceLevel` property. The problem will only arise if and when we want to create a separate package for the common, as you say. Can you think of an easy way to eliminate that property while still having the ability to set the engine trace level from the runner? Otherwise, let's leave it for a separate issue.

manfred-brands: If it is in the API, it should probably not be called InternalTraceLevel.
I only see an option splitting of a common.api but that might not be worth it if all itt does is hold an enum.
Probably leave it until another issue.

src/NUnitEngine/nunit.common.tests/Program.cs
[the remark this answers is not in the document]
CharliePoole: In fact, although  it's unrelated, I'll get rid of it in all the unit test packages.

CharliePoole:
I'll need to fix the conflicts and I want to check the defunct issue-1578 branch, which also created NUnit.Common to see if there is anything else to be added. Marked as WIP, briefly I hope.

PR #1612, opened 2025-01-25
Create new nunit.extensibility and nunit.extensibility.api assemblies

### LINKED ISSUE #1049: Separate assembly for NUnit extensibility
This feature was first developed as "Engine Extensibility" so the implementation is embodied in an engine service. This makes it next to impossible to use extensibility in other places, like in the console runner, third party runners and even in our own agents.

I propose that we create a separate assembly... NUnit.Extensibility to contain most of the implementation... the existing engine service would simply use an API to access the implementation details.

As a subsquent improvement (but not part of this issue) we could implement the IFrameworkDriverFactory extension in nunit.core using the API directly, without need for full access to engine services.

I'm calling this an idea for now and asking for @nunit/engine-team  discussion.

CharliePoole:
Fixes #1049


CharliePoole:
@manfred-brands As I told you on slack, I'm hoping you'll continue to review these major structural changes I'm making. This is another wide-ranging one and I think it will take some time, but the reviews are really productive. If you're worn out for the moment, I'll ask for volunteers

build.cake
```diff
@@ -25,7 +25,7 @@ PackageDefinition NUnitConsoleRunnerNet80Package;
 PackageDefinition NUnitEnginePackage;
 PackageDefinition NUnitEngineApiPackage;
 PackageDefinition NUnitConsoleRunnerChocolateyPackage;
-PackageDefinition NUnitConsoleZipPackage;
+//PackageDefinition NUnitConsoleZipPackage;
```
manfred-brands, line 28: Should this be removed?

src/Extensibility/nunit.extensibility.tests/nunit.extensibility.tests.csproj
```diff
@@ -0,0 +1,26 @@
...
+    <OutputType>Exe</OutputType>
+    <OutputPath>../../../bin/$(Configuration)</OutputPath>
+    <SignAssembly>true</SignAssembly>
+    <AssemblyOriginatorKeyFile>..\..\nunit.snk</AssemblyOriginatorKeyFile>
+    <DebugType>Full</DebugType>
+    <CheckEolTargetFramework>false</CheckEolTargetFramework>
+    <NoWarn>1505</NoWarn>
+  </PropertyGroup>
+
+  <ItemGroup>
```
manfred-brands, line 15: Please add Nullable.props and NUnit.Analyzers to suppress some errors.

src/Extensibility/nunit.extensibility.tests/ExtensionSelectorTests.cs
```diff
@@ -105,8 +105,8 @@ public void IsBetterVersionOfPrefersNoChangeIfFromWildcard()
         }
 
         private static IExtensionAssembly MockExtension(string assemblyName = "ExtensionSelectorTestsExtension",
-            Version? assemblyVersion = null,
-            Version? targetFramework = null,
+            Version assemblyVersion = null,
+            Version targetFramework = null,
```
manfred-brands, line 109: Please don't undo nullable enabled code.

src/Extensibility/nunit.extensibility.tests/ExtensionAssemblyTests.cs
```diff
@@ -2,18 +2,18 @@
...
+namespace NUnit.Extensibility
 {
     public class ExtensionAssemblyTests
     {
         private static readonly Assembly THIS_ASSEMBLY = Assembly.GetExecutingAssembly();
         private static readonly string THIS_ASSEMBLY_PATH = THIS_ASSEMBLY.Location;
-        private static readonly string? THIS_ASSEMBLY_NAME = THIS_ASSEMBLY.GetName().Name;
-        private static readonly Version? THIS_ASSEMBLY_VERSION = THIS_ASSEMBLY.GetName().Version;
+        private static readonly string THIS_ASSEMBLY_NAME = THIS_ASSEMBLY.GetName().Name;
+        private static readonly Version THIS_ASSEMBLY_VERSION = THIS_ASSEMBLY.GetName().Version;
```
manfred-brands, line 16: Don't undo nullable enabled code.

src/Extensibility/nunit.extensibility/ExtensionManager.cs
```diff
@@ -124,7 +115,7 @@ public virtual void FindExtensionPoints(params Assembly[] targetAssemblies)
                             string msg = string.Format(
```
manfred-brands, line 115: Did you forgot to convert this one to a formattable string?

src/Extensibility/nunit.extensibility/ExtensionManager.cs
```diff
@@ -95,12 +91,7 @@ public virtual void FindExtensionPoints(params Assembly[] targetAssemblies)
                 foreach (ExtensionPointAttribute attr in assembly.GetCustomAttributes(typeof(ExtensionPointAttribute), false))
                 {
                     if (_extensionPointIndex.ContainsKey(attr.Path))
-                    {
-                        string msg = string.Format(
-                            "The Path {0} is already in use for another extension point.",
-                            attr.Path);
-                        throw new NUnitEngineException(msg);
-                    }
+                        throw new Exception("The Path {attr.Path} is already in use for another extension point.");
```
manfred-brands, line 94: You forgot the leading `$` to make this a formattable string.

```suggestion
                        throw new Exception($"The Path {attr.Path} is already in use for another extension point.");
```

Note that MS doesn't like code throwing `Exception`.
It will raise [CA2201](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2201).
So we should either create an NUnitExtensibilityException or throw something like `InvalidOperation` or `Argument` or `NotSupport` exception.

src/Extensibility/nunit.extensibility/ExtensionNode.cs
```diff
@@ -24,13 +24,13 @@ public class ExtensionNode : IExtensionNode
...
-        /// <param name="targetFramework">The target framework of the extension assembly.</param>
-        public ExtensionNode(string assemblyPath, Version assemblyVersion, string typeName, IRuntimeFramework? targetFramework)
+        ///// <param name="targetFramework">The target framework of the extension assembly.</param>
+        public ExtensionNode(string assemblyPath, Version assemblyVersion, string typeName)
         {
             AssemblyPath = assemblyPath;
             AssemblyVersion = assemblyVersion;
             TypeName = typeName;
-            TargetFramework = targetFramework;
+            //TargetFramework = targetFramework;
```
manfred-brands, line 33: Remove.

src/NUnitConsole/nunit4-console/ConsoleRunner.cs
```diff
@@ -324,8 +325,8 @@ private void DisplayExtensionList()
                 {
                     _outWriter.Write(INDENT6 + "Extension: ");
                     _outWriter.Write(ColorStyle.Value, $"{node.TypeName}");
-                    if(node.TargetFramework != null)
-                        _outWriter.Write(ColorStyle.Value, $"(.NET {node.TargetFramework?.FrameworkVersion})");
+                    //if(node.TargetFramework != null)
+                    //    _outWriter.Write(ColorStyle.Value, $"(.NET {node.TargetFramework?.FrameworkVersion})");
```
manfred-brands, line 329: Remove.

src/NUnitEngine/nunit.common.tests/nunit.common.tests.csproj
```diff
@@ -12,8 +12,6 @@
     <DebugSymbols>true</DebugSymbols>
   </PropertyGroup>
 
-  <Import Project="$(MSBuildThisFileDirectory)/../../Nullable.props" />
-
```
manfred-brands, line 16: Why did I bother to all nullability if you remove it again?

manfred-brands (changes requested):
Overall looks good, except for two big issues:
1. Don't drop nullability. Not even for NUnit tests. The NUnit.Analyzer will fix most errors.
2. Come up with a better exception than `Exception` to throw.

CharliePoole:
Thanks for reviewing. To your general comments...

1. You're right... I was experimenting with removing it for tests because I have the impression that it's more trouble than it is worth in tests. That's one thing I'd like to chat about with you, so as to reach a common understanding. However, I did talk with Terje about it and he also favors making tests nullable so that should be the goal. 
2. I'm thinking that as well, I just failed to decide on a name.

build.cake
CharliePoole: Yes. It was indefinite because JetBrains uses it in their installation process for TeamCity. But as of a few days ago, we know that we can drop it for V4. They will use the nuget packages.

src/Extensibility/nunit.extensibility.tests/nunit.extensibility.tests.csproj
CharliePoole: I'll do it like in nunit.core.tests. However, this approach seems to be very busy. Maybe more should be put into Directory.Build.props.

manfred-brands: Nullable.props can be put in Directory.build.props, bit didn't you say you liked everything specified in individual csproj files?

We could create a `nunit.props` which includes everything needed for all test assemblies.

manfred-brands:
>  I have the impression that it's more trouble than it is worth in tests.

Why should unit test have a lower standard than the rest of the code?
The nunit.analyzers help a lot, suppressing warnings about uninitialised fields when these are set in `Setup` methods and when accessing nullable values after testing for `Is.Not.Null`.

CharliePoole:
Sorry, but I don't think that not using nullability means "lower standard". We made an effort to keep high standards before it existed. It could mean that maintaining a higher standard becomes easier, which is why I'm working at it now. When it first came out, I didn't believe the hype and since then I've been mostly working alone.

You convinced me to try it, to learn about it. That means you'll see me trying to find out if it's good in various places or not. Be happy for me. :-)

NUnit.Analyzers: I didn't know it did that. That's probably why adding the Nullability.props import without adding the analyzer caused so much trouble. I think this stuff needs to be more centralized, perhaps in Directory.Build.props.

src/Extensibility/nunit.extensibility.tests/ExtensionSelectorTests.cs
CharliePoole: Fixed

src/Extensibility/nunit.extensibility.tests/ExtensionAssemblyTests.cs
CharliePoole: Fixed

src/Extensibility/nunit.extensibility/ExtensionManager.cs
CharliePoole: Done... found a few more as well.

src/Extensibility/nunit.extensibility/ExtensionManager.cs
CharliePoole: Fixed the error. I'll go through this assembly and look at all `throw Exception` instances. NUnitExtensibilityException is good for a lot of cases, I think.

src/NUnitConsole/nunit4-console/ConsoleRunner.cs
CharliePoole: Yup. I was hoping to restore it but no dice for now. Maybe after all the changes are done and we have replaced RuntimeFramework with something more modern.

src/NUnitEngine/nunit.common.tests/nunit.common.tests.csproj
CharliePoole: Turns out this was a case of not having the NUnit.Analyzers and I didn't know that was the trick to make it work as a separate test assembly.

src/Extensibility/nunit.extensibility.tests/nunit.extensibility.tests.csproj
CharliePoole: I'm poor at expressing myself in short bits, like here.

I'm not entirely comfortable with "surprises" when editing the csproj file, due to not seeing the result. But I see the value of centralizing some things. There's a balance point to be found, probably, and I'd like to find it.

Directory.Build.Props works well for tests when all the tests are in some directory. Currently, we recognize tests in the build script by a naming pattern rather than locations. Maybe we should chat briefly sometime and decide on a direction that would not surprise me but make life easier for all.

CharliePoole:
@manfred-brands Waiting for you to re-review.

package-tests.cake
```diff
@@ -279,20 +279,20 @@ AddToBothLists(new PackageTest(
...
+//    new MockAssemblyExpectedResult("net-4.6.2"),
+//    KnownExtensions.NUnitV2ResultWriter));
 
-StandardRunnerTests.Add(new PackageTest(
-    1, "V2ResultWriterTest",
-    "Run mock-assembly under .NET 6.0 and produce V2 output",
-    "testdata/net6.0/mock-assembly.dll --result=TestResult.xml --result=NUnit2TestResult.xml;format=nunit2",
-    new MockAssemblyExpectedResult("netcore-6.0"),
-    KnownExtensions.NUnitV2ResultWriter));
+//StandardRunnerTests.Add(new PackageTest(
```
manfred-brands, line 290: Commented out code. Does something need fixing?

manfred-brands (approved):
One remaining question otherwise looks good.

package-tests.cake
CharliePoole: Because of changes in the API, extensions written for version 3 will need to be re-written, which is why almost all of them were already commented-out until I have a chance to look at them individually. For some reason I don't actually understand, the ResultWriter extension continued to work until this latest change. I decided to postpone it until the infrastructure changes are done.

Issue #1614, opened 2025-01-29
Broken Extensions discovery when installed with Chocolatey


mazharenko:
NUnit engine seems to always discover extensions as if installed with nuget.

Find the minimal reproducible example in [chocolib.zip](https://github.com/user-attachments/files/18587090/chocolib.zip). It contains nunit-console-runner and nunit-extension-teamcity-event-listener just as they are after installation via choco. 

```
> nunit3-console.exe --list-extensions 
NUnit Console Runner 3.19.1 (Release)
Copyright (c) 2022 Charlie Poole, Rob Prouse
Wednesday, January 29, 2025 4:43:05 PM

Runtime Environment
   OS Version: Microsoft Windows NT 6.2.9200.0
   Runtime: .NET Framework CLR v4.0.30319.42000

Installed Extensions
  Extension Point: /NUnit/Engine/NUnitV2Driver
  Extension Point: /NUnit/Engine/TypeExtensions/IService
  Extension Point: /NUnit/Engine/TypeExtensions/ITestEventListener
  Extension Point: /NUnit/Engine/TypeExtensions/IDriverFactory
  Extension Point: /NUnit/Engine/TypeExtensions/IProjectLoader
  Extension Point: /NUnit/Engine/TypeExtensions/IResultWriter
```

But when I rename the folder with the Teamcity listener to nunit.extension.teamcity-event-listener ([workaround.zip](https://github.com/user-attachments/files/18587110/workaround.zip)), it starts working:

```
> nunit3-console.exe --list-extensions
NUnit Console Runner 3.19.1 (Release)
Copyright (c) 2022 Charlie Poole, Rob Prouse
Wednesday, January 29, 2025 4:45:33 PM

Runtime Environment
   OS Version: Microsoft Windows NT 6.2.9200.0
   Runtime: .NET Framework CLR v4.0.30319.42000

Installed Extensions
  Extension Point: /NUnit/Engine/NUnitV2Driver
  Extension Point: /NUnit/Engine/TypeExtensions/IService
  Extension Point: /NUnit/Engine/TypeExtensions/ITestEventListener
    Extension: NUnit.Engine.Listeners.TeamCityEventListener(.NET 2.0) (Disabled)
      Version: 1.0.7.0
      Path: C:\Workspace\scratching\tmp\workaround\nunit.extension.teamcity-event-listener\tools\teamcity-event-listener.dll
  Extension Point: /NUnit/Engine/TypeExtensions/IDriverFactory
[...]

PR #1616, opened 2025-01-29
Fix determination of chocolatey

mazharenko:
This addresses #1614


CharliePoole:
@mazharenko Thanks for the fix!

It was obviously correct by inspection, so I merged it. Please verify it in your environment by installing from our MyGet feed. Version is 3.19.2-alpha.2. Unfortunately, our post-packaging tests only simulate installation under Chocolatey, which is why this problem escaped notice. I'll have to think about finding a way to actually use chocolatey in our CI builds, but that will take a bit of time.

Once you have verified that this works, I'll queue it up for including in 3.19.2

mazharenko:
Thank you! I can confirm that my tests now run once I upgraded nunit-console-runner from your chocolatey feed.

PR #1617, opened 2025-01-31
Move common settings to Directory.Build.props

### LINKED ISSUE #1589: Move common settings from individual .csproj files to Directory.Build.props
Items like `OutputPath` and `AssemblyOriginatorKeyFile` are necessary in every project.
We can reduce maintenance issues by moving these into `Directory.Build.props`

CharliePoole:
Fixes #1589 

Rather than changing the directory structure at this point, I created separate second-level `Directory.Build.props` files for Extensions, NUnitConsole and NUnitEngine. We may want to move things around to consolidate these in a separate issue.


src/Extensibility/Directory.Build.props
```diff
@@ -0,0 +1,7 @@
+<!-- Directory.Build.props for Extensibility -->
+<Project>
+
+    <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
```
manfred-brands, line 4: Technically correct, but in this case the same as the simpler:
```suggestion
    <Import Project="$(MSBuildThisFileDirectory)../Directory.Build.props" />
```

manfred-brands (approved):
Code is correct.
I have added one remark comment. It holds for all other similar references.

I have one further suggestion.
Define a property for the sign key hash in the main Directory.Build.props and move the internalsvisibleto from the .cs code to the .csproj where you can reference that key property.

```xml
<ItemGroup>
    <InternalsVisibleTo Include="$(AssemblyName).NUnit$(NUnitPublicKeyReference)" />
</ItemGroup>
```

This also works if signing and setting of the key property is done conditionally. At work we don't sign in Debug builds to ease debugging.

src/Extensibility/Directory.Build.props
CharliePoole: Done. I chose flexibility (in moving directories up and down the hierarchy) over simplicity. But in retrospect, that's a good choice if the Import is in a `.csproj` file, but not so much when in a `.props` file.

CharliePoole:
I like the idea of using `InternalsVisibleTo` item in the `.csproj` file rather than `AssemblyInfo.cs`.  I used a slightly different format for readability, i.e. 

```
  <ItemGroup>
    <InternalsVisibleTo Include="nunit.engine.tests, PublicKey=$(NUnitPublicKey)" />
    <InternalsVisibleTo Include="DynamicProxyGenAssembly2, PublicKey=$(NSubstitutePublicKey)" />
  </ItemGroup>
```

PR #1619, opened 2025-02-03
Isolate the agents from the engine

### LINKED ISSUE #1578: Isolate agents from the engine itself
This is a long-standing direction for V4, but one for which no actual issue had previously been written. This issue aims to capture the overall goal and steps needed beyond what has been done to date.

Isolation of the various agents that run tests from the engine itself is an important feature because it allows the two sides to evolve separately. Perhaps more important, it will allow the agents to _stop_ evolving when the runtimes they support reach EOL, while still allowing users to continue to use those agents if they need to.

Currently, agents do not reference the engine but they do reference `engine.core`, which is also referenced by the engine itself. We aim to eliminate `nunit.engine.core`. Types referenced by the engine will move into `nunit.engine` while those referenced by agents will move into a new assembly, `nunit.agent.core`. FWIW, I managed to do this with the TestCentric engine, only needing to duplicate a few types in the two surviving assemblies.

This issue has a soft dependency on #1049. It will be more convenient to first remove `ExtensionManager` and related classes from `nunit.engine.core` rather than moving those types twice.

Issue #909 has a similar soft dependency on this one.

issue #1539 is also related in that I would expect to create the new `nunit.agent.core` assembly as nullable in the first place.

CharliePoole:
Fixes #1578


src/NUnitCommon/nunit.agent.core.tests/nunit.agent.core.tests.csproj
```diff
@@ -0,0 +1,35 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net462;net8.0</TargetFrameworks>
+    <ExposedInternals>true</ExposedInternals>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)'=='net462' OR '$(TargetFramework)'=='net6.0' OR '$(TargetFramework)'=='net8.0'">
```
manfred-brands, line 8: There is no `net6.0` target anymore.

src/NUnitEngine/nunit.engine/nunit.engine.csproj
```diff
@@ -21,9 +21,11 @@
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\..\Extensibility\nunit.extensibility\nunit.extensibility.csproj" />
+    <ProjectReference Condition="'$(TargetFramework)'=='net8.0'"
+                      Include="..\..\NUnitCommon\nunit.agent.core\nunit.agent.core.csproj" />
```
manfred-brands, line 25: Engine references agent? But only for .net8.0.

manfred-brands (approved):
Looks good as far as I can see on phone.
A few TODOs.

src/NUnitCommon/nunit.agent.core.tests/nunit.agent.core.tests.csproj
CharliePoole: Yup. I'll make that change but wait to see what else you have. You may also notice that there are untested agent-core targets: netcoreap3.1 and net6.0. I keep them because I know I'll use them later for separate user-installable agents. I see no point of doing it for anything older than those.

src/NUnitEngine/nunit.engine/nunit.engine.csproj
CharliePoole: Yes... because the .NET 8.0 engine only runs in process, making it a sort of agent in itself. I want to eventually create an in-process agent, at which point we can drop the reference.

src/NUnitCommon/nunit.agent.core.tests/nunit.agent.core.tests.csproj
CharliePoole: I left it in for now as it does no harm.

PR #1620, opened 2025-02-05
Modify call sequence to agents so all arguments are named

### LINKED ISSUE #955: Modify arguments to agent process for 4.0
NUnit-Agent has an odd argument structure as compared to normal command-line apps. The first two args are positional and required... if they aren't present, the process just fails without explanation.

We felt OK doing this because the agent process is only started by TestAgency and the agent and agency are maintained in the same project. There's even a comment in the agent code explaining this logic. 

With 4.0, this will change in a few ways. With pluggable agents, agent and agency code may not be maintained in the same repo any longer. We are also likely to start agents in other ways than through test agencies.

I propose rationalizing the agent a bit by making all the accepted arguments named options: agentID for the first one, agencyUrl for the second. The code is already present for named arguments is already present.

I don't actually think this is breaking, but 4.0 it seems like a good point at which to make the change anyway.

At the same time, we may want to make the argument processing more like what we do in the console... accepting `/` to start an option on windows and sing-letter abbreviations for some of the options.

CharliePoole:
Fixes #955

This is part of #909 and also enables improvements in how we test agents.

### DIFF
--- a/src/NUnitCommon/nunit.agent.core/AgentOptions.cs
+++ b/src/NUnitCommon/nunit.agent.core/AgentOptions.cs
@@ -49,10 +49,9 @@ namespace NUnit.Agents
                         val = option.Substring(delim + 1);
                     }
 
-                    if (!VALID_OPTIONS.ContainsKey(opt))
+                    // Simultaneously check that the option is valid and determine if it takes an argument
+                    if (!VALID_OPTIONS.TryGetValue(opt, out bool optionTakesValue))
                         throw new Exception($"Invalid argument: {arg}");
-
-                    bool optionTakesValue = VALID_OPTIONS[opt];
 
                     if (optionTakesValue)
                     {
@@ -119,7 +118,7 @@ namespace NUnit.Agents
 
         private static bool IsOption(string arg)
         {
-            return arg.StartsWith("--");
+            return arg.StartsWith("--", StringComparison.Ordinal);
         }
     }
 }
--- a/src/NUnitCommon/nunit.agent.core/AgentOptions.cs
+++ b/src/NUnitCommon/nunit.agent.core/AgentOptions.cs
@@ -1,0 +1,125 @@
+﻿// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+using NUnit.Engine;
+using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace NUnit.Agents
+{
+    /// <summary>
+    /// All agents, either built-in or pluggable, must be able to 
+    /// handle the options defined in this class. In some cases,
+    /// it may be permissible to ignore them but they should never
+    /// give rise to an error.
+    /// </summary>
+    public class AgentOptions
+    {
+        static readonly char[] DELIMS = new[] { '=', ':' };
+        // Dictionary containing valid options with bool value true if a value is required.
+        static readonly Dictionary<string, bool> VALID_OPTIONS = new Dictionary<string, bool>();
+
+        static AgentOptions()
+        {
+            VALID_OPTIONS["agentId"] = true;
+            VALID_OPTIONS["agencyUrl"] = true;
+            VALID_OPTIONS["debug-agent"] = false;
+            VALID_OPTIONS["debug-tests"] = false;
+            VALID_OPTIONS["trace"] = true;
+            VALID_OPTIONS["pid"] = true;
+            VALID_OPTIONS["work"] = true;
+        }
+
+        public AgentOptions(params string[] args)
+        {
+            int index;
+            for (index = 0; index < args.Length; index++)
+            {
+                string arg = args[index];
+
+                if (IsOption(arg))
+                {
+                    var option = arg.Substring(2);
+                    var delim = option.IndexOfAny(DELIMS);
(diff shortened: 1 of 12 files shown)


CharliePoole:
@stevenaw If you don't have time for this right now, let me know and I'll ask someone else.

stevenaw:
Oh, sorry about that, I must've missed that the review came my way. I'll pull the code and review over the weekend. 
Thanks for following up @CharliePoole

src/NUnitCommon/nunit.agent.core/AgentOptions.cs
```diff
@@ -0,0 +1,125 @@
...
+                    var delim = option.IndexOfAny(DELIMS);
+                    var opt = option;
+                    string? val = null;
+                    if (delim > 0)
+                    {
+                        opt = option.Substring(0, delim);
+                        val = option.Substring(delim + 1);
+                    }
+
+                    if (!VALID_OPTIONS.ContainsKey(opt))
```
stevenaw, line 52: nitpick: Could this be simplified to something like the below to save the duplicate dictionary lookup?
```
if (!VALID_OPTIONS.TryGetValue(opt, out bool optionTakesValue))
   throw new Exception($"Invalid argument: {arg}");
```

src/NUnitCommon/nunit.agent.core/AgentOptions.cs
```diff
@@ -0,0 +1,125 @@
...
+        public bool DebugTests { get; } = false;
+        public bool DebugAgent { get; } = false;
+        public InternalTraceLevel TraceLevel { get; } = InternalTraceLevel.Off;
+        public string WorkDirectory { get; } = string.Empty;
+
+        public List<string> Files { get; } = new List<string>();
+
+        private static bool IsOption(string arg)
+        {
+            return arg.StartsWith("--");
```
stevenaw, line 122: nitpick: I think this `StartsWith` overload will be culture-sensitive by default. Could we pass a `StringComparison` in?
 
```suggestion
            return arg.StartsWith("--", StringComparison.Ordinal);
```

src/NUnitCommon/nunit.agent.core/AgentOptions.cs
```diff
@@ -0,0 +1,125 @@
...
+    /// All agents, either built-in or pluggable, must be able to 
+    /// handle the options defined in this class. In some cases,
+    /// it may be permissible to ignore them but they should never
+    /// give rise to an error.
+    /// </summary>
+    public class AgentOptions
+    {
+        static readonly char[] DELIMS = new[] { '=', ':' };
+        // Dictionary containing valid options with bool value true if a value is required.
+        static readonly Dictionary<string, bool> VALID_OPTIONS = new Dictionary<string, bool>();
```
stevenaw, line 20: question: will arg-parsing be case sensitive or case-insensitive? I'm unsure if a `StringComparer.OrdinalIgnoreCase` could be required as a parameter to this Dictionary constructor or not

src/NUnitCommon/nunit.agent.core/AgentOptions.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        static AgentOptions()
+        {
+            VALID_OPTIONS["agentId"] = true;
+            VALID_OPTIONS["agencyUrl"] = true;
+            VALID_OPTIONS["debug-agent"] = false;
+            VALID_OPTIONS["debug-tests"] = false;
+            VALID_OPTIONS["trace"] = true;
+            VALID_OPTIONS["pid"] = true;
+            VALID_OPTIONS["work"] = true;
```
stevenaw, line 30: question: The class-level comment indicates these arguments must be "well-known" to other consumers. Should we consider exposing the option names publicly in any way (not necessarily in this PR). Maybe something like this, but not necessarily exactly like this:
```
public class AgentOptions {
  public static class AgentOptionNames {
    public static readonly string AgentId = "agentId";
  }
}
```

Something like that, if even with private visibility, could be used internally by the class when we parse the values in the if/else block further down as well.

stevenaw (approved):
Thanks @CharliePoole . Looks good to me. I've left a few nitpicks and some questions. Nothing major, however I'm curious your thoughts on case (in)sensitivity of the option names.

src/NUnitCommon/nunit.agent.core/AgentOptions.cs
CharliePoole: Yes. The code will be less clear, I think, bu tI can add a comment.

src/NUnitCommon/nunit.agent.core/AgentOptions.cs
CharliePoole: Because NUnit is cross-platform, we have always made arguments case-sensitive.

src/NUnitCommon/nunit.agent.core/AgentOptions.cs
CharliePoole: That could be needed at some point. If so, it can sit alongside the constants we have for packages. Let's think about  this when launchers are introduced.

CharliePoole:
Thanks @stevenaw. Let's think about whether to standardiize option names in some way when we introduce pluggable agents with launchers.

PR #1624, opened 2025-02-09
Direct Agent Execution

### LINKED ISSUE #1623: Direct execution of agents
Allow execution of agents from the command-line with a simplified set of options and the path to a single test assembly file,

Output similar to the console runner will be produced and no options not already defined by our agents will be added. Executing in this way will run all tests in the specified assembly, displaying the results.

The primary purpose of this change is to allow easy testing of our pluggable agents. A secondary use may be found in debugging issues presented by users, where the console runner does not display sufficient information.

CharliePoole:
Fixes #1623

### DIFF
--- a/src/NUnitCommon/nunit.agent.core.tests/AgentDirectRunnerTests.cs
+++ b/src/NUnitCommon/nunit.agent.core.tests/AgentDirectRunnerTests.cs
@@ -19,7 +19,7 @@ namespace NUnit.Agents
             RunTestUnderTestBed(typeof(MockAssembly).Assembly.Location);
         }
 
-        private void RunTestUnderTestBed(string testAssembly)
+        private static void RunTestUnderTestBed(string testAssembly)
         {
             string agentAssembly = typeof(DirectTestAgent).Assembly.Location;
 
@@ -36,7 +36,7 @@ namespace NUnit.Agents
 
             Process? process = Process.Start(startInfo);
 
-            if (process != null)
+            if (process is not null)
             {
                 process.WaitForExit();
 
--- a/src/NUnitCommon/nunit.agent.core.tests/AgentDirectRunnerTests.cs
+++ b/src/NUnitCommon/nunit.agent.core.tests/AgentDirectRunnerTests.cs
@@ -1,0 +1,56 @@
+﻿// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Text;
+using NUnit.Framework;
+using NUnit.TestData.Assemblies;
+
+namespace NUnit.Agents
+{
+    public class AgentDirectRunnerTests
+    {
+        [Test]
+        public void RunAgentDirectly()
+        {
+            RunTestUnderTestBed(typeof(MockAssembly).Assembly.Location);
+        }
+
+        private void RunTestUnderTestBed(string testAssembly)
+        {
+            string agentAssembly = typeof(DirectTestAgent).Assembly.Location;
+
+#if NETFRAMEWORK
+            string agentExe = agentAssembly;
+#else
+            string agentExe = Path.ChangeExtension(agentAssembly, ".exe");
+#endif
+
+            var startInfo = new ProcessStartInfo(agentExe);
+            startInfo.Arguments = testAssembly;
+            startInfo.RedirectStandardOutput = true;
+            startInfo.UseShellExecute = false;
+
+            Process? process = Process.Start(startInfo);
+
+            if (process != null)
+            {
+                process.WaitForExit();
+
+                string output = process.StandardOutput.ReadToEnd();
+                int index = output.IndexOf("Test Run Summary");
+                if (index > 0)
+                    output = output.Substring(index);
+
+                Console.WriteLine(output);
+                Console.WriteLine($"Agent process exited with rc={process.ExitCode}");
+
+                if (index < 0)
+                    Assert.Fail("No Summary Report found");
+            }
+        }
+    }
+}
--- a/src/NUnitCommon/nunit.agent.core.tests/Program.cs
+++ b/src/NUnitCommon/nunit.agent.core.tests/Program.cs
@@ -5,9 +5,9 @@ using NUnitLite;
 
 namespace NUnit.Engine.Tests
 {
-    class Program
+    public class Program
     {
-        static int Main(string[] args)
+        public static int Main(string[] args)
         {
             return new TextRunner(typeof(Program).Assembly).Execute(args);
         }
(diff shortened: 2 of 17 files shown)


src/NUnitCommon/nunit.agent.core.tests/Runners/DomainManagerStaticTests.cs
```diff
@@ -16,11 +16,7 @@ public static class DomainManagerStaticTests
         static string path2 = TestPath("/test/bin/debug/test2.dll");
         static string path3 = TestPath("/test/utils/test3.dll");
 
-#if NETFRAMEWORK
-        const string STANDARD_CONFIG_FILE = "nunit.engine.core.tests.exe.config";
-#else
-        const string STANDARD_CONFIG_FILE = "nunit.engine.core.tests.dll.config";
-#endif
+        const string STANDARD_CONFIG_FILE = "nunit.agent.core.tests.exe.config";
```
manfred-brands, line 19: Compiling this branch, I see the previous behaviour:
```
$ find . -name 'nunit.agent.core.tests*.config'
./net462/nunit.agent.core.tests.exe.config
./net8.0/nunit.agent.core.tests.dll.config
```

manfred-brands (changes requested):
I don't think this new code is part of the testing as I don't see it in build.cake or package-tests.cake but I'm not clear on how the tester is tested.

CharliePoole:
The DirectTestAgent isn't part of the engine but of the tests. It just provides an exe to allow exercising the direct execution facility in nunit.agent.core. There's only one unit test that uses it right now but once the agents are in separate projects, all their tests will be run through it. We have never tested the agents in isolation before but this makes a start.

src/NUnitCommon/nunit.agent.core.tests/AgentDirectRunnerTests.cs
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        private void RunTestUnderTestBed(string testAssembly)
+        {
+            string agentAssembly = typeof(DirectTestAgent).Assembly.Location;
+
+#if NETFRAMEWORK
+            string agentExe = agentAssembly;
+#else
+            string agentExe = Path.ChangeExtension(agentAssembly, ".exe");
```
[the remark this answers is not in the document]
CharliePoole: It should continue to be supported. Some linux distros support exe but not all. Let's add an issue for this.

CharliePoole:
This has been dormant for a while, so I'm making it WIP while I rebase and decide how to move it  forward.

CharliePoole:
@manfred-brands

This is again ready for review. I took most of your suggestions and brought the PR up to date with the latest commits. Since that included the new analyzers, I fixed the analyzer problems. I also added a few package tests to actually make use of the feature.

For the moment, I have kept the code duplication for console output. I think some of this code will change once the agents are extensions in their own right. At that point, I'd like to allow it to evolve independently. We'll then have a better idea of exactly what parts can be common and eliminate the duplication.

src/NUnitCommon/nunit.agent.core.tests/Program.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+using System.Reflection;
```
manfred-brands, line 3: This using statement is not needed: Removed.

src/NUnitCommon/nunit.agent.core/AgentDirectRunner.cs
```diff
@@ -0,0 +1,325 @@
...
+                OutWriter.WriteLine(ColorStyle.SubHeader, DateTime.Now.ToString(CultureInfo.CurrentCulture.DateTimeFormat.FullDateTimePattern));
+                OutWriter.WriteLine();
+            }
+        }
+
+        private static TAttr? GetAttribute<TAttr>(Assembly assembly)
+            where TAttr : Attribute
+        {
+            return assembly?.GetCustomAttribute<TAttr>();
+        }
```
manfred-brands, line 90: This method no longer adds any simplification or abstraction: Removed.

src/NUnitCommon/nunit.common/XmlHelper.cs
```diff
@@ -1,8 +1,10 @@
 ﻿// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
 using System;
+using System.Collections.Generic;
 using System.Globalization;
 using System.Xml;
+using System.Xml.Linq;
```
manfred-brands, line 7: Not needed: Removed.

src/NUnitCommon/nunit.common/XmlHelper.cs
```diff
@@ -145,5 +149,32 @@ public static DateTime GetAttribute(this XmlNode result, string name, DateTime d
 
             return date;
         }
+
+        public static IEnumerable<XmlNode> NonNullChildNodes(this XmlNode node)
```
manfred-brands, line 153: Not used: Removed.

src/NUnitCommon/nunit.common/XmlHelper.cs
```diff
@@ -145,5 +149,32 @@ public static DateTime GetAttribute(this XmlNode result, string name, DateTime d
...
+        {
+            foreach (XmlNode childNode in node.ChildNodes)
+                if (childNode is not null)
+                    yield return childNode;
+        }
+
+        public static void ForEachChildNode(this XmlNode node, Action<XmlNode> action)
```
manfred-brands, line 160: Could simply call: `node.ChildNodes.ForEachNode(action)`: Updated,

manfred-brands (commented):
A few small items which I have fixed in a review commit.
The main issue I have is what I said before, why is the reporting different between direct agent execution and the ConsoleRunner.
Both call `runner.Run` and print the results.
Why can the `AgentDirectRunner` not create an `ResultReporter` for the report?

manfred-brands (approved):
Approved with the caveat of your message that the reporting will be re-visited.

CharliePoole:
Thanks for your review. I think it was important to get this odd bit of code in place before I start splitting out all the agents into separate projects. I also agree that something ought to be done about the output before that happens.

The duplication in this PR is duplication I just created _by design_. I copied some code from the console, solely as a way to get us  started. I want people to see this feature and give us feedback with ideas about where it could be used and how it should look in the future.

Once that happens, we might remove the duplication in several ways...
* Dropping the output entirely and relying on the console runner via some option.
* Refactoring so that the report class is somewhere in common and used by both
* Factoring out common pieces while keeping the rest specific to the individual application

If I had decided _not_ to have the duplication in the first place, I would be picking one of those things in advance.

If this sounds crazy to you, that's OK too. It's something I've done in a lot of places and I've had a lot of strange reactions at first, but it seems to pay off more than not. ;-)

CharliePoole:
Looks like that alpha.34 label was never removed. I'm a bit tired, so I'll mess directly with the labels on version4 branch in the morning when I'm fresh.

CharliePoole:
I got a few hours sleep and fixed the tags and did a push --tags -f on version4. Reran the build and it all looks good now. The problem was that the alpha.34 and alpha.35 tags were on detached commits. 

Latest on myget is now 4.0.0-alpha.35 and I think we're now good for future runs.

PR #1625, opened 2025-02-10
Fix error in use of --extensionDirectory

### LINKED ISSUE #1613: Extensions are no longer loading as is currently documented
NUnit Engine/API versions 3.19.1.

How should extensions be loaded now? With the latest nunit-console (3.19.1) and engines/api, our extensions are no longer loading. 

From the source code, you can see a major change here where the code is only looking at specific folders that match hardcoded patterns:
```
public void FindExtensionAssemblies(Assembly hostAssembly)
{
    log.Info($"FindExtensionAssemblies called for host {hostAssembly.FullName}");

    bool isChocolateyPackage = System.IO.File.Exists(Path.Combine(hostAssembly.Location, "VERIFICATION.txt"));
    string[] extensionPatterns = isChocolateyPackage
        ? new[] { "nunit-extension-*/**/tools/", "nunit-extension-*/**/tools/*/" }
        : new[] { "NUnit.Extension.*/**/tools/", "NUnit.Extension.*/**/tools/*/" };


    IDirectory startDir = _fileSystem.GetDirectory(AssemblyHelper.GetDirectoryName(hostAssembly));

    while (startDir != null)
    {
        foreach (var pattern in extensionPatterns)
            foreach (var dir in _directoryFinder.GetDirectories(startDir, pattern))
                ProcessDirectory(dir, true);

        startDir = startDir.Parent;
    }
}
```

Does the documentation around loading extensions need to be updated, or is there a way to get engine to allow for other folders to contain extensions?

CharliePoole:
Fixes #1613

### DIFF
--- a/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
+++ b/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
@@ -52,7 +52,7 @@ namespace NUnit.ConsoleRunner
 
         public ConsoleRunner(ITestEngine engine, ConsoleOptions options, ExtendedTextWriter writer)
         {
-            Guard.ArgumentNotNull(_engine = engine, nameof(engine));
+             Guard.ArgumentNotNull(_engine = engine, nameof(engine));
             Guard.ArgumentNotNull(_options = options, nameof(options));
             Guard.ArgumentNotNull(_outWriter = writer, nameof(writer));
 
@@ -71,7 +71,7 @@ namespace NUnit.ConsoleRunner
                 foreach (string extensionDirectory in extensionPath.Split(new[] { Path.PathSeparator }, StringSplitOptions.RemoveEmptyEntries))
                     _extensionService.FindExtensionAssemblies(extensionDirectory);
 
-            foreach (string extensionDirectory in _options.ExtensionDirectories)
+             foreach (string extensionDirectory in _options.ExtensionDirectories)
                _extensionService.FindExtensionAssemblies(extensionDirectory);
 
             _workDirectory = options.WorkDirectory;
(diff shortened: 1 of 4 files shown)


CharliePoole:
@veleek Essentially, this is a sort of spike to find out what works for the user. I already merged it to make it available. Two more stages will follow: (1) changes to our CI recipe to allow automated testing of this sort of change and (2) adding those tests and refining the fix based on feedback. 

For that reason, even though this PR is already merged, I'd like you to review it, which GitHub seems to allow. I'll also ask you to review the subsequent PRs in this series as they come along.

If you're not able to do this one, as usual, just let me know and I'll re-assign it.

veleek:
I will try to get this reviewed today or tomorrow.

CharliePoole:
Cool. Also, chime in on the issue about how this should really work after you've looked at it. I'm leaning toward option 2 myself.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -71,7 +71,7 @@ public ConsoleRunner(ITestEngine engine, ConsoleOptions options, ExtendedTextWri
                 foreach (string extensionDirectory in extensionPath.Split(new[] { Path.PathSeparator }, StringSplitOptions.RemoveEmptyEntries))
                     _extensionService.FindExtensionAssemblies(extensionDirectory);
 
-            foreach (string extensionDirectory in _options.ExtensionDirectories)
+             foreach (string extensionDirectory in _options.ExtensionDirectories)
```
veleek, line 74: Both changes in this file look like unintentiontal whitespace additions.

src/NUnitEngine/nunit.engine.core/Services/DriverService.cs
```diff
@@ -127,5 +118,24 @@ public override void StartService()
...
         }
+
+        private void InitializeDriverFactories()
+        {
+            _factories = new List<IDriverFactory>();
+
+            if (_extensionService != null)
+            {
+                foreach (IDriverFactory factory in _extensionService.GetExtensions<IDriverFactory>())
+                    _factories.Add(factory);
```
veleek, line 129: This could be switched to use `_factories.AddRange(...)` instead of manually iterating.

PR #1626, opened 2025-02-22
Further work on 1613

### LINKED ISSUE #1613: Extensions are no longer loading as is currently documented

CharliePoole:
Fixes #1613 by adding tests that use the `--extensionDirectory` option and checking the output. Updated to latest version of the cake recipe in order to do do this.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -1,5 +1,5 @@
 // Load the recipe 
-#load nuget:?package=NUnit.Cake.Recipe&version=1.3.1-alpha.1
+#load nuget:?package=NUnit.Cake.Recipe&version=1.4.0-alpha.1
 // Comment out above line and uncomment below for local tests of recipe changes
 //#load ../NUnit.Cake.Recipe/recipe/*.cake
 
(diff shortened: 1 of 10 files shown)


veleek (commented):
No notes from me on this.

CharliePoole:
@veleek Thanks for looking at it. I didn't actually intend to close this until after your review. I guess I'm working too many issues in parallel and getting confused. I'll reduce my WIP now!

PR #1627, opened 2025-02-23
Add --enable and --disable for extensions and fix --teamcity

### LINKED ISSUE #1570: Presence of a TEAMCITY_PROJECT_NAME environment variable makes builds fail if TeamCity service unavailable
With release 3.19.0 in, [this commit](https://github.com/nunit/nunit-console/pull/1517) leads to TeamCity CI builds failing for us with `Option --teamcity specified but the extension is not installed.` even though we do not specify this option and we do not use any of TeamCity's nunit-related features: we just happen to use nunit3-console to run tests.

- The full command line being used: `nunit3-console --noheader --noresult /path/to/tests.dll`
- Method of installation (e.g. NuGet, zip, msi): NuGet, but not relevant
- Version of the NUnit Engine/Console: 3.19.0

Can be reproduced easily by setting a non-empty `TEAMCITY_PROJECT_NAME` environment variable.

So:
- the error message says `--teamcity` is specified even though that is not the case, which is a bit confusing
- the [Options code](https://github.com/nunit/nunit-console/blob/main/src/NUnitConsole/nunit3-console/Options/DefaultOptionsProvider.cs#L15) treats the presence of a `TEAMCITY_PROJECT_NAME` environment variable as running under TeamCity. While correct, it also treats that as 'must use TeamCity extension', but that is for us not the case

I'd suggest that either the behavior is reverted, or modified to show a warning instead of an error (perhaps with a modified message stating 'TeamCity detected' or similar), or else made such that the user needs to explicitly request the TeamCity functionality (so by passing `--teamcity` or by another environment variable).

CharliePoole:
Fixes #1570


stevenaw (approved):
Thanks @CharliePoole No concerns from me.

Curious to hear your thoughts on of its worth issuing a warning or info of some kind of a user tries to "disable" an extension which couldn't be found. There would be no functional difference to the run either way, of course, but such a scenario could arise from a misconfiguration that a user may appreciate a notice about.

CharliePoole:
Thanks for reviewing @stevenaw .

I thought about a warning and didn't do it for several reasons.

1. We don't do warnings very well unless they pertain to individual tests. In that case we give a WARNING result. But general warnings can only be displayed in the output as text, probably at the start of the run.
2. Giving a warning doesn't work very nicely with the use case where the test may be run on various machines with or without the extension and you want the result to be the same on all machines.

So, I decided to leave it out on the principle that someone with a use case where a warning would really be helpful can always ask for a change.

stevenaw:
That makes a lot of sense. Thanks for the explanation.

PR #1642, opened 2025-03-12
feat: Add .NET9 agent

### LINKED ISSUE #1602: Add agent for .NET 9.0 to the version 3 build
This can be done soon and released with version 3.20.

mikkelbu:
Fixes #1602


CharliePoole:
Super! Thanks for getting this done. It's the last big piece needed for 3.20.

CharliePoole (changes requested):
Looks pretty good to me. I think your decision to keep all the dependent assemblies as is makes sense. We can upgrade any that need it later on.

Can you upgrade the agent unit tests by adding a .NET 9.0 build? We could also use some package tests that target 9.0. Both can be pretty minimal.

OTOH if you are out of time, this is good to merge, and one of us will add the tests later.

Thanks again!

Issue #1590, opened 2025-01-02
Enable more analyzers for code style consistency and prevent other common mistakes


manfred-brands:
We should enable both Microsoft.NetAnalyzers and StyleCop.Analyzers to align the code and keep it consistent.

Some files still contain tabs. We should get a warning and fix that. Some are even mixed mode (Options.cs)
Fields only set in the constructor should be marked `readonly`.
Properties only set in the constructor should only have a `get`, not a (private) `set`
Structures types should be marked `readonly` where possible.
Classes should be `sealed` if not meant to be extensible.
Use _expression_ bodies for properties if single-lined.
...

We can enable rules as you see fit.

CharliePoole:
@manfred-brands I'm in favor of adding the Microsoft.NetAnalyzers. Do you want to take a shot at setting it up with some set of rules to start with? All the stuff you mention is fine for me.

To a large extent, however, I don't like to have too many stylistic rules. To be clear, my experience, before I retired, was working with groups of highly competent developers who were able to read and code using a variety of styles without confusion. That may bias me and I'm happy to talk about it.

CharliePoole:
@manfred-brands I've lost track. Did we decide against dong any of this? I know we discussed it somewhere, but obviously not on the issue itself.

manfred-brands:
@CharliePoole I did find a local branch where I had done work. But I can't recall why I stopped. Let me rebase and check the status.

manfred-brands:
I now remember, it became a gigantic task, even if doing it rule by rule.

Split into:
* https://github.com/nunit/nunit-console/pull/1662 Add 'is null' analyzer.
* https://github.com/nunit/nunit-console/pull/1663 Add StyleCop analyzer.
* https://github.com/nunit/nunit-console/pull/1664 Add MS Net Analyzer.

CharliePoole:
Bummer! I didn't pay attention to the PR numbers and merged the last one first creating conflicts. Will start over. :-(

manfred-brands:
They are separate, I'll rebase them once approved.

CharliePoole:
I was going to say you're up early, but I see it's actually night there!

Should have left the merging to you in that case. I already did a forced push to remove the one I merged - I usually only do that on a feature branch, but it seemed simplest and it's not main (yet). :-)

Once the build passes, I'll go back to just reviewing and approving and you can rebase and merge in the morning.

PR #1662, opened 2025-04-13
CSharp is null analyzer 

manfred-brands:
Partly addresses #1590 

This analyzer is enabled in both nunit and nunit.analyzer projects.

Prefer `xxx is null` over `xxx == null` and `xxx is not null` over `xxx != null`


CharliePoole (approved):
Looks good. Thanks!

PR #1663, opened 2025-04-13
Add StyleCop.Analyzers

manfred-brands:
Partly addresses #1590 

Settings are taken from nunit framework project.

With the following exceptions:
1. Braces are optional as per your preferences.
2. The this. prefix is not defined yet.
    We can either leave it, or enforce the use of 'this.' or deny the the use of 'this.'.
    The latter has my preference.

### DIFF
--- a/src/NUnitEngine/nunit.engine.api/TestSelectionParserException.cs
+++ b/src/NUnitEngine/nunit.engine.api/TestSelectionParserException.cs
@@ -15,18 +15,24 @@ namespace NUnit.Engine
         /// <summary>
         /// Construct with a message
         /// </summary>
-        public TestSelectionParserException(string message) : base(message) { }
+        public TestSelectionParserException(string message) : base(message)
+        {
+        }
 
         /// <summary>
         /// Construct with a message and inner exception
         /// </summary>
         /// <param name="message"></param>
         /// <param name="innerException"></param>
-        public TestSelectionParserException(string message, Exception innerException) : base(message, innerException) { }
+        public TestSelectionParserException(string message, Exception innerException) : base(message, innerException)
+        {
+        }
 
         /// <summary>
         /// Serialization constructor
         /// </summary>
-        public TestSelectionParserException(SerializationInfo info, StreamingContext context) : base(info, context) { }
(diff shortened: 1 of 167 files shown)


src/NUnitEngine/nunit.engine.tests/Helpers/ProcessTester.cs
```diff
@@ -36,7 +37,8 @@ public static ProcessResult Run(ProcessStartInfo startInfo)
                 };
                 process.ErrorDataReceived += (sender, e) =>
                 {
-                    if (e.Data == null) return;
+                    if (e.Data == null)
+                        return;
```
CharliePoole, line 41: Another kind of thing where I find the rule overly pedantic, but at least it doesn't force adding parens.

src/NUnitEngine/nunit.engine.api/TestSelectionParserException.cs
```diff
@@ -15,18 +15,24 @@ public class TestSelectionParserException : Exception
...
 
         /// <summary>
         /// Construct with a message and inner exception
         /// </summary>
         /// <param name="message"></param>
         /// <param name="innerException"></param>
-        public TestSelectionParserException(string message, Exception innerException) : base(message, innerException) { }
+        public TestSelectionParserException(string message, Exception innerException) : base(message, innerException)
+        {
+        }
```
CharliePoole, line 29: I see a number of these. Depending on what happens when I try to put the empty braces on the same line, I may get annoyed at StyleCop for not being smart enough to see that there is nothing in the braces so don't worry. :-)

But let's give it a go anyway.

CharliePoole (approved):
Well, not much to argue about here except for the two items I noted. But rather than dispute the principles, let's see if anyone gets annoyed by them.

That's my main guideline for enforced rules on a team: they should not annoy anyone. :-)

src/NUnitEngine/nunit.engine.tests/Helpers/ProcessTester.cs
manfred-brands: If it starts to annoy you we can disable rule SA1501, but until them leave it.

PR #1664, opened 2025-04-13
Enable Minimum .NET Analyzers

manfred-brands:
Partly addresses #1590

Not sure why some of these rules are enabled in the minimum ruleset.

I did them per rule so we can decided if we want to have that rule or make our own set.


CharliePoole:
These all make sense to me as you have them. Thanks for this!

CharliePoole (commented):
Good to go. Thank you.

PR #494, opened 2018-11-08
Report failures in testfixture teardowns as failures

mikkelbu:
Failures in testfixture teardowns (e.g. failures in OneTimeTearDown
or Dispose) are now reported as failures, meaning that they will
affect the number of failures shown in the output of the console and
also affect the exit code of the console.

Fixes #483, fixes #464, and fixes #282.

PR #1669, opened 2025-04-23
Move common classes for console output to nunit.common

### LINKED ISSUE #1668: Move console report classes to nunit.common
ConsoleResultReporter and a number of dependencies are likely to be useful in other output situations, including the direct agent execution output. These types will be moved to the nunit.common assembly, which already contains a number of types used by the console runner.

CharliePoole:
Fixes #1668


src/NUnitCommon/nunit.common/ClientTestResult.cs
```diff
@@ -6,13 +6,13 @@
 using System.Xml;
 using NUnit.TextDisplay;
 
-namespace NUnit.ConsoleRunner
+namespace NUnit
```
manfred-brands, line 9: Prefer `NUnit.Common`

src/NUnitCommon/nunit.common/TextDisplay/ResultReporter.cs
```diff
@@ -0,0 +1,254 @@
...
+    {
+        public static void WriteHeader(ExtendedTextWriter writer)
+        {
+            Assembly entryAssembly = Assembly.GetEntryAssembly() ?? Assembly.GetExecutingAssembly();
+            var versionBlock = FileVersionInfo.GetVersionInfo(entryAssembly.ManifestModule.FullyQualifiedName);
+
+            var header = $"{versionBlock.ProductName} {versionBlock.ProductVersion}";
+
+            var configurationAttributes = entryAssembly.GetCustomAttribute<AssemblyConfigurationAttribute>();
+
```
manfred-brands, line 21: Left in when other code using the AssemblyConfigurationAttribute was removed?
Were you intending to add the `Configuration` to the _header_?

src/NUnitConsole/nunit4-console/TestEventHandler.cs
```diff
@@ -1,12 +1,10 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
-using System;
 using System.Xml;
 using NUnit.Common;
 using NUnit.Engine;
-using NUnit.TextDisplay;
 
-namespace NUnit.ConsoleRunner
+namespace NUnit.TextDisplay
```
manfred-brands, line 7: You changed the namespace, but left the code in the nunit4-console project.

src/NUnitCommon/nunit.common/ResultSummary.cs
```diff
@@ -204,10 +216,11 @@ private void Summarize(XmlNode node, bool failedInFixtureTearDown)
...
                     }
-                    if ((type == "SetUpFixture" || type == "TestFixture") && status == "Failed" && label == "Error" && site == "TearDown")
-                    {
-                        failedInFixtureTearDown = true;
-                    }
+                    // TODO: This seems wrong and causes failures. Leaving it commented while awaiting feedback.
+                    //if ((type == "SetUpFixture" || type == "TestFixture") && status == "Failed" && label == "Error" && site == "TearDown")
+                    //{
+                    //    failedInFixtureTearDown = true;
+                    //}
```
manfred-brands, line 223: This causes code to increment `ErrorCount` instead of `PassCount`.
You changed the `MockAssembly.Passed` expected count by adding: `PassedButFailedInTearDown`
Should a test be marked Error if the teardown fails?

manfred-brands (changes requested):
A few small items/questions.
Note that there is also a copy of `WriteHeader` in nunit4-netcore-console/Program.cs which could be replaced as well.

src/NUnitCommon/nunit.agent.core.tests/AgentDirectRunnerTests.cs
```diff
@@ -39,16 +41,10 @@ private static void RunTestUnderTestBed(string testAssembly)
...
-                string output = process.StandardOutput.ReadToEnd();
-                int index = output.IndexOf("Test Run Summary");
-                if (index > 0)
-                    output = output.Substring(index);
-
-                Console.WriteLine(output);
                 Console.WriteLine($"Agent process exited with rc={process.ExitCode}");
 
-                if (index < 0)
+                string output = process.StandardOutput.ReadToEnd();
```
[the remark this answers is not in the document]
CharliePoole: It was debug code, which I removed.

src/NUnitCommon/nunit.common.tests/ExceptionHelperTests.cs
```diff
@@ -4,11 +4,10 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Reflection;
-using NUnit.Common;
 using NUnit.Engine;
 using NUnit.Framework;
 
-namespace NUnit.ConsoleRunner
+namespace NUnit
```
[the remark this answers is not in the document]
CharliePoole: The common assemblies already used 'NUnit' as their default namespace. and only one class overrode that.all classes in the root directory used it. It causes surprising errors when helper classes like that contain commonly used extensions are in a namespace that is not referenced in a using statement, so I actually think they _do_ belong in 'NUnit'. 

Other commonly used classes are more a matter of judgement on a case by case basis but I wouldn't want to forbid use of the root namespace outright.

I'll take another a look at what's in the root and fix the ones that were not changed as well.

src/NUnitCommon/nunit.common.tests/TextDisplay/ResultReporterTests.cs
```diff
@@ -5,20 +5,36 @@
...
     public class ResultReporterTests
     {
         private XmlNode _result;
-        private ResultReporter _reporter;
-        private StringBuilder _report;
+
+        private StringBuilder _reportBuilder;
+        private ExtendedTextWrapper _writer;
+
+        private string ReportOutput => _reportBuilder.ToString();
```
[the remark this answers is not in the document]
CharliePoole: True. I'll make it lazy.

src/NUnitCommon/nunit.common.tests/TextDisplay/ResultReporterTests.cs
```diff
@@ -70,7 +88,7 @@ public void ReportSequenceTest()
 
             foreach (string title in reportSequence)
             {
-                var index = report.IndexOf(title);
+                var index = ReportOutput.IndexOf(title);
```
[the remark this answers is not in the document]
CharliePoole: Lazy will take care of this.

src/NUnitCommon/nunit.common.tests/TextDisplay/ResultReporterTests.cs
```diff
@@ -121,12 +139,10 @@ public void ErrorsAndFailuresReportTest()
...
 
-            var actualErrorFailuresReport = GetReport(_reporter.WriteErrorsFailuresAndWarningsReport);
+            ResultReporter.WriteErrorsFailuresAndWarningsReport(_result, _writer);
 
-            foreach (var ex in expected)
-            {
-                Assert.That(actualErrorFailuresReport, Does.Contain(ex));
-            }
+            foreach (var item in expected)
+                Assert.That(ReportOutput.Contains(item));
```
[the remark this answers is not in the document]
CharliePoole: Lazy again.

src/NUnitCommon/nunit.common.tests/TextDisplay/ResultReporterTests.cs
```diff
@@ -158,15 +174,15 @@ public void TestsNotRunTest()
...
-        [Test, Explicit("Displays failure behavior")]
-        public void WarningsOnlyDisplayOnce()
-        {
-            Assert.Warn("Just a warning");
-        }
+        //[Test, Explicit("Displays failure behavior")]
+        //public void WarningsOnlyDisplayOnce()
+        //{
+        //    Assert.Warn("Just a warning");
+        //}
```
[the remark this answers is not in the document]
CharliePoole: Ah yes. It was messing me up because the test was executed. If so, that's a framework regression. I'll make a note to check this separately and delete it.

src/NUnitCommon/nunit.common/ClientTestResult.cs
CharliePoole: Let me know what you think about my rationale above.

src/NUnitCommon/nunit.common/TextDisplay/ResultReporter.cs
CharliePoole: Probably a left-over but I'll check.

src/NUnitConsole/nunit4-console/TestEventHandler.cs
CharliePoole: I moved it and then put it back because I think each runner has to decide how to handle events. Will fix.

src/NUnitCommon/nunit.common/ResultSummary.cs
CharliePoole: I asked @mikkelbu for some background on that change, made seven years ago, as this was the part of the change that gave me the most trouble

ResultReporter was already using PassedInAttribute rather than Passed. I changed how the number is calculated so that Passed could be used but the values are still the same, and all tests are still passing.

@mikkelbu Can you take a look at the last commit, where I made that change?

src/NUnitCommon/nunit.common.tests/TextDisplay/ResultReporterTests.cs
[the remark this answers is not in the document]
CharliePoole: NO... Lazy isn't useful here as it would have to be done for each test anyway. I'll remove this.

src/NUnitCommon/nunit.common.tests/ExceptionHelperTests.cs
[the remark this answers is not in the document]
CharliePoole: I've done a fair bit of moving stuff around and deleted one class entirely. The root folder now contains classes in both the 'NUnit' and 'NUnit.Common' namespaces and the subfolders to not use 'Common' at all. This is definitely not the standard approach where folder name == namespace but it all seems right to me for an assembly used in common by multiple components.

Unless something particularly bothers you here, I think I'd like to start a discussion about standards. Ideally, we would have a statement of our preferred standards, categorized as DO, DONT, CONSIDER and AVOID. We could then have a final assembly-by-assembly code review before the 4.0 final release or one of the betas. We could also make sure our use of analyzers was consistent with the standards.

I'll send you a more detailed note about this idea.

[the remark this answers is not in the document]
manfred-brands: Ok.
We had a situation at work where some code was always available because it was in our root namespace.
It pulled in unexpected dependencies.

One other thing is that if there is a namespace (prefix) per project, it is easy to see what part of the code uses what part of a referenced project.
This is less relevant for projects only referenced in one repository, but important for projects used other packages.
If a package is updated or deprecated it is easier to find what depends on it or if a reference can be removed.

src/NUnitCommon/nunit.common/ResultSummary.cs
manfred-brands: Your update to ResultReportedTest now expects: Failed_Errors to be 1, previously it was 7. The 6 difference is the `PassedButFailedInTearDown` you added to `Passed`

The `PassedInAttribute` was used in other tests: NUnitFrameworkDriverTests and TestAgentRunnerTests.
If that is what comes back from the engine/framework then indeed the added code in ResultSummary is wrong.
If we want tests to be considered failed if the teardown fails, then this counting might need updating in the framework.

If we keep the counting we have now, you can remove the `bool failedInFixtureTearDown` parameter.

manfred-brands (approved):
Nothing more, except for some clean-up as a result of commented out code in the ResultSummary

CharliePoole:
> Nothing more, except for some clean-up as a result of commented out code in the ResultSummary

As a final check, I decided to run mock-assembly under the latest V3 and the current V4 builds. I got this

__Current V3 build__

Test Run Summary
  Overall result: Failed
  Test Count: 37, Passed: 17, Failed: 11, Warnings: 1, Inconclusive: 1, Skipped: 7
    Failed Tests - Failures: 1, Errors: 7, Invalid: 3
    Skipped Tests - Ignored: 4, Explicit: 3, Other: 0
  Start time: 2025-04-24 15:48:09Z
    End time: 2025-04-24 15:48:10Z
    Duration: 0.747 seconds

__Current V4 build__

Test Run Summary
    Overall result: Failed
    Test Count: 37, Pass: 17, Fail: 11, Warn: 1, Inconclusive: 1, Skip: 7
        Failed Tests - Failures: 1, Errors: 7, Invalid: 3
        Skipped Tests - Ignored: 4, Explicit: 3, Other: 0
    Start time: 2025-04-24 16:01:24Z
    End time: 2025-04-24 16:01:25Z
    Duration: 0.983 seconds

So, I'll go ahead and do the cleanup and then merge.

That said, I'm wondering if the whole business of mock-assembly containing it's own expected values was such a good idea after all. When I first created it that way, it was a whole lot simpler but over the years has taken on so many new wrinkles that we may want to think about replacing it. That would mean hard-coding the numbers into each test, so we would need to check out how many of them would be affected but it would eliminate the need for the tests to reference mock-assembly, which would be a plus.

Anyway, that would be for another time and another issue.

src/NUnitCommon/nunit.common/ResultSummary.cs
mikkelbu: > I asked @mikkelbu for some background on that change, made seven years ago, as this was the part of the change that gave me the most trouble

I'm sorry, but I cannot remember anything about this. I think I only had made a handful of PRs in the console before doing this change, so I think most of the logic was somewhat new for me.

> @mikkelbu Can you take a look at the last commit, where I made that change?

I can check it tomorrow night - so in 20 hours or so - and can also take a closer look at my old PR and see if I remember some more.

mikkelbu:
> What I think this means is that the change made in counting was only supposed to affect the summary report and results themselves. reported counts in test-result.xml. Do you recall any discussion about that? @mikkelbu ?

If I understand the question correctly (and compare it against the last sentence in https://github.com/nunit/nunit-console/pull/494#issuecomment-437884262 - copied in below) then it could be so, but I need to check the old PR to be sure.

> The only thing that could be confusing is that the XML already contains information about passed and failed tests which will not match what the console will report, but this will be an issue for both approaches. See e.g. the difference in XML for 5 tests and a failure in Setup/TearDownFixture, respectively.

CharliePoole:
@manfred-brands @mikkelbu 

OK... I've restored the original behavior. That is, when running MockAssembly under the console runner, TestResult.xml has 23 passed and 7 failed but the summary report shows only 17 passed and 11 failed due to 6 failures in higher-level teardowns. If we want to change this again at some point, that's OK but I didn't want to change the behavior as an unintended consequence of this issue.

I have changed the naming of many of the counts in MockAssembly to names that make more sense to me. Feel free to call for further changes. I introduced Passed_Raw, Errors_Raw and Failed_Raw, which represent the results of test cases _before_ any higher-level teardown is run. Simple Passed, Errors and Failed take the teardown failures into account.

Tests of the content of TestResult.xml use the XxxxRaw values. Tests of thhe console summary display use the simple values. This was not consistent earlier and it led to lots of confusion (for me)!

Please consider the three _Raw names as an invitation to come up with something better. For example, I thought about using PassedTestCaseButLaterFailed, etc. :-)

@mikkelbu I think at this point, the best use of any review you make is going to be planning for further changes. BTW I've been wondering how nunitlite displays its summary. Is it consistent with the console runner?

Issue #1660, opened 2025-04-12
Eliminate use of Remoting as a Transport between Engine and Agents


CharliePoole:
We currently use Remoting when communicating to a .NET Framework agent and TCP with a .NET Core agent.

Since the standard engine itself targets the .NET Framework, we are capable of using TCP in both environments. 

I propose we eliminate Windows Remoting and use TCP for all agents.

Although this would give us a single common Transport, I would keep the existing class structure, which permits us to add new transports to the engine.

@nunit/engine-team Thoughts on this?

manfred-brands:
I agree, it removes differences between frameworks and is less code to maintain but allowing for future new communications to be introduced.

CharliePoole:
OK, let's proceed with that approach. It's all internal, so a Feature rather than a Breaking Change.

mikkelbu:
I also agree with the above mentioned suggestion

CharliePoole:
Update: @veleek worked on this but created a PR for main rather than version4.

Because V4 has deviated so much from V3, simple cherrypicking of his commits didn't work very well, so I'm following along with his changes and have created a new branch for this issue. Currently, it's all working with one outstanding problem I intend to document separately. New PR soon.

veleek:
@CharliePoole let me know if you've made progress on this that you want me to pick up.

CharliePoole:
@veleek I've eliminated Windows Remoting but turned up a few issues. I've made this into a WIP PR so you and others can take a look at what I've done and the problems as well.

PR #1674, opened 2025-04-29
Add new interfaces and enable use of launchers internally

### LINKED ISSUE #1673: Convert Built-in agents to pluggable agents
This is part of issue #909.

IAgentLauncher will be the new extension interface for pluggable agents. It also introduces the struct TestAgentInfo, which gives information about each agent. Additional interfaces may also be needed. The first cut of the interface will be taken from that used in TestCentric and we'll develop it further from there.

Initially, IAgentInfo will be used internally only. A launcher will be created for each built-in agent. This is intended to facilitate converting agents, one by one, to pluggable extensions. Once that is complete, the code for loading these internal engines will be removed.

UPDATE

Having implemented all of the above in a PR, I realized that very little change is needed to go a great deal further and I have renamed this issue accordingly.

Both of the existing agents (net462 and net8.0)  will be converted to pluggable agents. They will be packaged and published as extensions and NUnit.ConsoleRUnner will include them as dependencies.

For the moment, they will continue to be built as a part of the console project. Creation of separate project repositories for each of them will then be a separate step and mostly a set of mechanical changes.

CharliePoole:
Fixes #1673

As explained in the issue, some of this is temporary, allowing us to use the same concept of a "launcher" for internal agents that will be used for pluggable agents. After all agents become pluggable extensions, the launcher code will disappear from  the engine itself. The plan for transitioning is...

1. Add one new pluggable agent, i.e. .NET 9.0.
2. Convert the .NET 8.0 agent to be pluggable.
3. Ditto for the .NET 4.6.2 agent.


CharliePoole:
@veleek I thought this wold be a good way for you to get  familiar with the v3 vs v4 approach to launching agents.

CharliePoole:
@nunit/engine-team I've greatly increased the scope of issue #1673 so I've asked several of you to review this. Those not named are, of course, also welcome to review or comment.

This is a key new feature and will put us very close to being able to release a beta. Whatever time you can give will be helpful.

src/NUnitEngine/nunit.engine/Services/AgentStore.cs
```diff
@@ -10,45 +10,57 @@ namespace NUnit.Engine.Services
...
-    internal sealed partial class AgentStore
+    internal sealed class AgentStore
     {
-        private readonly Dictionary<Guid, AgentRecord> _agentsById = new Dictionary<Guid, AgentRecord>();
+        private static readonly Logger log = InternalTrace.GetLogger(typeof(AgentStore));
+
+        private readonly object LOCK = new object();
+
+        private readonly Dictionary<Guid, AgentRecord> _agentIndex = new Dictionary<Guid, AgentRecord>();
+        private readonly Dictionary<Process, AgentRecord> _processIndex = new Dictionary<Process, AgentRecord>();
```
manfred-brands, line 20: You shouldn't need the dictionary on process. Process termination should be handled by the `LocalProcessAgent`

src/NUnitEngine/nunit.engine/Services/AgentStore.cs
```diff
@@ -10,45 +10,57 @@ namespace NUnit.Engine.Services
...
-            lock (_agentsById)
+            lock (LOCK)
             {
-                if (_agentsById.TryGetValue(agentId, out var record) && record.IsReady)
+                if (_agentIndex.TryGetValue(agentId, out var record)
+                    && record.Status == AgentStatus.Ready)
                 {
                     agent = record.Agent;
-                    return true;
+                    return agent is object;
```
manfred-brands, line 63: I strongly dislike `is object` to test for `not null`.
```suggestion
                    return agent is not null;
```

src/NUnitEngine/nunit.engine/Services/AgentStore.cs
```diff
@@ -72,17 +84,70 @@ public bool IsAgentProcessActive(Guid agentId, [NotNullWhen(true)] out Process?
             }
         }
 
-        public void MarkTerminated(Guid agentId)
+        public void MarkProcessTerminated(Process process)
```
manfred-brands, line 87: Shouldn't need this.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -31,57 +32,131 @@ public partial class TestAgency : ITestAgency, IService
...
+        /// </returns>
+        public IList<TestAgentInfo> GetAgentsForPackage(TestPackage targetPackage)
         {
-            if (_runtimeService is null || _availableRuntimeService is null)
-                throw new InvalidOperationException("TestAgency needs to be Started first");
+            Guard.ArgumentNotNull(targetPackage, nameof(targetPackage));
+
+            // Initialize lists with ALL available agents
+            var availableAgents = new List<TestAgentInfo>(GetAvailableAgents());
+            var validAgentNames = new List<string>(availableAgents.Select(info => info.AgentName));
```
manfred-brands, line 88:
```suggestion
            var validAgentNames = new HashSet<string>(availableAgents.Select(info => info.AgentName));
```

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -31,57 +32,131 @@ public partial class TestAgency : ITestAgency, IService
...
 
+            // Look at each included assembly package to see if any names should be removed
+            foreach (var assemblyPackage in targetPackage.Select(p => p.IsAssemblyPackage()))
+            {
+                // Collect names of agents that work for each assembly
+                var agentsForAssembly = new List<string>();
```
manfred-brands, line 94:
```suggestion
                var agentsForAssembly = new HashSet<string>();
```

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -31,57 +32,131 @@ public partial class TestAgency : ITestAgency, IService
...
+                    if (launcher.CanCreateAgent(assemblyPackage))
+                        agentsForAssembly.Add(launcher.AgentInfo.AgentName);
+
+                // Remove agents from final result if they don't work for this assembly
+                for (int index = validAgentNames.Count - 1; index >= 0; index--)
+                {
+                    var agentName = validAgentNames[index];
+                    if (!agentsForAssembly.Contains(agentName))
+                        validAgentNames.RemoveAt(index);
+                }
```
manfred-brands, line 105:
```suggestion
                validAgentNames.IntersectWith(agentsForAssembly);
```

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -31,57 +32,131 @@ public partial class TestAgency : ITestAgency, IService
...
+            // Initialize lists with ALL available agents
+            var availableAgents = new List<TestAgentInfo>(GetAvailableAgents());
+            var validAgentNames = new List<string>(availableAgents.Select(info => info.AgentName));
 
+            // Look at each included assembly package to see if any names should be removed
+            foreach (var assemblyPackage in targetPackage.Select(p => p.IsAssemblyPackage()))
+            {
+                // Collect names of agents that work for each assembly
+                var agentsForAssembly = new List<string>();
+                foreach (var launcher in _launchers)
```
manfred-brands, line 95: No point testing launchers which are already excluded:
```suggestion
                foreach (var launcher in _launchers.Where(l => validAgentNames.Contains(l.AgentInfo.AgentName)))
```

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -31,57 +32,131 @@ public partial class TestAgency : ITestAgency, IService
...
+
+            // Finish up by deleting all unsuitable entries form the List of TestAgentInfo
+            for (int index = availableAgents.Count - 1; index >= 0; index--)
+            {
+                var agentName = availableAgents[index].AgentName;
+                if (!validAgentNames.Contains(agentName))
+                    availableAgents.RemoveAt(index);
+            }
+
+            return availableAgents;
```
manfred-brands, line 116: Reads nicer to me:
```suggestion
            return availableAgents.Where(info => validAgentNames.Contains(info.AgentName)).ToList();
```

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -31,57 +32,131 @@ public partial class TestAgency : ITestAgency, IService
...
+        /// </summary>
+        /// <param name="package">The test package to be run</param>
+        /// <returns>An ITestAgent</returns>
+        /// <exception cref="ArgumentException">If no agent is available.</exception>
+        public ITestAgent GetAgent(TestPackage package)
+        {
             // Target Runtime must be specified by this point
             string runtimeSetting = package.GetSetting(EnginePackageSettings.TargetRuntimeFramework, string.Empty);
             Guard.OperationValid(runtimeSetting.Length > 0, "LaunchAgentProcess called with no runtime specified");
             bool runAsX86 = package.GetSetting(EnginePackageSettings.RunAsX86, false);
```
manfred-brands, line 149: Not used anymore, delete.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -31,57 +32,131 @@ public partial class TestAgency : ITestAgency, IService
...
-            var agentProcess = new AgentProcess(this, package, agentId);
-            var agentName = targetRuntime.Id + (runAsX86 ? "-x86" : string.Empty) + "-agent";
-            package.AddSetting("SelectedAgentName", agentName);
+            string agencyUrl = targetRuntime.FrameworkName.Identifier == ".NETFramework" ? RemotingUrl : TcpEndPoint;
+            var agentProcess = CreateAgentProcess(agentId, agencyUrl, package);
 
-            agentProcess.Exited += (sender, e) => OnAgentExit((Process)sender!, agentId);
+            agentProcess.Exited += (sender, e) => OnAgentExit(agentProcess);
 
             agentProcess.Start();
```
manfred-brands, line 158: As discussed this should be abstracted into a `LocalProcessAgent`

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -31,57 +32,131 @@ public partial class TestAgency : ITestAgency, IService
...
-            log.Debug("Launched Agent process {0} - see nunit-agent_{0}.log", agentProcess.Id);
+            log.Debug("Launched Agent process {0} - see testcentric-agent_{0}.log", agentProcess.Id);
```
manfred-brands, line 159: You did mention you got some of this from _testcentric_

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -195,6 +296,9 @@ public void StartService()
                 _runtimeService = ServiceContext.GetService<IRuntimeFrameworkService>();
                 _availableRuntimeService = ServiceContext.GetService<IAvailableRuntimes>();
 
+                _launchers.Add(new Net462AgentLauncher());
```
manfred-brands, line 299: Do we need a Mono runner? For .NET462 support on Linux

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -206,49 +310,71 @@ public void StartService()
...
-        //        {
-        //            var stopError = "Failed to stop the remote agent." + Environment.NewLine + ExceptionHelper.BuildMessageAndStackTrace(ex);
-        //            log.Error(stopError);
-        //            throw new NUnitEngineUnloadException(stopError, ex);
-        //        }
-        //    }
-        //}
+        #endregion
+
+        internal bool IsAgentProcessActive(Guid agentId, [NotNullWhen(true)] out Process? process)
```
manfred-brands, line 315: Drop the word `Process`. Abstract.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -206,49 +310,71 @@ public void StartService()
...
+        {
+            return _agentStore.IsAgentProcessActive(agentId, out process);
+        }
+
+        private void OnAgentExit(Process process)
```
manfred-brands, line 320: This handler should be in the `LocalProcessAgent`

manfred-brands (changes requested):
I would like `Process` to be abstracted, especially if you want to support `InProcess` and `RemoteProcess` agents. See further remarks embedded.

I assume that the existing `AgentProcess` and `AgentProcessTest` classes will be deleted once this is finished.

Not sure if we want to do #1660 first as this PR touches on cases where a different _transport_ is used.
If you keep RemotingTransport should all that code be in the `Net462Agent` as the only one using it.

CharliePoole:
Yes, I've been living with the tension created by GetAgent returning a Process, figuring I'd reconcile it at some point. I think this issue is the place to do it but it may be a commit or two away.

CharliePoole:
Regarding issue #1660, I had thought to do it first, but @veleek volunteered to look at it so we're working in parallel. I'm hoping to rebase on his changes at some point.

package-tests.cake
```diff
@@ -57,7 +57,7 @@ StandardRunnerTests.Add(new PackageTest(1, "Net462Test")
 AddToBothLists(new PackageTest(1, "Net80Test")
 {
     Description = "Run mock-assembly.dll under .NET 8.0",
-    Arguments = "testdata/net8.0/mock-assembly.dll",
+    Arguments = "testdata/net8.0/mock-assembly.dll --trace:Debug",
```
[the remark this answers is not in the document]
CharliePoole: Yes, to debug I have to change the script. I've been meaning to add an option to the script for applying trace:Debug to all the package tests from the command-line. Haven't gotten around to it yet.

src/NUnitEngine/nunit.engine.api/Extensibility/IAgentLauncher.cs
```diff
@@ -0,0 +1,39 @@
...
+        bool CanCreateAgent(TestPackage package);
+
+        /// <summary>
+        /// Returns an agent capable of running the specified package.
+        /// </summary>
+        /// <param name="agentId"></param>
+        /// <param name="agencyUrl"></param>
+        /// <param name="package"></param>
+        /// <returns></returns>
+        Process CreateAgent(Guid agentId, string agencyUrl, TestPackage package);
```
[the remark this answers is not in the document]
CharliePoole: Actually, RemoteProcess will be a process but with a few different implementation wrinkles. Local Agent will not be a process. But I think you are right that we need separate implementations for each agent type. I'll keep iterating toward that.

CharliePoole:
@manfred-brands I made a few off-the-cuff comments. I'll come back to these for actual action in the next commit.

src/NUnitEngine/nunit.engine.api/TestAgentInfo.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using System.Runtime.Versioning;
+
+namespace NUnit.Engine
+{
+    /// <summary>
+    /// The TestAgentInfo struct provides information about an
+    /// available agent for use by a runner.
+    /// </summary>
+    public struct TestAgentInfo
```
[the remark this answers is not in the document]
CharliePoole: Done.

src/NUnitEngine/nunit.engine/Services/AgentLaunchers/Net462AgentLauncher.cs
```diff
@@ -0,0 +1,85 @@
...
+        public Process CreateAgent(Guid agentId, string agencyUrl, TestPackage package)
+        {
+            // Should not be called unless we have previously checked CanCreateAgent
+            Guard.ArgumentValid(CanCreateAgent(package), "Unable to create agent. Check result of CanCreateAgent before calling CreateAgent.", nameof(package));
+
+            log.Debug($"CreateAgent: id={agentId} url={agencyUrl} package ={package.FullName}");
+
+            // Access other package settings
+            bool runAsX86 = package.GetSetting(EnginePackageSettings.RunAsX86, false);
+            bool debugTests = package.GetSetting(EnginePackageSettings.DebugTests, false);
```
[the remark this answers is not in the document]
CharliePoole: Now used

src/NUnitEngine/nunit.engine/Services/AgentLaunchers/Net80AgentLauncher.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        public Process CreateAgent(Guid agentId, string agencyUrl, TestPackage package)
+        {
+            // Should not be called unless we have previously checked CanCreateAgent
+            Guard.ArgumentValid(CanCreateAgent(package), "Unable to create agent. Check result of CanCreateAgent before calling CreateAgent.", nameof(package));
+
+            // Access other package settings
+            bool runAsX86 = package.GetSetting(EnginePackageSettings.RunAsX86, false);
+            bool debugTests = package.GetSetting(EnginePackageSettings.DebugTests, false);
```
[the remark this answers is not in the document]
CharliePoole: Now Used

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
CharliePoole: Are you seeing logs created with that name? If so, that's what I should fix.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
CharliePoole: We do. But that would be a function of the .NET Framework runner. Suport is currently commented out. I have a TODO for adding the support back.

src/NUnitEngine/nunit.engine/Services/AgentLaunchers/Net80AgentLauncher.cs
```diff
@@ -0,0 +1,97 @@
...
+            startInfo.UseShellExecute = false;
+            startInfo.CreateNoWindow = true;
+            startInfo.WorkingDirectory = Environment.CurrentDirectory;
+            startInfo.LoadUserProfile = loadUserProfile;
+
+            startInfo.FileName = "dotnet";
+            startInfo.Arguments = $"\"{agentPath}\" {agentArgs}";
+
+            // TODO: Remove the windows limitation and the use of a hard-coded path.
+            if (runAsX86)
```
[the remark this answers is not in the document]
CharliePoole: It's not machine dependent, but software. I.E. if any dependency requires running in an X86 process, it then the agent has to run under .NET x86. That said, I don't know how much use this gets but enough that MS supports it.

src/NUnitEngine/nunit.engine/Services/AgentLaunchers/Net462AgentLauncher.cs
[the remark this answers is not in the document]
CharliePoole: You're right. It should be!

src/NUnitEngine/nunit.engine/Services/AgentStore.cs
CharliePoole: Done

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
CharliePoole: Currently, there are two agents. Probably never more than four installed by default. Users could possibly install one or two more. Do we have any idea of the relative performance impact for small collections of agents?

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
CharliePoole: But `validAgentNames` contains all the names and doesn't exclude anything. Perhaps the naming is wrong.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
manfred-brands: Even if not, you point users to look at a non-existing one.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
CharliePoole: Actually, I think I need to rethink this API method. Why would I want a list of agents valid for any assembly in a multi-assembly package? Normally, I'd want to know valid agents of a given assembly.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
CharliePoole: Oh, sorry. I misread that was a change suggestion nunit => testcentric. I'll fix.

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
CharliePoole: Done

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
CharliePoole: Done

src/NUnitEngine/nunit.engine.tests/Services/AgentStoreTests.cs
```diff
@@ -147,7 +148,7 @@ public static void AgentIsNotRunningWhenTerminated()
 
             database.AddAgent(DummyAgentId, DummyProcess);
             database.Register(DummyAgent);
-            database.MarkTerminated(DummyAgentId);
+            database.MarkProcessTerminated(DummyProcess);
```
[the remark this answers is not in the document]
CharliePoole: No longer in code.

src/NUnitEngine/nunit.engine.tests/Services/AgentStoreTests.cs
```diff
@@ -32,7 +32,7 @@ public static void IdCannotBeReused()
             database.Register(DummyAgent);
             Assert.That(() => database.AddAgent(DummyAgentId, DummyProcess), Throws.ArgumentException.With.Property("ParamName").EqualTo("agentId"));
 
-            database.MarkTerminated(DummyAgentId);
+            database.MarkProcessTerminated(DummyProcess);
```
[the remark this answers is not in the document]
CharliePoole: Done

src/NUnitEngine/nunit.engine.tests/Services/AgentStoreTests.cs
```diff
@@ -60,7 +60,7 @@ public static void AgentMustNotRegisterAfterTerminating()
             var database = new AgentStore();
 
             database.AddAgent(DummyAgentId, DummyProcess);
-            database.MarkTerminated(DummyAgentId);
+            database.MarkProcessTerminated(DummyProcess);
```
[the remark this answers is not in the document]
CharliePoole: Done

src/NUnitEngine/nunit.engine.tests/Services/AgentStoreTests.cs
```diff
@@ -69,7 +69,8 @@ public static void AgentMustBeStartedBeforeTerminating()
         {
             var database = new AgentStore();
 
-            Assert.That(() => database.MarkTerminated(DummyAgentId), Throws.ArgumentException.With.Property("ParamName").EqualTo("agentId"));
+            Assert.That(() => database.MarkProcessTerminated(DummyProcess),
+                Throws.Exception.TypeOf<NUnitEngineException>().With.Message.EqualTo("Process terminated without registering an agent."));
```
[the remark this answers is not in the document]
CharliePoole: Done

src/NUnitEngine/nunit.engine.tests/Services/AgentStoreTests.cs
```diff
@@ -107,7 +108,7 @@ public static void AgentIsNotReadyWhenTerminated()
 
             database.AddAgent(DummyAgentId, DummyProcess);
             database.Register(DummyAgent);
-            database.MarkTerminated(DummyAgentId);
+            database.MarkProcessTerminated(DummyProcess);
```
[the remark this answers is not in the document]
CharliePoole: Done

src/NUnitEngine/nunit.engine/Services/AgentStore.cs
CharliePoole: Removed

CharliePoole:
@nunit/engine-team I've greatly increased the scope of issue #1673 so I've asked several of you to review this. Those not named are, of course, also welcome to review or comment.

This is a key new feature and will put us very close to being able to release a beta. Whatever time you can give will be helpful.

CharliePoole:
The second commit does two things:

1. Moves the implementation of the agent launcher into an abstract class. Our individual launchers only need to override a few members.

2. Applies most of @manfred-brands suggested code changes.

The exception to the code changes is the proposal that we make the launcher interface be about agents, abstracting from processes. This is where I'd like to go but it's a big change, so I'll take a shot at it in the next commit.

I can see two ways to do this:

1. Let the AgentLauncher start the process and then return an ITestAgent that wraps it, as is currently done within TestAgency. That means the launcher will also have to set up a handler for the Exited event. This implies that the AgentLauncher as it now exists would become a LocalAgentLauncher as suggested in the review.

2. Provide a way to get the process from an ITestAgent. TestAgency would get the process that way and then proceed just as it now does. This is not unreasonable since TestAgency only handles LocalProcess agents. This is probably a bit simpler, so I may try it first.

src/NUnitEngine/nunit.engine/Services/AgentLaunchers/AgentLauncherBase.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            string arguments = sb.ToString();
+
+            var startInfo = process.StartInfo;
+            startInfo.UseShellExecute = false;
+            startInfo.CreateNoWindow = true;
+            startInfo.WorkingDirectory = Environment.CurrentDirectory;
+            startInfo.LoadUserProfile = loadUserProfile;
+
+            switch (TargetRuntime.Identifier)
```
veleek, line 77: Why have this switch statement which is essentially logic specific to the subclasses of launcher instead of just isolating that logic to them.  Moving these out would also allow you to remove a bunch of stuff like AgentPath and AgentPathX86 out of this classe.

veleek: Maybe move this into an abstract `PopulateStartInfo` method?

src/NUnitEngine/nunit.engine.api/ITestAgentInfo.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Collections.Generic;
+
+namespace NUnit.Engine.Services
+{
+    /// <summary>
+    /// Objects implementing ITestAgentInfo by classes that can return information about
+    /// available agents. The interface is used by runners in order to provide information
+    /// to the user or to allow selecting agents.
+    /// </summary>
+    public interface ITestAgentInfo
```
veleek, line 12: I'm not a huge fan of TestAgentInfo not implementing ITestAgentInfo.  I feel like `ITestAgentInfoProvider` or something is more clear.

src/NUnitEngine/nunit.engine.api/ITestAgentInfo.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// <summary>
+        /// Gets a list containing <see cref="TestAgentInfo"/> for any available agents,
+        /// which are able to handle the specified package.
+        /// </summary>
+        /// <param name="package">A TestPackage</param>
+        /// <returns>
+        /// A list of suitable agents for running the package or an empty
+        /// list if no agent is available for the package.
+        /// </returns>
+        IList<TestAgentInfo> GetAgentsForPackage(TestPackage package);
```
veleek, line 28: Is there value in having two separate methods for this interface instead of just one with an optional TestPackage parameter.  In practice GetAgentsForPackage is always calling GetAvailableAgents and then filtering the result, so it could be easily changed to just not filter unless a package is provided.

src/NUnitEngine/nunit.engine/Services/AgentLaunchers/AgentLauncherBase.cs
```diff
@@ -0,0 +1,124 @@
...
+using TestCentric.Metadata;
+using System.IO;
+
+namespace NUnit.Engine.Services.AgentLaunchers
+{
+    public abstract class AgentLauncherBase : IAgentLauncher
+    {
+        protected abstract string AgentName { get; }
+        protected abstract TestAgentType AgentType { get; }
+        protected abstract FrameworkName TargetRuntime { get; }
```
veleek, line 22: Can you rename this to AgentRuntime to differentiate from the TestPackage TargetFrameworkName?

CharliePoole: Good idea!

src/NUnitEngine/nunit.engine.api/ITestAgentInfo.cs
CharliePoole: I wanted to emphasize the key difference, i.e. that GetAvailableAgents is stable for the test run. Probably should have made it a lazy property in order to signal that. Does that sound better?

CharliePoole: Made that change, initializing the property in the `StartService` method. Looks better to me.

src/NUnitEngine/nunit.engine/Services/AgentLaunchers/AgentLauncherBase.cs
CharliePoole: @veleek 
For this commit, I'm not doing any of the above but I have simplified setting up StartInfo quite a bit. Tell me what you think.

CharliePoole:
The third commit is mostly a major file reorganization. Launchers are moved to a separate project, which is how they will be when they are extensions. This creates some odd warnings due to incompatible references that are not actually used by the code. These will go away when I remove the launchers and agents.

I renamed `AgentLauncherBase` to `LocalProcessAgentLauncher', which is exactly what it is. We'll eventually have an `InProcessAgentLauncher` and probably a `RemoteProcessAgentLauncher`.

Made some other changes due to @veleek 's comments.

CharliePoole:
I'm moving away from this PR for a bit in order to work on the two independent agent projects. Once they are set up, I'll be able to remove the agent code from this project and have it depend on the two agent packages.

PR #1675, opened 2025-05-02
WIP: Issue 1660

veleek:


CharliePoole:
@veleek The branch you are merging adds three changes on top of the version4 branch from April 3. You are missing about 50 commits between then and now. To make matters worse, some of those changes moved files from one project to another, something that automatic merging doesn't handle very well.

You should always pull the branch you are going to work on so that the local copy of main or version4 is up to date before you add changes. The usual fix would be to update your local copy of version4 and then rebase issue-1660 over it. I tried that in a scratch copy of your branch and found the conflicts somewhat confusing. You might be able to dot hat better yourself, since you are familiar with the changes you made. Of your three commits, I think you can skip the first one, since it's already in version4.

When I can't use rebase or it's too hard to use it, I generally fall back on the approach of using reset and stash repeatedly and then re-applying each set of changes in the right place.

Let me know if I can help with this.

veleek:
Yup, that's why I marked it as WIP.  I branched from one of the v4 tags instead of the version4 branch because my fork didn't have version4 and I didn't notice until I tried to create this PR.  Doh!!

It was too late for me to fiddle around so I'll have to get this updated next time I'm available.

CharliePoole:
Too bad. It's happened to me too. Here's a hint for the future: make your feature branch in the NUnit-Console repository rather than under your own account. That way, it's simpler for multiple people to collaborate on it and the feature branch you create still remains isolated from any other branches being worked on.

CharliePoole:
@veleek If the changes don't get applied soon, we'll then be dealing with making the change across three different repositories, as the agents now have their own repos.

So, unless you're already working on a new PR, I'd like to go ahead and try to apply your own commits to current build, including the internal agents and then, assuming that works, to the new pluggable agents. That will put us in a place to eliminate the internal versions and simplify everything. Let me know, please.

PR #1676, opened 2025-05-11
WIP: Eliminate use of Windows Remoting

### LINKED ISSUE #1660: Eliminate use of Remoting as a Transport between Engine and Agents

CharliePoole:
Fixes #1660

Making this WIP PR so others can comment on some issues I've run into. See furthre comments...


CharliePoole:
The basic work is done, but there are two problems:

1. There either is or was a problem running multiple multiple assemblies. If any of the assemblies targets .NET Framework, one of the results is intermittently lost. I could make the problem disappear by specifying `--agents=1`, so I concluded that there is some underlying race condition when multiple agents run in parallel. I didn't really do anything to fix it in the third commit but merely added some code to help debug the problem, which then disappeared! I'm afraid that all I really did was to change the timings. The code for this is all in `AggregatingTestRunner`, where I could use some more eyes.

2. Using TCP for the .NET Framework assemblies caused another problem to surface, which also exists for .NET Core. but wasn't observed before. When a test package is transmitted using TCP, all package settings become strings. I've observed this with bool settings and expect to see it with ints, doubles, etc. as well. A real solution will require making our settings type safe, which they have never been in the past. That is, the code interpreting any settings is required to know the expected type of each setting when retrieving it. In our own tests, this has only affected us in a few places and I've managed to eliminate the errors for now. I'd like to take on the task of revising how we handle settings as a separate issue after this one is resolved.

CharliePoole:
Unless someone has any suggestions for change, I'll merge this as is in the next few days. It's clearly not ready for a production release, but is blocking some other tasks.

veleek:
I would need to dig into the issue for #2 to have some thoughts.  Is it getting serialized as JSON somewhere along the line and then deserialized as a `Dictionary<string, string>` or something.  While briefly looking through the code I noticed that the TcpTransport is using BinzaryFormatter which is obsolete (per https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md).  Since this NUnit.Common is a .NET Standard library it still exists here, but it should probably be replaced with something else.  

Is it this formatter which is causing the loss of type information?

veleek:
(Whoops, working through my e-mail, I see #1677 has more info)

CharliePoole:
@veleek BinaryFormatter is still used in main branch, but not version4. Probably better to leave it as is, at least for now, since the replacement is causing some side effects and main represents our current production release. Ideally, we'll get V4 out soon enough that we won't need to make further big changes to V3.

Issue #1680, opened 2025-05-21
The filter "tests" truncates the provided value at the first closing parenthesis.


east1k:
Hello everyone!

The filter "tests" truncates the provided value at the first closing parenthesis, which can lead to the following:
 - It breaks further use of the filter.
 - It returns more results than expected.
 - It returns fewer results than expected.

The bug appeared when we upgraded `nunit.consolerunner.netcore` from version `3.18.3` to version `3.20.0`.
I suspect that the bug itself was introduced in version `3.19.0` as a result of this commit: https://github.com/nunit/nunit-console/commit/1af35ed44618c64fa44d9d33028648ee879d7f06.

The bug reproduces in the following cases:
 - If a regex with parentheses is used, the parser may truncate the regex without an explicit error.

   <details>
   
   <summary>Example</summary>
   
   If I run the following command:
   ```
   nunit "test.dll" -where "test =~ '(namespace1|namespace2)\\.test1'" --explore
   ```
   
   The parser will truncate to `(namespace1|namespace2)`, and I will get all tests from these namespaces instead of `namespace1.test1` and `namespace2.test1`.
   
   </details>

 - If a regex without parentheses is used but with test arguments, the parser may truncate the regex without an explicit error.

   <details>
   
   <summary>Example</summary>
   
   If I run the following command:
   ```
   nunit "test.dll" -where "test =~ 'namespace\\.test1\\(1\\)|namespace\\.test2\\(2\\)'" --explore
   ```
   
   The parser will truncate to `namespace\\.test1\\(1\\)`, and I will get only one test instead of two.
   
   </details>

 - If a regex with parentheses is used, it may break the regex with an error.

   <details>
   
   <summary>Example</summary>
   
   If I run the following command:
   ```
   nunit "test.dll" -where "test =~ '(namespace\\.test1\\(1\\)|namespace\\.test2\\(2\\))'" --explore
   ```
   The parser will truncate to `(namespace\\.test1\\(1\\)`, resulting in an invalid regex that will fail when creating Regex:
   ```
   Unit.Engine.NUnitEngineException : An exception occurred in the driver while exploring tests.
     ----> System.Reflection.TargetInvocationException : Exception has been thrown by the target of an invocation.
     ----> System.Text.RegularExpressions.RegexParseException : Invalid pattern '(namespace\.test1\(1\)' at offset 22. Not enough )'s.
   
   --NUnitEngineException
   An exception occurred in the driver while exploring tests.
[...]

PR #1681, opened 2025-05-21
Fix issue https://github.com/nunit/nunit-console/issues/1680

east1k:
Changes:
 - If the filter is "tests" and the operator is regex, take the value without normalization.
 - If the filter is "tests" and the operator is comparison, take the value using "GetTestName".
 - GetTestName takes the value to the end, rather than truncating at the first closing parenthesis.
 - GetTestName removes extra spaces everywhere except in quoted arguments.
 - GetTestName ignores escaped quotes.


east1k:
@dotnet-policy-service agree

src/NUnitEngine/nunit.engine/Services/TestSelectionParser.cs
```diff
@@ -125,7 +125,9 @@ public string ParseFilterElement()
             {
                 case "test":
                     op = Expect(REL_OPS);
-                    rhs = GetTestName();
+                    rhs = op == MATCH_OP || op == NOMATCH_OP
+                        ? Expect(TokenKind.String, TokenKind.Word)
+                        : GetTestName();
```
CharliePoole, line 130: All tests pass without this change. Could you either remove it or add a test that shows it's required?

src/NUnitEngine/nunit.engine/Services/TestSelectionParser.cs
```diff
@@ -163,35 +165,15 @@ private Token GetTestName()
...
-                        default:
-                            sb.Append(ch);
-                            done = ch == ')';
-                            break;
-                    }
+                    if (ch == ' ' && !inQuotes)
+                        continue;
+                    sb.Append(ch);
+                    inQuotes = (!inQuotes && ch == '"') || (inQuotes && ch != '"') || (inQuotes  && ch == '"' && inEscape);
+                    inEscape = inQuotes && !inEscape && ch == '\\';
```
CharliePoole, line 176: Much cleaner!

CharliePoole (changes requested):
See inline comment. It's not clear to me why you want to avoid use of `GetTestName()` for regex expressions. With your other changes, it seems to work fine.

src/NUnitEngine/nunit.engine/Services/TestSelectionParser.cs
CharliePoole: Ah, of course! A regex __may__ be an actual test name, but is only required to __match__ a test name. So calling `TestTestName` really doesn't make sense. This means that the user has to be especially precise in formulating regular expressions and also has to be aware of how best to enter them at the command-line for each OS. But it has always been so with regex. :-)

No change needed. I'm going to merge this.

Issue #1500, opened 2024-10-15
Eliminate the NetCore runner and the NUnit dotnet tool


CharliePoole:
Discussion under #1493 have led to this decision. I'll excerpt some of it here so all the discussion can be in one place.

@manfred-brands commented:
> I'm still confused about the why we have all these different ways to run tests. Historically yes, we used nunit(gui) to run the test. Is this still needed? Now we only use dotnet test for CI and VS Test Explorer to run any tests on all my multi-targeting libraries. The only reason for a non-dotnet command is to allow running tests on a system with no SDK installed. That particular situation will be addressed with the new MS self-contained runners.

@CharliePoole commented:
> Regarding the various ways to run tests, you make a good point. I'm confused as well. Back when I was heavily involved in nunit, I worked on what I've been calling the "standard" console runner. It exists because of a perceived need to execute a number of different test assemblies, possibly with different runtime targets, in a single run, producing a single result file. Most of the complexity of the engine is there in order to deal with that requirement. If it ever ceased to be a need, then the runner and engine could be drastically simplified.

> I do think you're right to ask why we need the dotnet nunit tool at all. I know that xunit.net dropped their dotnet tool a long time ago. If the tool didn't exist, there would still be four other ways to run tests: the standard console runner, nunitlite, dotnet test and the test explorer. Since the tool is created by subtracting features from the standard console runner, it clearly doesn't do anything the standard runner does. @OsirisTerje Is this just a matter of not introducing a breaking change? If so, let's take do-as-little-as-possible approach right now and plan on getting rid of it in V4.

@OsirisTerje commented:
> I kind of like the ease of installing it, compared to the NUnit.Console, but when I look at the download numbers, they are very very low. Total is less than 200K. Last 6 weeks we're talking less than thousand per version.

> With the upcoming self-executing new MS Testing platform adding into the set of runners, I agree this one can go. Sad, but.....

@CharliePoole commented:
> FWIW, here's how I install NUnit.Console:  choco install nunit-console-runner. I'm of the opinion that, for executable applications, chocolatey has all other means of installation beat. I agree that dotnet tool install is very easy for the initial installation, but it doesn't get me any extensions, whereas I can do choco install nunit-extension-some-extension and it just works.

[...]

PR #1683, opened 2025-05-23
Eliminate unuused code for running more than one assembly per agent

CharliePoole:
We have a lot of legacy code going back to the practice of running multiple assemblies in the same process or even the same AppDomain, which we no longer support. In this PR, I have tried to clean out the legacy code without making other changes. I've added some TODOs in cases  where new issues will be created.

I've also limited this to fixing runners. There is more legacy code in certain services, for which I'll create a new issue.


manfred-brands (approved):
A few small issues, I added a commit to address those.

What was the reason to allow multiple dlls instead of simply requiring starting two separate instances? Is this to get aggregated logging?
Personally, I wouldn't think it is a useful feature. We would have a 1:1 mapping between a library and its tests.

CharliePoole:
Thanks for reviewing... I like this way of working by the way. It might be interesting to try solving some problems by alternating commits... like pairing but without necessarily working at the same time.

WRT your question, the console runner and .NET Framework engine had lots of ways of running things and still do. Because processes were considered more expensive, we ran in multiple domains or even multiple test assemblies in the same domain. We allowed running processes slightly later with each process potentially having multiple domains and sometimes multiple test assemblies per domain. It's good we are going to a simpler model in V4.

But when .NET came along, there was uncertainty about how people would want to use it. Frankly, I'm still uncertain about what features would be desirable. If we only do what dotnet test does, then there may be no reason to have a runner at all. Personally, I would prefer NOT to restore the features whose tests I commented out. OTOH we may want to have a complete .NET build of the V4 standard runner, launching processes to run multiple tests. For my own work, I prefer running one test assembly at a time while developing and reserving multi-test out of process execution for CI.

manfred-brands:
>  If we only do what dotnet test does, then there may be no reason to have a runner at all.

A good question. We haven't used nunit-console at my work, with 100+ repositories in CI with GitHub actions.

Even in multi-assembly repositories, dotnet test on the solution will run multiple assemblies in parallel.

What do you think nunit-console.can add?

manfred-brands:
> I like this way of working by the way. It might be interesting to try solving some problems by alternating commits

I use it at work as we work across the planet with colleagues is Australia, Europe and the US. For simple review changes it saves a day due to the timezone differences. Instead of the original developer making the changes and then asking for another review. If the changes are more extensive it depends, sometimes making part of the changes is faster than trying to explain what is required in words.

CharliePoole:
A quick top-of-my-head answer...

A combined report in NUnit3 or NUnit2 format, extensibility, running each test assembly under a separate runtime (unless dotnet test with the solution does that), running .net framework tests as well (and combining the results).

Also one less dependency in the chain, i.e.

dotnet test => nunit vs adapter => engine => framework
vs
console => engine => framework

That last, however could be changed if the adapter no longer depended on the engine.

You'll recall that I tried to eliminate the runner some months ago... https://github.com/nunit/nunit-console/issues/1500

Some users in that discussion used the engine directly rather than the console. Really, when we talk about the console runner, >90% of the code is actually in the engine. If we didn't publish the console, we would have to maintain a test runner for executing the engine anyway.

Issue #1101, opened 2022-01-20
Improvements to the NETCORE console runner


CharliePoole:
I'm writing this to replace issue #858. That issue has become cluttered with a lot of confusing comments... many of them my own. :-( The confusion largely arises because it wasn't always clear whether we were talking about the NETFX runner or the NETCORE runner... both of which have the same name!

In closing #858 I summarized the problem in a comment, which i'll repeat here...

1. The NETCORE console runner is (unfortunately) called nunit3-console just like the NETFX runner, but it has entirely different capabilities.

2. The NETCORE runner does not launch agent processes to isolate tests but runs them in-process.

3. The NETCORE runner has some support for engine extensions, but of course only for extensions, which target .NET Core or .NET Standard. I only know of one such extension... our V2 Result Writer. All our other extensions are NETFX only.

4. The NETCORE runner itself targets .NET Core 3.1 and is set up to allow rollover to new major versions if 3.1 is not installed. That means the tests it runs have to target .NET Core 3.1 or lower.

5. Since there is no VS project loader extension compatible with .NET Core, the NETCORE runner can only run assemblies entered at the command-line.

(Everyone... please correct me if any of the above seems wrong!)

I'd like to use this issue to restart the discussion about our NETCORE runner (or runners) and define more clearly where we want to go in coming releases. Once we have made our decisions, this Epic issue will be used to gather all the sub-issues we create to implement them.

CharliePoole:
Here is a list of topics I think we need to discuss and make decisions on. I have tried to think very broadly about this and not limit myself to what we are already doing. Feel free to do the same as you suggest additions or changes.

1. Is there only one NETCORE runner, do we need two of them? I'm thinking we may have a need for a .NET Core build of the __full__ console runner in the future. After all, .NET FX won't be around forever. OTOH, the runner that is used as a dotnet global or local tool might be expected (by users) to have different capabilities.
2. Should the NETCORE runner - or at least the one used as a tool - have a different name?
3. What capabilities should that runner have? Should it launch processes? Support extensions? Handle projects in addition to assemblies? Solutions?
4. If the runner used as a tool doesn't launch processes, shouldn't it then target the latest release of .NET?
5. Again... if we are not launching processes... do we need to use the engine at all? Would it be simpler to call the framework directly as nunitlite does?

CharliePoole:
@nunit/engine-team Ideas and comments? This seems like a really key decision to make.

jpactor:
[...]

PR #1701, opened 2025-06-17
Bundle .NET 9.0 agent with Runner

### LINKED ISSUE #1700: Bundle .net 9.0 agent with runner
This is a newly created pluggable agent, which we will bundle with the next release.

CharliePoole:
Fixes #1700

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -40,7 +40,8 @@ BuildSettings.Packages.AddRange(new PackageDefinition[] {
                 "nunit.extensibility.dll", "nunit.extensibility.api.dll", "nunit.common.dll",
                 "nunit.engine.api.dll", "testcentric.metadata.dll"),
             HasDependency("NUnit.Extension.Net462PluggableAgent", "4.0.0-alpha.6"),
-            HasDependency("NUnit.Extension.Net80PluggableAgent", "4.0.0-alpha.4")
+            HasDependency("NUnit.Extension.Net80PluggableAgent", "4.0.0-alpha.4"),
+            HasDependency("NUnit.Extension.Net90PluggableAgent", "4.0.0-alpha.1")
         },
         symbols: new PackageCheck[] {
             HasDirectory("tools").WithFiles(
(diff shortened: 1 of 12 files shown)


CharliePoole:
I've asked four of you to "review" this PR. We really only need a line-by-line code review from one of you, so please somebody say you'll do that. :-)

What I'd like from everyone is more of a conceptual review around the issue of the .NET Core Runner, now that we are supporting .NET 9.0 in the standard runner.

Since the .NET Core runner targets .NET 8.0, it can't handle .NET 9.0 tests in general. Some of my simple tests run but others don't. I've added comments to `package-tests.cake` to indicate that. 

SO...

What should we do?
* Change the netcore runner to target 9.0? (Makes it hard for those who are still on 8.0)
* Add in a  temporary .Net 9.0 runner, as we once did for .NET 8.0. (I dislike temporary packages!)
* Include both runners in one package. (But I don't think one package can host two dotnet tool commands.)

As to timeframe... I'm looking for a solution that will work for us until we have a .NET Core runner that launches agents, equivalent to our present .NET Framework runner.

CharliePoole:
BTW, all the above could apply as well to our 3.20 release, which includes a .NET 9.0 agent but only supports 8.0 in the .NET Core runner. The same options would work although we could choose different approaches for V3 and V4.

mikkelbu:
> Change the netcore runner to target 9.0? (Makes it hard for those who are still on 8.0)

We follow LTS, so we are still on .NET8 and will stay on that until .NET10 is released. Can you be more precise what exactly "Makes it hard for those who are still on 8.0" means? (I've not been following this topic closely)

> Add in a temporary .Net 9.0 runner, as we once did for .NET 8.0. (I dislike temporary packages!)

I guess this would be quite temporary as .NET10 will be released in less than 5 months.

> Include both runners in one package. (But I don't think one package can host two dotnet tool commands.)

If it is not possible, then I guess I lean towards the temporary .Net 9.0 runner, but this is more a feeling than any deep insight, so hopefully some of the others have some more qualified input

CharliePoole:
@mikkelbu Thanks for the review. The new agent itself has package tests... taken from the .NET 8.0 runner with .Net 9.0 tests added in. The 9.0 tests were also added here, where a few failed, leading me to raise the general question.

For the general question, I think my best bet may be to do nothing until some user asks for something. :-) Like you, it's not the runner I use in any of my own work, even though I'm maintaining it, so it's hard to guess what people will want.

Probably, old issue #1101 is the best place to resolve the general question anyway, so I'll just go ahead and merge this.

PR #1710, opened 2025-06-23
Fix Invalid XML from TestFixtureBuilder

### LINKED ISSUE #491: TestFilterBuilder builds invalid XML
(Updated)

This implementation mishandles surrogate pairs. Control characters other than tab, carriage return and line feed cannot be represented in XML and should throw, IIRC. (See https://github.com/nunit/nunit3-vs-adapter/issues/484.)

https://github.com/nunit/nunit-console/blob/af89a437153955fb3aea0ea168965f596a6f3066/src/NUnitEngine/nunit.engine/Services/TestFilterBuilder.cs#L82-L90

`GetFilter()` should be using `XmlWriter.Create` to a `StringWriter`. This will not only get us out of the responsibility of implementing the XML spec, it will probably also result in some significant performance gains. (Judging by the chained `string.Replace` calls.)

(Test names containing control characters cannot be represented in XML without first encoding using base64 or something custom. This will need to be dealt separately with in the framework.)

@nunit/engine-team Does this seem right to you? Can we let @MatthewBeardmore get started on this if he is interested?

CharliePoole:
Fixes #491


stevenaw:
Thanks @CharliePoole 
This looks alright from my phone but not with enough confidence to give a formal approval. 
Thanks for fixing this long-standing issue!

CharliePoole:
@veleek Have you had a chance to look at this?

CharliePoole:
This is a minor code change, so I'll go ahead and merge without the review.

Issue #1706, opened 2025-06-21
Continued errors running multiple test assemblies in parallel


CharliePoole:
This was thought to be fixed by PR #1702 but the problem still turns up intermittently when more than one assembly is executed without specifying --agents=1, i.e. when they are run in parallel.

@manfred-brands 
Your fix in #1702 looked like this...

```
if (_launchers is null)
{
    lock(_launchersLock);
    (
        if (_launchers is not null)
            return _launchers; // DoubleCheck locking

        // Code to initialize the list of launchers
    }
}
```

I'm slightly suspicious of the double-check, even though it makes sense to me. That is, the first runner that gets past the lock will cause the list to be completely initialized, therefore, we should just use that list. Still, a multi-threaded test would give us complete confidence.

Assuming it's not the double-check (which I would bet on) then I guess we need to consider that it's a problem with how we handle the TCP messages providing the test results. In that case, we have multiple instances of `TestAgentTcpProxy`, one per assembly being run. Each proxy should be using a separate socket, on which it sends commands and receives results. I'm wondering if the agents are all using the correct socket, i.e. not the socket on which they originally contacted the agency but the one being used to send and receive commands.

Assigning this to you as a followup to #1702 but let me know if you think it's outside your wheelhouse.

CharliePoole:
I'm currently overriding failed tests in these package tests to merge fixes. I think I'll go back to using `--agents=1` so that's not necessary.

PR #1711, opened 2025-06-24
Issue1706 parallel test run

manfred-brands:
Trying to reproduce #1706


CharliePoole:
I removed the last remaining `--agents=1` from the package tests. I can't get it to fail either locally or in CI so I'll merge it in as soon as everything passes.

manfred-brands:
But you saw builds failing which is the reason you put in the `--agents=1` again?
Otherwise we are back at the code form #1702

CharliePoole:
No, it was always --agents=1, even before the commit you reverted, which is why the reversion didn't take it out.

IIRC, the first errors I saw came when a PR was merged, so it could be that there was a glitch in the merge. I re-ran your PR a couple of times, since the failure had seemed to be intermittent. It keeps passing.

PR #1722, opened 2025-07-02
Exclude some not run tests from console report

### LINKED ISSUE #431: Suppress "skipped tests" output in console-runner for Explicit or Ignore tests
NUnit console: "nunit3-console.exe" v3.8.0

We use console-runner in a test script as part of git processes like `git rebase --exec` and `git bisect`. Often we'll have tests marked `Explicit` (they are slow tests, we don't need them for the git process, but we'll run them once we finished the process) or `Ignore` (we try and avoid this, but it happens if there is an issue open to fix it, but we can't get to it _right now_).

The issue is that the `Tests Not Run` output swamps the console (or the output log which is just the console output written to a file) when the script is run. All we really want to see on these test outputs is Green/Red (and if red then what went wrong). In this case the tests that were marked Explicit or Ignore are not important, and actually make it harder to go through the output log. The test summary also gives the total Skipped, Explicit and Ignore counts which is enough to note that there are tests that are not being run.

Things I tried:

- Use `--where=EXPRESSION` to exclude tests marked Explicit or Ignore (doesn't work)
- Use `--noresult` (doesn't work)
- Use `--trace=Error` (doesn't work)

The only other workaround I could think of is to mark every Explicit or Ignore test also with a Category that can be excluded using `--where`, but we never remember to do this when it matters, and the ongoing argument has been that if we have already added one attribute that should be enough to remember to do.

TL;DR for the console runner is there a way to, or can we have a parameter that would, suppress the "Tests Not Run" output.

------------

Sample test script:  <details><summary>(expand for full details).</summary>
<p>

```bash
#!/bin/sh
# exit on error
set -e

# ... do git stuff here to skip deliberately red commits in a red-green two-step ...

# ... do build scripts here ...

# Run unit tests
"C:\Program Files\NUnit\nunit3-console.exe" \
    //noresult \
    //stoponerror \
    ./src/A.Test.Project/A.Test.Project.csproj

status=$?
if [ $status -ne 0 ]
    then
        echo "Test failure, status $status."
        exit $status
fi

echo "Tests passed."
```

</p></details>

Repro steps:

1. Have a test dll, mark a test as Explicit and another as Ignore
2. Run the console runner using the shell script
3. The output includes a "Tests not run" list with the Explicit and Ignore test listed.

----------

CharliePoole:
Fixes #431

### DIFF
--- a/src/NUnitCommon/nunit.agent.core/AgentDirectRunner.cs
+++ b/src/NUnitCommon/nunit.agent.core/AgentDirectRunner.cs
@@ -40,7 +40,7 @@ namespace NUnit.Agents
                 var xmlResult = runner.Run(null, TestFilter.Empty).Xml;
                 var summary = new ResultSummary(xmlResult);
 
-                ResultReporter.ReportResults(summary, OutWriter);
+                new ResultReporter().ReportResults(summary, OutWriter);
 
                 var pathToResultFile = Path.Combine(_options.WorkDirectory, "TestResult.xml");
                 WriteResultFile(xmlResult, pathToResultFile);
--- a/src/NUnitCommon/nunit.common.tests/TextDisplay/ResultReporterTests.cs
+++ b/src/NUnitCommon/nunit.common.tests/TextDisplay/ResultReporterTests.cs
@@ -3,6 +3,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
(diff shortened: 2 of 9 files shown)


CharliePoole:
See description of the change in the issue #431

veleek (approved):

PR #1723, opened 2025-07-02
Implement a test run timeout

### LINKED ISSUE #437: Command line should provide an option for run timeout
The current `--timeout` parameter when invoking command line sets the per-test timeout. Can we also have an option for the full timeout for the test run similar to `RunConfiguration.Test
SessionTimeout` from vstest?

CharliePoole:
Fixes #437

### DIFF
--- a/src/NUnitConsole/nunit4-console/ConsoleOptions.cs
+++ b/src/NUnitConsole/nunit4-console/ConsoleOptions.cs
@@ -68,11 +68,10 @@ namespace NUnit.ConsoleRunner
             get { return WhereClause is not null; }
         }
 
+        public int TestRunTimeout { get; private set; } = -1;
+        public bool TestRunTimeoutSpecified => TestRunTimeout >= 0;
         public int DefaultTestCaseTimeout { get; private set; } = -1;
-        public bool DefaultTestCaseTimeoutSpecified
-        {
-            get { return DefaultTestCaseTimeout >= 0; }
-        }
+        public bool DefaultTestCaseTimeoutSpecified => DefaultTestCaseTimeout >= 0;
 
         public int RandomSeed { get; private set; } = -1;
         public bool RandomSeedSpecified
@@ -253,7 +252,10 @@ namespace NUnit.ConsoleRunner
                     }
                 });
 
-            this.Add("testCaseTimeout=", "Set timeout for each test case in {MILLISECONDS}. May be overridden with TimeoutAttribute.",
+            this.Add("testRunTimeout=", "Set timeout for the entire test run in {MILLISECONDS}",
+                v => TestRunTimeout = parser.RequiredInt(v, "--testRunTimeout"));
+
+            this.Add("testCaseTimeout=", "Set default timeout for each test case in {MILLISECONDS}. May be overridden with TimeoutAttribute.",
                 v => DefaultTestCaseTimeout = parser.RequiredInt(v, "--testCaseTimeout"));
 
             this.Add("seed=", "Set the random {SEED} used to generate test cases.",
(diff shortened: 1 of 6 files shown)


src/NUnitConsole/nunit4-console/ConsoleOptions.cs
```diff
@@ -68,11 +68,10 @@ public bool WhereClauseSpecified
             get { return WhereClause is not null; }
         }
 
+        public int TestRunTimeout { get; private set; } = -1;
```
veleek, line 71: Would it be reasonably easy to use nullable values for these properties instead of needing a separate `BlahSpecified` property?  Or would that require a bunch of additional work on the args parsing side.

src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
```diff
@@ -429,9 +435,20 @@ private TestEngineResult RunTests(ITestEventListener? listener, TestFilter filte
 
             _eventDispatcher.OnTestEvent(startRunNode.OuterXml);
 
+            if (_testRunTimeout > 0)
+            {
+                _testRunTimer = new Timer(_testRunTimeout);
+                _testRunTimer.Elapsed += OnTestRunTimeout;
+                _testRunTimer.AutoReset = false;
+                _testRunTimer.Enabled = true;
```
veleek, line 443: I prefer using object initializers where possible.  

```suggestion
                _testRunTimer = new Timer(_testRunTimeout)
                {
                    AutoReset = false,
                    Enabled = true,
                };
                _testRunTimer.Elapsed += OnTestRunTimeout;
```

src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
```diff
@@ -429,9 +435,20 @@ private TestEngineResult RunTests(ITestEventListener? listener, TestFilter filte
...
+            }
+
             long startTicks = Stopwatch.GetTimestamp();
 
-            TestEngineResult result = PrepareResult(GetEngineRunner().Run(_eventDispatcher, filter)).MakeTestRunResult(TestPackage);
+            TestEngineResult result = GetEngineRunner().Run(_eventDispatcher, filter);
+            result = result.MakeTestRunResult(TestPackage);
+            result = PrepareResult(result);
+            //TestEngineResult result = PrepareResult(GetEngineRunner().Run(_eventDispatcher, filter)).MakeTestRunResult(TestPackage);
```
veleek, line 451: Remove this comment

src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
```diff
@@ -429,9 +435,20 @@ private TestEngineResult RunTests(ITestEventListener? listener, TestFilter filte
...
+                _testRunTimer.AutoReset = false;
+                _testRunTimer.Enabled = true;
+            }
+
             long startTicks = Stopwatch.GetTimestamp();
 
-            TestEngineResult result = PrepareResult(GetEngineRunner().Run(_eventDispatcher, filter)).MakeTestRunResult(TestPackage);
+            TestEngineResult result = GetEngineRunner().Run(_eventDispatcher, filter);
+            result = result.MakeTestRunResult(TestPackage);
+            result = PrepareResult(result);
```
veleek, line 450: This looks like a change in the logic.  The previous code is

1. Call `GetEngineRunner().Run(...)`
2. Call `PrepareResult( ... )` on it.
3. Call `MakeTestRunResult(TestPackage)` on it.

In the new code steps 2 and 3 are swapped.  No idea if this has an impact.

src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs
CharliePoole: I like it too... I was experimenting with ordering. Will update.

Issue #1724, opened 2025-07-02
Intermittent Failure in TcpServerTests


CharliePoole:
I've observed a recent failure in the test `MultipleClientConnections()` which connects 3 clients to the server and verifies that they all connected.

In the context of some other failures we have noted, this one could be very helpful since it is such a limited test:
making the connection and nothing more. Unfortunately, I have not yet been able to make it fail locally.

CharliePoole:
It turns out that the test was not testing what it should. The connections were being done synchronously rather than asynchronously. The only failure was in not recording the last connection in rare cases.

I've revised the tests so that they now fail routinely.  In this case, a good thing. :-)

I'll be creating a WIP PR in order to get more eyes on this. I first want to be sure that the test itself has no faults. Once that's sure, we can get on to fixing the bug.

PR #1725, opened 2025-07-03
Add Tests to demonstrate issue #1724

CharliePoole:
Although this is failing, I'd like to have the revised test reviewed for correctness in testing what it intends to test correctly in an asynchronous environment. In fact, that's the whole purpose of the PR at this stage.

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/Transport/Tcp/TcpServerTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Transport/Tcp/TcpServerTests.cs
@@ -2,8 +2,10 @@
 
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Net.Sockets;
 using System.Threading;
+using System.Threading.Tasks;
 using NUnit.Framework;
 
 namespace NUnit.Engine.Communication.Transports.Tcp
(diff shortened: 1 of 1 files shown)


src/NUnitEngine/nunit.engine.tests/Transport/Tcp/TcpServerTests.cs
```diff
@@ -28,42 +30,58 @@ public void StopServer()
...
-            Assert.That(_serverConnections.Count, Is.EqualTo(num), $"Should have received {num} connection events");
+            Assert.That(_serverConnections.Count, Is.EqualTo(numClients), $"Should have received {numClients} connection events");
 
-            for (int i = 0; i < num; i++)
+            for (int i = 0; i < numClients; i++)
             {
                 Assert.That(_serverConnections[i].Connected, $"Server is not connected to client {i + 1}");
                 Assert.That(clients[i].Connected, Is.True, $"Client {i + 1} is not connected to server");
             }
         }
```
manfred-brands, line 79: Need a disconnect for each client (and a dispose for the worker)

src/NUnitEngine/nunit.engine.tests/Transport/Tcp/TcpServerTests.cs
```diff
@@ -28,42 +30,58 @@ public void StopServer()
...
+
+        private void RunTask(TcpClient client)
```
manfred-brands, line 81: Was this supposed to be used in the DoWork above?

manfred-brands (commented):
A few comments.

src/NUnitEngine/nunit.engine.tests/Transport/Tcp/TcpServerTests.cs
CharliePoole: Left over from the code I replace.

CharliePoole:
The tests are now passing, with the final errors only removed after re-inserting the `Thread.Sleep(1)` check I had initially removed as an experiment. I'm going to merge this since I'm hopeful that the subject of the issue, i.e. the unreliability of the test, has been removed. I'll think about how to add some further stress testing as well, with actual runs being performed in parallel.

manfred-brands:
> re-inserting the `Thread.Sleep(1)` check

And that might still be flaky, depending on agent speed.

The issue is that the `ClientConnected` handler is executed from a separate thread in TcpServer, after the client might be finished.
So there is a time after `RunWorkerCompleted` when the handler is run.

This is only for the test, so possibly add a counter and event to the client connected as well.

PR #1775, opened 2025-11-14
PackageSettings improvements

### LINKED ISSUE #1774: Improvements to PackageSettings and SettingDefinitions APIs
These new API classes were introduced to make package settings type-aware. They are lacking a few features...

1. Because we support both known settings, defined in `SettingDefinitions` and unknown settings of Type string, int or bool, it's possible for a user to add a known setting with an incorrect Type. The methods of `PackageSettings` that permit this are `Add(string name, string value)`, `Add(string, name, bool value)` and `Add(string name, int value)`. These should give an error if a known setting name is used with an incorrect Type.
2. To support item 1, `SettingDefinitions` needs a static lookup method to provide the `SettingDefinition` associated with a name.
3. PackageSettings needs methods to remove a setting when required.

CharliePoole:
Fixes #1774

### DIFF
--- a/src/NUnitCommon/nunit.common/PackageHelper.cs
+++ b/src/NUnitCommon/nunit.common/PackageHelper.cs
@@ -165,12 +165,11 @@ namespace NUnit.Engine
             {
                 var name = xmlReader.Name;
                 var value = xmlReader.Value;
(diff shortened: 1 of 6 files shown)


rowo360 (approved):
There's not much for me to comment on here. That's all fine and solves the problem.

CharliePoole:
Thanks. I'll have more. :-)

Issue #1780, opened 2025-11-23
TestAssemblyResolver not working for .NET 10.0.0 (FindBestVersionDir)


IRebo:
Im using the current version3 branch and console runner net462 (not netcore).

Trying to run nunit test which references windows forms fails with :
 

> Unable to load one or more of the requested types.
> Could not load file or assembly 'System.Windows.Forms, Version=10.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'. The system cannot find the file specified.
>   ----> Could not load file or assembly 'System.Windows.Forms, Version=10.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'. The system cannot find the file specified.

The reason is that TestAssemblyResolver is refusing to use "C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App\10.0.0\" for an assembly with version 10.0.0.0.

The function FindBestVersionDir will try to compare `new System.Version("10.0.0") >= new System.Version("10.0.0.0") `which returns false.  Because the revision number in the 10.0.0 is not defined the version is considered lower than any version that has a revision number.

IRebo:
My hacky solution that works :

```
if (TryGetVersionFromString(Path.GetFileName(subdir), out version))
{
    Version myTargetVersion = targetVersion;
    if (version.Revision == -1) {
     myTargetVersion = new Version(targetVersion.Major, targetVersion.Minor, targetVersion.Build);
    }

    if (version >= myTargetVersion)
        if (bestVersion.Major == 0 || bestVersion > version)
            bestVersion = version;
}
```

CharliePoole:
@IRebo I released 3.22.0 without checking on this one but I think it may be fixed. Can you verify?

CharliePoole:
Yes, the new comparison will use the highest available Major+Minor version that is greater than the corresponding Major+Minor you are trying to reference. Exact matches should always occur for a test assembly you just compiled and it's direct references, but not necessarily for indirect references or for tests being run on a different machine.

Feel free to create a new issue if more problems arise. :-)

PR #1781, opened 2025-11-24
Enhanced assembly resolution for windows and aspnetcore

### LINKED ISSUE #1779: `TypeInitializationException` in `TestAssemblyResolver` Targeting .NET 10
I'm currently seeing an issue running tests with NUnit targeting the release bits of .NET 10. I'm using the following packages:

```xml
<PackageVersion Include="Microsoft.NET.Test.Sdk" Version="18.0.1" />
<PackageVersion Include="NUnit" Version="4.4.0" />
<PackageVersion Include="NUnit.Analyzers" Version="4.11.2" />
<PackageVersion Include="NUnit3TestAdapter" Version="6.0.0-beta.1" />
```

I have a `global.json` in my repo defined like this:

```json
{
  "test": {
    "runner": "Microsoft.Testing.Platform"
  }
}
```

My project is set up like this:

```xml
<PropertyGroup>
	<AllowUnsafeBlocks>true</AllowUnsafeBlocks>
	<EnableNUnitRunner>true</EnableNUnitRunner>
</PropertyGroup>
```

Note that I don't have `TestingPlatformDotnetTestSupport ` set based on [this recommendation](https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-with-dotnet-test#microsofttestingplatform-mtp-mode-of-dotnet-test).

When I run my tests, either in VS 2026 or at the command line with `dotnet test`, I see this error:

```
 NUnit.Engine.NUnitEngineException: An exception occurred in the driver while loading tests.
 ---> System.TypeInitializationException: The type initializer for 'NUnit.Engine.Internal.TestAssemblyResolver' threw an exception.
 ---> System.ArgumentNullException: Value cannot be null. (Parameter 'path1')
   at System.ArgumentNullException.Throw(String paramName)
   at System.IO.Path.Combine(String path1, String path2, String path3)
   at NUnit.Engine.Internal.TestAssemblyResolver..cctor()
   at NUnit.Engine.Internal.TestAssemblyResolver..ctor(AssemblyLoadContext loadContext, String testAssemblyPath)
   at NUnit.Engine.Internal.TestAssemblyLoadContext..ctor(String testAssemblyPath)
   at NUnit.Engine.Drivers.NUnitNetCore31Driver.Load(String assemblyPath, IDictionary`2 settings)
   at NUnit.Engine.Runners.DirectTestRunner.LoadDriver(IFrameworkDriver driver, String testFile, TestPackage subPackage)
   at NUnit.Engine.Runners.DirectTestRunner.LoadDriver(IFrameworkDriver driver, String testFile, TestPackage subPackage)
   at NUnit.Engine.Runners.DirectTestRunner.LoadPackage()
   at NUnit.Engine.Runners.DirectTestRunner.EnsurePackageIsLoaded()
   at NUnit.Engine.Runners.DirectTestRunner.Explore(TestFilter filter)
   at NUnit.Engine.Runners.MasterTestRunner.Explore(TestFilter filter)
   at NUnit.VisualStudio.TestAdapter.NUnitEngine.NUnitEngineAdapter.Explore(TestFilter filter) in C:\repos\nunit\nunit3-vs-adapter\src\NUnitTestAdapter\NUnitEngine\NUnitEngineAdapter.cs:line 95
   at NUnit.VisualStudio.TestAdapter.NUnit3TestExecutor.RunAssembly(String assemblyPath, IGrouping`2 testCases, TestFilter filter, String assemblyName) in C:\repos\nunit\nunit3-vs-adapter\src\NUnitTestAdapter\NUnit3TestExecutor.cs:line 311
  from bin\Debug\net10.0\CslaGeneratorSerialization.Analysis.Tests.dll (net10.0|x64)
```

CharliePoole:
DotNetHelper.cs (class `DotNet`) is enhanced to directly return all or selected runtimes.

`TestAssemblyResolver` no longer needs to know where dotnet is installed and its `AdditionalDirectoriesStrategy` is now replaced by `AdditionalRuntimesStrategy` and derived classes `WindowsDesktopStrategy` and `AspNetCoreStrategy`.

Once this is merged and @JasonBock has tested it I'll close #1779 manually.

Reviewing the code, I can see a lot of other areas for improvement, but that's for another day and probably for V4.

### DIFF
--- a/src/NUnitEngine/nunit.engine.core/DotNetHelper.cs
+++ b/src/NUnitEngine/nunit.engine.core/DotNetHelper.cs
@@ -84,7 +84,7 @@ namespace NUnit.Engine
             foreach (string line in DotnetCommand("--list-runtimes", x86: x86))
             {
                 string[] parts = line.Trim().Split([' '], 3);
-                yield return new RuntimeInfo(parts[0], parts[1] + ".0", parts[2].Trim(['[', ']']));
+                yield return new RuntimeInfo(parts[0], parts[1], parts[2].Trim(['[', ']']));
             }
         }
 
--- a/src/NUnitEngine/nunit.engine/Services/AgentProcess.cs
+++ b/src/NUnitEngine/nunit.engine/Services/AgentProcess.cs
@@ -69,7 +69,7 @@ namespace NUnit.Engine.Services
 
                 if (runAsX86)
                 {
-                    if (OS.IsWindows)
+                    if (!OS.IsWindows)
                         throw new Exception("Running .NET Core as X86 is currently only supported on Windows");
 
                     if (!File.Exists(StartInfo.FileName))
--- a/src/NUnitEngine/nunit.engine.core/Internal/PathUtils.cs
+++ b/src/NUnitEngine/nunit.engine.core/Internal/PathUtils.cs
@@ -144,10 +144,10 @@ namespace NUnit.Engine.Internal
 
             // if lengths are the same, check for equality
             if ( length1 == length2 )
-                return string.Compare( path1, path2, OS.IsWindows ) == 0;
+                return string.Compare( path1, path2, RunningOnWindows ) == 0;
 
             // path 2 is longer than path 1: see if initial parts match
-            if ( string.Compare( path1, path2.Substring( 0, length1 ), OS.IsWindows) != 0 )
+            if ( string.Compare( path1, path2.Substring( 0, length1 ), RunningOnWindows) != 0 )
                 return false;
             
             // must match through or up to a directory separator boundary
@@ -174,10 +174,12 @@ namespace NUnit.Engine.Internal
         /// <exception cref="ArgumentNullException"><paramref name="path"/></exception>
         public static bool IsFullyQualifiedPath(string path )
         {
-            return OS.IsWindows
+            return RunningOnWindows
                 ? IsFullyQualifiedWindowsPath(path)
                 : IsFullyQualifiedUnixPath(path);
         }
+
+        private static bool RunningOnWindows => DirectorySeparatorChar == '\\';
 
         /// <summary>
         /// Returns a value that indicates whether the specified file path is fully qualified or not on Windows operating systems.
@@ -275,7 +277,7 @@ namespace NUnit.Engine.Internal
 
         private static bool PathsEqual(string path1, string path2)
         {
-            if (OS.IsWindows)
+            if (RunningOnWindows)
                 return path1.ToLower().Equals(path2.ToLower());
             else
                 return path1.Equals(path2);


src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyResolver.cs
```diff
@@ -22,59 +23,57 @@ internal sealed class TestAssemblyResolver : IDisposable
 
         private readonly AssemblyLoadContext _loadContext;
 
-        private static readonly string INSTALL_DIR;
-        private static readonly string WINDOWS_DESKTOP_DIR;
-        private static readonly string ASP_NET_CORE_DIR;
+        private readonly IEnumerable<DotNet.RuntimeInfo> _windowsDesktopRuntimes;
```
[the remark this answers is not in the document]
manfred-brands: No they are not. The Strategy get their own list in the base class. The latter is nice and extendable.

src/NUnitEngine/nunit.engine/Services/RuntimeLocators/NetCoreRuntimeLocator.cs
```diff
@@ -15,100 +13,14 @@ public static IEnumerable<RuntimeFramework> FindRuntimes(bool x86)
         {
             List<Version> alreadyFound = new List<Version>();
 
-            foreach (string dirName in GetRuntimeDirectories(x86))
+            foreach (var runtime in DotNet.GetRuntimes("Microsoft.NETCore.App"))
```
manfred-brands: This seems to be doing partly what `TestAssemblyResolver` does, but for a different runtime category.
This one doesn't consider the WindowsDektop or AspNetCore runtimes.

src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyResolver.cs
```diff
@@ -22,59 +23,57 @@ internal sealed class TestAssemblyResolver : IDisposable
...
             
-            if (isWindowsDesktop && Directory.Exists(WINDOWS_DESKTOP_DIR))
-                ResolutionStrategies.Add(new AdditionalDirectoryStrategy(WINDOWS_DESKTOP_DIR));
-            if (isAspNetCore && Directory.Exists(ASP_NET_CORE_DIR))
-                ResolutionStrategies.Add(new AdditionalDirectoryStrategy(ASP_NET_CORE_DIR));
+            if (tryWindowsDesktopFirst)
+                ResolutionStrategies.Add(new WindowsDesktopStrategy());
+            if (tryAspNetCoreFirst)
+                ResolutionStrategies.Add(new AspNetCoreStrategy());
+
```
manfred-brands: Why is there no strategy for `Microsoft.NETCore.App`?

src/NUnitEngine/nunit.engine.core/DotNetHelper.cs
```diff
@@ -2,65 +2,113 @@
...
+                Runtimes.Add(new RuntimeInfo(parts[0], parts[1], parts[2].Trim(['[', ']'])));
+            }
+        }
+
+        /// <summary>
+        /// Get the correct install directory, depending on whether we need X86 or X64 architecture.
+        /// </summary>
+        /// <param name="x86">Flag indicating whether the X86 architecture is needed</param>
+        /// <returns></returns>
+        public static string GetInstallDirectory(bool x86) => x86
```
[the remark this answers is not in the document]
manfred-brands: Nothing in this solution uses it.
Maybe `DotnetCommand` should depending on my other comments in `NetCoreRuntimeLocator`

src/NUnitEngine/nunit.engine.core/DotNetHelper.cs
```diff
@@ -2,65 +2,113 @@
...
+                yield break;
             }
 
-            return _x86InstallDirectory;
+            while (!process.StandardOutput.EndOfStream)
+                yield  return process.StandardOutput.ReadLine();
         }
+
+#if NETFRAMEWORK
+        private static bool IsWindows => Path.DirectorySeparatorChar == '\\';
```
manfred-brands: I like this condition as I don't think any other OS flipped the path separator, but in other places `Environment.OSVersion.Platform == PlatformID.Win32NT` is used.

src/NUnitEngine/nunit.engine/Services/AgentProcess.cs
```diff
@@ -73,7 +73,7 @@ public AgentProcess(TestAgency agency, TestPackage package, Guid agentId)
                     if (Path.DirectorySeparatorChar != '\\')
                         throw new Exception("Running .NET Core as X86 is currently only supported on Windows");
 
-                    var x86_dotnet_exe = Path.Combine(DotNet.GetX86InstallDirectory(), "dotnet.exe");
+                    var x86_dotnet_exe = Path.Combine(DotNet.X86InstallDirectory, "dotnet.exe");
                     if (!File.Exists(x86_dotnet_exe))
```
manfred-brands: Suggest to move this into the `DotNet` class.

src/NUnitEngine/nunit.engine.core/Internal/TestAssemblyResolver.cs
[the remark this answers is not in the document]
CharliePoole: You're right... These are now redundant.

manfred-brands (changes requested):
Main unknown for me is the overlapping functionality of TestAssemblyResolver in finding Runtimes and the NetCoreRuntimeLocator. Both looking for separate runtimes.
The first is used to find assemblies referenced by a test assembly, the latter seems to be use to find an Agent.

CharliePoole:
@manfred-brands Thanks for the comments. I did several commits to remove unused code then squashed and force-pushed. I'll go through the remaining comments next.

src/NUnitEngine/nunit.engine/Services/RuntimeLocators/NetCoreRuntimeLocator.cs
CharliePoole: TestAssemblyResolver is associated with TestAssemblyLoadContext and is used in the agent. It isn't known to the engine at all.

The RuntimeLocators are specific to the engine and, as you say, is used to select agents. Agents (at least current as well as any I am thinking of) are not specific to WindowsDesktop or AspNetCore but only to the specific runtime target. 

Essentially, the engine needs less info to select an agent. The agent then needs a bit more info to correctly load the tests.

src/NUnitEngine/nunit.engine.core/DotNetHelper.cs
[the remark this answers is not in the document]
CharliePoole: I'll look at that.

src/NUnitEngine/nunit.engine.core/DotNetHelper.cs
```diff
@@ -2,65 +2,113 @@
...
+        private const string X64_SUBKEY1 = @"SOFTWARE\dotnet\SetUp\InstalledVersions\x64\sharedHost\";
+        private const string X64_SUBKEY2 = @"SOFTWARE\WOW6432Node\dotnet\SetUp\InstalledVersions\x64\";
+        private const string X86_SUBKEY1 = @"SOFTWARE\dotnet\SetUp\InstalledVersions\x86\InstallLocation\";
+        private const string X86_SUBKEY2 = @"SOFTWARE\WOW6432Node\dotnet\SetUp\InstalledVersions\x86\";
 
-        public static string GetInstallDirectory(bool x86) => x86
-            ? GetX86InstallDirectory() : GetX64InstallDirectory();
+        public static readonly string X64InstallDirectory;
+        public static readonly string X86InstallDirectory;
+        public static readonly List<RuntimeInfo> Runtimes;
```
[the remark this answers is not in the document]
manfred-brands: I prefer encapsulation over various pieces of code directly accessing fields.
Especially `Runtimes` as that is a mutable instance, some other code could add/remove entries and suddenly things don't work anymore.
Probably less important because you control the whole eco-system here, but in NUnit.Framework end-user code could call into this if not encapsulated.

src/NUnitEngine/nunit.engine/Services/RuntimeLocators/NetCoreRuntimeLocator.cs
manfred-brands: Thanks, but if the engine instantiated an x86 agent, the TestAssemblyResolver should only look at x86 versions.
That decision should not be up to the callers, but determined by the current runtime using `System.Environment.Is64BitProcess`

src/NUnitEngine/nunit.engine.core/DotNetHelper.cs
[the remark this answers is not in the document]
CharliePoole: I agree. Because the engine needs to discover runtimes that do not match it's process bitness, I think we should always use the x86 argument when asking for either runtimes or the install directory. Agents can use their own bitness, but specifying it in the call isn't such a hardship. And users should not even see any of this.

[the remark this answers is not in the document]
manfred-brands: I agree.

src/NUnitEngine/nunit.engine.core/DotNetHelper.cs
CharliePoole: Good idea.

src/NUnitEngine/nunit.engine.core/DotNetHelper.cs
```diff
@@ -2,65 +2,120 @@
...
+        public static string GetInstallDirectory(bool x86) => x86
+            ? _x86InstallDirectory.Value : _x64InstallDirectory.Value;
 
-            return _x64InstallDirectory;
+        /// <summary>
+        /// Get the correct dotnet.exe, depending on whether we need X86 or X64 architecture.
+        /// </summary>
+        /// <param name="x86">Flag indicating whether the X86 architecture is needed</param>
+        /// <returns></returns>
+        public static string GetDotnetExecutable(bool x86) => Path.Combine(GetInstallDirectory(x86), "dotnet.exe");
```
manfred-brands: Not that only windows has a `.exe` suffix, on Linux, the executable is just `dotnet`.

manfred-brands: I fixed this.

manfred-brands:
@CharliePoole Why does the code build with `dotnet build NUnitConsole.sln` but not with `build.cmd`?

Castle.Core seems to reference a non-existing version of System.ComponentModel.TypeConverter

src/NUnitEngine/nunit.engine.core/DotNetHelper.cs
```diff
@@ -2,65 +2,120 @@
...
-        private static string _x86InstallDirectory;
-        public static string GetX86InstallDirectory()
+        private static IEnumerable<RuntimeInfo> GetAllRuntimes(bool x86)
         {
-            if (_x86InstallDirectory == null)
-                _x86InstallDirectory = Environment.GetEnvironmentVariable("DOTNET_ROOT_X86");
+            foreach (string line in DotnetCommand("--list-runtimes", x86: x86))
+            {
+                string[] parts = line.Trim().Split([' '], 3);
+                yield return new RuntimeInfo(parts[0], parts[1], parts[2].Trim(['[', ']']));
```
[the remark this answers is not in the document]
CharliePoole: OK. Let's go with that and see if it resolves #1780 as well.

manfred-brands:
> @manfred-brands Thanks for all your help on this one. It's miles better than my first cut now.

Sorry that I broke the build, you fixed it before I had a chance to get back to it.

PR #1785, opened 2025-11-29
Handle prerelease dotnet runtimes correctly

CharliePoole:
Corrects the fix to already closed issue #1779, which causes an error when a prerelease version is encountered in the output of `dotnet --list-runtimes`. The error no longer occurs.

### DIFF
--- a/src/NUnitEngine/nunit.engine.core/DotNetHelper.cs
+++ b/src/NUnitEngine/nunit.engine.core/DotNetHelper.cs
@@ -6,6 +6,8 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Reflection;
+using System.Xml.Linq;
 
 namespace NUnit.Engine
 {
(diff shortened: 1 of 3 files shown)


src/NUnitEngine/nunit.engine.core/DotNetHelper.cs
```diff
@@ -36,18 +38,43 @@ public static class DotNet
...
+                int dash = PackageVersion.IndexOf('-');
+                IsPreRelease = dash > 0;
+
+                if (IsPreRelease)
+                {
+                    Version = new Version(PackageVersion.Substring(0, dash));
+                    PreReleaseSuffix = PackageVersion.Substring(dash + 1);
+                }
+                else
+                    Version = new Version(packageVersion);
```
manfred-brands: Inconsistent. Uses `PackageVersion` property in `if` and `packageVersion` parameter in `else`.
More important `PreReleaseSuffix` will be `null` in the `else` case.

manfred-brands (approved):
The code changes are fine.

I found an [article](https://learn.microsoft.com/en-us/dotnet/core/versions/selection#control-roll-forward-behavior) about runtime selection in .net

Indirectly it says:
_The .NET Runtime supports loading libraries that target older .NET Runtime versions. An app that's upgraded to a newer major .NET Runtime version can reference libraries and NuGet packages that target older .NET Runtime versions. It's unnecessary to simultaneously upgrade the target runtime version of all libraries and NuGet packages referenced by the app._

Which means that finding runtime based upon referenced assembly versions might not be correct, you also don't want to switch runtimes for each assembly you are trying to load. Currently the `DotNetHelper.FindBestRuntime` is stateless meaning it determines the runtime directory for every assembly passes to `AssemblyLoadContext.Resolving` over and over again. As an application can only use 1 runtime, either the first one is THE one, or you have to re-evaluate all previously loaded assemblies from another runtime.

Maybe you might have to allow specifying a major runtime.
That would allow users to test an suite compiled for say .net8.0 using the .net10 runtime or see if there are incompatibilities between 9.0.4 and 9.0.11 runtimes.

CharliePoole:
@OsirisTerje If you have any comments post-merge, I'll deal with them in the release process. You can test using 3.21.0-alpha.4

PR #1806, opened 2025-12-25
Eliminate double-loading of assemblies

### LINKED ISSUE #1807: Add support for MTP
To support MTP we must check whether the assembly is pre-loaded. If it is, we should use that assembly in it's own load context and continue to seek dependencies in that context. If we resolve any dependencies, we must load them into that context as well.

If the assembly is not pre-loaded, we create our own AssemblyLoadContext and continue to use that as before.

We expect this change to resolve nunit/nunit3-vs-adapter#1389 and possibly other issues involving pre-loaded assemblies.

CharliePoole:
Fixes #1807


package-tests.cake
```diff
@@ -476,15 +476,15 @@ public static class PackageTests
...
-        StandardAndZipLists.Add(new PackageTest(1, "AppContextBaseDirectory_NET80")
-        {
-            Description = "Test Setting the BaseDirectory to match test assembly location targeting .NET 8.0",
-            Arguments = "testdata/net8.0/AppContextTest.dll",
-            ExpectedResult = new ExpectedResult("Passed")
-            {
-                Assemblies = new ExpectedAssemblyResult[] { new ExpectedAssemblyResult("AppContextTest.dll", "netcore-8.0") }
-            }
-        });
+        //StandardAndZipLists.Add(new PackageTest(1, "AppContextBaseDirectory_NET80")
```
manfred-brands: Can the commented out  code be deleted

manfred-brands (approved):
Looks good to me, checks of assembly is loaded before loading it.

PR #1843, opened 2026-07-08
Pass all framework settings to the framework

### LINKED ISSUE #1820: TestPackage setting 'TestParametersDictionary' is not applied to test run
This bug report is about the **TestCentric GUI test runner** using the NUnit V4 engine.

One of the essential changes is the way package settings are handled. In this context it turns out that the values set in the TestPackage for setting `SettingDefinitions.TestParametersDictionary` are not applied to the test run.

I'm using this kind of test:

But although the parameters are correctly set in the TestPackage settings, the TestContext.Parameters dictionary is always empty during a test run. I have the impression that the parameters are correctly passed to the NUnit engine/agent, but somehow get lost in between. 

NUnit Engine version: 4.0.0-beta.1.39

CharliePoole:
Fixes #1820

### DIFF
--- a/src/NUnitCommon/nunit.agent.core/Runners/TestAgentRunner.cs
+++ b/src/NUnitCommon/nunit.agent.core/Runners/TestAgentRunner.cs
@@ -26,6 +26,23 @@ namespace NUnit.Engine.Runners
 
         private IFrameworkDriver? _driver;
 
+        private static SettingDefinition[] _frameworkSettingDefinitions = [
+            SettingDefinitions.DefaultTimeout,
+            SettingDefinitions.DefaultCulture,
+            SettingDefinitions.DefaultUICulture,
+            SettingDefinitions.InternalTraceWriter,
+            SettingDefinitions.LOAD,
+            SettingDefinitions.NumberOfTestWorkers,
+            SettingDefinitions.RandomSeed,
+            SettingDefinitions.StopOnError,
+            SettingDefinitions.ThrowOnEachFailureUnderDebugger,
+            SettingDefinitions.SynchronousEvents,
+            SettingDefinitions.DefaultTestNamePattern,
+            SettingDefinitions.TestParameters,
+            SettingDefinitions.TestParametersDictionary,
+            SettingDefinitions.RunOnMainThread
+        ];
+
         protected AppDomain? TestDomain { get; set; }
 
         // Used to inject DriverService for testing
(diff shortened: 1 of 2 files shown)


CharliePoole:
@rowo360 Since you reported this on V4 I'd like you to review. At the moment there are neither unit nor package tests. Can you try it out for your own example as well?

src/NUnitCommon/nunit.agent.core/Runners/TestAgentRunner.cs
```diff
@@ -26,6 +26,23 @@ public abstract class TestAgentRunner : ITestEngineRunner
...
+            SettingDefinitions.LOAD,
+            SettingDefinitions.NumberOfTestWorkers,
+            SettingDefinitions.RandomSeed,
+            SettingDefinitions.StopOnError,
+            SettingDefinitions.ThrowOnEachFailureUnderDebugger,
+            SettingDefinitions.SynchronousEvents,
+            SettingDefinitions.DefaultTestNamePattern,
+            SettingDefinitions.TestParameters,
+            SettingDefinitions.TestParametersDictionary,
+            SettingDefinitions.RunOnMainThread
```
rowo360, line 43: The SettingDefinitions.cs file contains also a section named 'Settings Used by both the Engine and the Framework' - do we need to consider those settings as well? (DebugTests, PauseBeforeRun, InternalTraceLevel and WorkDirectory)

rowo360 (approved):
Looks fine from code perspective - all framework related settings defined in the assemblyPackage are added to the dictionary and passed to the driver, now 👍 
Nevertheless, I will also check the entire use case - I'll do that tomorrow.

rowo360:
Unfortunately, that didn't quite work out completely.
It worked fine using NUnit console. I deployed the new version of 'nunit.agent.core.dll' to the agent folder first. Afterwards I started a test run adding the command line argument "--param=MyParameter=test123". My test checking this parameter works successfully now. So, I think this issue can be closed now.

Unfortunately, the use case doesn't work yet using TestCentric. I assume there must be an additional issue. I can observe that we pass all Settings to the test run in the TestCentric code, but NUnit doesn't receive all of them. And therefore cannot process and add them to the dictionary. Did they get lost somehow during serialization? 
If you can provide a hint, I can try to debug a further on - currently I don't know where to look best.
Here are two screenshots:

CharliePoole:
Thanks @rowo360 I'll get this merged and released.

I'm not surprised to see it doesn't work readily in any runner other than the console runner because the console runner and agent core are built together. We'll continue to work the TestCentric side once I release updated versions of engne, agent.core and the three agents themselves.

PR #1849, opened 2026-07-16
Remove TypeExtensionAttributes from the engine API

### LINKED ISSUE #1847: Eliminate use of TypeExtensionPointAttribute in the engine API
TypeExtensionPointAttribute provides a shortcut for defining extension points on an interface rather than with each host assembly that supports the extension. The extension path is not specified on the attribute and a standard path is used by the host.

Although convenient in some situations, this creates a dependency between the assembly containing the interface and extensibility api assembly. It's preferable in many cases for the hosting assembly to define its extension points directly, specifying the path to be used. This allows multiple hosts to use different paths if desired.

This change will reduce, but not eliminate, the dependency of the engine API on the extensibility API. Once complete, we can assess whether complete elimination of the dependency is feasible.

CharliePoole:
Fixes #1847


CharliePoole:
It turns out that most attributes had already been removed. Two remaining uses were removed in this PR.

PR #1858, opened 2026-07-28
Eliminate use of StopRun(true)

### LINKED ISSUE #1856: Eliminate forced stop from the API

CharliePoole:
Eliminates use of `StopRun(bool)` in the engine and agent core and fixes #1856.

`StopRun(false)` is replaced by `RequestStop()`, which is intended to make it clear that the test may not stop. The caller may only request it and the result depends on the test framework, possibly its version and the test code itself.

`StopRun(true)` is replace by `ForcedStop()`. If the test framework provides any facilities for such a stop, they are used. Otherwise, any non terminated processes are canceled by the engine.

The initial commit makes the necessary replacements but code to kill agent processes is not yet implemented.

### DIFF
--- a/src/NUnitCommon/nunit.agent.core.tests/Runners/TestAgentRunnerExceptionTests.cs
+++ b/src/NUnitCommon/nunit.agent.core.tests/Runners/TestAgentRunnerExceptionTests.cs
@@ -115,18 +115,7 @@ namespace NUnit.Engine.Runners
                 .Do(x => { throw new NUnitEngineException("Message"); });
 
             var ex = Assert.Throws<NUnitEngineException>(() => _runner.ForcedStop());
-            Assert.That(ex.Message, Is.EqualTo("Message"));
-        }
-
-        [Test]
-        public void StopRun_Throws_ArgumentException()
-        {
-            _driver.When(x => x.ForcedStop())
-                .Do(x => { throw new ArgumentException("Message"); });
-
-            var ex = Assert.Throws<ArgumentException>(() => _runner.ForcedStop());
-            Assert.That(ex.
-                Message, Is.EqualTo("Message"));
+            Assert.That(ex?.Message, Is.EqualTo("Message"));
         }
     }
 }
--- a/src/NUnitCommon/nunit.agent.core.tests/Runners/TestAgentRunnerExceptionTests.cs
+++ b/src/NUnitCommon/nunit.agent.core.tests/Runners/TestAgentRunnerExceptionTests.cs
@@ -111,7 +111,7 @@ namespace NUnit.Engine.Runners
         [Test]
         public void StopRun_Passes_Along_NUnitEngineException()
         {
-            _driver.When(x => x.StopRun(Arg.Any<bool>()))
+            _driver.When(x => x.ForcedStop())
                 .Do(x => { throw new NUnitEngineException("Message"); });
 
             var ex = Assert.Throws<NUnitEngineException>(() => _runner.ForcedStop());
@@ -119,15 +119,14 @@ namespace NUnit.Engine.Runners
         }
 
         [Test]
-        public void StopRun_Throws_NUnitEngineException()
+        public void StopRun_Throws_ArgumentException()
         {
-            _driver.When(x => x.StopRun(Arg.Any<bool>()))
+            _driver.When(x => x.ForcedStop())
                 .Do(x => { throw new ArgumentException("Message"); });
 
-            var ex = Assert.Throws<NUnitEngineException>(() => _runner.ForcedStop());
-            Assert.That(ex.InnerException, Is.Not.Null);
-            Assert.That(ex.InnerException, Is.InstanceOf<ArgumentException>());
-            Assert.That(ex.InnerException.Message, Is.EqualTo("Message"));
+            var ex = Assert.Throws<ArgumentException>(() => _runner.ForcedStop());
+            Assert.That(ex.
+                Message, Is.EqualTo("Message"));
         }
     }
 }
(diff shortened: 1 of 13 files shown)


CharliePoole:
@manfred-brands 
I'm keeping this as a draft until a few things are resolved...

1. I'd like to see how you resolve nunit/nunit#5373 and accommodate the engine fix to whatever you do in the driver code.
2. Deciding whether to Kill test processes from the engine in response to a forced stop and, if so, where to do it.
3. Deciding whether to continue to fabricate TestSuite results in the engine after a forced stop or to expect clients (runners) to deal with it. In the PR, that code is currently removed using conditional compilation, so we can test the adapter with it.

It would be be useful if you'd take a look at the PR as it is now and let me know what you think.

manfred-brands:
> 1. I'd like to see how you resolve [Eliminate forced stop from the API nunit#5373](https://github.com/nunit/nunit/issues/5373) and accommodate the engine fix to whatever you do in the driver code.

I have raised https://github.com/nunit/nunit/pull/5374

I left the signature for the FrameworkController the same (similar what you did earlier).
That way existing runners should still work (unless they set _force_ to `true`)

CharliePoole:
On the other items...

> 2. Deciding whether to Kill test processes from the engine in response to a forced stop and, if so, where to do it.

I'll have the agent just terminate itself when it receives a `ForcedStop()` call.

> 3. Deciding whether to continue to fabricate TestSuite results in the engine after a forced stop or to expect clients (runners) to deal with it. In the PR, that code is currently removed using conditional compilation, so we can test the adapter with it.

For now, I'm going to turn the tracker code back on.

CharliePoole:
@manfred-brands This is ready for review. I've left a few things for a later issue/PR...

1. Somebody up the hierarchy should probably wait to see if the job was cancelled. This is hard to test under the console runner. My GUI has menu options for Stop and Kill, so I'll test it there once it's using the latest NUnit code. I'm thinking that the best place for this is probably in each runner, but we'll see.
2. Right now, I'm guessing a bit about what will be in the final 5.0 release of the framework, so changes may be needed when that comes out.

src/NUnitCommon/nunit.agent.core.tests/Runners/TestAgentRunnerExceptionTests.cs
```diff
@@ -111,23 +111,22 @@ public void Run_Throws_NUnitEngineException()
...
+            _driver.When(x => x.ForcedStop())
                 .Do(x => { throw new NUnitEngineException("Message"); });
 
             var ex = Assert.Throws<NUnitEngineException>(() => _runner.ForcedStop());
             Assert.That(ex.Message, Is.EqualTo("Message"));
         }
 
         [Test]
-        public void StopRun_Throws_NUnitEngineException()
+        public void StopRun_Throws_ArgumentException()
```
manfred-brands: This method is looks identical to `StopRun_Passes_Along_NUnitEngineException`.
Proofing that `_runner.ForcedStop()` calls `_driver.ForcedStop()`

src/NUnitCommon/nunit.agent.core/Drivers/NUnitFrameworkDriver.cs
```diff
@@ -35,6 +38,7 @@ public NUnitFrameworkDriver(AppDomain testDomain, string id, AssemblyName nunitR
             Guard.ArgumentNotNull(nunitRef);
 
             ID = id;
+            _nunitVersion = nunitRef.Version.ShouldNotBeNull();
 
             if (nunitRef.Version >= MINIMUM_NUNIT_VERSION)
```
manfred-brands: Can use the new field.

```suggestion
            if (_nunitVersion >= MINIMUM_NUNIT_VERSION)
```

src/NUnitCommon/nunit.agent.core/Drivers/NUnitFrameworkDriver.cs
```diff
@@ -152,8 +159,19 @@ public void RunAsync(ITestEventListener? listener, string filter) =>
...
+        /// <summary>
+        /// Force the current test run to stop, killing threads or processes if necessary.
+        /// If no tests are running, the call is ignored.
+        /// </summary>
+        public void ForcedStop()
+        {
+            if (_api.ForcedStopSupported)
+                _api.ForcedStop();
+            else
+                Process.GetCurrentProcess().Kill();
```
manfred-brands: Is that not the same as calling `Exit`?
```suggestion
                Environment.Exit(1);
```

manfred-brands (approved):
A few small suggestions.

CharliePoole:
Thanks. Good suggestions. I'll put them in and merge in the morning.

CharliePoole:
@manfred-brands FYI, added a new exit code for the agent, CANCELLED_BY_FORCED_STOP = -6;

PR #1859, opened 2026-08-04
Remove NUnit.Agent.Core from this repo and use a reference

CharliePoole:


NUnitConsole.sln
manfred-brands, line 1: Could be much simpler if migrating to .slnx file.
There is a `dotnet sln migrate` command for it.

manfred-brands (approved):
Instead of having to repeat the version, even with the msbuild properties, you could use central package management and only specify the version in `Directory.packages.props`
See https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management

See also remark on .slnx

Both can be done separately from this PR.

CharliePoole:
@manfred-brands 
Thanks for the review. Both of those items are on my radar but, as you say, seem better done separately. I'll add issues for these.

NUnit.Engine.Api and NUnit.Agent.Core use `.slnx` and I only had to upgrade NuGet to make it work.

Experimenting with central package management a few years ago, I ran into problems with multiple package sources (NU1507 warning). However I see that Terje has made it work without having to use package source mapping, so we should give that a try.

PR #1861, opened 2026-08-17
Fix --list-extensions showing wrong target framework for extensions

### LINKED ISSUE #839: Error with --list-extensions listing for extension that target .NET Standard
From https://github.com/nunit/teamcity-event-listener/pull/75#issuecomment-732876065

The TeamCity extension which targets .NET Standard 2.0 was displayed in the --list-extensions output as targeting .NET 4.0. 

I presume this was because the console was running on .NET 4.7ish, and the console is instead (incorrectly) displaying the extensions runtime. So two issues:

1. We should be displaying an extensions target framework, rather than the current runtime framework.
2. We should look at when .NET 4.0 was displayed instead of .NET 4.7. I suspect that an extension which targets .NET 4.7 would also display as .NET 4.0, which is incorrect.

HarnageaGabriel:
Fixes #839

## Problem
`--list-extensions` never displayed a target framework for installed extensions. The user-visible symptom reported in #839 was that a TeamCity extension targeting .NET Standard 2.0 appeared to be running under .NET 4.0/4.7 — the runtime the console itself was running on, not the extension's own declared target.

## Root cause
`ExtensionAssembly` (`src/NUnitCommon/nunit.extensibility/ExtensionAssembly.cs`) already reads an extension assembly's *own* `TargetFrameworkAttribute` correctly via `TestCentric.Metadata`, exposed as `FrameworkName`. But that value was never surfaced anywhere — `ConsoleRunner.DisplayExtension` (`src/NUnitConsole/nunit4-console/ConsoleRunner.cs`) never printed a framework field at all.

The natural fix would be adding `TargetFramework` to `NUnit.Extensibility.IExtensionNode` and threading it through `ExtensionNode`. However, `nunit4-console` only ever sees extension nodes through the console's `IExtensionService`/`IExtensionNode`, which resolve (transitively, through `nunit.engine`'s `NUnit.Engine.Api` package reference) to the **externally published, version-pinned** `NUnit.Extensibility.Api` NuGet package rather than this repo's local project source. Extending the local interface doesn't reach the console without also bumping and republishing that package — out of scope for this fix.

## Fix
`ConsoleRunner` now reads each listed extension's declared target framework directly from its own assembly file (`node.AssemblyPath`), using the same `TestCentric.Metadata` technique already used by `ExtensionAssembly`. This reports what the extension itself declares, independent of the console's own runtime.

Note: the issue also speculated about a secondary bug where an extension targeting .NET 4.7 might display as .NET 4.0 — that would be in `TestCentric.Metadata`'s own attribute/runtime-version parsing (an external dependency), not in this repo, so it isn't addressed here.

## Testing
- `dotnet build` on the affected projects (0 warnings, 0 errors).
- Verified via a standalone harness against `net462` and `netstandard2.0` builds of the repo's own `FakeExtensions` test data that the new code path correctly reports `.NETFramework,Version=v4.6.2` vs `.NETStandard,Version=v2.0` for each assembly's own target, rather than the console's runtime.
- Wasn't able to get `--extensionDirectory` to pick up the loose `FakeExtensions` build locally to screenshot full CLI output (addin discovery/packaging format), so verification was done at the metadata-reading level directly. `dotnet test` for `nunit4-console.tests` currently fails to run in this environment due to a pre-existing, unrelated NUnit3TestAdapter/`nunit.engine.api v3.0.0.0` binding issue (reproduces identically on `main` without this change).

### DIFF
--- a/src/NUnitConsole/nunit4-console/ConsoleRunner.cs
+++ b/src/NUnitConsole/nunit4-console/ConsoleRunner.cs
@@ -7,6 +7,7 @@ using NUnit.Engine;
 using NUnit.Engine.Extensibility;
 using NUnit.Extensibility;
 using NUnit.TextDisplay;
+using TestCentric.Metadata;
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -412,6 +413,9 @@ namespace NUnit.ConsoleRunner
             _outWriter.Write(INDENT8 + "Version: ");
             _outWriter.WriteLine(ColorStyle.Value, node.AssemblyVersion.ToString());
 
+            _outWriter.Write(INDENT8 + "Framework: ");
+            _outWriter.WriteLine(ColorStyle.Value, GetTargetFrameworkDisplayName(node.AssemblyPath));
+
             _outWriter.Write(INDENT8 + "Status: ");
             _outWriter.WriteLine(ColorStyle.Value, node.Status.ToString());
 
@@ -432,6 +436,32 @@ namespace NUnit.ConsoleRunner
                         _outWriter.Write(ColorStyle.Value, " " + val);
                     _outWriter.WriteLine();
                 }
+            }
+        }
+
+        // Reads the target framework declared by the extension's own assembly metadata,
+        // rather than the runtime framework of this console, which is what an extension
+        // node's own runtime type may otherwise appear to reflect. See issue #839.
+        private static string GetTargetFrameworkDisplayName(string assemblyPath)
+        {
+            try
+            {
+                var resolver = new DefaultAssemblyResolver();
+                resolver.AddSearchDirectory(Path.GetDirectoryName(assemblyPath));
+                resolver.AddSearchDirectory(Path.GetDirectoryName(System.Reflection.Assembly.GetExecutingAssembly().Location));
+                var parameters = new ReaderParameters { AssemblyResolver = resolver };
+
+                using var assemblyDefinition = AssemblyDefinition.ReadAssembly(assemblyPath, parameters);
+
+                if (assemblyDefinition.TryGetFrameworkName(out string frameworkName))
+                    return frameworkName;
+
+                var runtimeVersion = assemblyDefinition.GetRuntimeVersion();
+                return $".NETFramework,Version=v{runtimeVersion.Major}.{runtimeVersion.Minor}";
+            }
+            catch (Exception)
+            {
+                return "Unknown";
             }
         }
 
--- a/build.cake
+++ b/build.cake
@@ -1,5 +1,5 @@
 // Load the recipe
-#load nuget:?package=NUnit.Cake.Recipe&version=2.0.0-beta.4.3
+#load nuget:?package=NUnit.Cake.Recipe&version=2.0.0-beta.4.9
 // Comment out above line and uncomment below for local tests of recipe changes
 //#load ../NUnit.Cake.Recipe/src/NUnit.Cake.Recipe/content/*.cake
 
--- a/src/NUnitConsole/nunit4-console/nunit4-console.csproj
+++ b/src/NUnitConsole/nunit4-console/nunit4-console.csproj
@@ -31,4 +31,8 @@
     <ProjectReference Include="..\..\NUnitCommon\nunit.common\nunit.common.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="TestCentric.Metadata" Version="$(TestCentricMetadataVersion)" />
+  </ItemGroup>
+
 </Project>


CharliePoole:
@HarnageaGabriel I approved your PR to run so we can see how it works. I'd actually prefer a unit test as well, but lets hold off on that.

I agree with what you stated about the "natural fix", which is what I would really prefer...

> The natural fix would be adding TargetFramework to NUnit.Extensibility.IExtensionNode and threading it through ExtensionNode. However, nunit4-console only ever sees extension nodes through the console's IExtensionService/IExtensionNode, which resolve (transitively, through nunit.engine's NUnit.Engine.Api package reference) to the externally published, version-pinned NUnit.Extensibility.Api NuGet package rather than this repo's local project source. Extending the local interface doesn't reach the console without also bumping and republishing that package — out of scope for this fix.

I think we could make this work but let's see how the present fix works first.

CharliePoole:
@HarnageaGabriel Your build has run afoul of ongoing changes to the build process. The recipe used by all the repos I manage has been updated, but not all the repos have been updated to work with it. I'll get a change in so it builds correctly and then retry your PR.

CharliePoole:
@HarnageaGabriel I just pushed a new version of the cake recipe, which may help. Change your build.cake to use version 2.0.0-beta.4.9 of the recipe. No guarantees, I'm afraid, since you are doing something new by accessing the extensibility framework from within the console runner. It's exactly the purpose for which I made NUnit.Extensibility a separate assembly, but it hasn't been tried out yet.

If that doesn't work, I'll pull your changes down and play with the code myself.

HarnageaGabriel:
Thanks for the pointer — bumped build.cake to NUnit.Cake.Recipe 2.0.0-beta.4.9 and pushed. Let me know if it needs another retry or if you'd rather pull the branch down yourself.

src/NUnitConsole/nunit4-console/ConsoleRunner.cs
```diff
@@ -412,6 +413,9 @@ private void DisplayExtension(IExtensionNode node)
             _outWriter.Write(INDENT8 + "Version: ");
             _outWriter.WriteLine(ColorStyle.Value, node.AssemblyVersion.ToString());
 
+            _outWriter.Write(INDENT8 + "Framework: ");
+            _outWriter.WriteLine(ColorStyle.Value, GetTargetFrameworkDisplayName(node.AssemblyPath));
```
CharliePoole, line 417: In an ideal solution, this would simply use something like `node.TargetFrameworkName` and method `GetTargetFrameworkDisplayName` wouldn't be needed. OTOH, we could deploy this immediately and its availability would assist in debugging problems we tend to run into with loading extensions.

CharliePoole (approved):
This works great now that the build is fixed. We can merge it or work for a better solution involving the extensibility API. Or we can do both. See general discussion for more about this choice.

CharliePoole:
@HarnageaGabriel

This is all passing with the use of the (interim) recipe version 2.0.0-beta.4.9. Nice work!

Theoretically speaking, we have three possible paths here.
1. Merge this
2. Continue to work toward a solution where you merely need to examine a field in the API to get the framework.
3. Do 1 and then do 2.

Not merging what we have seems silly to me so I discount 2 and I'll go ahead and merge this.

So the question remains... would you like to work on the "improved solution" going forward? You pointed out some of the difficulties with the API in your analysis, but I think they are all soluble. Equally important, they are difficulties we will eventually need to solve anyway, so why not now?

Let me know and we can start discussing the next steps.

Charlie

@OsirisTerje I wanted you to take a look at this issue because the solution @HarnageaGabriel provided is the same sort of thing that the adapter may need to do at some point - not for the same reason, but for some as yet unknown problem with extensions.

HarnageaGabriel:
@CharliePoole I'm interested in taking a crack at the improved solution — threading TargetFramework through IExtensionNode properly. Could you point me at the repo where NUnit.Extensibility.Api actually lives/gets published from? Want to understand the versioning/release process before proposing changes there, and happy to loop in @OsirisTerje if there's overlap with the adapter's needs. No fixed timeline on my end, just want to get the scope right.

CharliePoole:
@HarnageaGabriel That's great. For a fuller discussion of the overall process, we should probably get out of this thread. :-) Send me an email at charliepoole AT gmail DOT com and I'll respond with more info.

As to the repo... it's this one! look at `src/NUnitCommon/nunit.extensibility` and `src/NUnitCommon/nunit.extensibility.api`. It's possible that these will be moved to a separate repo, but for now they are here. Two other packages, `NUnit.Engine.Api` and `NUnit.Agent.Core` have been moved to separate repos and the extensibility packages may be moved as well, but only after some other pending changes are made.

As for the adapter, that's a bit further in the future, but I can imagine it needing to use the extensibility interface directly rather than through the engine, so I wanted to point this issue out.

PR #1867, opened 2026-08-19
Add AssemblyTargetFramework to IExtensionNode

### LINKED ISSUE #1866: Add TargetFramework of the extension assembly to IExtensionNode
PR #1861 fixing issue #839 demonstrated the desirability of adding the TargetFramework of an extension to `IExtensionNode`. Because it was not available, the console runner had to use it's own  access to extensibility in order to pull out the target framework of the loaded extension. While this works well and is an excellent demonstration of how a non-engine component can access extensibility, it really should not be necessary. If we were to add console extensions in addition to engine extensions, then we'd use NUnit.Extensibility directly but this is really an engine extension we're talking about.

The interface already has `AssemblyPath` and `AssemblyVersion` properties, so let's add `AssemblyTargetFramework`. At the same time the code added to the runner by #1861 can be simplified.

HarnageaGabriel:
## Summary
- Adds `AssemblyTargetFramework` (`System.Runtime.Versioning.FrameworkName`) to `IExtensionNode`, alongside the existing `AssemblyPath`/`AssemblyVersion`, populated the same way from `ExtensionAssembly.FrameworkName` in `ExtensionNode`'s constructor.
- Simplifies the workaround added by #1861 (fixing #839): `ConsoleRunner.DisplayExtension` previously had to independently re-read each extension assembly's metadata via Mono.Cecil/TestCentric.Metadata (`GetTargetFrameworkDisplayName`) just to display its target framework, since the interface didn't expose it. That method is now removed and the console reads `node.AssemblyTargetFramework` directly. The now-unused `TestCentric.Metadata` package reference is dropped from `nunit4-console.csproj` (still used elsewhere in the solution, so the central `PackageVersion` entry is untouched).

Fixes #1866

## Test plan
- [x] `dotnet cake -t Test` (via `build.cmd -t Test`) — full Build + Test succeeded (300/300 net462 non-skipped tests passed, 313/313 net8.0 non-skipped tests passed).
- [x] `build.cmd -t Package` fails locally with `MSB4041` on `nunit.extensibility.api.csproj`, but this reproduces identically on a clean checkout of `main` with no changes applied — pre-existing local MSBuild-toolchain issue (old .NET Framework MSBuild 4.8 invoked by the Package task can't parse the SDK-style `<Project Sdk=...>` project), unrelated to this change.

### DIFF
--- a/src/NUnitCommon/nunit.extensibility.api/IExtensionNode.cs
+++ b/src/NUnitCommon/nunit.extensibility.api/IExtensionNode.cs
@@ -2,6 +2,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Runtime.Versioning;
 
 namespace NUnit.Extensibility
 {
@@ -70,6 +71,11 @@ namespace NUnit.Extensibility
         /// The version of the assembly implementing this extension.
         /// </summary>
         Version AssemblyVersion { get; }
+
+        /// <summary>
+        /// The target framework of the assembly implementing this extension.
+        /// </summary>
+        FrameworkName AssemblyTargetFramework { get; }
 
         /// <summary>
         /// Gets a collection of the values of a particular named property.
--- a/src/NUnitCommon/nunit.extensibility/ExtensionNode.cs
+++ b/src/NUnitCommon/nunit.extensibility/ExtensionNode.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.Versioning;
 using TestCentric.Metadata;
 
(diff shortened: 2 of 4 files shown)


CharliePoole (approved):
Looks good to merge. Thanks!

CharliePoole:
@HarnageaGabriel You didn't request a review but I guessed you were done and reviewed it. You can merge or continue working if you wanted to add anything. In that case just re-request a review when you're done.

In future, unless you prefer a particular person, you can just request a review from the engine core team and  we'll all get notified.

CharliePoole:
@HarnageaGabriel Ping!

PR #1869, opened 2026-08-20
Make Warning the default for InternalTraceLevel

### LINKED ISSUE #1792: Make Warning the default setting for internal trace logging
Historically, the console runner has used "Off" as the default for internal trace logging.

An argument can be made that "Error" or even "Warning" is a better choice and that turning all logging off should require a user decision.

@nunit/engine-team @nunit/core-team What do you think?

CharliePoole:
Fixes #1792


manfred-brands (approved):
Look good, no comments

Issue #1890, opened 2026-09-05
Change console runner --framework option syntax?


CharliePoole:
The current console runner implementation includes the `--framework` option, which allows the user to force execution to occur under a specific target runtime. The syntax used is

```
--framework RUNTIME [- VERSION]
```

Where RUNTIME must be one "net", "mono", "netcore" and VERSION is a 2- or 3-part version number.

This syntax is odd and unique to NUnit. I'm tempted to change it so that the argument is simply a valid TFM, like `net462` or net10.0.

This would leave two problems open...

1. How to deal with the variant where the version is NOT specified. In the past, this meant "any version", which is exactly what we do when the argument is not specified.
2. How to deal with mono, which is, of course, an alternative implementation of the .NET Framework.

My own leaning is to drop support for the missing version and require a full TFM as an argument. In the case of mono, the same TFM is used as for the Microsoft implementation. The only need to distinguish them occurs when Mono is installed under the Windows OS. I'm thinking about dealing with this corner case by adding a separate --mono option.

@nunit/engine-team @nunit/core-team What are your thoughts about this?

manfred-brands:
Developers are used to `--framework` and a TFM.

Regarding `mono`, I don't know if anyone would use mono on windows, but in case you want to allow it, then `--mono` sounds like a good option. On non-windows systems the .net framework TFM should imply mono.

CharliePoole:
@nunit/engine-team @nunit/core-team 

After some review and experimentation, I'm expanding this issue to include ensuring that the `--framework` option works _as we want it to work_. I say that becuse reading the code and the docs, I can see some potential for confusion.

The docs say this about the option...

> FRAMEWORK type/version to use for tests. Examples: mono, net-4.5, v4.0, 2.0, mono-4.0. __NOTE:__ Only use this option if you are sure all tests will run under the specified runtime. The option does not select a particular build from a project with multiple runtime targets.

I read this to say that you will run under the framework specified provided that all your tests are capable of doing so. That makes sense to me, although it doesn't say what will happen if all your tests are not capable of running as specified. It doesn't say that you have to have the runtime installed, although that seems obvious, or that you have to have installed an agent for that runtime, which may be less obvious to some people.

Here's what I propose...

1. If your syntax is wrong (invalid TFM for example) we give an error message and exit.
2. If your syntax is right but that runtime is not present, we give a different error message and exit.
3. If you have the runtime but no agent, we give an error message and exit.
[...]

PR #1891, opened 2026-09-05
Improvements to Console Runner under .NET 8.0

### LINKED ISSUE #946: Allow test engine to be created inside a running test on .NET Core

CharliePoole:
Partial fix of #1101 
Fixes #946


CharliePoole:
As it stands, this PR is a first cut at running agents under the NetCore build of the console runner.

Current limitations...
1. The runner still runs under .NET 8.0. Conversin to use .NET 10.0 is coming.
2. There is only one agent that currently works under the NetCore console... the .NET 10 agent. Other agents must be converted so that he launcher may load under .NET.
3. I haven't been able to successfully bundle agents with the runner, so they must be installed separately. For the CI build, this is done in package-tests.cake.
4. Currently, the runner only works on Windows.

src/NUnitEngine/nunit.engine.tests/Helpers/ShadowCopyUtils.cs
```diff
@@ -25,16 +24,21 @@ from assemblyName in assemblyNames
...
                 var dependency = Assembly.ReflectionOnlyLoad(dependencyName.FullName);
-
-                if (!dependency.GlobalAssemblyCache && r.Add(Path.GetFullPath(dependency.Location)))
-                {
+                if (dependency.GlobalAssemblyCache)
+                    continue;
+#else
+                // ReflectionOnlyLoad isn't available on .NET Core / .NET 5+. Load the assembly into the
+                // default context for inspection instead, and skip assemblies without a physical location.
+                var dependency = Assembly.Load(dependencyName);
```
manfred-brands: There is a specific [`MetadataLoadContext`](https://learn.microsoft.com/en-us/dotnet/standard/assembly/inspect-contents-using-metadataloadcontext

src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
 using System.Collections.Generic;
+using System.Diagnostics;
```
manfred-brands: This extra using does not co alongside more code, so was it put in accidentally?

src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs
```diff
@@ -233,32 +232,40 @@ private RuntimeFramework GetCurrentFramework()
...
+                {
+                    runtime = Runtime.NetCore;
+                }
             }
-            else if (major == 4 && Type.GetType("System.Reflection.AssemblyMetadataAttribute") is not null)
-            {
-                minor = 5;
-            }
+            else
+                throw new Exception("Platform is not recognized");
```
manfred-brands: MS suggests to use 'NotSupportedException'

src/NUnitEngine/nunit.engine/Runtime.cs
```diff
@@ -86,6 +89,15 @@ private class NetFrameworkRuntime : Runtime
             public override string FrameworkIdentifier => FrameworkIdentifiers.NetFramework;
 
             public override string ToString() => "Net";
+
+            public override string GetTFM(Version version)
+            {
+                string v = version.ToString(3);
+                if (v.EndsWith(".0"))
+                    v = v.Substring(0, v.Length - 2);
+                return "net" + v.Replace(".", string.Empty);
```
manfred-brands: Would this be close to:
```csharp
string v = v.ToString(v.Build == 0 ? 2 : 3);
return "net" + v.Replace(".", string.Empty);
```

src/NUnitEngine/nunit.engine/Runtime.cs
```diff
@@ -86,6 +89,15 @@ private class NetFrameworkRuntime : Runtime
             public override string FrameworkIdentifier => FrameworkIdentifiers.NetFramework;
 
             public override string ToString() => "Net";
```
manfred-brands, line 90: We just changed this in the framework that "NET" means .NET5+ and we use "NETFramework" for .NET Framework.

src/NUnitEngine/nunit.engine/Runtime.cs
```diff
@@ -116,6 +128,11 @@ private class NetCoreRuntime : Runtime
 
             public override string ToString() => "NetCore";
```
manfred-brands, line 126: In NUnit Framework this is now just "NET" for .NET5+

manfred-brands (approved):
I few small remarks, but mostly it is removing conditional compilation.

src/NUnitEngine/nunit.engine.tests/Helpers/ShadowCopyUtils.cs
CharliePoole: Thanks... I'll use that.

src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
CharliePoole: Right! Left over from some debug code I removed.

CharliePoole:
For the moment, I'm leaving classes `Runtime` + `RuntimeFramework` unchanged. They use a a non-standard id for each runtime, which is also used as the argument to the console `--framework=` option, i.e. `--framework=net-4.6.2` or `--framework=` netcore-8.-0`.

Issue #1890 suggests replacing that nomenclature with a TFM. You commented on it and I was waiting to see if anyone else had comments before doing it. I'll do a PR for that issue and replace `RuntimeFramework` with a new class that is entirely based on `FrameworkName` and TFM. 

I'll apply your other suggestions and merge this. Once I have all four agents working under the .NET 8.0 runner, I'll be able to activate all the package tests for it and start to deal with the limitations I listed.

CharliePoole:
@manfred-brands A question for you...

Have you built a dotnet tool containing package dependencies? I can't seem to make it work. When the package is installed, the dependencies are not automatically pulled into the install directory, where they need to be in order to be located by the extension service.

Right now, my package tests are passing by explicitly loading the .NET 10 pluggable agent. But that won't help users who install the runner and find it doesn't work because there are no agents.

manfred-brands:
> Have you built a dotnet tool containing package dependencies?

I pack using.csproj properties instead of a manual .nuspec file.

Are those dependencies tools themselves?
NuGet only installs libs with your tool.

> By default, adding a NuGet package as a dependency to a .NET tool project only bundles library (.dll) dependencies. If those dependencies are tools themselves, they won't automatically execute or register in the path

The suggested approach adding a dotnet tool manifest to your package (.config/dotnet-tools.json) which contains the tool dependencies.

> The .NET CLI will read the manifest and instantly download and register all the tools.

However those will likely all be in their own subdirectory.

CharliePoole:
I'm packing via the csproj for the netcore runner. The standard runner is using a nuspec because I wasn't able to make it work any other way.

> NuGet only installs libs with your tool.

That got me thinking. All of our extensions have traditionaly held their code under a "tools" directory, even though none of them are dotnet tools or even executables. The pluggable agents contain an executable, of course, but the "extension" part is actually the launcher, which runs that executable.

The extension service even uses a file pattern that includes "tools" to locate installed extensions. 

> The suggested approach adding a dotnet tool manifest to your package (.config/dotnet-tools.json) which contains the tool dependencies.

> > By default, adding a NuGet package as a dependency to a .NET tool project only bundles library (.dll) dependencies. If those dependencies are tools themselves, they won't automatically execute or register in the path

> However those will likely all be in their own subdirectory.

Fortunately, that's exactly what I want, provided I can find the subdirectory!

src/NUnitEngine/nunit.engine/Runtime.cs
CharliePoole: Postponing

src/NUnitEngine/nunit.engine/Runtime.cs
CharliePoole: Postponing

